PR #1, opened 2018-03-27
enable presolving in ctest

mattmilten:


src/HApp.cpp
```diff
@@ -956,14 +956,16 @@ double presolve(HModel &mod, double &time)
...
-      std::cout << "Infeasible" << std::endl;
-    else if (status == HPresolve::Unbounded)
-      std::cout << "Unbounded" << std::endl;
-    else
-      std::cout << "Unknown, status=" << status << std::endl;
-    return 0;
+      if ( status == HPresolve::Infeasible )
+        mod.problemStatus = 1;
+      else if ( status == HPresolve::Unbounded)
+        mod.problemStatus = 2;
```
mattmilten: Don't you want to use an enumerator here instead of assigning a number?

mattmilten:
Nice to see that this test also succeeds now!

src/HApp.cpp
galabovaa: I do, but that should be changed in HModel and would require a few changes across the code and is not related to presolve so I thought it is better done on another branch

mattmilten: Oh, well, I used `LP_Status_` now. Should be fine and reads more nicely.

PR #2, opened 2018-03-27
Dev ig

galabovaa:


mattmilten:
This is a pretty big merge. In general, I'd suggest to have smaller feature branches that only contain changes for a specific improvement/bug fix and are merged separately. This makes code review much easier. I know that this is not always possible.

PR #81, opened 2019-02-02
Ems format

galabovaa:


galabovaa:
reading in .ems files 
with a test
extract model name from file name

PR #85, opened 2019-02-07
Less toolchain specific bits

zhihaoy:


feldmeier:
many toolchain specific bits are being removed as part of the windows branch.

PR #99, opened 2019-02-18
use pkg-config to find Osi, etc

svigerske:
Reads the osi.pc, etc., files to figure out compiler and linker flags to build against Osi lib and Osi test lib.
Also uses pkg-config to look for COIN-OR's Data/Sample and Data/Netlib, used in unittest.

OSI_ROOT can be set.
If not set, but Osi is installed in the system (in a path where pkg-config will looks into), the Osi interface would also be build.


svigerske:
https://cmake.org/cmake/help/latest/module/FindPkgConfig.html

PR #101, opened 2019-02-19
revise flags to link against Osi again

svigerske:
- a previous version that used OSI_LINK_LIBRARIES did not work with older
  cmake versions, as this variable wasn't available then
- now use OSI_LIBRARIES and OSI_LIBRARY_DIRS, which are available with
  cmake 3.10 at least

Compare https://cmake.org/cmake/help/v3.10/module/FindPkgConfig.html and https://cmake.org/cmake/help/latest/module/FindPkgConfig.html

Also duplicate the code that adds OSI_ROOT to the pkg-config search path in check/CMakeLists.txt in case that this file is executed by cmake separately.


svigerske:
a89b0bb0ef3 removed the use of target_link_directories

PR #116, opened 2019-02-27
Readme

galabovaa:


galabovaa:
Oh so sorry! I will correct that :) I should have known better, no H pronounced either and words are spelled as they sound!

PR #216, opened 2019-10-31
handling reduced to empty case.

galabovaa:


galabovaa:
Most of Highs.cpp is formatting only.

PR #217, opened 2019-11-06
Ipx today

galabovaa:


feldmeier:
src/interior_point/basiclu/src/lu_timer.c uses <time.h> which is toolchain/OS dependent

feldmeier:
Still breaks some versions of gcc on linux, some clang builds, and all macos builds. Even when it works it throws a lot of warnings

galabovaa:
I am working on it right now.

PR #273, opened 2020-03-25
presolve max problem

galabovaa:


jajhall:
This is all very well for presolve, but what about postsolve? When a max problem has been solved, dual feasibility for non-equalities is negated. Hence dual values for variables at lower or upper bounds in the HighsSolution will not have the same sign as when a min problem has been solved. This has to be propagated into the "optimal" solution that you return. As with presolve, it can be achieved by considering yourself to be working with an LP with negated costs - flip the model costs and flip the dual values you're given. Then, when you've finished postsolve, flip the dual values that you deduce. 

For debugging, here are six models. There are two versions of each: *_objsense_max.mps and *_max.mps. The first is the LP as a max problem - achieved by means of the OBJSENSE keyword in an MPS file. This sets lp.sense_ to -1. The second is the LP as a max problem - achieved by negating the cost coefficients. Presolve, solve, postsolve and (any) resolve that's required [only happens in the case of dcp1. However, to read the *_objsense_max.mps you need the MPS file read updates in my Gomory branch!

[dcp1_max.mps.gz](https://github.com/ERGO-Code/HiGHS/files/4404014/dcp1_max.mps.gz)
[dcp1_objsense_max.mps.gz](https://github.com/ERGO-Code/HiGHS/files/4404015/dcp1_objsense_max.mps.gz)



[etamacro_max.mps.gz](https://github.com/ERGO-Code/HiGHS/files/4403896/etamacro_max.mps.gz)
[etamacro_objsense_max.mps.gz](https://github.com/ERGO-Code/HiGHS/files/4403897/etamacro_objsense_max.mps.gz)
[afiro_max.mps.gz](https://github.com/ERGO-Code/HiGHS/files/4403898/afiro_max.mps.gz)
[afiro_objsense_max.mps.gz](https://github.com/ERGO-Code/HiGHS/files/4403899/afiro_objsense_max.mps.gz)

galabovaa:
Try again :) I can't read the objsense* files but on the max* I get the same objective as soplex, as a sanity check.

jajhall:
"To read the *_objsense_max.mps you need the MPS file read updates in my Gomory branch"

PR #288, opened 2020-04-03
run the C API unit tests

mlubin:
It appears that these tests were not running as a part of `ctest`.


galabovaa:
Thank you, Miles, that was long overdue!

PR #279, opened 2020-03-30
DOC: Clarifying notes on Windows compilation.

mckib2:
Ref Issue #270. 

- Only modifies README; adds notes for Windows build using MSYS2

PR #289, opened 2020-04-03
FIX: add HConfig.h for Windows build.

mckib2:
Some `#define` statements are missing which cause errors in the build on Windows 10 VM.  I've added include of cmake-generated `HConfig.h` in two files (one of which was run through clang-format, so has a lot of cosmetic changes).

Build process (from command line, not MSYS2):
```
git clone https://github.com/mckib2/HiGHS
cd HiGHS
git checkout windows-build-fix
cmake -E make_directory ./build
cd build
cmake .. -DOPENMP=OFF
cmake --build . --config RELEASE
```

`ctest -C RELEASE` runs with no errors.

### DIFF
--- a/app/RunHighs.cpp
+++ b/app/RunHighs.cpp
@@ -11,6 +11,7 @@
  * @brief HiGHS main
  * @author Julian Hall, Ivet Galabova, Qi Huangfu and Michael Feldmeier
  */
+#include "HConfig.h"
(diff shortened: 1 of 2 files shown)


mckib2:
@galabovaa This should resolve discussion in #279

galabovaa:
Thank you, merging :)

PR #292, opened 2020-04-03
fixed typo: IPN_ON to IPX_ON

feldmeier:


feldmeier:
When I spotted this, I was wondering if there is any way to not compile IPX at the moment? The if-statement after the typo will always override me specifying -DIPX_ON=off, right?

PR #299, opened 2020-04-05
added get/set for bool option and get for string option to C API

feldmeier:
for getter of string option the user needs to provide a pointer to sufficient allocated memory

### DIFF
--- a/src/interfaces/highs_c_api.h
+++ b/src/interfaces/highs_c_api.h
@@ -101,6 +101,14 @@ int Highs_passLp(
 /*
  * @brief
  */
+int Highs_setHighsBoolOptionValue(void* highs,  //!< HiGHS object reference
+                                 const char* option,  //!< name of the option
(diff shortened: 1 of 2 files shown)


jajhall:
Thanks!

src/interfaces/highs_c_api.h
```diff
@@ -146,6 +162,14 @@ int Highs_getHighsDoubleOptionValue(void* highs,  //!< HiGHS object reference
                                     double* value        //!< value of option
 );
 
+/*
+ * @brief
+ */
+int Highs_getHighsStringOptionValue(void* highs,  //!< HiGHS object reference
+                                    const char* option,  //!< name of the option
+                                    char* value          //!< pointer to allocated memory to store value of option
```
mlubin, line 170: Given the potential for a buffer overflow, I think there should be a big warning on this function. Alternatively provide a method to query the length of the string before querying its value.

PR #303, opened 2020-04-09
Remove malloc

galabovaa:
Fixed issue with windows build in debug mode. Removed lu_timer all togehter.


jajhall:
I might have preferred the lu_timer calls to have been commented out so that a timer can be reinstated more easily in future - perhaps using the HiGHS timer. However, there's still Lukas' repo to refer to

PR #314, opened 2020-04-12
Added cleanBounds to resolve minor inconsistencies in bounds after presolve and return error if large

jajhall:
Added method to remove minor inconsistencies in bounds after presolve and return error if large


jajhall:
Uncontroversial modification to HighsLpUtils and call from highs.run() after presolve

PR #336, opened 2020-04-29
Dev presolve

galabovaa:
PresolveComponent class added, as an extension of the new HighsComponent: basically a place holder for data, which helps make it clearer how presolve interacts with Highs. 

Mostly what was in PresolveInfo before but was getting very  confusing.

The idea is that a HighsComponent has: 

- Options (struct)
- Data (which will be kept after run(), say) (struct)
- Info (which the user will want to access (read)(struct)

And the methods in the Component operate on these three. 

Julian,


jajhall:
I've got a rough idea of what you've done, and can believe that it's more sensible than before, and will look through the code as it executes to understand better how it works. It may also help me refine my ideas on how to refactor the simplex part of HiGHS.

However, I note that it fails one of the tests

jajhall:
Still fails

PR #341, opened 2020-05-01
Fix phase one

jajhall:
HiGHS is now forced to move to Phase 2 on dual phase 1 optimality, even if infeasible. This is because no conclusions on the primal LP can be deduced. After re-perturbing the costs, have to shift any dual infeasibilities for free variables (since it's not done in phase 2 rebuild) and go to dual phase 2 to determine whether dual unboundedness implies that the LP is primal infeasible. If dual phase 2 finds a primal feasible point then the shifts are removed and primal phase 2 will identify whether the LP is primal unbounded.

Checked on primal unbounded LPs gas11 and adlittle_max, and issue334 - which is both primal and dual infeasible. Still gets the right answers for the unbounded unit tests, too.

### DIFF
--- a/check/TestLpSolvers.cpp
+++ b/check/TestLpSolvers.cpp
@@ -232,10 +232,10 @@ void testSolvers(Highs& highs, IterationCount& model_iteration_count,
   int from_i = (int)SimplexStrategy::SIMPLEX_STRATEGY_MIN;
   int to_i = (int)SimplexStrategy::SIMPLEX_STRATEGY_NUM;
   for (int i = from_i; i < to_i; i++) {
-    if (!have_omp) {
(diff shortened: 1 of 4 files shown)


mckib2:
Issue with Windows debug build is a timeout:
```
42/47 Test #42: gas11--presolve=off ..............***Timeout 1500.02 sec
```

The Windows release build runs in under a second, so there does seem to be an issue with debug build

mckib2:
Running on Widows 10 VM this error is reported:

image

jajhall:
Thanks for this. The code is only used for printing/debugging, so I'll decide what to do in the light of your comments and merge to master tonight.

PR #343, opened 2020-05-02
Fix phase one ( #341) and Fix free list size

jajhall:
With  #341 failing because the Windows debug build was failing, this combines #341 and the fix to code for assessing free list size that was causing the build fail.

Fix phase one ( #341)

HiGHS is now forced to move to Phase 2 on dual phase 1 optimality, even if infeasible. This is because no conclusions on the primal LP can be deduced. After re-perturbing the costs, have to shift any dual infeasibilities for free variables (since it's not done in phase 2 rebuild) and go to dual phase 2 to determine whether dual unboundedness implies that the LP is primal infeasible. If dual phase 2 finds a primal feasible point then the shifts are removed and primal phase 2 will identify whether the LP is primal unbounded.

Checked on primal unbounded LPs gas11 and adlittle_max, and issue334 - which is both primal and dual infeasible. Still gets the right answers for the unbounded unit tests, too.

Fix free list size

The dodgy setting of freeListSize = *freeList.end(); has been removed. freeListSize was only ever used for reporting and debugging, and now the analysis of freeList has been moved to a method in HDmplexDebug.

### DIFF
--- a/check/TestSpecialLps.cpp
+++ b/check/TestSpecialLps.cpp
@@ -374,7 +374,7 @@ TEST_CASE("test-special-lps", "[TestSpecialLps]") {
   issue280(highs);
   issue282(highs);
   issue285(highs);
-  //  issue295(highs);
+  issue295(highs);
   issue306(highs);
   issue316(highs);
   mpsUnbounded(highs);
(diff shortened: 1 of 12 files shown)


jajhall:
Little code has really changed in these two merges, and it's well tested

PR #351, opened 2020-05-06
Feature/call from java

potehin143:


feldmeier:
Before merging this request: are we sure there are no licensing implications? I see a lot of Oracle copyright notices in some of those files.

galabovaa:
> Before merging this request: are we sure there are no licensing implications? I see a lot of Oracle copyright notices in some of those files.

I saw it is GPL2 which is OK, I believe. @jajhall Can you double check please?

Another idea I had is to somehow keep interfaces like this one in another repo in ERGO-Code/ and have HiGHS use them only optionally. The C interface should be build always since it is useful and quick. Then we avoid the licensing issue all together. Would that be OK with you @feldmeier and @jajhall?

I am using static linking for experiments and discussed with @jajhall that we should have an option to build static libs. Then will the current Python interface work, Michael? As it is calling a dynamic lib? So you know if Position Independent Code slows down execution times?

My idea was that a "fast build" should have most features the users will need to solve but not all interfaces.

mckib2:
GPL could be problematic for us. @mdhaber would this be fine if we exclude these files from our fork?

galabovaa:
> GPL could be problematic for us. @mdhaber would this be fine if we exclude these files from our fork?

It is simpler to keep our master free of problematic code. Thanks, @mckib2

jajhall:
I feel that we must keep such interfaces in another repo in ERGO-Code/. As well as the licensing issue that this interface raises, there are other issues relating to "ownership" and support. 

Although multi-contributor open source is a great thing, and we want to encourage it with HiGHS, there needs to be a central core that is (1) Purely MIT (2) Written by the HiGHS team. This allows us to give simple, clear answers when we're asked - as we have had to recently - is there any third party code, where was the code written and who owns it. HiGHS isn't a "simple, free open-source project": companies pay money for its development and we're looking to charge for support. Thus we can't have a "it's free so take-it-or-leave-it" approach.

It's already proving to be slightly awkward that IPX doesn't handle all of the unit tests posed by the Python guys, and that we can't expect quick fixes (if any) from Lukas.  I'm also conscious that, even though IPX is a good and reliable piece of code, we can't give proper support for it.

galabovaa:
Thanks, I agree with everything. Maybe we can we write a note that IPX is an optional feature and not fully supported? Meaning not usable through interfaces too?

jajhall:
> The C++ code should have an equivalent to Highs_passLp. See if passModel is what you need? In Highs.h / Highs.cpp

The C++ code only has passModel. Currently this only takes a HighsLp argument but, in future, we'll have structs for other model types, with multiple passModel methods taking the various structs as arguments. In the C API, we can't overload in the same way, so need to differentiate according to the model name.

mckib2:
@galabovaa 

> I am using static linking for experiments and discussed with @jajhall that we should have an option to build static libs. 

FYI, in scipy we are building basiclu and ipx as static libraries.  This is a limitation of Python's build system

PR #355, opened 2020-05-07
Add status to char* methods to C API

jajhall:
These methods were requested by a "client", and I've done my best to create them. However, they involve char* and strings, and I know that these are tricky in C. I only got them to return const char* Please @feldmeier have a look at them to see whether they are as good as they can be. In particular, if you think that it's advantageous, can they return char* rather than const char*. [I guess there may be a situation where someone wants to change the value after it's returned, or declare a char* and set it to the value returned by one of these methods according to some conditional statement.]


feldmeier:
Will take a look at it tonight!

jajhall:
> The code looks fine to me.
> It is not easily possible to return a non-const C string (.c_str() points to internal data structures of std::string, so we would have to create a copy, and then somehow make sure that the memory is properly deallocated).

Thanks, that was the impression I'd got from a quick trawl of Stackexchange this morning, although the details were beyond me.

PR #356, opened 2020-05-07
Fast build

galabovaa:


galabovaa:
The version numbers are defined by the project() command which is equivalent. 

Thank you, merging.

PR #361, opened 2020-05-10
ENH: Toggle crossover and add push counts to IPX info

mckib2:
@jajhall had discussed that it might be useful to turn crossover on/off to actual tell the utility of it, how much it costs to run, etc.  Crossover iterations are also always reported as 0 currently, so I've tried to do my best to report a value there.

- Adds a HiGHS option for turning off/on crossover (default true)
- Adds Primal/Dual push counts to IPX info
- Reports Primal/Dual push counts as crossover iterations

Primal/Dual push counts seem to me like they should be forwarded to HiGHS as `crossover_iteration` count, so I have assigned `iteration_counts.crossover` to be the combined push counts.

One unit test is failing due to the change in the way crossover iterations are reported and I do not understand the unit test well enough to comment on how to fix it.

### DIFF
--- a/src/ipm/IpxWrapper.h
+++ b/src/ipm/IpxWrapper.h
@@ -559,6 +559,8 @@ HighsStatus solveLpIpx(const HighsOptions& options, HighsTimer& timer,
   // Determine the run time allowed for IPX
   parameters.time_limit = options.time_limit - timer.readRunHighsClock();
   parameters.ipm_maxiter = options.ipm_iteration_limit - iteration_counts.ipm;
+  // Determine if crossover is to be run or not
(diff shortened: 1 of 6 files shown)


jajhall:
Thanks for having a go at this. I'm building my own copy of your branch and, once I've reformatted it, I'll see what changes you made and identify why the unit test fails. Once the code is merged into master I'll let you know so that you can pull the changes into your branch.

jajhall:
I've extended what you started - introduce an option run_crossover [default true]

When false, crossover isn't run, and if the IPX solution is optimal or imprecise the primal values are returned. Crossover can only be not run if presolve is not used, as presolve requires a basic solution.

PR #363, opened 2020-05-12
Crossover iteration count

jajhall:
Introduces an option run_crossover [default true] requested by @mckib2 When false, crossover isn't run, and if the IPX solution is optimal or imprecise the primal values are returned. Crossover can only be not run if presolve is not used, as presolve requires a basic solution.


mckib2:
Thanks @jajhall !  I had a chance to try this out and am getting correctly reported crossover iterations and can turn it on/off

PR #366, opened 2020-05-20
Dev presolve

galabovaa:
Presolve options:

- Max iterations
- Selection and order of presolvers
- Time limit

Time limit also added to free format mps parser


jajhall:
MPS file reader timeout is not what's needed. It only works if options.time_limit is set to a finite value, but the default is Inf. Even if options.time_limit is set, it will be to something "large" like 3600 s, and HiGHS shouldn't be silent for this length of time just because the user's given it an empty file. Surely the MPS reader can recognise that it's reached EOF and fail on this unless ENDATA has been read. 

**Presolve timeout is a mess.** 

After a long time in VScode, I now see that you have your own "timer" in the PresolveTimer class, which contains a reference to the main HiGHS timer_ (good) OK, and then its own start_time and time_limit, as well as records for each rule, including their HiGHS Timer clock id and information on what the rule achieves. For me it's more logical to give yout "timer" a different name so as (1) it's not confused with the HiGHS timer and (2) it contains more than just timing. Looks more like analysis to me! 

When you set your time_limit, you set it to options_.time_limit - fair enough - but the value presolve_.options_.time_limit set in Highs::runPresolve() [as a (horrible) difference between options_.time_limit and the HiGHS run time when runPresolve was started] isn't used!

When you set your start_time at the start of Presolve(int print), you use your timer.getTime - which is the current wall time.
 
Now, when you do your timeout test, you subtract your start_time from the current wall time and compare with the time_limit (which is an absolute time).

The whole thing would be much simpler (and correct) if you got rid of presolve_.options_.time_limit [set in Highs::runPresolve()], got rid of your start_time and had timer.getTime call timer_.readRunHighsClock() rather than timer_.getWallTime(). This means that you're relating (your) time_limit directly to the RunHighs clock - as in the simplex solver. [IPX has to look at a "time left" limit since it has its own timing mechanism that starts from zero when IPX is called.]

Your max_iterations is fine and, I guess, you intend to set its value using something from the HiGHS options.

The flexible ordering of the rules is fine

Having said all this, since the MPS and Presolve timeouts don't break master, I'm happy to approve the pull request in the interests of making progress.

galabovaa:
The MPS timeout is needed if the limit is small but the user accidentally gave it a problem which is large. From the user's point of view time_limit=3 should mean HiGHS does not run for longer than 3 but reports an error message. Many reasonably sized problems take longer than 3s to finish.

I saw the default time limit is inf() and figured in that case there shouldn't be a time out in the MPS parser. I don't understand what your consideration is here. 

PresolveAnalysis and PresolveTimer are quick hacks from initial presolve time testing we had to do earlier this year. They will both be removed. In this PR I only added the time_limit option to presolve options. 

For presolve experiments I won't be using the current HighsOptions. The implementation is messy and will be replaced with something simpler along the way. So I will be setting the Presolve options from calls from other CPP codes or Julia - so we have some progress on the HiGHS system / Julia side.

galabovaa:
@jajhall I forgot to say that presolve needs its own starting time because for larger problems initialising the LP form MPS and copying also take time to be excluded 

So in fact we have 
- highs init time (read matrix)
- presolve init time (copy matrix)
and then presolve start time

to have an accurate measure

Whether the time limit should be in simplex I am not convinced: rather I think the time limit should be overall from the user's point of view

jajhall:
> The MPS timeout is needed if the limit is small but the user accidentally gave it a problem which is large. From the user's point of view time_limit=3 should mean HiGHS does not run for longer than 3 but reports an error message. Many reasonably sized problems take longer than 3s to finish.

Agreed. There needs to be something to stop the MPS read running forever. 
 
> I saw the default time limit is inf() and figured in that case there shouldn't be a time out in the MPS parser. I don't understand what your consideration is here.

Yes, if options.time_limit is Inf, then there's no reason for the MPS reader to have a time out.

However, the MPS reader must be coded so it can't get into an infinte loop. In particular, if it's given an empty .mps file it should identify EOF before reading ENDATA and return. As it stands, unless the user has set options.time_limit to a finite value, it will just hang. Even if options.time_limit is set to something like 1000 the MPS file reader shouldn't run for 1000 seconds if it's given an empty file.
 
> PresolveAnalysis and PresolveTimer are quick hacks from initial presolve time testing we had to do earlier this year. They will both be removed. In this PR I only added the time_limit option to presolve options.

The ability to analyse the performance of the presolve in terms of time and actions must be retained. We don't want  to have to write all the code from scratch when we get an example of presolve behaviour that we don't understand. Just as I'm accumulating stuff in HSimplexDebug that will be enabled when options.highs_debug_level is positive, I suggest that the same is done for presolve.

> For presolve experiments I won't be using the current HighsOptions. The implementation is messy and will be replaced with something simpler along the way. So I will be setting the Presolve options from calls from other CPP codes or Julia - so we have some progress on the HiGHS system / Julia side.

OK Seems like a good way to experiment with a replacement system for handling the main HighsOptions

jajhall:
> @jajhall I forgot to say that presolve needs its own starting time because for larger problems initialising the LP form MPS and copying also take time to be excluded
> 
> So in fact we have
> 
>     * highs init time (read matrix)
> 
>     * presolve init time (copy matrix)
>       and then presolve start time
> 
> 
> to have an accurate measure

Some of what you want is captured in the three "hard-coded" clocks [in addition to the HiGHS clock] that are set up when an instance of HighsTimer is created. They are

- Presolve
- Solve
- Postsolve

There's not just "read matrix" in HiGHS "init", unless you mean more than reading a data file. Very few users will get HiGHS to read from a data file. With the current HighsTimer, it's easy to collect all "before (possible) presolve" time into one clock. Given that Postsolve on its own never takes time, it's almost not worth having it, but it can be left for completeness. Time solving after postsolve can be accumulated in the Solve clock, other clocks that would be of interest to us, and possibly users, are Crash, Crossover and Model modification.

However, ultimately, the solve time for an LP - as collected by Mittelman includes everything: read MPS, presolve, solve, postsolve and solve after postsolve.

Even Gurobi quotes this aggregated time, with elapsed time during solve including the time for MPS read and presolve:

Running /opt/gurobi901/linux64/bin/gurobi.sh GurobiSimplex.py
Using license file /home/jajhall/gurobi.lic
Academic license - for non-commercial use only
Read MPS format model from file ml.mps
Reading time = 1.40 seconds
HUAWEI: 1142385 rows, 1065289 columns, 3811311 nonzeros
Changed value of parameter Method to 1
   Prev: -1  Min: -1  Max: 5  Default: -1
Gurobi Optimizer version 9.0.1 build v9.0.1rc0 (linux64)
Optimize a model with 1142385 rows, 1065289 columns and 3811311 nonzeros
Model fingerprint: 0x1a48c12a
Coefficient statistics:
  Matrix range     [3e-01, 1e+03]
  Objective range  [3e-06, 5e+07]
  Bounds range     [0e+00, 0e+00]
  RHS range        [3e-14, 2e+08]
Presolve removed 1041205 rows and 951877 columns
Presolve time: 3.71s
Presolved: 101180 rows, 113455 columns, 438313 nonzeros

Iteration    Objective       Primal Inf.    Dual Inf.      Time
       0   -1.1872002e+40   1.336405e+34   1.187200e+10      4s
   62241    9.7732890e+12   0.000000e+00   3.276341e+07      6s
Extra 809 simplex iterations after uncrush
   63050    9.7732890e+12   0.000000e+00   0.000000e+00      6s

Solved in 63050 iterations and 6.04 seconds


> 
> Whether the time limit should be in simplex I am not convinced: rather I think the time limit should be overall from the user's point of view

The time limit _is_ overall from a user's point of view. The HiGHS clock is started and stopped when HiGHS methods are called, and the simplex code compares the HiGHS clock time with value of options.time_limit.

galabovaa:
Thanks, I now got what you meant about the timer. I will do it.

galabovaa:
Of course, the functionality will be retained, by "will be removed" I meant "will be replaced with neater mechanisms, like the HighsOptions case"

PR #379, opened 2020-06-07
Kkt check and other 

galabovaa:
> Highs.clear() and PresolveComponent.clear()

> Basis in postsolve

> New KKT checker (the old one may be removed from the main repo later on)


jajhall:
Looks fine, but overlapped slightly with some things I've been looking at to do with numerics - see https://github.com/ERGO-Code/HiGHS/tree/presolveNumerics

We should talk about this, and a few issues that I've spotted before you continue

PR #384, opened 2020-07-09
Create debug highs solution

jajhall:
Code for checking a HiGHS basis and solution [JAJH's KKT Check] has been moved from HighsSolution to HighsSolutionDebug, controlled by the value of high_debug_level, so that it's clearer that it isn't used to compute anything necessary for the solver. It did need one new method to compute the primal/dual infeasibilities after IPX


check/TestSpecialLps.cpp
```diff
@@ -405,16 +405,109 @@ void almostNotUnbounded(Highs& highs) {
...
+  if (require_model_status == HighsModelStatus::OPTIMAL) {
+    REQUIRE(objectiveOk(info.objective_function_value, optimal_objective));
+  }
+  status = highs.resetHighsOptions();
+  REQUIRE(status == HighsStatus::OK);
+
+  reportSolution(highs);
+  */
+}
+TEST_CASE("LP-272", "[highs_test_special_lps]") {
```
galabovaa, line 470: Thanks!

jajhall:
> Good to go! I have a maybe unrelated question:
> 
> Has getObjectiveValue() disappeared from HiGHS? How do I get the objective value after solve?

Yes, it's disappeared, since it's available via getHighsInfo. Although it would introduce duplication, I'd be OK with restoring it since Highs_getObjectiveValue is in the Julia, FORTRAN and C# interfaces. However, so is Highs_getIterationCount, which is ambiguous now that we've got IPX so ought to go.

No-one's using the FORTRAN interface and, to our knowledge, only SINTEF [Quiet since November, anyway] is using the C# interface and JuliaOpt is using the Julia interface [Friends who can be told that he API has changed]

PR #391, opened 2020-07-28
Dual postsolve

galabovaa:


jajhall:
Gives expected result on all 154 sub-1000 LPs. Iterations after postsolve for only 

- pilot87 (149) due to a single primal infeasibility in postsolve
- watson_1 (6) due to 6 dual infeasibilities after reinversion and recomputing primal and dual values - ie not a postsolve error

Also for the following - the only 1000+ second LP that had iterations after postsolve

- rail02 (223) due to 1067 dual infeasibilities after reinversion and recomputing primal and dual values - ie not a postsolve error

jajhall:
Good to go

PR #399, opened 2020-08-28
C API and static analysis fixes

mckib2:
Small fixes

- C API example in master not compiling in master using gcc 7.5 or clang-10
    - Fix: add `const` qualifiers to array sizes
- IPX Basis class was trying to declare as default an implicitly deleted move constructor -- explicitly delete and make note
- Clang static analysis notes that control flow can be simplified by removing last `else if` (if I understood the control flow correctly)


mckib2:
There's more pressing static analysis fixes than these right now than these small ones.  Will close

PR #428, opened 2020-12-09
Bug scipy

galabovaa:


jajhall (approved):
Works for me with issue425.mps. Also double-checked by adding this problem to TestSpecialLps in my branch: got the failure and then no failure after merging bug-scipy into my branch!

PR #429, opened 2020-12-09
cmake file fix

galabovaa:
disable optimisation when in DEBUG mode


jajhall (approved):
Who am I to disagree!

PR #430, opened 2020-12-09
Mip

lgottwald:
One unit test fails but I don't know how to fix it, maybe you can have a look @galabovaa


jajhall:
> One unit test fails but I don't know how to fix it, maybe you can have a look @galabovaa

It relates to a unit test and code written by @jajhall so I'll investigate

jajhall:
An assert is still failing for FAST_BUILD on Windows with small_mip.mps

jajhall (approved):
Despite the assert with FAST_BUILD on Windows, I'm happy with the merge. It needs to happen, and shouldn't be delayed just because we can't find a way to debug on Windows.

PR #433, opened 2020-12-11
MIP followup

lgottwald:


jajhall (approved):
I trust that the MIP changes are OK to approve, so I'll approve the merge since it gets the fixed dual simplex bug (when breakIndex<0) and the fix to -nan objective reporting into master

lgottwald:
Yes, I believe the MIP changes are fine to merge. When the current set of features I am working on for MIP is finished I will start doing larger scale debug runs. When that point is reached we should start being more careful with MIP merges, but at the moment it is fine to assume that the changes are not influencing the early state of the solver negatively.

PR #435, opened 2020-12-27
Mip changes based on ekk_simplify

lgottwald:
These are changes for the MIP solver that are based on the ekk_simplify branch. The changes are explained in the long commit message. This includes the feasibility pump heuristic which uses primal simplex calls as it changes the objective function between LP solves and therefore gives some additional test coverage for the ekk primal simplex.


jajhall (approved):
Merge leaves simplex behaviour unchanged, so happy to accept it

PR #437, opened 2020-12-29
Jcc

galabovaa:
Changes for building updated Julia interface: no ipx or mip code updated yet
 > ipx update on a separate branch
 > mip update after mip solver is included in library and installs correctly


jajhall (approved):
Can't comment on the technical details, but it sounds like progress!

PR #441, opened 2021-01-17
update .lp filereader to latest version (include fixes for recognizin…

feldmeier:
…g illegal file format such as long lines


jajhall (approved):
Thanks Michael. These improve the LP file reader, so should be in master

PR #451, opened 2021-02-08
Remove Julia interface and point to HiGHS.jl

odow:


jajhall (commented):
This makes sense to me, since it means that we have only one Julia interface to maintain. I'll leave the final approval to @galabovaa

galabovaa (approved):
LGTM

PR #452, opened 2021-02-08
Update to2021

jajhall:
Updates the -2020 to -2021 in the file headers and applies formatting to methods added to the C API by @odow


odow:
Oops. For future reference, how do I run the formatter?

jajhall:
clang-format -style=file -i File.h

jajhall:
> clang-format -style=file -i File.h but @galabovaa is the expert

In the directory from which this is run there is also a file .clang-format containing the following, that may well be used to drive clang-format 

# We'll use defaults from the LLVM style, but with 4 columns indentation.
BasedOnStyle: Google
---
Language: Cpp
# Force pointers to the type for C++.
DerivePointerAlignment: false
PointerAlignment: Left
IncludeBlocks: Regroup

Issue #449, opened 2021-02-08
C API improvements


odow:
Some suggestions that arose from implementing the Julia wrapper: https://github.com/jump-dev/HiGHS.jl/pull/22

- [ ] Expose `getDualRay` https://github.com/ERGO-Code/HiGHS/blob/47d1a2eef636181c638783416d8212ed0920bc57/src/lp_data/Highs.cpp#L907
- [ ] Expose `getPrimalRay` https://github.com/ERGO-Code/HiGHS/blob/47d1a2eef636181c638783416d8212ed0920bc57/src/lp_data/Highs.cpp#L913-L914
- [ ] Expose `changeCoef` https://github.com/ERGO-Code/HiGHS/blob/47d1a2eef636181c638783416d8212ed0920bc57/src/lp_data/Highs.cpp#L1420
- [ ] Implement(?) and expose `getPrimalDualStatus`. I couldn't find this, only: https://github.com/ERGO-Code/HiGHS/blob/47d1a2eef636181c638783416d8212ed0920bc57/src/lp_data/Highs.cpp#L1758-L1760
- [ ] Implement `changeColsCostByRange` similar to https://github.com/ERGO-Code/HiGHS/blob/47d1a2eef636181c638783416d8212ed0920bc57/src/interfaces/highs_c_api.cpp#L354-L358
- [ ] Provide way to pass an option string and get the associated option type (e.g., -1 for not supported, 0 for Bool, 1 for Int, 2 for double, 3 for string).

jajhall:
changeColsCostByRange now implemented in 

HiGHS/src/interfaces/highs_c_api.cpp

Still only in branch ekk-master, but should be merged into master soon!

Issue #389, opened 2020-07-26
Build and verify examples in github actions


matbesancon:
It might be a bit too much work for some example scripts, but compiling and verifying results for the C example might ensure it is always in sync with HiGHS itself

PR #453, opened 2021-02-08
Add functionality to programmatically return the type of an option.

odow:
Suggested by point 6 in #449. This is useful for wrappers who do not want to store the string-to-type mapping (like [HiGHS.jl does](https://github.com/jump-dev/HiGHS.jl/blob/edc8bc310d43411348130c49cceb702979e9fd57/src/MOI_wrapper.jl#L325-L340)).

* I don't really know my way around the code, so hopefully I haven't missed anything obvious. 

* I've tested the C by calling the new function from `call_highs_from_c.c`. I had to make some other changes to get it to compile; I can split the second commit into a separate PR if you'd prefer. Seems like https://github.com/ERGO-Code/HiGHS/issues/389 is worthwhile.

* I ran the formatter on the source, but not on `call_highs_from_c`, because the diff was very large.

### DIFF
--- a/check/TestOptions.cpp
+++ b/check/TestOptions.cpp
@@ -291,11 +291,18 @@ TEST_CASE("highs-options", "[highs_options]") {
   return_status = highs.writeHighsOptions("Highs.set");
   REQUIRE(return_status == HighsStatus::OK);
 
+  HighsOptionType highs_option_type;
+
   bool get_mps_parser_type_free;
   return_status = highs.getHighsOptionValue("mps_parser_type_free",
                                             get_mps_parser_type_free);
   REQUIRE(return_status == HighsStatus::OK);
   REQUIRE(get_mps_parser_type_free == false);
+
+  return_status =
+      highs.getHighsOptionType("mps_parser_type_free", highs_option_type);
+  REQUIRE(return_status == HighsStatus::OK);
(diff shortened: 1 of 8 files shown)


galabovaa (approved):
LGTM!

jajhall (commented):
Looks OK in general to me, but why change declarations such as
double colcost[numcol] = {2.0, 3.0};
to
double colcost[2] = {2.0, 3.0};

odow:
> why change declarations such as

It's been a while since I wrote C, but I thought you could only initialize vectors inline as `double colcost[] = {...}` or `double colcost[2] = {...}`.

feldmeier:
> 
> 
> > why change declarations such as
> 
> It's been a while since I wrote C, but I thought you could only initialize vectors inline as `double colcost[] = {...}` or `double colcost[2] = {...}`.

No, that's one of the differences between C and C++. C99 allows [variable length arrays](https://en.wikipedia.org/wiki/Variable-length_array#C99) (with lifetime limited to its scope in contrast to malloc'ed arrays),  Not that it matters much in this case

odow:
I think for variable length arrays, the length needs to be inferred as constant. I've pushed an updated version.

jajhall:
Thanks for the insight into const dimensions. Looks ready to merge now 

We don't apply formatting to the code in src/interfaces - because it's not all c++ @galabovaa ?

Issue #402, opened 2020-09-03
MPS read problem


jajhall:
HiGHS can't read 

[it_22_10_dec.mps.gz](https://github.com/ERGO-Code/HiGHS/files/5167701/it_22_10_dec.mps.gz)

because the COLUMNS section looks like this

COLUMNS
    x(1) OBJROW -32.32035516598602 c_8 50911075
    x(1) c_9 -7343692 c_38 57759781
    x(1) c_39 -10114706 c_68 17442552
    
The column names are so short that it reads the first few characters of the row name as part of the column name and thinks that there are spaces in the column names. If I remember correctly, the FF MPS reader assumes that names are at least 8 characters. It's reasonable for a user to think that in free format the COLUMNS section could look like this. Only the first four characters in a line are sacrosanct. As a pointer, Clp manages to read this file as well as handling spaces in names.

With all the names at least 8 characters, the MPS file - dumped out of Clp - is

[it_22_10_dec_fixed.mps.gz](https://github.com/ERGO-Code/HiGHS/files/5167700/it_22_10_dec_fixed.mps.gz)

It was HDual minimizing the objective of the presolved LP that highlighted the failure of presolve to "pass on" the maximization sense to the presolved LP.

The original MPS file that I had is 

[it_22_10_hex.mps.gz](https://github.com/ERGO-Code/HiGHS/files/5167699/it_22_10_hex.mps.gz)

Unlike Clp, HiGHS, Gurobi and Soplex can't read it because the values are in Hex. Although the exactness of Hex is cool, fixing HiGHS so that it an read Hex is not a priority, although it might be interesting to see how Clp achieves this, given that atof doesn't work.

jajhall:
Fixed

PR #462, opened 2021-02-16
Mps bug

galabovaa:
Addressing https://github.com/ERGO-Code/HiGHS/issues/402


jajhall (commented):
INDICATORS etc are not bound types, but different sections. If the BOUNDS section is deleted from this file
[allsections.mps.txt](https://github.com/ERGO-Code/HiGHS/files/5995042/allsections.mps.txt)
then the MPS read crashes.
Note that this file contains all the MPS sections that are listed in
https://www.gurobi.com/documentation/9.1/refman/mps_format.html
including LAZYCONS that appears between ROWS and COLUMNS
If the MPS reader hits any of these unsupported sections then it should fail. Surely it shouldn't call exit(1) in case Highs::readModel is called from a user's executable rather than the HiGHS exe. It should pass back a failure that is (eventually) interpreted as a model status of MODEL_ERROR, with readModel returning HighsStatus::Error

galabovaa:
Code out of date, closing for the moment.

PR #466, opened 2021-02-17
Update num_nz in getCols

### LINKED ISSUE #465: ENH: update num_nnz in getCols regardless
`getCols` only updates `num_nnz` if `col_matrix_index` or `col_matrix_value` is not `NULL`. 
https://github.com/ERGO-Code/HiGHS/blob/2b89bd90391bb60edf252e0cbc7ea40f59798cf9/src/simplex/HighsSimplexInterface.cpp#L562-L569

It should update `num_nnz` regardless, so that you can call it once to get the number of non-zeros, allocate the correct arrays, and then call it again (ala Gurobi and CPLEX).

This is useful for the Julia interface: https://github.com/jump-dev/HiGHS.jl/pull/38#discussion_r577231422

odow:
Closes #465

### DIFF
--- a/check/TestCAPI.c
+++ b/check/TestCAPI.c
@@ -84,9 +84,37 @@ void options() {
   Highs_destroy(highs);
 }
 
+void test_getColsByRange() {
+    void* highs = Highs_create();
+    Highs_addCol(highs, -1.0, 0.0, 1.0, 0, NULL, NULL);
+    Highs_addCol(highs, -1.0, 0.0, 1.0, 0, NULL, NULL);
+    int aindex[2] = {0, 1};
+    double avalue[2] = {1.0, -1.0};
+    Highs_addRow(highs, 0.0, 0.0, 2, aindex, avalue);
+    int num_cols;
+    int num_nz;
+    int matrix_start[2] = {-1, -1};
+    Highs_getColsByRange(highs, 0, 1, &num_cols, NULL, NULL, NULL, &num_nz,
+                         matrix_start, NULL, NULL);
+    assert( num_cols == 2 );
+    assert( num_nz == 2 );
+    assert( matrix_start[0] == 0 );
+    assert( matrix_start[1] == 1 );
+    int matrix_indices[2] = {-1, -1};
+    double matrix_values[2] = {0.0, 0.0};
+    Highs_getColsByRange(highs, 0, 1, &num_cols, NULL, NULL, NULL, &num_nz,
+                         matrix_start, matrix_indices, matrix_values);
+    assert( matrix_indices[0] == 0 );
+    assert( matrix_indices[1] == 0 );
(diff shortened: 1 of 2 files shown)


jajhall (commented):
I can see that the test in TestCAPI.c is useful, but I don't see what is achieved in the change to HighsSimplexInterface.cpp. Note that refactoring means that HighsSimplexInterface.cpp will soon no longer exist in master, but I can propagate the change to the new file that carries out the same actions - if I understand what they achieve!

odow:
I wanted something like the Gurobi `GRBgetvars`: https://www.gurobi.com/documentation/9.0/refman/c_getvars.html

> Retrieve the non-zeros for a set of variables from the constraint matrix. Typical usage is to call this routine twice. In the first call, you specify the requested set of variables, with NULL values for vbeg, vind, and vval. The routine returns the number of non-zero values for the specified variables in numnzP. That allows you to make certain that vind and vval are of sufficient size to hold the result of the second call.

The key bit is
```
     int num_cols;
     int num_nz;
     int matrix_start[2] = {-1, -1};
     Highs_getColsByRange(highs, 0, 1, &num_cols, NULL, NULL, NULL, &num_nz,
                          matrix_start, NULL, NULL);
     assert( num_nz == 2 );
     # Now we know what num_nz is, so we can allocate arrays for `matrix_indices` and `matrix_value`.
```

At present, `num_nz` is only updated if `matrix_indices` or `matrix_value` is non-null. That means you have to be defensive and allocate a vector that is `getNumRows` long.

jajhall:
> I wanted something like the Gurobi `GRBgetvars`: https://www.gurobi.com/documentation/9.0/refman/c_getvars.html
> 
> > Retrieve the non-zeros for a set of variables from the constraint matrix. Typical usage is to call this routine twice. In the first call, you specify the requested set of variables, with NULL values for vbeg, vind, and vval. The routine returns the number of non-zero values for the specified variables in numnzP. That allows you to make certain that vind and vval are of sufficient size to hold the result of the second call.
> 
> The key bit is
> 
> ```
>      int num_cols;
>      int num_nz;
>      int matrix_start[2] = {-1, -1};
>      Highs_getColsByRange(highs, 0, 1, &num_cols, NULL, NULL, NULL, &num_nz,
>                           matrix_start, NULL, NULL);
>      assert( num_nz == 2 );
>      # Now we know what num_nz is, so we can allocate arrays for `matrix_indices` and `matrix_value`.
> ```
> 
> At present, `num_nz` is only updated if `matrix_indices` or `matrix_value` is non-null. That means you have to be defensive and allocate a vector that is `getNumRows` long.

Great: now I understand completely!

jajhall (approved):
Makes sense now!

jajhall:
> Makes sense now!

I guess that there's an analogous change to make to getRows in HiGHS. I'm on it.

PR #485, opened 2021-03-01
Add a link to the highs rust binding

lovasoa:
This updates the readme to add a link to the HiGHS rust binding.


jajhall (approved):
Happy to have Rust listed as another language from which HiGHS can be called!

PR #486, opened 2021-03-01
Update highs_lp_solver.f90

ogmundur:


src/interfaces/highs_lp_solver.f90
```diff
@@ -77,7 +77,7 @@ function Highs_passLp ( h, n, m, nz, cc, cl, cu, rl, ru, as, ai, av) result ( s
       integer ( c_int ) :: s
     end function Highs_passLp
 
-    function Highs_setOptionValue ( h, o, v ) result( s ) bind ( c, name='Highs_setOptionValue' )
+    function Highs_setOptionValue ( h, o, v ) result( s ) bind ( c, name='Highs_setHighsOptionValue' )
```
odow, line 80: Perhaps the function name should also be changed to match? That will make it easier to maintain/document/etc. (Because it was broken, no one can have used this, so it should be okay to change.)

ogmundur: Agreed, that makes sense

jajhall (commented):
The F90 interface hasn't been tested beyond Highs_call - because we didn't have people asking us about it. I'm now extending the example call_highs_from_fortran.f90 to test other aspects of the interface. I'm happy to do this, as I was only a FORTRAN programmer for over 30 years!

jajhall:
I'm working my way through the Fortran API, adding things that are missing, and fixing things that are wrong, so please wait until you see what I come up with - by the end of 2 March - before creating your own fixes.

jajhall (commented):
Yes, it's wrong, but having "Highs" in the (multiply-overloaded) methods setHighsOptionValue of the Highs C++ class is unnecessary. They will become setOptionValue in the first formal release. The naming consistency of the C API and FORTRAN API will mean that names like Highs_setHighsIntOptionValue will become Highs_setIntOptionValue

jajhall:
I've got a first draft of an extended F90 example and updated F90 API to HiGHS. There are some features that I don't know how to add. For example, in line 201 of examples/call_highs_from_fortran.f90, I can get the value of the option simplex_scale_strategy, using Highs_getIntOptionValue, but can't set it using Highs_setIntOptionValue. My skills in interfacing Fortran and C are very limited. There are other (similar?) issues with Highs_getDoubleInfoValue and Highs_getIntInfoValue. If it's obvious to @ogmundur how to do this, it would be very helpful. Otherwise I can turn to @feldmeier as the local interface guru.

jajhall:
Oh, the new code is in https://github.com/ERGO-Code/HiGHS/tree/f90-interface !

ogmundur:
Great, thanks. I'll take a look at it tomorrow!

jajhall:
> Great, thanks. I'll take a look at it tomorrow!

Thanks. Note the following in relation to functions and their names

- Highs_setOptionValue - keep function and name
- Highs_getIterationCount - deprecated since it's ambiguous now that HiGHS has an interior point solver. The C API has the method Highs_getIterationCounts that return void, with a second parameter being pointer to the set of all HiGHS iteration counts. The Fortran API should have a corresponding function
- Highs_set(get)Bool/Int/Double/StringValue - should all exist, and call the C method of the same name (ie without Highs in the name, as these methods are deprecated

jajhall:
I've modified the C API so that it has the methods without "Highs" in the name (after the initial Highs_) as well as the deprecated methods with Highs in the name - so you can identify Fortran functions (without Highs in the name) with the corresponding C methods

ogmundur:
Could it be, that all that is missing is the "VALUE" parameter for the input value "v" in Highs_setIntOptionValue? In my local copy this fixes the issue in line 201 of examples/call_highs_from_fortran.f90 (see my latest pull request).

As a stickler, I also added intent(in) to the input value but this isn't necessary.

jajhall:
Thanks for adding the "VALUE" parameter, I needed to learn this for all the other things I've done today. My biggest issue was the need to ensure that the strings passed from call_highs_from_fortran.f90 were null-terminated. 
The only thing that I know doesn't work is get/setStringOptionValue, as illustrated in the final part of call_highs_from_fortran.f90. This file illustrates a lot of use of HiGHS!

New version in https://github.com/ERGO-Code/HiGHS/tree/f90-interface

PR #488, opened 2021-03-02
Add Highs_runQuiet to Fortran interface

ogmundur:
I missed this option from the Fortran interface so I added it in my local copy. In my application it appears to work correctly in my limited number of tests.

### DIFF
--- a/src/interfaces/highs_lp_solver.f90
+++ b/src/interfaces/highs_lp_solver.f90
@@ -326,7 +326,12 @@ module highs_lp_solver
       type(c_ptr), VALUE :: h
       integer(c_int) :: nnz
     end function Highs_getNumNz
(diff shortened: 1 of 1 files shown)


jajhall (commented):
RunQuiet is deprecated, and will disappear in the first formal release of HiGHS, which will have a more transparent logging system, with RunQuiet (as well as setHighsLogfile and setHighsOutput) deleted. 
This system will soon be available in master, with RunQuiet, setHighsLogfile and setHighsOutput preserved for the time being. 
So, just keep runQuiet in your Fortran API for now.

jajhall:
Highs_runQuiet is now in the F90 API - but will disappear when the new logging system migrates to master

PR #490, opened 2021-03-03
Update highs_lp_solver.f90

ogmundur:
Fix to correctly pass integer option from Fortran to HiGHS


jajhall (approved):
Thanks: this probably gives me the clue I need to complete the rest

PR #496, opened 2021-03-17
Add writeSolutionPretty to the C API

lovasoa:
There was no way to get the prettily-formatted solution from the C API

This keeps backwards compatibility by creating a new function.
﻿
There is no other way to get a solution wil the column names besides writeSolutionPretty


jajhall (approved):
Indeed, something useful in the C API

PR #497, opened 2021-03-17
Mention JS in the list of available interfaces

lovasoa:
I compiled HiGHS to WASM in order to be able to use it from a web browser.


jajhall (approved):
Sounds good!

PR #500, opened 2021-04-08
Fix getModelStatus in Fortran interface

ogmundur:
The VALUE parameter must be added to the scaled_model argument so it gest passed correctly to Highs_getModelStatus (otherwise a 0 gets passed by reference which will be a nonzero value which will be interpreted as true)


jajhall:
Correction has been incorporated independently

Issue #493, opened 2021-03-05
bug: segmentation fault


kyler1990:
when use the ipm, some exception log happended as below:

exception log:
```
LP       : scpm1_lp
Rows     : 5000
Cols     : 500000
Nonzeros : 6250000
Presolve : Reductions: rows 5000(-0); columns 500000(-0); elements 6250000(-0)
INFO   : Solving the presolved LP
INFO   : IPX model has 5000 rows, 500000 columns and 6250000 nonzeros
Input
    Number of variables:                                500000
    Number of free variables:                           0
    Number of constraints:                              5000
    Number of equality constraints:                     0
    Number of matrix entries:                           6250000
    Matrix range:                                       [1e+00, 1e+00]
    RHS range:                                          [1e+00, 1e+00]
    Objective range:                                    [1e+00, 1e+02]
    Bounds range:                                       [1e+00, 1e+00]
Preprocessing
    Dualized model:                                     no
    Number of dense columns:                            0
    Range of scaling factors:                           [1.00e+00, 1.00e+00]
IPX version 1.0
Interior Point Solve
 Iter     P.res    D.res            P.obj           D.obj        mu     Time
   0   1.76e+00 1.22e+02   1.88315564e+04 -6.01335657e+07  2.75e+02       0s
   1   4.21e-01 7.87e+00  -4.29423998e+04 -2.84760116e+07  6.64e+01       1s
   2   7.30e-02 5.21e+00   1.50500005e+05 -2.08104746e+07  2.73e+01       1s
   3   7.30e-08 4.44e+00   1.17082577e+06 -1.83520070e+07  2.08e+01       2s
   4   3.20e-13 2.83e-01   4.33474449e+05 -1.00919977e+06  1.48e+00       2s
   5   2.31e-13 1.31e-01   8.49038248e+04 -5.18162815e+05  6.08e-01       2s
   6   7.64e-14 7.92e-03   3.31518488e+04 -6.32209947e+04  9.61e-02       3s
   7   7.46e-14 1.48e-03   6.43844068e+03 -1.20950025e+04  1.85e-02       3s
   8   1.90e-14 1.01e-03   4.62931832e+03 -9.32209761e+03  1.39e-02       4s
   9   1.78e-14 6.50e-04   3.26726556e+03 -6.59243879e+03  9.81e-03       5s
  10   6.66e-15 3.36e-04   2.44443824e+03 -3.88469546e+03  6.30e-03       6s
[...]

PR #508, opened 2021-05-18
Master-prep-ekk into master

jajhall:
It had to happen sometime!


jajhall:
@lgottwald and @jajhall have been effectively using this branch for weeks, so no review is required.

jajhall:
@odow @mckib2 Don't try to switch to this immediately! 
This is phase 1 of merging in some other significant changes in the next week or so. Once this is done, there is a document that will guide users through the enhancements and changes. As discussed elsewhere, there are a few changes that will break some current uses of HiGHS.

odow:
No problem. Please ping me when we should make a new release of the Julia binaries! Exciting stuff.

mckib2:
Thanks @jajhall !  We had seen some strange errors in our MacOS build similar to https://github.com/ERGO-Code/HiGHS/issues/493#issuecomment-795135134 that we were not able to debug so we are still currently on an older version of HiGHS.  When this release happens it might be a good chance to resume testing and catch back up so we can finally provide a ranging interface

PR #510, opened 2021-05-20
Revised and corrected the Python LP interface file, example file and added a MIP interface file

jajhall:
Also spotted that Highs_callMip in the C API shouldn't be returning duals or a basis!

### DIFF
--- a/examples/call_highs_from_python.py
+++ b/examples/call_highs_from_python.py
@@ -1,14 +1,42 @@
-file = "../src/interfaces/highs_lp_solver.py"
+file = "../HiGHS/src/interfaces/highs_lp_solver.py"
 exec(compile(open(file).read(), file, 'exec'))
 
-cc = (1.0, -2.0)
-cl = (0.0, 0.0)
-cu = (10.0, 10.0)
-ru = (2.0, 1.0)
-rl = (0.0, 0.0)
-astart = (0, 2)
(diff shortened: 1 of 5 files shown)


examples/call_highs_from_python.py
```diff
@@ -1,14 +1,42 @@
...
+astart = (0, 3)
+aindex = (0, 1, 2, 0, 1, 2)
+avalue = (2.0, 3.0, 2.0, 2.0, 4.0, 1.0)
+
+# Find the continuous solution of the LP
+return_code, col_value, col_dual, row_value, row_dual , col_basis, row_basis = Highs_lpCall(colcost, collower, colupper,
+                                                                                            rowlower, rowupper,
+                                                                                            astart, aindex, avalue)
+
+print return_code, col_value, col_dual, row_value, row_dual, col_basis, row_basis
```
mckib2, line 29: These `print` statements will not work with Python 3 (the current and supported version) and should have parentheses, i.e.
```python
print(return_code, col_value, col_dual, row_value, row_dual, col_basis, row_basis)
```

jajhall: Thanks. I'm not a Python programmer and wanted to get rid of the unnecessary parentheses that were printed. But I was just using "python" on Linux. I'll reinstate the parentheses and be careful to check with python3 in future!

PR #511, opened 2021-05-22
Create highs model

jajhall:
The Highs class now has its model data in a HighsModel instance, consisting of a HighsLp instance and a HighsHessian instance. If there is a Hessian matrix, then callSolveQp() is called by run(), paving the way for the QP solver to be integrated. When a HighsModel is passed via passModel, any Hessian is checked and normalised by computing (Q+Q^T)/2. Unit test for the HighsModel class is in check/TestHighsModel.cpp


jajhall:
Although this is quite a change at a high level, it clearly doesn't break anything that was working before, and is necessary for integration of the QP solver.

PR #518, opened 2021-06-10
Julia interface fixes

galabovaa:


jajhall (approved):
So, in the end, there were lots of fixes required in things that I'd written!

I'm puzzled by the switch from fabs to abs in src/simplex/HEkkDebug.cpp as there are lots of other uses of fabs elsewhere in HiGHS. I see that abs() is OK in this method since it's double = abs(double), but I remember the use of abs() causing problems elsewhere. No matter, I approve the change!

PR #528, opened 2021-06-29
Install

galabovaa:
> installing headers in /include/highs 

> Installing fast_build also 

> Installation works on all platforms

> Linking with C++ code tested on Linux and Mac


jajhall (approved):
I trust you on the build stuff!

PR #534, opened 2021-07-13
Triangular hessian

jajhall:
Defines the QP interface to HiGHS to be Hessian triangle (lower triangle column-wise); extends internal debugging to QPs; adds/fixes Python interface issues, adds C API functions requested by @odow; fixes some QP solver bugs, stops IPX being used in unit tests with 64-bit integers


src/interfaces/highs_c_api.cpp
```diff
@@ -84,18 +92,124 @@ HighsInt Highs_mipCall(const HighsInt numcol, const HighsInt numrow,
...
+      if (copy_row_value) rowvalue[i] = solution.row_value[i];
+      if (copy_row_dual) rowdual[i] = solution.row_dual[i];
+      if (copy_row_basis) rowbasisstatus[i] = (HighsInt)basis.row_status[i];
+    }
+  }
+  return (HighsInt)status;
+}
+
+HighsInt Highs_lpDimMpsRead(
+    //			    const char* filename,
```
odow, line 172: Is the filename meant to be hard-coded as `ml.mps` below?

src/interfaces/highs_c_api.cpp
```diff
@@ -84,18 +92,124 @@ HighsInt Highs_mipCall(const HighsInt numcol, const HighsInt numrow,
...
+
+HighsInt Highs_lpDataMpsRead(
+    //			     const char* filename,
+    const HighsInt numcol, const HighsInt numrow, HighsInt* sense,
+    double* offset, double* colcost, double* collower, double* colupper,
+    double* rowlower, double* rowupper, HighsInt* astart, HighsInt* aindex,
+    double* avalue) {
+  Highs highs;
+  highs.setOptionValue("output_flag", false);
+  const char* filename = "ml.mps";
```
odow, line 193: Ditto here.

src/interfaces/highs_c_api.cpp
```diff
@@ -84,18 +92,124 @@ HighsInt Highs_mipCall(const HighsInt numcol, const HighsInt numrow,
...
+  *numnz = lp.Astart_[lp.numCol_];
+  return (HighsInt)status;
+}
+
+HighsInt Highs_lpDataMpsRead(
+    //			     const char* filename,
+    const HighsInt numcol, const HighsInt numrow, HighsInt* sense,
+    double* offset, double* colcost, double* collower, double* colupper,
+    double* rowlower, double* rowupper, HighsInt* astart, HighsInt* aindex,
+    double* avalue) {
```
odow, line 190: Rather than having these two separate methods, `lpDataMpsRead` and `lpDimMpsRead`, the Gurobi approach is to only have the `Data` version, and allow `NULL` to be passed as the vectors. 

Then the approach is:
 - call it will `NULL` and obtain `numcol`, `numrow`, and `numnz`
 - allocate vectors based on the size
 - call it again with the allocated vectors

Thoughts on that approach?

jajhall: I'm sure that there's a better approach. I had to create something to allow an MSc student to read MPS files into Python! 
Passing the file name as a string from Python was beyond me, but it's clearly not so hard in C!

odow: > Passing the file name as a string from Python was beyond me

Yeah I'm no help on that front, unfortunately.

Although a quick google suggests: https://nesi.github.io/perf-training/python-scatter/ctypes
image

jajhall: Thanks. We lack local knowledge of Python, and it's an important interface, even if we're in SciPy!

mckib2: I don't know everything about ctypes, but happy to help out if I can.  I've seen more and more requests for modeling package wrappers for HiGHS, but it's just not something we're interested in for SciPy (we're not that kind of package).

For string encoding, you can also use a more convenient syntax:
```python
b'myCoolFile.mps' == 'myCoolFile.mps'.encode()
```

@jajhall You might also be interested in some comparisons we've been doing for LP solvers in Python-land for a SciPy conference.  I present tomorrow, I will share slides with you afterward.  We're comparing against Python interfaces for Google's GLOP, SCIP, CLP, and others

odow: > I present tomorrow, I will share slides with you afterward. We're comparing against Python interfaces for Google's GLOP, SCIP, CLP, and others

I'd also be very interested in this.

PR #537, opened 2021-07-19
Mip symmetry

lgottwald:
contains a lot improvements and bugfixes.
Most notably:
-Symmetry detection and handling of symmetries by orbital fixing
-Conflict analysis
-Dominated column presolver for MIP
-improved cMIR cut generation
-improved performance of domain propagation


lgottwald:
Julia interface also builds now

jajhall (commented):
All good for me, but Julia interface needs to build before merge can be approved.

PR #539, opened 2021-07-24
Fix multiply_modM61 and  Merge branch 'fix-performance-issue' 

WTFHCN:
I find a mistake in HighsHash.h

multiply_modM61 result is wrong 

The specific derivation process and the details of the modification , you can see the pdf .(PS: I dont know how put Mathematical formula on here )
[README.pdf](https://github.com/ERGO-Code/HiGHS/files/6872524/README.pdf)

beause Linux can use **__int128_t**,so I make a test.cpp for test **multiply_modM61** (this functional)
PS: I dont know how to write your test ,so used a simple method to test .
```c++
#include <iostream>
#include <random>
#include <cassert>
using u32 = unsigned int;
using u64 = unsigned int64_t;

void write128(__int128_t x)
{
    if (x > 9)
        write128(x / 10);
    putchar(x % 10 + '0');
}
void write64(u64 x)
{
    if (x > 9)
        write64(x / 10);
    putchar(x % 10 + '0');
}

static constexpr u64 M61() { return u64{0x1fffffffffffffff}; };
static constexpr u64 M29() { return u64{0x1fffffff}; };
/// compute a * b mod 2^61-1
static u64 multiply_modM61(u64 a, u64 b)
{
    u64 ahi = a >> 32;         // < 2^29
    u64 bhi = b >> 32;         // < 2^29
    u64 alo = a & 0xffffffffu; // < 2^32
    u64 blo = b & 0xffffffffu; // < 2^32

    // compute the different order terms with adicities 2^64, 2^32, 2^0
    u64 term_64 = ahi * bhi;             // < 2^58
    u64 term_32 = ahi * blo + bhi * alo; // < 2^62
    u64 term_0 = alo * blo;              // < 2^64

    // now extract the upper 61 and the lower 61 bits of the result a * b
    u64 ab61 = (term_64 << 3) + (term_32 >> 29) + (term_0 >> 61);
    u64 ab0 = ((term_0 & M61()) + ((term_32 & M29()) << 32));

    // ab0 may be greater than 2^61
    if (ab0 > M61())
        ab61++;
    ab0 &= M61();
    // ab61 may be greater than (2^61 - 1)
    if (ab61 >= M61())
        ab61 -= M61();
    // finally take the result modulo M61 which is computed by exploiting
    // that M61 is a mersenne prime, particularly, if a * b = q * 2^61 + r
    // then a * b = (q + r) (mod 2^61 - 1)
    u64 result = ab0 + ab61;
    if (result >= M61())
[...]

PR #540, opened 2021-07-26
fix performance issue with counting number of binary implications

lgottwald:
Small change that showed up as performance issue in some benchmark instance. The counting of binary implications in cliques of size two took very long when there where many cliques of size two. Now the implications are counted more efficiently and do not need top loop over cliques of size two. Also some calls to counting the implications are avoided by writing the comparison function that made the calls during sorting in a way that first evaluates the first sorting criterion and only counts the implications when the first criterion is equal.


lgottwald:
I added some smaller other bug fixes to this branch too, so that they all go into master quickly

WTFHCN:
According to [PR](https://github.com/ERGO-Code/HiGHS/pull/539)  there is a Overflow occurred problem

jajhall (approved):
Sounds fine to me. Ivet's on holiday, so I'll approve

PR #541, opened 2021-07-26
Wtfhcn/bug multiply mod m61

WTFHCN:
I think I am better in understanding and code logic


jajhall:
Thanks for this. Your suggested change has influenced a modification made by our MIP developer

PR #544, opened 2021-07-27
improve MIP reporting

lgottwald:
Improve reporting of the MIP solver. Uses a table without bars and lines take a bit less space. Information is more organized. Prints information about the model size after presolve and about variable types.

### DIFF
--- a/src/mip/HighsMipSolver.cpp
+++ b/src/mip/HighsMipSolver.cpp
@@ -264,6 +264,8 @@ restart:
                                << nTreeRestarts);
 
       if (doRestart) {
+        highsLogUser(options_mip_->log_options, HighsLogType::kInfo,
+                     "\nRestarting search from the root node\n\n");
         mipdata_->performRestart();
         goto restart;
       }
--- a/src/mip/HighsMipSolver.cpp
+++ b/src/mip/HighsMipSolver.cpp
@@ -265,7 +265,7 @@ restart:
 
       if (doRestart) {
         highsLogUser(options_mip_->log_options, HighsLogType::kInfo,
-                     "\nRestarting search from the root node\n\n");
+                     "\nRestarting search from the root node\n");
         mipdata_->performRestart();
         goto restart;
       }
--- a/src/mip/HighsMipSolver.cpp
+++ b/src/mip/HighsMipSolver.cpp
@@ -161,13 +161,7 @@ restart:
           options_mip_->mip_pool_soft_limit)
         mipdata_->conflictPool.performAging();
 
-      if (mipdata_->dispfreq != 0) {
-        if (mipdata_->num_leaves - mipdata_->last_displeave >=
-            std::min(mipdata_->dispfreq,
-                     3 + int64_t(0.1 * mipdata_->num_leaves)))
-          mipdata_->printDisplayLine();
-      }
-
+      mipdata_->printDisplayLine();
       // printf("continue plunging due to good esitmate\n");
     }
     search.openNodesToQueue(mipdata_->nodequeue);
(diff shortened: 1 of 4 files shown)


jajhall (changes requested):
Looks a lot cleaner and more comprehensible. Clearly you've narrowed some columns - resulting in compiler warnings like the one below? Also, it's still rather chatty. Solving bell5 produces rather a lot of output for just 0.6s of execution. Clearly there should be the first line of results after a header output, and the "culmination" line, but maybe no other lines unless a second has elapsed - until 100 seconds - and then every 10 seconds etc.
As before, and with simplex, keep the more verbose output for the benefit of developers. (ie when option.highs_log_dev is positive. And, as with Gurobi, no need for the output to be deterministic.
 
[bell5.txt](https://github.com/ERGO-Code/HiGHS/files/6884760/bell5.txt)


/home/jajhall/HiGHS/src/mip/HighsMipSolverData.cpp:795:44: note: directive argument in the range [-9223372036854, 9223372036854]
  795 |       std::snprintf(printString.data(), 8, "%ldm", val / 1000000);
      |                                            ^~~~~~
In file included from /usr/include/stdio.h:867,
                 from /usr/include/c++/9/cstdio:42,
                 from /usr/include/c++/9/ext/string_conversions.h:43,
                 from /usr/include/c++/9/bits/basic_string.h:6493,
                 from /usr/include/c++/9/string:55,
                 from /home/jajhall/HiGHS/src/./lp_data/HConst.h:20,
                 from /home/jajhall/HiGHS/src/./mip/HighsCliqueTable.h:20,
                 from /home/jajhall/HiGHS/src/./mip/HighsMipSolverData.h:19,
                 from /home/jajhall/HiGHS/src/mip/HighsMipSolverData.cpp:13:
/usr/include/x86_64-linux-gnu/bits/stdio2.h:67:35: note: ‘__builtin___snprintf_chk’ output between 3 and 16 bytes into a destination of size 8
   67 |   return __builtin___snprintf_chk (__s, __n, __USE_FORTIFY_LEVEL - 1,
      |          ~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   68 |        __bos (__s), __fmt, __va_arg_pack ());
      |        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
At global scope:

lgottwald:
The output for bell5 being verbose is just due to restarts happening. At the point of the restart HiGHS cannot know that the model is solved almost without tree search. So it starts over with reporting for the root node, which is more verbose than reporting in the tree. Also the ouput is deterministic right now, so for a tiny model it will print a bit more frequently in the beginning, but as the model solves the time between the prints will increase and the output is less chatty than during root node evaluation. Also if the root node evalutation uses more separation rounds the gaps between output lines will increase. But when a restart occurs then the output behavior within the root node is the same as for the first time, which I could maybe adapt a bit. But the output for the following tree search will use the already increased print frequency, e.g. if already a lot of nodes had been processed before a restart the tree search will keep increasing the print frequency based on the nodes of all runs.

I'd prefer the output deterministic for now as it is more convenient for development. Its often good to see from the output when solution path of two versions have diverged which I could not see if the output was non-deterministic.

jajhall (approved):
Of course, dev output must be deterministic (and not mention the time) so that diff can be used - I rely on it myself.
It's the user output that can be non-deterministic
I take your point on bell5. I know that there are far more reasons to report special events in MIP

lgottwald:
I just thought about the determinism part and I think I found a solution. I will just keep the points where a line could be printed deterministic but skip lines if the last print was less then 5 seconds ago. That way I can simply remove the check and replace it with something deterministic if I need it.

lgottwald:
Can you now check again if the chattiness is satisfactory?

lgottwald:
Okay I am now also happy with reporting frequency in case of restarts, it now prints a small note when a restart happens and only reports on the final model size after the restart. Printing is now non-deterministic, but the points at which highs would print are deterministic. When an improving solution was found a line is always printed.

PR #545, opened 2021-07-27
Fix small details

WTFHCN:
make Fast modexp can accept e=0

### DIFF
--- a/src/util/HighsHash.h
+++ b/src/util/HighsHash.h
@@ -111,9 +111,11 @@ struct HighsHashHelpers {
   }
 
   static u64 modexp_M61(u64 a, u64 e) {
-    u64 result = 1;
+    // the exponent need to be greater than zero
+    assert(e > 0);
+    u64 result = a;
 
-    while (e) {
+    while (e != 1) {
       // square
       result = multiply_modM61(result, result);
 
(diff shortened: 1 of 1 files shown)


src/util/HighsHash.h
```diff
@@ -111,9 +111,9 @@ struct HighsHashHelpers {
   }
 
   static u64 modexp_M61(u64 a, u64 e) {
-    u64 result = a;
+    u64 result = 1;
```
lgottwald: It is intended to save one multiplication mod M61 as this case is never encountered for its inteded use within the hash functions. But it should probably have an assert then, so if you add the assert instead of making it allow `e == 0` I'll merge the MR as it also fixes a comment.

lgottwald:
Thanks for your merge request. The reason I do not want to treat the case is the additional multiplication. Consider the hash function running for each nonzero of a problem matrix. There are problems with several millions of nonzeros which would mean for each nonzero two additional calls to `multiply_mod_M61()` when we detect parallel rows and columns, whereas the hash functions would never call the function with the exponent less than 1.

It would be good to document the behavior regarding an exponent of 0 though and add asserts. If you want to update this I can merge your MR.

WTFHCN:
Indeed, if I write this way, it will affect efficiency . So i follow your suggestion that document the behavior regarding an exponent of 0 though and add asserts. (want to update).

lgottwald (approved):

WTFHCN:
I see that there is a problem with the check code format, it should be caused by my vscode automatic formatting, can you help me simply change it back

lgottwald:
I fixed in directly in master. But its weird that the pull request said that all checks had passed.

PR #546, opened 2021-07-27
Mip scaling and missing presolve reduction

lgottwald:
I added a presolve reduction to the postsolve datastructure that allows to transform a column by a linear transformation `x = a x' + b` .

Its only used for MIP and covers the following cases:
-Transforms integral variables such that their lower bound is 0 and the upper bound is >= 1. This includes shifting of integer variables to become binary. Binary variables are exploited in more places of the MIP solver, e.g. probing, and can be part of cliques. Hence this reduction can be advantageous even if the problem is not reduced.
-Transform continuous variables by an appropriate scale for which they can be marked to implied integer variables. This can give a huge boost in performance if it occurs, as the new variables integrality status can be exploitet by the whole MIP machinery, e.g. for cutting planes, branching, etc.
-Reduce the range of integer variables by exploiting divisibility of integers in equations, e.g. transform an integer x with `x = a x' + b`  where `a > 1` and a,b are integers.

The latter two are described in the paper https://www.researchgate.net/publication/337127039_Presolve_Reductions_in_Mixed_Integer_Programming
under euclidean and modular inverse reduction.

Also I added scaling of rows that contain continuous and integer variables so that the largest absolute coefficients of continuous variables is 1.0 (or as close as possible with a power of two scale). Integral columns in rows are not considered for the largest absolute coefficient.
The scaling is only on the rows and only considers continuous columns for equilibration for several reasons:
1. Integers might have big-M coefficients and we do not want to equilibrate a row by such a huge coefficient.
2. During the tree search integers will eventually end up fixed the deeper we go into the tree. Hence the model rows remain on an appropriate scale for sub problems encountered during the search if we only consider the continuous columns.
3. Continuous columns could be scaled too, but this gave very mixed results on different instances and therefore I disabled the code block. I may add the scaling later for numerically challenging instances as an option that is disabled by default.
Edit: A simple equilibration scaling of continuous columns after the rows have been scaled seems to work well when it does something at all. Columns with small coefficients can hurt cutting plane performance because the cut generation removes coefficients that small relative to the largest coefficient in a cut.
4. Integral rows are not scaled as their integrality can be exploited for gomory cuts from basis indices of such rows when they have a fractional value. Also their logical slack variables get stronger coefficients in cut generation.


lgottwald:
I fine-tuned reporting of objective values in the MIP solver a bit. It was sometimes misleading. E.g. on snp-02-004-104 the objective was on the order of 1e9 so the `%.9g` format specifier printed the objective values rounded to an integral value printed in non-scientific notation. If the value printed is integral I would expect it to not be rounded to that value if printed in non-scientific notation. Therefore the reporting now takes into account the magnitude of the objective values and prints 10 digits (including fractional part) for values up to an integral part of 9999. Then it stays at this precision level of 6 digits after the decimal point until the magnitude of 1e7. From magnitude 1e8 to 1e11 it starts decreasing the precision after the fractional point so that it prints only 2 digits after the fractional point with magnitude 1e11. From magnitude 1e12 it switches to scientific notation with 9 significant digits.

In summary it now never prints more then 9 significant digits of the fractional part which is the default threshold for small_matrix_value, even for small objective values. As the magnitude increases it prints at least 2 decimal digits after the decimal point and switches to scientific notation when that is not possible anymore. Hence an integral value that is printed in decimal point notation is now really integral within a reasonable tolerance depending on the magnitude. The integral part is never rounded in decimal point notation.

jajhall (approved):
Sounds good to me.

PR #551, opened 2021-07-29
Mip full orbitopes

lgottwald:
This implements some improvements for symmetry handling. So far no special structure is exploited from the symmetries that are found. With this MR a detection of so called full orbitopes is implemented. Full orbitopes have the structure that a set of binary variables in the problem can be arranged in a matrix for which the columns can be freely permuted, i.e. the group acting on the columns of the matrix is the symmetric group. The full orbitope is the polytope associated with the set of matrices where the columns lexicographically non-increasing. A special case of full orbitopes is also detected, so called packing orbitopes in which each row of the orbitope matrix can only have a single entry set to 1.

For full orbitopes without the partitioning structure there is an algorithm for fixing all variables of the full orbitope that are fixed in the face of the orbitope that intersects the current set of fixed variables, e.g. from branching. The algorithm is described here and is implemented in this MR
https://arxiv.org/pdf/1710.10189.pdf

Packing orbitopes are more restrictive and allow for potentially more fixings to be made. There is an algorithm for determining the set of fixed variables in the current face as for full orbitopes. It is described here:
https://opus4.kobv.de/opus4-zib/files/942/ZR_06_48.pdf

For both fixing algorithms a dynamic version is implemented. The dynamic version reorders the rows of the orbitope by following the branching decisions. In the first paper above the dynamic version is shown to perform much better and the infrastructure to always keep the knowledge about what bound changes are branching decisions was already added with the general symmetry handling. For packing orbitopes the dynamic algorithm is an extension of mine and is not described in the second paper, but the arguments of why it is correct are similar. It is always correct to consider a subset of the rows of the orbitope matrix. The order that is used is always the same for related subproblems as new rows are only considered when they are first branched on and then have their order fixed in all children of that subproblem. Another advantage of the dynamic algorithm is that we can dynamically decide whether the packing structure is present on all rows that are considered so far. So if only some rows have the set packing structure we can exploit the stronger fixing algorithm until the first branching on a row that doesn't have the structure occurs.

Finally for packing orbitopes we add an additional restriction to the choice of branching columns. Since for a packing orbitope there can only be a single 1 in each row we restrict the branching variable to belong have the smallest index within its row of the orbitope matrix. The reason is that together with the dynamic fixing algorithm "stupid" decisions would be made. E.g. the upper triangle of a packing orbitope would be immediately fixed to zero in the static fixing algorithm as it can never be 1 in a lexicographically non-increasing column order. Since the dynamic version only considers rows that have been branched on it could try to branch on a variable that would immediately lead to an empty set when intersected with the packing orbitope. Always branching on the first non-fixed variable in the row that is chosen for branching avoids this.

Another extension that is added with this MR is that history observations about variables that are used for branching, e.g. pseudocost, are always added for all variables that are in the same row of an orbitope. If a variable is not in an orbitope then the observation is added as before.

The structures added here are quite important for the solvability if they are present in a problem instance. One case that is missing is the detection and handling of partitioning orbitopes which must have exactly one variable set to 1 in each row (i.e. =1 where packing is <= 1). This I will add at a later time in a separate MR.


jajhall (approved):
Sounds impressive! Well done!

PR #552, opened 2021-07-30
General tuning

lgottwald:
Some general tuning and bug fixes, for now of symmetry detection, as it takes a while on some instances and is a little slower than the bliss package used by SCIP.

In the process I encountered a bug too which led to many symmetries not being detected.

One thing that showed up in the profiler is the hash calculations. For speeding that up I simply increased the chunk size of the part that uses a fast hashing routine from 64 bytes to 256 bytes which requires adding more random constants. As the random constants are now there I also used them in the other places where more of them could be useful.

The chunksize of the hashing showed a significant improvement of symmetry detection runtimes.

Now one other thing that showed up in the profiler is sorting. For this I wanted to ask if I can add pattern defeating quicksort to the repository under external. It is here https://github.com/orlp/pdqsort and only consists of a single header which introduces a drop-in replacement for std::sort. The package does not have any dependencies and a permissive license. Using the pdqsort function also showed significant improvement in the runtime of the symmetry routine, but I have not added the header yet. If I add the header I would replace all calls to std::sort in HiGHS by pdqsort.


lgottwald:
For performance difference of pdqsort here is an example on fiball:
with std::sort 11.8s with pdqsort 11s which is a ~7% improvement on total runtime just from faster sorting where sorting is just a smaller (but significant) part of the total time spent.

The reason pdqsort is so fast is that it detects many patterns in the input with some clever book keeping that introduces alsmost no overhead. Pdqsort can then achieve linear time for many patterns in the input data, e.g. when there are many blocks of equal elements which is often the case in symmetry detection.

lgottwald:
The CI fails are probably differing iteration counts due to small changes in the random generator.

lgottwald:
Another note on the pdqsort: We do not need to install this header if we only use it in .cpp files as it has no attached library, nothing needs to be linked. All the defined sort functions are templates as with std::sort

lgottwald:
I added the pdqsort header and license file to external and replaced all calls to std::sort by pdqsort within one commit so that it can easily be reverted if you have concerns. But I figure it is a free small performance increase as sorting is not that rare.

lgottwald:
I tuned the symmetry detection routine to better exploit sparsity in the partition refinement process and in now runs in ~5 seconds on fiball. Now I am in the process of debugging a problem that popped up on graph20-20-1rand of miplib2017 which has an orbitope with 37 columns and 52 rows where something goes wrong in the fixing algorithm.

lgottwald:
Found the bug(s). The fixing algorithm for full orbitopes was far from correct, probably coded it too late at night and the bug did not show up as the bug in the symmetry detection prevented many cases of full orbitopes being detected.

lgottwald:
fix several warnings and bugs now in this MR too

jajhall (approved):
Fair enough, given the permissive license.

PR #553, opened 2021-08-02
fix bug in presolve that can lead to wrong fixings of variables in equations

lgottwald:
This fixes a bug in presolve (I think only for MIP)


jajhall (approved):
Fine!

PR #554, opened 2021-08-03
Small change

lgottwald:
This introduces some smaller changes and some cleanup of unused functions.

I realized that the hash table implementation does not support 64 bit and I implemented it before HiGHS added 64bit compile mode. For 64bit support the hash functions needed slight adjustment to output 64bit instead of 32bit which only comes with a negligible increase in operations. The hash table now always uses 64bit as its number of entries cannot be so easily bounded. E.g. in the clique table cliques of size two are stored in a hash table and there can be quatratically many. So even when HiGHS is compiled in 32 bit mode and columns, rows, and number of nonzeros all fit with 2^31-1 the hash table could still overflow for moderate numbers of columns if the conflict graph is very dense and many cliques of size two are added. E.g. with 2^17=131072 binary variables the complete conflict graph would already have more than 2^31 - 1 edges. Still not likely that so many cliques are added but its not an astronomically large number of columns and with always using 64bits for the hash table we never need to worry about overflows.

This does not impact LP as much as with the HighsRandom changes. Random numbers are unaffected. But a small impact is there occasionally as presolve uses the hashing for random tie breaking which is why the unit test LP iterations needed adjustment.


lgottwald:
Also added some changes to the restart policy after some observations on different instances. The restarts occur when the tree is estimated to be large which is estimated based on the tree weight, where a node at depth d gets weight 2^-d (root has depth 0) and the weight of pruned leave nodes is summed up to get en estimate of the completion of the search which is between 0 and 1 and will be exactly 1 when all leave nodes have been explored.

To estimate the tree size we require a series of consecutive estimations to be large. Initially, before any restart was triggered, we require 10 consecutive estimations and for every 1000 leave nodes explored 1 more. This base number is multiplied by 2^k where k is the number of restarts that have already been made due estimating a large tree. Hence, we are less and less likely to perform a restart the more leaves we already have explored in the current tree and the more restarts have been made. This protects we will not end up in a restart loop as we will require exponentially longer sequences where the tree is estimated to be huge.

For computing the tree size estimate we take a start point where we remember the current number of nodes `N0` and the current pruned tree weight `P0` and then at a later point we estimate the tree size as
`N0 + (N - N0) * (1-P0)/(P-P0)` for the current number of nodes `N` and the current pruned tree weight `P`. So we assume that the final tree size will have an offset of N0 and then increase proportionally to the pruned tree weight and nodes processed since N0. This estimate must be large over several checks.

Now before this MR the checks where sometimes being too close to each other after very few nodes which was problematic in the following case: If the sub problems from the queue came from deeper parts of the tree and could quickly be pruned then the estimate has changed very little as deep nodes do not influence the tree weight much and the estimation points where close together because after every node from the queue the tree size was estimated. This led to bad decisions to do a restart when the tree was not going to be huge. It is fixed by requiring at least 100 nodes between measurements so that initially the first restart can occur after 1000 nodes. This was already the intended interval as the search would always try to process 100 nodes in a depth first manner before taking the next node from the queue, but for the situation above this was not the case as the search in the subproblems coming from the queue were quickly exhausted.

Another problem was that the limit for a tree being huge was too strict, so that huge trees were not recognized as such. Before this MR a factor of 1000 over the current tree size was required whereas now we require a factor of 50. 50 is still not too small as it would throw away the tree after the estimation indicated several times that less than 2% of the tree have been explored. The factor being large before was an artifact due to the other problem above to prevent too many restarts.

This changes fixed issues on 2 instances where on one instance too few restarts were made and on the other one an unnecessary restart was made when the tree was good. Effects on other instances did not seem negative.

lgottwald:
Also there was a bug in the tree weight calculation and a missing header in cmake which I now fixed

jajhall (approved):
OK by me, but we need to check with Ivet that the Julia build is still OK

lgottwald:
@galabovaa can you check whether the Julia build works on this branch?

galabovaa:
Julia builds OK! Sounds good!

PR #556, opened 2021-08-05
add command line options for random seed and solution file

lgottwald:
I added command line options for specifying a solution file and specifying a random seed. The latter is now used for additional testing in the ctest by running all instances with seeds 1 to 5 in addition to the default seed of 0. This increases the probability of catching bugs a little and the instances solve quickly enough so that it is not an issue to run them a bit more often.

### DIFF
--- a/check/CMakeLists.txt
+++ b/check/CMakeLists.txt
@@ -198,6 +198,11 @@ set(mipInstances
 set(settings
     "--presolve=off"
     "--presolve=on"
+    "--random_seed=1"
+    "--random_seed=2"
+    "--random_seed=3"
+    "--random_seed=4"
+    "--random_seed=5"
 #    "--parallel=on"
     )
 
--- a/src/lp_data/HighsLpUtils.cpp
+++ b/src/lp_data/HighsLpUtils.cpp
@@ -1808,16 +1808,16 @@ void writeSolutionToFile(FILE* file, const HighsLp& lp, const HighsBasis& basis,
     fprintf(file, " Basis\n");
     fprintf(file, "Columns\n");
     for (HighsInt iCol = 0; iCol < lp.num_col_; iCol++) {
-      if (have_value) fprintf(file, "%g", use_col_value[iCol]);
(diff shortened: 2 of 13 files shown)


galabovaa:
We are trying to keep the command line options to a minimum and anything not particularly required by a basic user should be left to HighsOptions. 

Is there a way for you to add / read these parameters from the option_file instead of adding explicit command line options?

lgottwald:
I think these are very basic options. random seed is annoying to specify with potions files for testing and development and writing out a solution is kind of a basic functionality for a solver.
The random seed is quite important for testing and development and benchmarking in MIP and I would often need to create an options file just for that when I want to scan for a bug with 5 different seeds. I need to either create 5 files or create a script that seds the option file but then I cannot easily see where which seed was used.

I only added those two options because I figured it should be kept minimal but I think just those two more are very nice to have.

lgottwald:
And I think it is not nice if ctest needs to write out an options file. The functionality that ctest uses should be by definition basic as the ctests run on easy instances to test basic functionality.

lgottwald:
So essentially I'd like to cover all normal testing benchmarking and running functionality for default HiGHS without needing to create an options_file. If I do not want to alter any option and simply run default HiGHS with some random seed and obtain a solution I should not need an options file I think. Option files can get very messy if I am forced to need a lot of them for simple things like the seed. So it greatly simplifies my testing/development workflow to just have this one single command line option for the seed.

For me personally I would not need the solution file option but I think I would expect it from a user perspective, which is why I added it. If I simply download HiGHS for the first time to test it on a problem that I have and want to look at the solution I would need to check the documentation for the first thing I want to do. If the solution file option is there I would succeed with just running --help. Not writing out the solution is actually our developer perspective, because we do not care about solving a particular instance but just the feasibility/optimality of the solution. Users of HiGHS would usually want a solution.

galabovaa:
OK that sounds very reasonable. I will check it out too.

galabovaa (approved):

lgottwald:
One last question would be if we want to rather write out the pretty solution when the command line option is specified?

Issue #561, opened 2021-08-11
Presolve fails on feasibility problem (QPLIB_8906)


feldmeier:
The following assertion fails when solving the feasibility problem of QPLIB_8906 (all objective coefficients zero, no Hessian involved):

highs: /home/s1613957/PhD/HiGHS_master/src/presolve/HighsPostsolveStack.cpp:572: void presolve::HighsPostsolveStack::DuplicateColumn::undo(const HighsOptions&, HighsSolution&, HighsBasis&): Assertion `basis.col_status[col] == HighsBasisStatus::kBasic' failed.
Aborted (core dumped)

[8906_feasibility.lp.txt](https://github.com/ERGO-Code/HiGHS/files/6967368/8906_feasibility.lp.txt)

PR #563, opened 2021-08-11
fix postsolve case in duplicate columns with basis status zero

lgottwald:
this should fix issue #561

### DIFF
--- a/src/presolve/HighsPostsolveStack.cpp
+++ b/src/presolve/HighsPostsolveStack.cpp
@@ -543,30 +543,38 @@ void HighsPostsolveStack::DuplicateColumn::undo(const HighsOptions& options,
     // do postsolve using basis status if a basis is available:
     // if the merged column is nonbasic, we can just set both columns
     // to the corresponding basis status and value
-    if (basis.col_status[col] == HighsBasisStatus::kLower) {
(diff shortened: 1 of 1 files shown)


feldmeier:
This fixes the issue #561 for me. 
It leaves the (unrelated) question why a feasible LP with zero objective is unbounded, but I can work around that.

feldmeier (approved):

lgottwald:
But the LP in the issue is not unbounded? I get
```
Presolving model
591 rows, 4128 cols, 8283 nonzeros
590 rows, 3864 cols, 7747 nonzeros
Presolve : Reductions: rows 590(-248); columns 3864(-1359); elements 7747(-7811)
Solving the presolved LP
Using EKK dual simplex solver - serial
  Iteration        Objective     Infeasibilities num(sum)
          0     0.0000000000e+00 Pr: 531(341.788); Du: 0(7.62279e-10) 0s
        599     0.0000000000e+00 Pr: 0(0) 0s
Solving the original LP from the solution after postsolve
Model   status      : Optimal
Simplex   iterations: 599
HiGHS run time      :          0.06
```

lgottwald:
Ah ok, good then. Should presolve declare an instance unbounded that should not be please open another issue and assign to me

PR #567, opened 2021-08-25
Small improvements

lgottwald:
Stricter limit on the number of tableau rows used for cut generation. Previously only the number of cuts was limited to 1000. This could lead to all tableau rows of fractional basis variables being tested for cut generation if the cut generation always failed, e.g. due to numeric safe-guards or other limits imposed on the cuts. Now the cut generation for tableau based cuts is limited to only try the tableau rows of the 200 most fractional basis indices for cut generation. While leading to slightly more nodes on average runtime and number of solved instances improves.


lgottwald:
I'll close this one as it is contained in the other pull request I just opened

PR #568, opened 2021-08-25
Add binary download location to README

### LINKED ISSUE #527: Access from PuLP in Python
Hi,

I am very glad to have realized HiGHS solver over the last two weeks, since I see that it performs consistently compared to Cbc and at times does better than it. All that I miss is if I could access it from Python using the nice PuLP modelling language. Is there any work in this regard that I can use or contribute to?

Thanks,
Sam

### LINKED ISSUE #557: HiGHS.so for MacOS and Linux
I was using a Python program that uses HiGHS, and the readme said that I need to use HiGHS.so .  Where is the download for this?

odow:
Closes #557
Closes #527

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,18 @@ Documentation
 
 The rest of this file gives brief documentation for HiGHS. Comprehensive documentation is available via https://www.highs.dev.
 
+Download
+--------
+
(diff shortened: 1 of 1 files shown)


jajhall (approved):
Great! Thanks!

PR #569, opened 2021-08-30
Update SolutionToFile format

WTFHCN:
Recently, I am doing some work on connecting open source solvers and Pyomo, and I have done Pulp and Pyomo to connect MindOpt before. You encountered a small problem during the docking of HiGH (using the command line docking instead of the Python interface docking). The value of your variable is connected with the dual variable, which makes it impossible to analyze the value of the variable. I made some small changes. , Changed your format so that the program can read your value. In about a week, I will do a good job of connecting Pyomo to HIGHS, and then I want to show you the effect.


jajhall (approved):
Well spotted! These are fine by me

PR #570, opened 2021-08-30
Mip path mixing sepa

lgottwald:
It does not make a big difference but looks like a slight improvement on affected instances. It is a small extension of the already existing aggregation heuristics which produces a sequence of row aggregations where it tries to project out continuous variables with nonzero distance to their bounds in each step. This sequence of aggregations is then relaxed to form a so-called mixing set for which inequalities are separated. Using such a path to form a mixing set can be used to derive several special purpose cuts like flow path cuts and the path mixing separator is implemented to mimic the generation of flow path inequalities but is more general. I.e. it can also generate cuts when general integer variables are involved as it just relies on the MIR cuts derived from the individual aggregated rows of the path and mixes them into a single cut. When certain structure is present the cuts can be expected to be strong as they use information from several rows of the MIP and not just from a single row. The resulting inequalities can have a higher MIR-rank than 1, i.e. they would require several rounds of separation where cuts are generated from cuts to be found by the normal cMIR separation using single-row relaxations.

### DIFF
--- a/src/mip/HighsCutGeneration.cpp
+++ b/src/mip/HighsCutGeneration.cpp
@@ -537,13 +537,25 @@ bool HighsCutGeneration::cmirCutGenerationHeuristic(double minEfficacy,
 
       if (solval[i] > feastol) {
         double delta = abs(vals[i]);
-        if (delta <= 1e-4) continue;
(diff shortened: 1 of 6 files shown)


jajhall (approved):
Fine by me

Issue #571, opened 2021-09-02
HIGHS could not recognize Free MPS files


WTFHCN:
I ran into a problem that HIGHS could not recognize Free MPS files. For example:
```
* Source:     Pyomo MPS Writer
* Format:     Free MPS
*
NAME unknown
OBJSENSE
 MAX
ROWS
 N  x3
 L  c_u_x4_
 L  c_u_x5_
 L  c_u_x6_
COLUMNS
     x1 x3 40
     x1 c_u_x4_ 1
     x1 c_u_x5_ 1
     x1 c_u_x6_ 2
     x2 x3 30
     x2 c_u_x5_ 1
     x2 c_u_x6_ 1
RHS
     RHS c_u_x4_ 40
     RHS c_u_x5_ 80
     RHS c_u_x6_ 100
BOUNDS
 LO BOUND x1 -1
 UP BOUND x1 1
 LO BOUND x2 0
ENDATA

```

jajhall:
Interesting, I'll find out why

WTFHCN:
https://github.com/Pyomo/pyomo/issues/2115 I also asked this question in the Pyomo community

jajhall:
So, I see what happens when you try to load this file into HiGHS. Note that it's not a standard MPS file. Gurobi and Soplex read it, but not Clp, so it's not just HiGHS!

I'll let you know when the HiGHS MPS reader can handle it

jajhall:
This issue will be cleared when #573 is approved and merged

PR #572, opened 2021-09-03
Pyomo highs

WTFHCN:
Hello, I helped you write a Solver that connects Pyomo to HIGHS. For specific use, please refer to the test I wrote.

But some problems
+ MIP problem is not added
+ MPS file cannot be used, please refer to my last question https://github.com/ERGO-Code/HiGHS/issues/571
+ The branch inherited my previous changes to the answer file https://github.com/ERGO-Code/HiGHS/pull/569

I hope you can make some suggestions.


odow:
@WTFHCN this might be more suited to a contribution in the Pyomo repository: https://github.com/Pyomo/pyomo/tree/main/pyomo/solvers/plugins/solvers

This is what its being done with the PuLP interface: https://github.com/coin-or/pulp/pull/477
And the JuMP interface is in a separate repo https://github.com/jump-dev/HiGHS.jl

mckib2:
I will echo @odow here -- a better design choice is to have this interface live somewhere else where you can control it.  Multiple packages are using HiGHS as a backend including several Python modules.  This would also place a burden on HiGHS maintainers/developers to maintain an interface they didn't write and don't have a stake in

jajhall:
Thanks for your comments @odow and @mckib2 . PR #569 has been merged since it fixes an error (missing spaces)  in the code that prints a solution file, and I've just fixed issue #571 - which only related to recognition of very short names in free format MPS, not free format itself. 

So, @WTFHCN , please follow the suggestions of @odow and @mckib2 and add your interface as a contribution to the Pyomo repository. That would also seem to be a better way of publicising to Pyomo users that there is an interface to HiGHS. 

Hence I'll not approve this PR to be merged into master.

When you say that there is a "problem" that the MIP problem is not added, is that just because there's no interface, or some problem with HiGHS? The HiGHS MIP solver is exceptionally powerful now - it's overtaken SCIP on one of the two MIPLIB 2017 measures (problems solved in 3600s) - and will be of great value to open source users of modelling languages and interfaces like Pyomo. So, let's keep talking!

Finally, many thanks for the work you're doing and for keeping us on our toes!

WTFHCN:
OK! I will continue to try to optimize the code structure and add MIP questions!

odow:
> it's overtaken SCIP on one of the two MIPLIB 2017 

Are these benchmarks public?

jajhall:
Not yet. We want to improve solution speed a bit more first so that we beat SCIP on that, too.

jajhall:
Thanks for the heads-up on ECOLE. I've seen a few packages using SCIP - some probably illegally. If nothing else, SCIP is the only MIP solver in Google's OR-Tools. They have a disclaimer, but I know that OR-Tools is being used commercially for MIP work. Google will be offering HiGHS as a solver in due course.

The HiGHS benchmark results are public. It's only a the limited documentation that's preventing us from promoting the solvers more.

PR #573, opened 2021-09-04
Mps short names

jajhall:
#571 is another issue where short row and column names are fooling the free format MPS reader into thinking that it's a file in fixed format with spaces in the (row) names. The Netlib example that I always thought had strict spaces in its names is d6cube, but they are just very odd row names like "     1  " that can be read as "1" in free format. So, I've added code that, when a name with spaces might previously have been identified - because the MPS line is (eg) "x1 x3 40" - if the second word "x3" is found in the row names, it's accepted as being such. Otherwise the file is given to the fixed format reader and fails.

### DIFF
--- a/src/io/HMPSIO.cpp
+++ b/src/io/HMPSIO.cpp
@@ -197,7 +197,7 @@ FilereaderRetcode readMps(const HighsLogOptions& log_options,
     highsLogUser(log_options, HighsLogType::kWarning,
                  "COLUMNS section entries contain %8" HIGHSINT_FORMAT
                  " with row not in ROWS  "
-                 "  section: ignored",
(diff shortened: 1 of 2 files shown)


jajhall:
I tried some pathological MPS files like 

x-special/nautilus-clipboard
copy
file:///home/jajhall/HiGHS/build64/odd_names.mps

that is fine with fixed format MPS. Gurobi recognises that it's got three rows, but doesn't distinguish them (or the second column) in the rest of the file. HiGHS free format MPS reader recognises that there is a space in the row name "r 0" and reads the faile assuming that it's in fixed format. No sane person would put spaces into names of an MPS file in free format and expect it to be read!

PR #575, opened 2021-09-09
Qpsimplifications

feldmeier:
Main change: removed explicit nullspace matrix, replaced all operations with nullspace matrix by ftran/btran operations. 
Shaves off ~10% of computation time on some instances

Minor changes here and there, e.g. avoid creating some temporary vectors and buffer them instead. Also some duplicate calculations (ftran/btran) are avoided now.


jajhall (approved):
Looks good to me

PR #577, opened 2021-09-13
Mip aggregation heuristic

lgottwald:
This makes some changes to the row aggregation heuristic for cut separation, and additionally contains 1 bugfix and a small change regarding the timing when heuristics are called. The aggregation heuristic's behavior was not as intended at several places leaving some room for improvements. E.g. the where some cases where I used code like 'success = success || generateCut()' where the `generateCut()` function was not called if success was true due to short circuit evaluation. Also the heuristic was not checking whether a row is part of the current aggregation which could lead to adding a row twice. The changed code gives much better dual bounds on quite some problems and allows to solve some that where not solved before, e.g. the instance lotsize from the benchmark set which none of the other listed solvers solves within two hours. The benchmark test set is still running right now to check the performance impact of this changes.

Also I added some working limits to avoid blowing up the clique table with too many entries. This avoids some cases where the MIP solver got stuck in the clique merging routine, but I need to look into another instance where it still happens.

### DIFF
--- a/src/mip/HighsCliqueTable.cpp
+++ b/src/mip/HighsCliqueTable.cpp
@@ -524,15 +524,17 @@ void HighsCliqueTable::doAddClique(const CliqueVar* cliquevars,
         case 0:
           // clique empty, so just mark it as deleted
           cliques[cliqueid].start = -1;
+          cliques[cliqueid].end = -1;
           freeslots.push_back(cliqueid);
-          break;
+          return;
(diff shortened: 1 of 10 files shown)


jajhall (approved):
Nice to see that you've picked up on the short circuit evaluation! Congratulations on adding lotsize to the set of solved instances. Naturally I trust you on these modifications

PR #578, opened 2021-09-20
Improve pkg-config experience

### LINKED ISSUE #504: 'pkg-config --cflags highs' returns wrong output
```
$ pkg-config --cflags highs
-I/usr/local/ 
```

chrhansk:
These are some small fixes which should aid with using highs as a library via pkg-config. This should fix #504, which is why I did not open a seperate issue beforehand.


jajhall:
Thanks for this. I'll get it reviewed by the person who handles the HiGHS software engineering

PR #581, opened 2021-09-23
Mip working limits

lgottwald:
Adds some working/iteration limits and code to cleanup dead entries in the clique table. Prevents several cases where HiGHS gets stuck currently.


jajhall (approved):
Good that you've spotted and fixed this

PR #582, opened 2021-09-27
Write solution mittelmann

lgottwald:


lgottwald:
What would be left todo would maybe be to bump the version to 1.1.1 and to change the written solution slightly, so that instead of `# Columns` as first line it has `=obj= <the solution's objective value>`

jajhall:
> What would be left todo would maybe be to bump the version to 1.1.1 and to change the written solution slightly, so that instead of `# Columns` as first line it has `=obj= <the solution's objective value>`

Agree on the version number, but maybe it's clearer to have a # Objective comment and keep the # Columns comment so that no-one confuses `=obj= <the solution's objective value>` with a column value.

Hard to be sure when you don't know how this file is being used

lgottwald:
I think it is supposed to be passed to the solution checker of the MIPLIB which is contained in this archive:
http://miplib.zib.de/downloads/miplib2017-testscript-v1.0.3.zip

It expects the objective value to be given as `=obj= <objective value>`.
I need to check if it can deal with `# Columns` as comment properly.

jajhall:
Ah, I see, then the '=obj=' line can be added and the comments can be removed. If it's '=obj=", and not the name of the objective in the MPS file, then clearly the only ambiguity comes if someone calls a column '=obj='!

lgottwald:
Yes. I just checked the solution checker and it indeed fails to check the solution if the =obj= line is missing...

lgottwald:
I adapted the solution format to use =obj= and expect primal values only

lgottwald:
And also tested that it is accepted by the miplib solution checker

lgottwald:
Now I also moved the version and think this is ready for merging

jajhall:
As it stands, HIGHS_VERSION_PATCH is incorrect in HConfig.h, as it is set separately in CMakeLists.txt (not in 
project(HIGHS VERSION 1.1 LANGUAGES CXX C)
so I've done this.

jajhall (approved):
Assuming the final two CI tests are OK, this is good to go - creating v1.1.1!

PR #583, opened 2021-10-03
Parser and presolve fixes

lgottwald:
Fixes an issue where presolve uses dense columns in doubleton equations to do substitutions leading to excessive presolve time on sgpf5y6. Also fixes an issue in the MPS parser leading to warnings on istanbul-no-cutoff and to the miplib solution checker rejecting the solution due to column names that don't match. The column names differ because the " character is trimmed sometimes from the names.


jajhall (approved):
Fine

Issue #580, opened 2021-09-23
OsiInterface building fails


felixlen:
Currently building the OSI interface fails because `lp_` cannot be found in `highs`:
```
src/interfaces/OsiHiGHSSolverInterface.cpp:340:23: error: no member named 'lp_' in 'Highs'
  return this->highs->lp_.num_col_;
```
and because in l. 554 and 582 of the same file the return type of the functions `addRow` and `addCol` is no longer `bool` but `HighsStatus`:
```
bool success = this->highs->addRow(rowlb, rowub, vec.getNumElements(),
                                     vec.getIndices(), vec.getElements());
```

If I replace every occurrence of `highs->lp_` with `highs->model_.lp_` and change the return type in l. 554 and 582 to `HighsStatus` and the subsequent `assert` and `if (success)` to a comparison with `HighsStatus::kOk`, I am able to compile HiGHS again, however run into a segfault in the unit test of the OSI interface.

From https://github.com/ERGO-Code/HiGHS/issues/525 I understand that the OSI interface is currently not maintained and will only be restored later. However, it is built as soon as the OSI library is found (`src/CMakeLists.txt:369`), so that building HiGHS fails for every user having OSI installed.

I suggest considering to just deactivate building the OSI interface for the time it is unmaintained to improve the experience for users having the OSI library installed on their system.

jajhall:
Yes, I perfectly understand, and had not considered this. I'm sure that it's putting off some users from using HiGHS.

@galabovaa Is this easy to do?

jajhall:
> Yes, I perfectly understand, and had not considered this. I'm sure that it's putting off some users from using HiGHS.
> 
> @galabovaa Is it easy to deactivate building the OSI interface for now?

Do we only have to comment out the following line in CMakeLists.txt @galabovaa ?

pkg_check_modules(OSI osi)

jajhall:
#704 should mean that OsiInterface now builds. The OSI interface still needs updating since some of the unit tests fail. This is covered by #525

PR #592, opened 2021-10-15
add option HiGHSOSI and fix option 'OPENMP'

tosttost:
This fixes https://github.com/ERGO-Code/HiGHS/issues/580.
In order to build the OSI interface you can simply add "-DHiGHSOSI=ON" when running cmake. 

Currently there is an option "OPENMP" but if you set it to OFF in a build environment that supports OpenMP 3.0 it is ignored and OpenMP is used. I fixed it such that if OPENMP is set to OFF, it is not used. If set to ON the check for OpenMP 3.0 is performed and if successful OpenMP is activated.
This is useful if you are compiling for use on another machine that does not support OpenMP, i.e. cross compiling using mingw. (mingw supports OpenMP 3.0 but the resulting binaries are linked dynamically against libgomp.dll. Unless you ship libgomp.dll together with the binary it will run on most windows machines). 
Additionally some of the stripped down linux distributions used by cloud providers like AWS do not have OpenMP preinstalled. Therefore any HiGHS binary built on a 'normal' linux machine won't run on such systems unless you explicitly install OpenMP.

tbd: would it be better to print a warning message if OPENMP=ON but OpenMP 3.0 is not found?


tosttost:
Error found: 2 places in the qpsolver use std::thread. 
At least my version of gcc require to link explicitly against pthreads if you use the builtin C++ threading functionality. AFAIK clang behaves similar and this is expected behaviour, see e.g. https://github.com/compiler-explorer/compiler-explorer/issues/1428
I suspect -fopenmp causes gcc to link against pthreads and therefore fixing the problem by coincidence.

Someone with a (hypothetical) version  of gcc/clang that is able to compile c++11 but does not support openmp 3.0 will see the same linker error in master. I don't think such a version of gcc/clang exists.

PR #607, opened 2021-11-01
Taboo restrictions

jajhall:
Vast amount of work to merge into master. 

- Solving the unscaled LP using scaled NLA
- Removal of HighsModelObject and reduction of number of copies of the constraint matrix
- Eliminated unnecessary refactorizations with simplex
- Added saving and recovery of hot start for simplex - when "continuing" with current state is undesirable. Includes fast refactorization using stored pivot sequence.
- Added freezing of current basis, with subsequent simplex iterations performed using product form updates, enabling efficient return to frozen basis.
- Avoids small "late pivots" in simplex, which can identify primal infeasibility in dual simplex via construction of proof constraint. latter incorporates quad precision evaluation of matrix-vector product.
- Cycling generally resolved by means of taboo row (column) in dual (primal) simplex.
- Ability to solve LP by forming its dual and using simplex
- Solution file now writes out primal solution and any dual solution or basis separately. Option to compute and write out ranging information.


lgottwald (approved):
I approve for now and would just suggest to remove all debug prints so that master runs without extra output.

If new problems come up we can just do follow-up merge requests

PR #608, opened 2021-11-08
Taboo restrictions infeas proof changes

lgottwald:


jajhall (approved):
Maybe there are important differences between this and the code you've got to do the same thing in the MIP solver, but if they really are equivalent, it would be nice if there was a single method used by both.

lgottwald:
It is not exactly the same right now, but the differences are not well-founded by tests and I would rather plan for a follow-up merge request to expose the proof constraints computed within the LP solver via API functions so that the MIP solver essentially just uses it instead of recomputing it.

The code in the MIP solver uses a different criterion for dropping small values in row_ep and differs in scaling right now, but I would rather remove that part based on my current understanding and current tests.

jajhall:
> It is not exactly the same right now, but the differences are not well-founded by tests and I would rather plan for a follow-up merge request to expose the proof constraints computed within the LP solver via API functions so that the MIP solver essentially just uses it instead of recomputing it.
> 
> The code in the MIP solver uses a different criterion for dropping small values in row_ep and differs in scaling right now, but I would rather remove that part based on my current understanding and current tests.

OK, thanks, I'll leave this in your hands. I'll move on to the issue of incomplete/rectangular matrix factorization

PR #609, opened 2021-11-09
Cleanup master after merge

lgottwald:
- Clean up some prints in simplex
- Add handling of new status code kUnkown in MIP solver
- Adjust invalid infeasibilitity count/num to be kHighsInf/kHighsIInf
- Adjust basisTransfer function to maybe reduce likelihood of rank deficiencies by preferring rows to be basic first, then structural continuous columns, and last structural integral columns


jajhall (approved):
I wouldn't worry about prioritising rows on grounds of avoiding singularities, particularly if a basis can be constructed with the same "structural essence" as a basis previously returned by the simplex solver

PR #611, opened 2021-11-09
Highs task simplex

lgottwald:
This adds a work stealing scheduler mostly based on https://link.springer.com/chapter/10.1007%2F978-3-319-14313-2_18
It only uses standard C++11 parallel features internally.

All functionality is exposed in parallel/HighsParallel.h
There needs to be one call to highs::parallel::initialize_scheduler(threadnum) before any functionality is used.
After that there is a `highs::parallel:spawn(<task>)` and `highs::parallel::sync()` call that is the core functionality. It provides a fork-join parallelism model. Every call to spawn needs to be matched with one call to sync which can be thought of like a stack. After the call to sync the task of the latest call to spawn is guaranteed to be finished. As a wrapper around this basic functionality there is a highs::parallel::for_each function which provides a parallel for-loop mechanism. For synchronization there is a highs::parallel::mutex which can be used like std::mutex, but will yield control to the scheduler in case the critical section cannot be entered. There also is a parallel/HighsSpinMutex.h which should only be used with great care, as it spin waits in a busy loop until the mutex becomes available, which should be reserved for the scheduler itself.

Additionally the branch adapts options regarding parallelization for HiGHS. The options `highs_min_threads`/`highs_max_threads` are replaced by `simplex_min_concurrency`/`simplex_max_concurrency`. Additionally there is a new option to control the number of threads, `highs_threads`, which is 0 by default meaning the number of threads is chosen automatically. Currently it will use half the number of CPUs of the system. Reason for using only half the number of CPUs is that most modern CPUs support hyperthreading with 2 logical CPUs per physical core.

The parallel simplex code is adapted to use the new work-stealing scheduler and OpenMP functionality is removed from the code and CMakeLists.txt.

Additionally several compiler warnings are fixed that pop up on my machine, probably due to using newer compiler versions where some of the previous suppression options did not work.


lgottwald:
builds seem to mostly pass now, but the fast-build-debug for windows somehow gets stuck

jajhall:
This is an epic achievement! Fast-build has the only test of the parallel code, so I guess that on Windows the parallel work has caused an error. I'll look at it.

lgottwald:
I caught one error in the parallel code already that popped up with valgrind on the adlittle test. Also I improved/fixed several things in the cmake build files and added suppression flags for the many warnings that made it impossible to follow the compiler output on windows.

lgottwald:
@feldmeier There was a bug caught by an MSVC warning where you did not return a value from a function. I added a todo for you and am return a signaling_NaN() now. This is within the expected behavior of your function before since the compiler may assume that the if condition you are checking before is always true since when it is false you are invoking currently invoking undefined behavior.
Also there are some more warnings in MSVC that you may want to look at:
```
D:\a\HiGHS\HiGHS\src\qpsolver\solver.cpp(69): warning C4702: unreachable code [D:\a\HiGHS\build\src\libhighs.vcxproj]
D:\a\HiGHS\HiGHS\src\qpsolver\solver.cpp(72): warning C4702: unreachable code [D:\a\HiGHS\build\src\libhighs.vcxproj]
D:\a\HiGHS\HiGHS\src\qpsolver\solver.cpp(73): warning C4702: unreachable code [D:\a\HiGHS\build\src\libhighs.vcxproj]
D:\a\HiGHS\HiGHS\src\qpsolver\solver.cpp(75): warning C4702: unreachable code [D:\a\HiGHS\build\src\libhighs.vcxproj]
D:\a\HiGHS\HiGHS\src\qpsolver\solver.cpp(84): warning C4702: unreachable code [D:\a\HiGHS\build\src\libhighs.vcxproj]
D:\a\HiGHS\HiGHS\src\qpsolver\solver.cpp(87): warning C4702: unreachable code [D:\a\HiGHS\build\src\libhighs.vcxproj]
```

lgottwald:
@jajhall The windows build now seems to pass, however, that is because I noticed how incredibly slow it runs comapred to the linux debug builds and found out that one can add `-D_ITERATOR_DEBUG_LEVEL=0` which is done in line 132 of the CMakeLists.txt now.
`
add_definitions(-D_ITERATOR_DEBUG_LEVEL=0)
`
I would still like to see if you can comment this part out and see where it fails on windows. The definition disables the expensive checks that are run on every STL container iterator. The debug builds on windows indeed do run much faster now and I also was able to significantly speed-up the runtime of the cmake configuration step by not checking for Fortran if the compiler is MSVC. The check whether a Fortran compiler exists for some reason takes over a minute on Windows before it fails.

Besides there are several improvements and fixes for the build system and code bugs fixed that where discovered by warnings.

jajhall (approved):
Well done! I'll see what happens on a real Windows machine, as requested.

PR #616, opened 2021-11-27
some changes for MIP

lgottwald:
Some changes that accumulated in a working branch. In HEkkPrimal.h there was a runtime error reported by the undefined behavior sanitizer caused by reading the value of use_hyper_chuzc. Since this is a bool it must be 0 or 1 but apparently it was read when there was some garbage value stored due to not being initialized. I added an initialization with value `false` and hope that is alright.


jajhall (approved):
The change to HEkkPrimal.h is fine. I've fixed the error in a different branch, I'm sure.

PR #619, opened 2021-11-29
Ipm postsolve

jajhall:
Performs primal postsolve when presolved LP is solved with IPM and no cross-over, plus a lot of profiling code in IPX


lgottwald:
I will add primal/dual postsolve to this too and will also enable the sparsification presolver, as it also allows primal dual postsolve but not to retain a basis currently.

jajhall:
Maybe I misunderstand you but, by default, presolve doesn't know if the problem is to be solved with IPM with/without crossover, or simplex. That said, it's easy for me to set a flag before calling presolve once I know what solver options are set

lgottwald:
Ah ok, no you did not misunderstand. Knowing whether a basis is required would be good and allow some additional presolve steps for IPM. So setting a flag would be nice.

jajhall:
Just create a HiGHS option value whose default value runs presolve as it is now. I'll set it locally in Highs::run(), but it could be set by a user who is only using Highs::presolve()

lgottwald:
Okay, I now adapted postsolve to allow primal/dual postsolve without a basis. This can be done by calling undo with a basis that has basis.valid = false or by using the undoPrimalDual function without passing in a basis.

The code should still be run through sanity checks. Also it is not yet called as I don't know where you currently set up the presolve and postsolve calls for the case of IPM without crossover.

Also I added an option lp_presolve_requires_basis_postsolve which is true by default. If set to false HPresolve will allow some additional presolve steps (matrix sparsification, almost parallel rows). In the future I might adapt presolve to use even more steps when given a use case where no crossover is required and performance is not good enough.

jajhall:
Thanks. I'm currently working on a use case where the IPM solver is annoyingly slow - for the PyPSA people who may be able to fund work on it. I'll let you know if the additional presolve features improve performance

lgottwald:
Also I remember from some conversations that tighter column bounds may be helpful for IPM solvers, so you can also test whether use_implied_bounds_from_presolve=true helps performance.

jajhall:
So, for the PyPSA LPs, sparsify removes nothing - I've double-checked that it's being tried! - and use_implied_bounds_from_presolve=true slows down solution! Such is life with LPs

lgottwald:
Okay, then its not doing anything for this LP. Have you had a chance to set up the primal-dual postsolve yet? I would like to see some testing for the code I added and I might still need to do some debugging.

jajhall:
> Okay, then its not doing anything for this LP. Have you had a chance to set up the primal-dual postsolve yet? I would like to see some testing for the code I added and I might still need to do some debugging.

In looking through the changes that you made, I was reminded that unless crossover is run IPX returns only a primal solution. It shouldn't be hard to identify the dual solution, but I'll have to do this first. IPX can't handle boxed constraints so adds a variable for each.

jajhall:
Profiling code now remains in only ipm-timer. After IPX without crossover, minimal truncations of primal and dual values tare made to achieve primal and dual feasibility, at the expense of primal and dual residuals. There's no avoiding the loss of something when complementarity after IPX without crossover allows primal or dual infeasibilities that are modest, but larger than the tolerances. Tested on the easy 100 LP problems with no nasty surprises

jajhall:
Other new things in this PR are

- The "Alien basis" handling, requiring HFactor to handle a rectangular matrix. This allows the guessed basis completion in the MIP solver to be removed. For debugging, an optional parameter has been added to Highs::setBasis so that the origin of the simplex basis can be tracked. Hence changes when setBasis is called in the MIP solver.
- Handling of redundant equations and dependent free columns added to presolve, but not (yet) the bail-out if the rectangular factorization takes too long due to fill-in, but this only affects bdry2
- Suppression of printing in TestHighsParallel.cpp
- Some diagnostics (data gathered in basis.cc and reported via reportBasisData) retained in IPX - for an issue that I will want to come back to in future. No overhead, and no printing by default.
- The draft code to take action when cycling is detected, and for preventing repeated factorization of singular basis matrices when back-tracking. These are related. This is generally only active in a few MIPs that we can't solve anyway (eg glass4) but cannot be fully debugged until the strange asserts in HFactor are investigated. Until then, the singularity, at least, isn't being identified properly.
- Last references to OpenMP (in comments and reporting) removed

lgottwald:
So the changes that I made in postsolve are now already tested and work for providing a primal-dual solution after postsolve of a non-basic solution?

jajhall:
Yes, on my 100 easy LPs, but that should be enough to spot logical errors. The accuracy of what's produced - sometimes only O(1e-3) residuals - isn't fantastic, but that's down to what IPX gives us

lgottwald:
residuals you mean regarding the kkt constraints?

lgottwald (approved):

jajhall:
> residuals you mean regarding the kkt constraints?

Yes. I have the choice of leaving primal and dual values that don't satisfy complementarity, or making the least truncations - primals to bounds and duals to zero - so that they do. Naturally this reduces the extent to which the KKT equations are satisfied. In due course I'll look at alleviating this by running IPX to greater default accuracy when crossover is not used. However, for now, there's no external demand - unless you want better accuracy when you use IPX without crossover. Of course, you always were accepting a primal solution for which the dual solution (not extracted from IPX) violated complementarity.

PR #620, opened 2021-11-30
Small change

lgottwald:
I did some testing and clang yields 3-4% better performance on my machine (AMD) and on oronsay (Intel). I added a small bit to the CMakeLists.txt that will use clang over gcc when it is available.


jajhall (approved):
OK

PR #622, opened 2021-12-06
Misc changes

lgottwald:
A small adjsutment on the infeasibility proof to consider the sum of all unbounded entries in  row_ap and some change to avoid compilation failures in https://github.com/JuliaPackaging/Yggdrasil/pull/4028#issuecomment-986292158


jajhall (approved):
Fine by me

PR #624, opened 2021-12-09
passLp in c# interface working again

### LINKED ISSUE #621: C# wrapper API is not correctly calling highs.dll
Method Highs_passLp of wrapper class HighsLpSolver uses invalid signature for dll call. Correct function in C api has signature

HighsInt Highs_passLp(void* highs, const HighsInt numcol, const HighsInt numrow,
                      const HighsInt numnz, const HighsInt a_format,
                      const HighsInt sense, const double offset,
                      const double* colcost, const double* collower,
                      const double* colupper, const double* rowlower,
                      const double* rowupper, const HighsInt* astart,
                      const HighsInt* aindex, const double* avalue)

whereas C# assumes that it expects
int Highs_passLp(void* highs, int numcol, int numrow, int numnz, double[] colcost,
   double[] collower, double[] colupper, double[] rowlower, double[] rowupper, int[] astart, int[] aindex, double[] avalue);

Arguments a_format, sense and offset are missing, therefore calling Highs_passLp from .net causes access violation. To fix the issue one would probably introduce enum HighsMatrixFormat and adapt c# signature.

feldmeier:
Fixes #621, more changes might be necessary to update other C# API functions. C# example in /examples folder is now running successfully.

### DIFF
--- a/examples/call_highs_from_csharp.cs
+++ b/examples/call_highs_from_csharp.cs
@@ -12,8 +12,11 @@ class Program {
       int[] astart = {0, 2};
       int[] aindex = {0, 1, 0, 1};
       double[] avalue = {1, 2, 1, 3};
+      HighsObjectiveSense sense = HighsObjectiveSense.kMinimize;
+      double offset = 0;
+      HighsMatrixFormat a_format = HighsMatrixFormat.kColwise;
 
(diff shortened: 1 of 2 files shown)


jajhall (approved):
Thanks. This tells me better what to do for all the rest

PR #628, opened 2021-12-13
refine basis status of fixed nonbasic vars to be dual feasible after …

lgottwald:
…var is unfixed

As this is quite a small change I would think it is better to set the basis status for fixed non-basic variables to be kUpper when the dual is negative and kLower otherwise. When then a basis is reused that contained fixed variables for an LP where one of the variables is unfixed dual feasibility is retained without requiring bound flips to happen first.


jajhall (commented):
I've just made equivalent edits to HEkk.cpp on favour-dual-phase2. However, for consistency, it also has to be done when getting a HighsBasis from the IPX solution. As I'm half way through this edit, and don't want to have to remove my edits from HEkk.cpp I won't approve this PR

lgottwald:
Ah okay, then nevermind.
I was just in the process of adding some changes I made on a working branch in separate smaller branches and this was one of them, but most are to MIP things.

PR #631, opened 2021-12-17
FIX: remove always-false assert statements

### LINKED ISSUE #630: BUG: primal solver dies when solving simple unbounded LPs
When built in debug, we run into assert statements of the form `assert(1 == 0);` for simple unbounded LPs.

The two instances I see are:
- https://github.com/ERGO-Code/HiGHS/blob/2da8f7d4d35965d4c88a1b5a69ee772ebd1e9208/src/simplex/HEkkPrimal.cpp#L2146
- https://github.com/ERGO-Code/HiGHS/blob/2da8f7d4d35965d4c88a1b5a69ee772ebd1e9208/src/simplex/HEkkPrimal.cpp#L2161

It appears to me that these were a developer's rough-n-dirty break-points and can probably be safely removed.  I will make a PR, please do say if this is not the case.

mckib2:
- closes #630 

Removes always-false `assert` statements that prevent the solution to some unbounded LPs when built in debug mode.

### DIFF
--- a/src/simplex/HEkkPrimal.cpp
+++ b/src/simplex/HEkkPrimal.cpp
@@ -2143,7 +2143,6 @@ void HEkkPrimal::basicFeasibilityChangePrice() {
   const bool quad_precision = false;
   if (use_col_price) {
     // Perform column-wise PRICE
-    assert(1 == 0);
     ekk_instance_.lp_.a_matrix_.priceByColumn(quad_precision,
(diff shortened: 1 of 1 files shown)


jajhall:
Fixed by https://github.com/ERGO-Code/HiGHS/commit/ab302f1c1dd7a81a5abc50fc86f852afc115424f

PR #634, opened 2021-12-18
Fixes for correct infeasibility claim

lgottwald:


lgottwald:
Not that this goes into you working branch and not into master. The first change I sent you an email about.

The second one is a minor change in rare cases where a basis of a node closer to the root node is now preferred to be used when solving the next node from the node queue. Previously it could happen that several nodes of the higher parts of the tree where pruned due to domain propagation and then the basis deeper down the tree of the last feasible node's parent would be used.

lgottwald:
I added a third change to use the set overload of changeColsBounds to avoid scanning all columns everytime one bound is changed.

jajhall (approved):
OK

Issue #633, opened 2021-12-18
MIP Highs::run() returns status -1 for optimal solution


mckib2:
When non-zero integral constraints are given the `Highs::run()` function returns a `HighsStatus` of `-1` for this simple problem taken from MATLAB's intlinprog documentation:

![image](https://user-images.githubusercontent.com/28734319/146629878-7079cb12-751a-4eb4-9e10-c94716ce558d.png)

If useful, I can provide an MPS file describing this problem, but I'm hoping it will be a simple fix to set the status correctly.  Printed to the console is the following:
```
Presolving model
2 rows, 3 cols, 6 nonzeros
0 rows, 1 cols, 0 nonzeros
0 rows, 0 cols, 0 nonzeros
Presolve: Optimal

Solving report
  Status            Optimal
  Primal bound      -12
  Dual bound        -12
  Solution status   feasible
                    -12 (objective)
                    0 (bound viol.)
                    0 (int. viol.)
                    0 (row viol.)
  Timing            0.00 (total)
                    0.00 (presolve)
                    0.00 (postsolve)
  Nodes             0
  LP iterations     0 (total)
                    0 (strong br.)
                    0 (separation)
                    0 (heuristics)
```

The console output leads me to believe this problem was actually solved, but the status says otherwise.

_EDIT: as discovered in investigations below, the incorrect status is reported when `highs_debug_level >= HighsDebugLevel_kHighsDebugLevelCheap`_

mckib2:
With presolve off, the following output is achieved will with status -1 reported:
```
resolve is switched off

Solving MIP model with:
   2 rows
   3 cols (1 binary, 0 integer, 0 implied int., 2 continuous)
   6 nonzeros

( 0.0s) Starting symmetry detection
( 0.0s) No symmetry present

Solving root node LP relaxation

Presolving model
2 rows, 3 cols, 6 nonzeros
2 rows, 3 cols, 6 nonzeros
Presolve : Reductions: rows 2(-0); columns 3(-0); elements 6(-0)
Solving the presolved LP
Using EKK dual simplex solver - serial
  Iteration        Objective     Infeasibilities num(sum)
          0    -2.1200077156e+01 Pr: 2(5.65) 0s
          1    -1.2000000000e+01 Pr: 0(0) 0s
Solving the original LP from the solution after postsolve
Model   status      : Optimal
Simplex   iterations: 1
Objective value     : -1.2000000000e+01
HiGHS run time      :          0.00

[...]

PR #635, opened 2021-12-19
ENH: don't generate log file by default

### LINKED ISSUE #586: The Highs C library systematically tries to create a Highs.log file
The highs C library seems to always create a `Highs.log` file.

When explicitly setting the `log_file` option to the empty string, the `Highs.log` file remains empty, but it is always created.

Is there an easy way to prevent the usage of the library from having any side effect ? When building a larger software on top of libraries, one generally wants to control input/output centrally rather than having each library writing to stdout and creating its own files.

mckib2:
- closes #586

All usages of `options.log_file_stream` appear to check for NULL, so this should be safe and all tests pass locally.  It avoids the creation of a log file if the `log_file` option is the empty string.  It also sets the `log_file` option to be the empty string by default.

### DIFF
--- a/src/lp_data/HighsOptions.h
+++ b/src/lp_data/HighsOptions.h
@@ -601,7 +601,7 @@ class HighsOptions : public HighsOptionsStruct {
     records.push_back(record_string);
 
     record_string = new OptionRecordString(kLogFileString, "Log file", advanced,
-                                           &log_file, "Highs.log");
+                                           &log_file, "");
     records.push_back(record_string);
 
     record_bool =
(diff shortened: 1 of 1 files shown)


jajhall:
I can see how this avoids creating the log file by default, there's a lot more to controlling the way HiGHS logs to a file. Since you seem keen to have this fixed, I'll work on it this week.

mckib2:
Thanks @jajhall !  Yes, this is the last thing holding us getting up to date with the latest/greatest HiGHS.  Appreciate you taking a look especially with holidays ongoing -- I finally have time off from my day job to work on this and I expect you are looking forward to time off yourself!

jajhall:
I've fixed this, and #633, which will close when #645 is passed. If you want to experiment in the meantime, check out the modify-logging branch

PR #643, opened 2021-12-20
make the dualUB bailout more reliable

lgottwald:
This changes the mechanism by which the exact dual objective is computed. I think the current one would often fail to compute a valid dual objective under cost perturbation, because it does not consider the cases where the unperturbed duals may flip sign so that the opposite bound should be used for the exact dual objective. It also ignores that basic variables might get a nonzero dual value in the unperturbed space and also need to be considered with their corresponding bound for computing a valid exact dual objective.

Additionally, when the status now is determined to be kObjectiveBound for the scaled LP, then the unscaled LP is solved without cost perturbation to make the verification of whether the objective bound is reached more reliable.

This prevents many cases of wrong claims of reaching the dual objective bound and gets rid of most cases where the solver terminates with a dual infeasible solution. It may still happen that the solver terminates with dual infeasibilities but then they are guaranteed to be rather simple as the dual infeasible solution still provides a proof that the objective bound is reached. I.e. when there is a dual infeasibility for a nonbasic variable it must be removable with a bound flip and when there is one for a basic variable it must be bounded in the direction of its dual value. I think the latter would mean that the column can be chosen as leaving column in a primal simplex pivot?

With this change, however, there will be no bailout in cases where unbounded variables have a too large dual to reliably compute a finite bound and return -inf as the exact dual objective. In such a case instead of continuing the solve to optimality it could be better to clean up any cost perturbations with primal simplex starting in phase 2 by using bound shifts for reaching primal feasibility, but this should happen within HEkkDual, not within HApp as in the other branch.

### DIFF
--- a/src/mip/HighsLpRelaxation.cpp
+++ b/src/mip/HighsLpRelaxation.cpp
@@ -463,11 +463,11 @@ bool HighsLpRelaxation::computeDualProof(const HighsDomain& globaldomain,
 
     if (std::abs(val) <= mipsolver.options_mip_->small_matrix_value) continue;
 
-    bool removeValue = std::abs(val) <= mipsolver.mipdata_->feastol ||
-                       globaldomain.col_lower_[i] == globaldomain.col_upper_[i];
+    bool removeValue = std::abs(val) <= mipsolver.mipdata_->feastol;
 
     if (!removeValue &&
-        mipsolver.variableType(i) == HighsVarType::kContinuous) {
+        (globaldomain.col_lower_[i] == globaldomain.col_upper_[i] ||
+         mipsolver.variableType(i) == HighsVarType::kContinuous)) {
       if (val > 0)
         removeValue =
             lpsolver.getSolution().col_value[i] - globaldomain.col_lower_[i] <=
--- a/src/mip/HighsLpRelaxation.cpp
+++ b/src/mip/HighsLpRelaxation.cpp
@@ -796,7 +796,10 @@ HighsLpRelaxation::Status HighsLpRelaxation::run(bool resolve_on_error) {
       if (info.max_dual_infeasibility <= mipsolver.mipdata_->feastol)
         return Status::kUnscaledDualFeasible;
 
-      return Status::kUnscaledInfeasible;
+      if (scaledmodelstatus == HighsModelStatus::kOptimal)
+        return Status::kUnscaledInfeasible;
+
+      return Status::kError;
     case HighsModelStatus::kIterationLimit: {
       if (!mipsolver.submip && resolve_on_error) {
         // printf(
--- a/src/mip/HighsLpRelaxation.cpp
+++ b/src/mip/HighsLpRelaxation.cpp
@@ -597,10 +597,10 @@ void HighsLpRelaxation::storeDualInfProof() {
 
     if (removeValue) {
       if (val < 0) {
-        if (globaldomain.col_upper_[i] == kHighsInf) return false;
+        if (globaldomain.col_upper_[i] == kHighsInf) return;
         upper -= val * globaldomain.col_upper_[i];
       } else {
-        if (globaldomain.col_lower_[i] == -kHighsInf) return false;
+        if (globaldomain.col_lower_[i] == -kHighsInf) return;
         upper -= val * globaldomain.col_lower_[i];
       }
 
--- a/src/mip/HighsLpRelaxation.cpp
+++ b/src/mip/HighsLpRelaxation.cpp
@@ -563,6 +563,8 @@ void HighsLpRelaxation::storeDualInfProof() {
     }
   }
 
+  const HighsDomain& globaldomain = mipsolver.mipdata_->domain;
+
   for (HighsInt i = 0; i != lp.num_col_; ++i) {
     HighsInt start = lp.a_matrix_.start_[i];
     HighsInt end = lp.a_matrix_.start_[i + 1];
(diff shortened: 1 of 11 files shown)


jajhall:
Thanks. I'll look through this, but my instinct is that it's a clear improvement on what we had before. In particular it brings objective bound bail-out into the environment of the unscaled LP. Unless I misunderstand your point, basic variables can't acquire a "nonzero dual value in the unperturbed space" since they are always zero with exact arithmetic, being c_B - B^T\pi, where \pi = B^{-T}c_B and c_B is the vector of basic costs, perturbed or not.

lgottwald:
Ah your right. I thought it may happen that basic columns get nonzero cost due to the dual values being computed in the perturbed space, but the function already makes sure to compute the duals with the unperturbed cost. There could still be errors of the system solve though so I would leave the dual computation of the basic columns there.

Also I am looking into a case in the debugger right now where the result is kObjectiveBound but the MIP solver comes to the conclusion that the gap is ~ 2000 when it recomputes the dual proof, which indicates something could be wrong.

jajhall:
Values of c_B - B^T\pi are currently either not computed or zeroed explicitly. Since they are measures of how accurately the nonbasic duals have been computed, for me it's right to ignore them, as they are of the same order as the errors in the nonbasic duals. 

Otherwise, yes, I'm sure that there is room for refinement in the kObjectiveBound case as it's not been stress tested until now.

As for using primal simplex, I now see your logic. I'd never considered "abusing" the primal simplex solver by forcing it to shift large infeasibilities like this and work in phase 2, but I can see that it works!

lgottwald:
Okay, then I'll change the code to ignore the basic duals for now. I thought it might be good to include them by the same logic as when the primal infeasibility proof is computed. The correspondence would be to treat the objective function just like a row and the dual solution as row_ep vector where the objective row gets weight -1.0 and as upper bound the objective bound.

jajhall:
In deciding what code changes to look at, can you confirm that the objbound-cleanup branch is redundant? I'm puzzled, because this is where the primal and dual simplex solvers can be forced externally to use phase 2, which I thought was part of your strategy for making the dualUB bailout more reliable.

lgottwald:
The objbound-cleanup is not redundant, but I am unsure about whether the changes there are all good and I think it probably would make more sense to do cleanup of dual infeasibilities within HEkkDual just as it is done in the case when there model status is optimal and the cost perturbations are removed.

This branch adresses a part of the problem from a different angle by avoiding to have a dualUB bailout in the first place when the kObjectiveBound is not reached in the unperturbed space.

The objbound-cleanup I kept around for you as a reference as it resolves the dual infeasible points returned on beasleyC3 but is a bit hacky and I don't fully understand all consequences of the changes.

lgottwald:
Also this branch may make the changes in objbound-cleanup redundant for most cases, because given that the duals of basic columns are zero it is now guaranteed that the basis can be made dual feasible with just bound flips when the status is set to kObjectiveBound and the dual solution after bound flips should prove that the objective bound is reached.

So what should happen now in this branch is, that the cleanup call in the unscaled space will start with a possibly dual infeasible basis when the status is kObjectiveBound but does not use cost perturbation. Hence it should go to phase 2 directly having force_phase_2 = false and perform the necessary flips. After that it will determine again whether the dualUB bailout is really reached.

The case where the primal cleanup would possibly help is when there is a non-boxed variable for which the perturbation flipped the dual signs. In that case the exact dual objective could be repeatedly computed as -kHighsInf and there are unnecessary calls to compute it every few iterations which always fail and still solve the LP to optimality. In such a case it would likely be better to first attain dual feasibility by primal phase 2 cleanup and then proceed with dual simplex without cost perturbation until the objective bound is reached.

lgottwald:
```
Failing highsAssert("HEkk::computeFactor: lpFactorRowCompatible")
assert(HEkk::computeFactor: lpFactorRowCompatible) failed ...
```
Now happens quickly on this branch on the instance milo-v12-6-r2-40-1

jajhall:
Noted. I'm having to deal with some other work business at the moment, and have a meeting at 15:00. Until what you're doing is stable, if I have time for HiGHS I've got another little job to get on with.

lgottwald:
That's alright. The changes in this branch I would consider stable after some testing now. In the process I discovered a few minor anomalies in the MIP solver that I also changed here. Also this pull request goes into your correct-infeasibility-claim branch and not into master.

jajhall:
I've made progress, identifying that if phase 2 is forced, I only have to consider primal (dual) infeasibilities when deciding whether to use cost (bound) perturbation in dual (primal) simplex.

I also don't see why the unscaled LP needs to be solved after the scaled model has identified a model status of HighsModelStatus::kObjectiveBound and there are no unscaled dual infeasibilities. 

I've also changed the primal infeasibility calculations in HEkkDualRHS.cpp so that they are done the same way as in the rest of the simplex solvers. I consider whether value < lower-tolerance rather then lower-value > tolerance. When inconsistent there was an infeasibility in HEkkDualRHS.cpp that was 1e-6+eps, but 1e-6 elsewhere, leading to an LP with a feasible solution having a valid infeasibility proof corresponding to the row of the spurious infeasibility. Now rocI-4-11 runs to an assert in the MIP solver (after 360s). 

What I was wanting to look at all along was a case where primal simplex is called after scaled_model_status == HighsModelStatus::kObjectiveBound, and one occurs for `bin/highs --random_seed=1 ../check/instances/dcmulti.mps`

jajhall:
So, I've followed through an example when primal iterations were required to clean up dual infeasibilities in the unscaled LP, and it looks fine. It cleaned up the dual infeasibilities, then dual phase 2 was called because there was a significant primal infeasibility. Interestingly, it moved to a primal feasible point, but spotted that the objective bound was satisfied before trying the next simplex iteration and identifying optimality.

There's still an assert to catch an example where there are dual infeasibilities when computing the exact dual objective, but that shouldn't prevent this PR.

jajhall (approved):
All looks good, and my refinements to prevent bound/cost perturbation when forcing phase 2 are simple.

PR #646, opened 2021-12-22
Fix master qp mods conflicts

jajhall:
This resolves the horrible merge conflicts in #629 and corrects one error in Highs.cpp that I spotted

### DIFF
--- a/check/TestQpSolver.cpp
+++ b/check/TestQpSolver.cpp
@@ -4,7 +4,7 @@
 #include "Highs.h"
 #include "catch.hpp"
 
-const bool dev_run = false;
+const bool dev_run = true;
 const double double_equal_tolerance = 1e-5;
 
 TEST_CASE("qpsolver", "[qpsolver]") {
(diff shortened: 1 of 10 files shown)


jajhall:
Little to review, but I've only run it through a few unit/instance tests, so please check that your non-trivial use of the QP solver is still OK through the changed code.

feldmeier:
Thanks! I'll run some tests tomorrow

feldmeier:
Thanks again. I had to resolve a couple more conflicts (e.g. when calling readMps). tests and instances run fine. I merged this into master-qp-mods, and will create a new merge request to get all the changes into master at once.

Interestingly, the behaviour (as in number of iterations) changed on some instances with this merge.

jajhall:
Good. I guess that changes to the simplex solver mean that you're getting a different initial feasible solution sometimes. 
[Seasons Greetings 2021.pdf](https://github.com/ERGO-Code/HiGHS/files/7776291/Seasons.Greetings.2021.pdf)

PR #647, opened 2021-12-24
add option to set a limit on the gap for terminating the MIP solver w…

lgottwald:
…hen it is reached


DillonJ:
This is great! Thanks @lgottwald!

jajhall (approved):
Fine - and a happy user!

PR #651, opened 2021-12-26
WIP; improve behavior of primal heuristics

lgottwald:
- small changes to how heuristics are called in the beginning of the tree search to safeguard against spending too much time in heuristics
- fix cases where not all LP iterations spent in heuristics are counted in the statistics
- do not restrict the heuristic search space by removing symmetric solutions prior to creating a sub-MIP


jajhall (approved):
I trust you!

PR #652, opened 2021-12-27
Parallel compiler warnings

mckib2:
Context: for SciPy, there is a push to turn on the `-Werror` flag when switching to our new build system (meson).  HiGHS currently has just a few warnings that might have upstream solutions.

Here is an attempt at solving the current build warnings for `-Wclass-memaccess` we see using GCC.  The issue is that lambdas are not necessarily trivially copyable, so I understand that the current code using unions and memcpy to avoid heap allocations may be undefined behavior and vary compiler to compiler (or at the very least, it seems like it's abusing union/struct packing).

This PR's patch is obviously not the solution as we don't want to bump minimum compiler support to only those implementing C++17 standards, but it's the best I could come up with in an hour or so and demonstrates the kind of solution that is hopefully possible.  This seems to be an application that `std::launder` was made for, so I wonder if there will ever be a clean solution without C++17 compiler support.  The other thing to do would just add the `-Wno-class-memaccess` flag for these sources.

cc @rgommers 

EDIT: well the CI results maybe suggest my solution wasn't as good as I thought it was :)

### DIFF
--- a/src/parallel/HighsTask.h
+++ b/src/parallel/HighsTask.h
@@ -63,7 +63,7 @@ class HighsTask {
     static void do_call(const char* functorBytes) {
       // declaring F f would fail as type F is not default constructible
       FunctorSpace s{};
-      std::memcpy(&s.f, functorBytes, sizeof(F));
+      std::memcpy(reinterpret_cast<char*>(&s.f), functorBytes, sizeof(F));
       s.f();
     }
   };


lgottwald:
But the compiler warning is bogus since there is a static_assert that assures that the lambda is trivially copyable. If there is ever a lambda created that captures a non-trivially copyable type then the compilation will fail due to a static assert.

The solution with the memcpy is what I added to avoid a different solution that caused a strict aliasing warning of gcc. The issue there was that I used a base class with a virtual function to call a stored lambda. Since the base class was created in the storage space provided for the task, there is an aliasing violation that I could not avoid. It is apparently not guaranteed that the pointer returned by placement new points to the beginning of the memory and simply casting the pointer to the memory storing the inherited class instance into a pointer to the base class type to call the virtual function is not allowed. This solution with the memcopy avoids this at the expense of requiring that the lambda is trivially copyable, which is done with a static assert.

I can see if I can find a different solution as it seems the static assert failed for one user, but I would strongly advise to not compile code from other projects with the Werror flag. We can never guarantee that there is no warning with new compiler versions or with compilers we have not tested.

rgommers:
> can see if I can find a different solution as it seems the static assert failed for one user, but I would strongly advise to not compile code from other projects with the Werror flag. We can never guarantee that there is no warning with new compiler versions or with compilers we have not tested.

We understand this - the `-Werror` is never added by default in source releases or the master branch of our Git repo, only in a single CI job. That's actually quite valuable to guard against more warnings (some of which important) creeping in over time.

lgottwald:
Can you try instead of this solution to do a cast of the memcpy argument to `char*`? I found some post on stackoverflow that indicates that this might lift the warning and it is standard compilant too. I.e. change
```
std::memcpy(&s.f, functorBytes, sizeof(F));
```

to 

```
std::memcpy(reinterpret_cast<char*>(&s.f), functorBytes, sizeof(F));
```

Please let me know if that fixes the warning. Besides, your current solution is not standard compliant as the class type of the functor might have different alignment rules. By using the union type with the dummy member it is avoided to need to call the constructor while at the same time the stack space for storing the functor will have the proper alignment guarateed.

mckib2:
> Please let me know if that fixes the warning. Besides, your current solution is not standard compliant as the class type of the functor might have different alignment rules. By using the union type with the dummy member it is avoided to need to call the constructor while at the same time the stack space for storing the functor will have the proper alignment guarateed.

That did indeed silence the warning on my machine, running through CI now!

I tend to be a little stronger than Ralf on usage of Werror -- I would prefer to have it always on in all CI jobs to catch warnings as they are introduced (and if annoying, simple to turn off locally when developing).  HiGHS doesn't have this problem as it's a newer, well-maintained code base, so you may feel differently, but large legacy codebases are difficult and expensive to resolve all warnings and if you can do it once, you never want to do it again.  Turning on Werror is a good way to enforce good behavior and I have found effective for scaling maintaining burden.

lgottwald (approved):

PR #657, opened 2022-01-03
make sure to call objective bound refinement when status first appear…

lgottwald:
…s in unscaled LP

Just a minor change handling an edge case I once observed to happen. When the kObjectiveBound status in rare cases first appears during the refinement call then simplex can again terminate with dual infeasibilities due to perturbation. With the change it is checked again if the status after the refinement LP is kObjectiveBound for the first time and solve is called again with primal simplex in case of dual infeasibilities.


lgottwald:
this also goes into the branch correct-infeasibility-claim which I will do a bit more testing on

jajhall (approved):
Sure. I see that this can happen, and that this modification fixes it

PR #658, opened 2022-01-03
Correct infeasiblilty claim

lgottwald:


lgottwald:
FYI: I added a bugfix for the bad basis change code you added. There was a possibly incorrect bad_basis_change_num being used which could lead to infinite looping.

lgottwald:
There where two bugs related to the bad basis change code. The first one used a wrong basis change number so that it could happen that the wrong record was updated to have taboo=true when it already existed. The second bug did not update the row_out value so that the same variable_out could be repeatedly be selected.

Both could lead to an infinite loop repeatedly selecting the same bad basis change.

Now there only seems to left to fix the HFactor assert that fails on glass4. With random_seed = 1 it fails after just a few seconds.

lgottwald:
I found the issue with the HFactor and now fixed it properly. The factor compatibility in Highs.cpp is now always called with the unpresolved LP's number of rows. Previously the function would sometimes return true when the original LP was not solved in cases where no postsolve was called.

lgottwald:
The setting of row out is then not needed at that place, but in HEkk::isBadBasisChange the list is scanned without the check for row_out, so there the change is important.

jajhall:
Sure. This taboo business is getting very complicated!

lgottwald:
Indeed, I only found out why simplex was looping by stepping through it in the debugger. But now these intricate bugs should be gone and the taboo basis change records are maintained as intended. The fix could of course also change the loop in isBadBasisChange to also check the row_out value instead so that the basis change becomes allowed again when it uses a different row_out. Do you think this is better?

jajhall:
We've introduced something that leads to a unit test failing. Let me investigate

lgottwald:
Indeed that seems to be better. There are much fewer cases where the LP is solved to undetermined status.

lgottwald:
I think I fixed the unit_test already. I was clearing the ekk_instance too often but now adapted the check for row compatibility to use the unpresolved number of rows in Highs.cpp which only clears when necessary.

jajhall:
> Indeed that seems to be better. There are much fewer cases where the LP is solved to undetermined status.

OK I was being cautious in avoiding a particular variable exchange to avoid cycling, regardless of the row

jajhall:
> I think I fixed the unit_test already. I was clearing the ekk_instance too often but now adapted the check for row compatibility to use the unpresolved number of rows in Highs.cpp which only clears when necessary.

OK, but please run the unit tests before pushing a commit to Git

lgottwald:
Hm, you might be right for the case of cycling. The cases where I did the change is when cycling is not detected, I believe, since the check on the hash values was not indicating cycling but the basis change is marked as bad. Maybe we leave it to be less restrictive for now and I'll watch out if any cycling that simplex does not break out of seems to occur on any instance.

jajhall:
Indeed, your cycling detection was creating many false positives, which his why I only recognise cycling if it's detected in consecutive iterations. That said, many of the false positives were linked to basis changes that occurred when bound flips etc were being performed unnecessarily due to small dual infeasibilities after cost perturbation and then being reversed. Of course, we're avoiding all this now

lgottwald:
Yes, things seem to be way more stable now. Except for the prints I don't see any anomalous behavior in this branch anymore on the MIP's I tested, so from my side this is good to be merged now. I just pushed one more change to MIP that now simply trusts the status kObjectiveBound of the LP solver when it is returned just as is done for kInfeasible.

jajhall:
> Yes, things seem to be way more stable now. Except for the prints I don't see any anomalous behavior in this branch anymore on the MIP's I tested, so from my side this is good to be merged now. I just pushed one more change to MIP that now simply trusts the status kObjectiveBound of the LP solver when it is returned just as is done for kInfeasible.

That's good, but I'm conscious of not having run any extensive LP tests for a while. In particular, I was reminded yesterday that there are 99 simplex iterations after postsolve for `shell` - not that this is a recent phenomenon. It's an LP that's very easy numerically, so I don't understand why postsolve isn't exact.

lgottwald:
The assert in HEkkDual should not be triggered, I think, because the infeasibilities are not visible when cost perturbation is present. When the cost perturbation is removed as it happens for the dual values that are computed in the function itself without cost perturbation, then dual infeasibilities may come up. The function now simply uses the bound value that is indicated by the dual value sign of a column to compute the dual objective. So it implicitly may perform bound flips or conclude that the objective bound is not yet reached as there is a nonzero dual with wrong sign on an unbounded column.

Now that the primal_simplex cleanup is performed, it might be sensible to adapt this so that instead of saying that the dual UB bailout is not yet reached primal cleanup is performed first. Ideally that would happen within HEkkDual and be moved out of HApp, but for now we can leave things as they are as it seems to work and already improves over what was there before.

For the postsolve iterations I can look into what happens, especially when the LP is simple this should be easy to find out with the debug code I added to presolve. Also now that there is the force_phase2 flag I think this should also be used for solving the LP from the postsolved basis. When there are dual infeasibilities after postsolve that are slightly larger than the squared dual feasibility tolerance dual simplex will start flipping bounds again and destroy the basis recovered from postsolve.

lgottwald:
The postsolve iterations on shell are due to the removal of dependent equations. It seems the code determines non-redundant equations as redundant which leads to the primal infeasibilities in the original space.

lgottwald:
I forgot to return true in isBadBasisChange for one of the case distrinctions which lead to some looping again. Hopefully things are fixed now. Will test again on the instances that are prone to LP troubles.

lgottwald:
I added an assert in HFactor that fails on this branch on instance umts and would otherwise cause a segfault.
```
      0       0         0   0.00%   29991111.29384  36647756          18.16%      306     32    639      1311     3.0s
highs: src/util/HFactor.cpp:588: void HFactor::buildSimple(): Assertion `BcountX < b_index.size()' failed.
Aborted (core dumped)
```
I think I also fixed it but once you are back at work you should review the change in HFactor. The problem occured when an alien basis with many more basic variables than rows was factorized. I think there was an error in the calculation of `basis_matrix_limit_size` where the number of rows was added when instead the max of number of rows and number of basic variables needed to be added.

jajhall:
Quite possible that I've missed one of these when allowing a rectangular basis

jajhall:
> The postsolve iterations on shell are due to the removal of dependent equations. It seems the code determines non-redundant equations as redundant which leads to the primal infeasibilities in the original space.

I've spotted what causes it. Quite an oversight on my part, and a non-trivial fix, but it should be entertaining!

jajhall (approved):
I think this is ready to be merged into master. Today I cleaned up some of the "example-seeking" asserts and use of printf in logging. Over the weekend, I fixed the issue with shell identifying redundant equations incorrectly. It's been hard to fully test all that I've done, as one particular case doesn't occur with all of the LPs in my test set. I have managed to create a toy example to verify the code's correctness.

lgottwald:
Great. I added one more bugfix discovered in another branch and will merge once the test passes.

PR #666, opened 2022-01-12
revise MIPGap options as tolerances

lgottwald:
This removes the `mip_gap_limit` option and introduces the options `mip_rel_gap` and  `mip_abs_gap`.

The new options are not handled like a limit where the solver just stops but rather as a tolerance to determine optimality. That means the solver does not check the whether the gap reaches the desired tolerance, but rather uses the specified tolerances to adjust the limit on the upper bound for pruning nodes during the search. Hence, when a node's subproblem cannot contain a solution that is improving regarding to the tolerances it is simply dropped.

This has the effect that HiGHS will terminate with 0% gap when the status is optimal and always report the lower bound as equal to the upper bound. The commercials seem to determine the lower bound value differently. Particularly they report the actual minimal LP objective as computed for the subproblems it seems. In HiGHS the lower bound is computed to be the lowest LP objective of an unprocessed node. Since always all nodes are processed when optimality is determined the lower bound will be equal to the upper bound.

Making sure that the lower bound is reported as in the commercials would require a bit more changes, so I think I will rather modify the solving report to something like this:
```
Solving report
  Status            Optimal
  Primal bound      386522.173195
  Dual bound        386522.173195 (tolerance: 19326.1086)
  Gap               0.00% (tolerance: 5.00%)
  Solution status   feasible
                    386522.173195 (objective)
                    9.26547727431e-11 (bound viol.)
                    3.73479880966e-15 (int. viol.)
                    4.02366140406e-10 (row viol.)
  Timing            41.23 (total)
                    0.17 (presolve)
                    0.00 (postsolve)
  Nodes             1
  LP iterations     27922 (total)
                    0 (strong br.)
                    859 (separation)
                    11031 (heuristics)
```

Then it is clear that the reported lower bound and gap values are subject to the specified tolerances.

### DIFF
--- a/check/CMakeLists.txt
+++ b/check/CMakeLists.txt
@@ -11,6 +11,10 @@ target_include_directories(Catch INTERFACE ${HIGHS_SOURCE_DIR}/src/ipm/basiclu/s
 target_include_directories(Catch INTERFACE ${HIGHS_SOURCE_DIR}/src/ipm/ipx/include)
 target_include_directories(Catch INTERFACE ${HIGHS_SOURCE_DIR}/src/ipm/basiclu/include)
 
+FILE(WRITE ${CMAKE_BINARY_DIR}/testoptions.txt
+"mip_rel_gap=0.0
+mip_abs_gap=0.0")
+
 # Make test executable
 set(TEST_SOURCES
     TestAlienBasis.cpp
@@ -282,7 +286,7 @@ foreach(instance ${mipInstances})
     set(inst "${HIGHS_SOURCE_DIR}/check/instances/${name}.mps")
 
     foreach(setting ${settings})
-        add_test(NAME ${name}${setting} COMMAND $<TARGET_FILE:highs> ${setting}
+        add_test(NAME ${name}${setting} COMMAND $<TARGET_FILE:highs> ${setting} --options_file ${CMAKE_BINARY_DIR}/testoptions.txt
               ${inst})
 
         set_tests_properties (${name}${setting} PROPERTIES
--- a/check/TestAlienBasis.cpp
+++ b/check/TestAlienBasis.cpp
@@ -206,6 +206,10 @@ TEST_CASE("AlienBasis-rectangular-completion", "[highs_test_alien_basis]") {
   double solution_error = 0;
   for (HighsInt iRow = 0; iRow < num_row; iRow++)
     solution_error += std::abs(rhs.array[iRow] - solution[iRow]);
+  printf("AlienBasis-rectangular-completion: solution_error = %g\n",
+         solution_error);
+  fflush(stdout);
+
   REQUIRE(solution_error < 1e-8);
 }
 
--- a/check/TestMipSolver.cpp
+++ b/check/TestMipSolver.cpp
@@ -99,7 +99,7 @@ TEST_CASE("MIP-integrality", "[highs_test_mip_solver]") {
   REQUIRE(info.objective_function_value == optimal_objective);
 
   REQUIRE(info.mip_node_count == 1);
-  REQUIRE(info.mip_dual_bound == -6);
+  REQUIRE(fabs(info.mip_dual_bound + 6) < double_equal_tolerance);
   REQUIRE(info.mip_gap == 0);
 }
 
--- a/check/TestMipSolver.cpp
+++ b/check/TestMipSolver.cpp
@@ -100,7 +100,7 @@ TEST_CASE("MIP-integrality", "[highs_test_mip_solver]") {
 
   REQUIRE(info.mip_node_count == 1);
   REQUIRE(fabs(info.mip_dual_bound + 6) < double_equal_tolerance);
-  REQUIRE(info.mip_gap == 0);
+  REQUIRE(std::fabs(info.mip_gap) < 1e-12);
 }
 
 TEST_CASE("MIP-nmck", "[highs_test_mip_solver]") {
--- a/src/lp_data/HighsOptions.h
+++ b/src/lp_data/HighsOptions.h
@@ -348,8 +348,9 @@ struct HighsOptionsStruct {
   HighsInt mip_pscost_minreliable;
   HighsInt mip_report_level;
   double mip_feasibility_tolerance;
+  double mip_rel_gap;
+  double mip_abs_gap;
   double mip_heuristic_effort;
-  double mip_gap_limit;
 #ifdef HIGHS_DEBUGSOL
   std::string mip_debug_solution_file;
 #endif
@@ -697,9 +698,17 @@ class HighsOptions : public HighsOptionsStruct {
     records.push_back(record_double);
 
     record_double = new OptionRecordDouble(
-        "mip_gap_limit",
-        "limit to stop the MIP solve when the gap is at most this value",
-        advanced, &mip_gap_limit, 0.0, 0.0, kHighsInf);
+        "mip_rel_gap",
+        "tolerance on relative gap, |ub-lb|/|ub|, to determine whether "
+        "optimality has been reached for a MIP instance",
+        advanced, &mip_rel_gap, 0.0, 1e-4, kHighsInf);
+    records.push_back(record_double);
+
+    record_double = new OptionRecordDouble(
+        "mip_abs_gap",
+        "tolerance on absolute gap of MIP, |ub-lb|, to determine whether "
+        "optimality has been reached for a MIP instance",
+        advanced, &mip_abs_gap, 0.0, 1e-6, kHighsInf);
     records.push_back(record_double);
 
     // Advanced options
--- a/src/mip/HighsMipSolver.cpp
+++ b/src/mip/HighsMipSolver.cpp
@@ -118,7 +118,9 @@ restart:
     bool considerHeuristics = true;
     while (true) {
       if (considerHeuristics && mipdata_->moreHeuristicsAllowed()) {
-        search.evaluateNode();
+        if (search.evaluateNode() == HighsSearch::NodeResult::kSubOptimal)
+          search.currentNodeToQueue(mipdata_->nodequeue);
+
         if (search.currentNodePruned()) {
           ++mipdata_->num_leaves;
           search.flushStatistics();
@@ -143,7 +145,9 @@ restart:
       if (mipdata_->domain.infeasible()) break;
 
       if (!search.currentNodePruned()) {
-        search.dive();
+        if (search.dive() == HighsSearch::NodeResult::kSubOptimal)
+          search.currentNodeToQueue(mipdata_->nodequeue);
+
         ++mipdata_->num_leaves;
 
         search.flushStatistics();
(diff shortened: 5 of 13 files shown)


lgottwald:
Also the solving report above is the improvement on the model in #660 which is a considerable improvement. Indeed when the gap is set to zero the lower bound  after almost an hour is `372744.337092` which means with `mip_rel_gap=0.05` it would still be correct when a solution with value `394671.0811` is declared as optimal. Indeed It happened in one test that HiGHS terminated with a solution value of `390228.3818` as optimal before I adapted the code in the primal heuristics to always use `mip_rel_gap=0.0`. This is important since for the main search we still want to be able to find solutions that improve the current solution by less then the specified tolerances, since then we can again tighten their objective value by the tolerances for pruning.

lgottwald:
The failing tests are likely due to using the default values `mip_rel_gap=1e-4` and `mip_abs_gap=1e-6` which are similar to the values Gurobi uses by default.

jajhall (approved):
This seems to be a sensible compromise between reflecting the behaviour of commercial solvers and being clear mathematically.

jajhall:
Wise for the instance tests to use use mip_rel_gap=0.0 and mip_abs_gap=0.0, but should have a unit test to check that positive values lead to consistent values of mip_gap.

lgottwald:
Yes, I think this needs some adjustment in the test result verification though. Also the solver only failed for dcmulti which had large enough primal bounds to get some inaccuracy with mip_rel_gap equal to 1e-4. Maybe we could add a single test for dcmulti to check that the reported primal_bound is smaller or equal to the optimal one divided by (1-mip_rel_gap)?

DillonJ:
This is really awesome!

```
Solving report
Status            Optimal
Primal bound      387466.331569
Dual bound        387466.331569 (tolerance: 19373.3166)
Gap               0.00% (tolerance: 5.00%)
Solution status   feasible
387466.331569 (objective)
2.85353962681e-11 (bound viol.)
2.9721644363e-16 (int. viol.)
2.85353962681e-11 (row viol.)
Timing            92.56 (total)
0.23 (presolve)
0.00 (postsolve)
Nodes             1
LP iterations     32825 (total)
9112 (strong br.)
860 (separation)
12352 (heuristics)
```

Solving takes 91 seconds for me - but I have a suspicion that the conversion to MPS is resulting in us both seeing a slightly different problem - for example, when writing the mps file, I get a lot of warnings saying that variable x appears in bounds but not in the columns section and is ignored. I'm not sure if that accounts for the difference or if you can afford better hardware than me :-) 

Incidentally, the only HiGHS option I set is `mip_rel_gap`, so everything else will be default.

jajhall:
> Solving takes 91 seconds for me - but I have a suspicion that the conversion to MPS is resulting in us both seeing a slightly different problem - for example, when writing the mps file, I get a lot of warnings saying that variable x appears in bounds but not in the columns section and is ignored. I'm not sure if that accounts for the difference or if you can afford better hardware than me :-)

I'm puzzled: these sorts of warnings appear in HiGHS when _reading_ an MPS file, not writing it. 

Please would you send me the MPS file, too?

lgottwald:
The hardware I have is indeed quite fast. I think I had no warnings when reading the MPS though

DillonJ:
> 
> I'm puzzled: these sorts of warnings appear in HiGHS when _reading_ an MPS file, not writing it.
> 
Actually, I should point out that I am using Julia JuMP's `write_to_file()` to write the mps file from Julia - which is also how I'm interacting with HiGHS

jajhall:
> > I'm puzzled: these sorts of warnings appear in HiGHS when _reading_ an MPS file, not writing it.
> 
> Actually, I should point out that I am using Julia JuMP's `write_to_file()` to write the mps file from Julia. I am also using Jump to interact with HiGHS (within the Spine/SpineOpt framework).

Ah, I see. It sounds as if you've got variables with zero cost and zero constraint matrix column, but finite bounds. Obviously these don't affect the solution or optimal objective value and, if they are ignored, this explains why @lgottwald is not seeing the warnings when reading the MPS file. If they weren't ignored, HiGHS would put out warnings when encountering them in the MPS file.

jajhall:
> > Please would you send me the MPS file, too?
> 
> Sure, I sent it to [highsopt@gmail.com](mailto:highsopt@gmail.com). Please let me know if I should send it somewhere else.

Thanks. Although we've resolved the "warnings" issues, it's always good to get more models. If it makes you feel better, it took 82 seconds to solve it on my 3-y-o machine - @lgottwald has the best machine on the team! - and Gurobi took 7s, so we've still a way to go!

DillonJ:
Interestingly, I just ran highs.exe directly on the mps file and I get:

```
Solving report
  Status            Optimal
  Primal bound      386522.173195
  Dual bound        386522.173195 (tolerance: 19326.1087)
  Gap               0.00% (tolerance: 5.00%)
  Solution status   feasible
                    386522.173195 (objective)
                    9.26547727431e-11 (bound viol.)
                    3.73479880966e-15 (int. viol.)
                    4.02366140406e-10 (row viol.)
  Timing            74.88 (total)
                    0.26 (presolve)
                    0.00 (postsolve)
  Nodes             1
  LP iterations     27922 (total)
                    0 (strong br.)
                    859 (separation)
                    11031 (heuristics)
```

Which replicates your results... but when I run in Julia, I get:

```
Solving report
Status            Optimal
Primal bound      387466.331569
Dual bound        387466.331569 (tolerance: 19373.3166)
Gap               0.00% (tolerance: 5.00%)
Solution status   feasible
387466.331569 (objective)
2.85353962681e-11 (bound viol.)
2.9721644363e-16 (int. viol.)
2.85353962681e-11 (row viol.)
Timing            92.56 (total)
0.23 (presolve)
0.00 (postsolve)
Nodes             1
LP iterations     32825 (total)
9112 (strong br.)
860 (separation)
12352 (heuristics)
```
You will notice that the solve time is 20 seconds longer and the solution is not as good. One difference between the two is that the mps writer is discarding the redundant variables whereas the problem sent by Julia will have them still in it. Does this indicate potential room for improvement in the presolver? Perhaps there are other differences though.

jajhall:
> You will notice that the solve time is 20 seconds longer and the solution is not as good. One difference between the two is that the mps writer is discarding the extraneous variables whereas the problem sent by Julia will have them still in it. Does this indicate potential room for improvement in the presolver? Perhaps there are other differences though.

The "better solution" is an artefact of using a positive value of `rel_mip_gap`. The presolve will remove the extraneous variables almost instantaneously, so that's not an issue.

DillonJ:
> > You will notice that the solve time is 20 seconds longer and the solution is not as good. One difference between the two is that the mps writer is discarding the extraneous variables whereas the problem sent by Julia will have them still in it. Does this indicate potential room for improvement in the presolver? Perhaps there are other differences though.
> 
> The "better solution" is an artefact of using a positive value of `rel_mip_gap`. The presolve will remove the extraneous variables almost instantaneously, so that's not an issue.

But both are using the same value of mip_rel_gap - in theory the problem should be the same, the only difference is  
  Julia -> MOI model -> HiGHS   
vs  
  Julia -> MOI model -> mps file -> HiGHS

jajhall:
> But both are using the same value of mip_rel_gap - in theory the problem should be the same, the only difference is
> Julia -> MOI model -> HiGHS
> vs
> Julia -> MOI model -> mps file -> HiGHS

Mathematically, yes, but the MPS file introduces modifications to the problem data. For example you have a cost value of 0.000864769 in the MPS file. Maybe the model has this exact value, but I suspect that it's a truncation. In the matrix, you have 

image

Although most of your matrix coefficients are +1 or -1, I'm sure you've got plenty that are truncated. All these perturbations can lead to radically different behaviour when the simplex implementation solves LP subproblems since they affect how ties are broken. It'll get the right objective value, but maybe a very different solution (since LPs typically have non-unique solutions). All this will lead to the MIP solver taking very different actions.

lgottwald:
When the model is formulated slightly differently, even if it is just randomly permuted but the mathematically same model, then it may still be that the solver goes on a different solution path. With the mip_rel_gap being 0.05 the primal solution that is determined as optimal can lie within 5% of the optimal objective value which is at best around ~373000-374000 as I know from running this particular model for some time with mip_rel_gap=0.0. Hence an objective value up to ~393000 lies still within 5% gap to the optimal solution and when the solver goes on a different solution path, e.g. due to permuting the problem or changing the random_seed option, then a different solution value may be returned and also the solver may be more or less lucky and solve the model slower or faster. It is not unusual that the difference is considerable and solving time sometimes can even be of a different magnitude. This is also an effect that is observable on other solvers and also commercial ones. It was named as performance variability, I believe first in the literature for MIP in this talk:
https://coral.ise.lehigh.edu/mip-2008/talks/danna.pdf

lgottwald:
Regarding potential for improvement there is still a lot for this particular instance as I can see from the solver behavior. Particularly the heuristic limits could be adjusted as the root heuristics take a bit much time (not easily possible via options right now) and strong branching effort could be reduced, e.g. by setting the option mip_pscost_minreliable to 0.

jajhall:
Strange source of error: I'll investigate

lgottwald:
An alternative could be to revert to the previous solution, but display the lower bound as the tightened upper bound at most. Then HiGHS will terminate faster but always with the maximal possible gap being reported.

jajhall:
> It is still work in progress. I can also say that performance with this solution is worse than the previous one but lower bound and gap is printed accurately. The reason for this is that for getting an accurate lower bound it is important to not use the tightened upper bound for things like reduced cost fixing and conflict analysis. Highs now simply stores the nodes within a separate part of the nodequeue so that the search terminates as optimal when the regular part of the queue is empty but the nodes are still kept as open nodes to get an accurate lower bound.

Thanks, but one failure was in TestAlienBasis - even in Linux! I didn't reproduce this, but have added a print statement. The error I did get was for 
REQUIRE(info.mip_gap) == 0);
and this is now
REQUIRE(std::fabs(info.mip_gap) < 1e-12);

jajhall:
> It is still work in progress. I can also say that performance with this solution is worse than the previous one but lower bound and gap is printed accurately.

Maybe the old version could be the default, when  a "lower bound analysis" option is false, but the new version can be obtained if it's set true. Forget this if it's too complicated

lgottwald:
I think the new solution has gotten a bit too complicated an also is slower. Also the lower bound will often be just at the gap limit. I think for now I will revert to the old solution but adapt the display of the lower bound to be at most the adjusted upper bound. ALso because performance differs so much.

jajhall:
That's OK. We'll just have to let users know that it's all in the interests of performance, and document what is being reported.

lgottwald:
I am a bit undecided right now and will do a bit more testing. Performance difference is not super clear on different instances. Also it is hard to compare since it depends on which solution is found when.

lgottwald:
I will keep using the newer solution. Internally it adds a lot of ground work to allow reoptimization when only the objective function (or the limits on the objective like mip_*_gap) have changed. This feature will come almost naturally when I am finished with making the MIP solver reentrant. It also may be a very useful feature for use within heuristics as it allows to call them with initially a large mip_*_gap value and then adapt it to become smaller and smaller as long as the heuristic finds improving solutions.
Performance is also not so much worse than with the other solution and depends more on finding a good enough solution instead of which implementation is used. Since this implementation will help heuristics to become better it is likely that it closes any performance gap in the future. Also, when reoptimization is possible HiGHS becomes a nice advantage when it comes to uses within other solvers. E.g. in a column generation approach HiGHS can solve the pricing problems using reoptimization after only changing the objective or when it is used to solve MIP relaxations of an MINLP it may be useful too.

PR #668, opened 2022-01-12
Python Interface

michaelbynum:
This PR starts a Python wrapper for the `Highs()` class using pybind11. Not all functionality is there yet, but it is a reasonable start.

I tried putting the `highspy` directory in `src/interfaces`, but I could not get the license into the wheel that way. I'm sure it is possible, but I couldn't figure it out.

The installation requires the `find_library` function from Pyomo, which is not ideal, but the `ctypes.util.find_library` function does not provide the path to the library on linux. I can try to reimplement a simplified version if needed.


jajhall (approved):
Great!

PR #673, opened 2022-01-15
after restart resolve first LP without cutting planes

lgottwald:
Seemed to have some advantages and also allows the previously added cutting planes in the LP are cleaned up again by selecting only those that are nonbasic after adding them to the LP without cuts at once and resolve. When restarting from a different node in the tree the basis is also not optimal for the restarted LP even if the basis transferred perfectly through presolve.

Will also do some performance checks first before merging.


lgottwald:
Performance here is also ok, this can also be merged

lgottwald:
also contains the other changes already

PR #674, opened 2022-01-16
Add a guard to avoid defining `__STDC_FORMAT_MACROS` twice

rgommers:
This resulted in a (harmless but scary-looking) build warning in SciPy, because we build a Python extension with both HiGHS
and NumPy in it, and NumPy headers define this macro with a value of `1`. Both defines are common and it doesn't really matter, but the compiler sees this as a redefinition.

### DIFF
--- a/src/util/HighsInt.h
+++ b/src/util/HighsInt.h
@@ -20,7 +20,9 @@
 #include <stdint.h>
 
 #ifdef __cplusplus
+#ifndef __STDC_FORMAT_MACROS
 #define __STDC_FORMAT_MACROS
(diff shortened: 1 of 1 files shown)


jajhall (commented):
This looks utterly uncontroversial

lgottwald:
Yes, can be merged, although the warning is weird. I would not expect `#include <stdint.h>` to bring in said macro definition but guarding against it certainly does not hurt and if it silences a warning that's reason enough.

lgottwald (approved):

Issue #669, opened 2022-01-14
Possible bug in presolver when setting zero coefficients


chrhansk:
I think that there is a bug in the HiGHS presolver. The behavior is triggered by [this](https://gist.github.com/chrhansk/8ec7e471c9d58c1eb02796c380c0daca) example, which yields the following error message:

    ERROR:   LP matrix packed vector 0, entry 2, is duplicate index 2
    highs: /src/HiGHS/src/lp_data/Highs.cpp:799: HighsStatus Highs::run(): Assertion `assessLp(reduced_lp, options_) == HighsStatus::kOk' failed.

The problem disappears when I remove all calls to `Highs_changeCoeff(highs, i, j, 0.000000);`, i.e., those with zero coefficient values. This may be a division by zero (or just a coincidence).

jajhall:
Thanks for this. I've reproduced the behaviour that you report and am investigating.

jajhall:
Setting explicit zero coefficients in the matrix is not permitted, and any zero (or small) entries in the matrix when Highs_passLp is called are removed (with a warning message). However, you introduce them via Highs_changeCoeff, and I'd forgotten to check the value in this method. I guess that the presolve assumes that there are no zero coefficients in the matrix, and this leads to a matrix with duplicate indices (highly illegal!) being passed to the solver.

chrhansk:
First of all, thank you for the quick reply.

I am using HiGHS to solve relaxations of nonlinear problems. These relaxations of course depend on the primal point, I solve them in sequence using the HiGHS C API using a single instance and updating the problem using calls to `Highs_changeCoeff`, `Highs_changeColCost` etc. 

Based on a series of problems it can always happen that non-zero coefficients must be zeroed in a subsequent iteration. If I cannot change a coefficient to zero by using `Highs_changeCoeff`, is there another function in the C API to zero out coefficients, or alternatively to clear all coefficients and to start from scratch? Or should I rather create a new HiGHS instance for each problem? Since the problems are closely related as the algorithm converges, the bases will likely coincide and the solutions will be closely related.

PR #677, opened 2022-01-18
Fix 669

jajhall:
Two issues here. 

In reproducing the behaviour in #669 I gave lp_.a_matrix_.start_ a size of lp_.num_col_+2, triggering an assert in HPresolve.cpp:1946. It can't be illegal for any model vector to have greater size than necessary. This behaviour can be reproduced by setting issue669a = true in line 229 of TestSpecialLps.cpp

The other issue comes from not checking that |value| is sufficiently large in Highs::changeCoeff. This is fixed

### DIFF
--- a/check/SpecialLps.h
+++ b/check/SpecialLps.h
@@ -162,6 +162,28 @@ class SpecialLps {
     require_model_status = HighsModelStatus::kInfeasible;
   }
 
+  void issue669Lp(HighsLp& lp, HighsModelStatus& require_model_status) {
+    lp.model_name_ = "issue669";
+    lp.num_col_ = 27;
+    lp.num_row_ = 9;
+    vector<double> zero_vector_col;
+    vector<double> zero_vector_row;
+    vector<HighsInt> zero_vector_start;
+    for (int iCol = 0; iCol < lp.num_col_; iCol++) zero_vector_col.push_back(0);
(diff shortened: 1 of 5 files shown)


lgottwald:
Why should it not be illegal to have a wrong size for the nonzero vector? The model is in packed format and reading the number of nonzeros from the vector size is the easiest way and there is not reason to not resize the vector properly wherever it is created. The vector is already a dynamic buffer and the internal buffer size is decoupled from the logical size. Resizing it comes for free and having a wrong size seems to me like a logical error. If it was an array that needed reallocation that would be different but the vectors internal buffer is not touched from calling resize to a smaller size. It is more like having a field in the class that is called numNnz and then set it to a larger value.

jajhall:
To me, it's sufficient for a user to supply correct LP data, and allocating more space than is necessary isn't a problem. 
The number of nonzeros in the matrix is defined as start_(num_col), if defined column-wise, and start_(num_row), if defined row-wise, not the size of index_/value_.
If you depend on the size of index_/value_ being equal to the number of nonzeros in the matrix, then I can resize them, but if we require this of users, we're sure to get complaints if we return an error!

jajhall:
The fix to Highs::changeCoeff needs to zero any nonzero entry if a new small value is passed

lgottwald:
Thanks! And yes, I agree, this is not worth breaking the interface over.

lgottwald (approved):

PR #681, opened 2022-01-20
Fix uninitialized warnings

Smit-create:
Warnings:
```
[1474/1557] Compiling C++ object scipy/optimize/_highs/libhighs.a.p/.._..__lib_highs_src_mip_HighsImplications.cpp.obj
../scipy/_lib/highs/src/mip/HighsImplications.cpp: In member function 'bool HighsImplications::runProbing(HighsInt, HighsInt&)':
../scipy/_lib/highs/src/mip/HighsImplications.cpp:167:46: warning: 'implicsdown' may be used uninitialized in this function [-Wmaybe-uninitialized]
  167 |       if (implicsup[u].column < implicsdown[d].column)
      |  
```
Related issue: https://github.com/rgommers/scipy/issues/120

### DIFF
--- a/src/mip/HighsImplications.cpp
+++ b/src/mip/HighsImplications.cpp
@@ -154,10 +154,10 @@ bool HighsImplications::runProbing(HighsInt col, HighsInt& numReductions) {
       return true;
 
     // analyze implications
-    const HighsDomainChange* implicsup;
-    const HighsDomainChange* implicsdown;
-    HighsInt nimplicsup;
-    HighsInt nimplicsdown;
+    const HighsDomainChange* implicsup = nullptr;
+    const HighsDomainChange* implicsdown = nullptr;
+    HighsInt nimplicsup = 0;
+    HighsInt nimplicsdown = 0;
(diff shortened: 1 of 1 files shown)


lgottwald:
On what compiler does this warning occur? I suspect some quite old gcc.

The warning is clearly false positive and I think the code is clearer and more readable without the initialization. The initialization makes it seems as if the behavior of the function `getImplications()` called right afterwards actually included a case where things are left uninitialized which is not ever the case. Without the initialization the code is self-documenting due to the method names and also correct.
gcc/clang compilers on ubuntu 20.04 LTS do not issue this warning, so for any remotely recent compiler the false positive has been fixed.

jajhall:
Given the comment above of @lgottwald we won't accept this PR

mckib2:
> On what compiler does this warning occur? I suspect some quite old gcc.

SciPy [toolchain roadmap](https://docs.scipy.org/doc/scipy/reference/dev/toolchain.html) says we support as old as GCC 4.8, but I believe this recently got bumped up to 5.x.

PR #685, opened 2022-01-22
Dse weight preservation

jajhall:
Maintains DSE weights to a great extent. Still some scope for more, but this is a big step.
Ctest passes, and expensive debugging code has been removed, so worth trying this branch on test MIPs


lgottwald:
That's great! I'll run tests overnight

jajhall:
I'll also want to look through to see whether there are still some places where I switch to Devex unnecessarily

lgottwald:
I will send you the logs tomorrow then

lgottwald (approved):

lgottwald:
Do you still want to do more changes to this or shall we merge to master for now?

jajhall:
I've already spotted one expensive but infrequent check that Ieft in, and am doing some tidying up of redundant code. Plus I've had a thought about how to identify whether I'm missing any situations where DSE are lost unnecessarily, so I'll have a look for these today. Depending on how much I change, I'll either just make the merge or run the small MIP test set. I should also run my medium LP test set as a sanity check.

lgottwald:
Okay. I finished the follow-up test now too to check the improvements to the tableau separator and performance, as anticipated, looks great. I'll create a merge request that should be merged after this branch is merged.

jajhall:
Removed redundant code and cleaned up. Dual simplex Devex is now only used if specified explicitly, or if no DSE weights are known and basis is near optimal. If not near optimal, DSE weights are computed from scratch if necessary.

PR #686, opened 2022-01-23
Mip tableau cuts

lgottwald:
Contains the changes from the DSE-preservation branch to use the weights in the tableau separator and also revises the separator on several fronts.


jajhall (approved):
I trust you on this!

PR #687, opened 2022-01-24
stop probing when there are many implications

lgottwald:
Small change preventing a long runtime of presolve in rare case


lgottwald:
Ah okay, thanks for clarifying

jajhall (approved):
Fine

PR #688, opened 2022-01-24
Freeze basis in strong branching

lgottwald:
Based on performance obersvations this seems to be the correct fix. Once you give the ok I will do a performance run on this branch but some preliminary tests show that it is likely good.

### DIFF
--- a/src/mip/HighsLpRelaxation.h
+++ b/src/mip/HighsLpRelaxation.h
@@ -127,6 +127,9 @@ class HighsLpRelaxation {
       if (freezeBasisId != -1) {
         lpsolver->unfreezeBasis(freezeBasisId);
         freezeBasisId = -1;
+        lpsolver->setOptionValue("output_flag", true);
+        lpsolver->run();
+        lpsolver->setOptionValue("output_flag", false);
       }
     }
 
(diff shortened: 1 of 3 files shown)


lgottwald:
To avoid confusion: The fix is only the one line in commit 09252c63ec7bf862c418442a350e93d0dc941855

jajhall:
OK This looks believable.

lgottwald:
Meaning the fix looks reasonable and you would approve for performance tests?

lgottwald:
Ah okay, thanks for clarifying

jajhall:
I've fixed an error that prevented the invertible representation of the unfrozen basis from being used, and the unnecessary call to run() in line 129 of HighsLpRelaxation.h is commented out. I'll bear in mind the fix to src/simplex/HSimplexNla.cpp

jajhall:
So, I'm pretty sure that I've fixed the freeze basis NLA issue - using a different fix in the place you suggested. See fix-freeze-basis. blend2 and ctest run correctly without assertions 

Hence your branches freeze-basis-in-strong-branching and freeze-basis-in-strong-branching-fix2 are redundant. 

However, in strong branching there's a sudden jump in the dual steepest weight errors. Combining freeze basis with DSE weight updating is not a situation that was easy to code for without examples. I'll investigate it today.

Issue #680, opened 2022-01-19
Windows linking issue


jajhall:
A user has contacted me with the following issue

> I have downloaded the file HiGHS.v1.1.1x86_64_w64_mingw32_cxx11.tar.gz and ‘unzipped’ the files to my C:\HiGHS directory.
> However, on trying to execute highs.exe I was the informed that the following files could not be found viz:

> libgcc_s_seh-1.dll
> libgomp-1.dll
> libstdc++-6.dll
> libwinpthread-1.dll libstdc++-6.dll

> I downloaded copies of each of these files from the internet and placed them in the c:\HiGHS directory, but when I now attempt to run highs.exe I get the message
> highs.exe Application Error
> The application was unable to start correctly (0xc000007b).
> Click OK to close the application.

I'd ask @galabovaa, but she's busy working on her thesis until the end of February

Perhaps the solution is obvious to one of our other users

NPC:
@jajhall I'm not (yet) experienced with HiGHS, but I got the support libraries [here](https://github.com/JuliaBinaryWrappers/CompilerSupportLibraries_jll.jl/releases/download/CompilerSupportLibraries-v0.5.1%2B0/CompilerSupportLibraries.v0.5.1.x86_64-w64-mingw32-libgfortran5.tar.gz), and now it works for me.

jajhall:
> @jajhall I'm not (yet) experienced with HiGHS, but I got the support libraries [here](https://github.com/JuliaBinaryWrappers/CompilerSupportLibraries_jll.jl/releases/download/CompilerSupportLibraries-v0.5.1%2B0/CompilerSupportLibraries.v0.5.1.x86_64-w64-mingw32-libgfortran5.tar.gz), and now it works for me.

Thank you so much @NPC

jajhall:
Thanks @NPC, the user is now liking HiGHS OK!

Issue #590, opened 2021-10-15
Windows *.lib file for HiGHS.


jajhall:
I have downloaded your Windows binary files.

I have interfaced all commercial- and community-based LP, QP, MILP, etc. solvers to our modeling software and I would like to do the same for HiGHS.

All of the other solvers provided the *.lib file for their solvers but I do not see this file extension for HiGHS.

Would it be possible to get this file type somehow?

odow:
This Windows release:
https://github.com/JuliaBinaryWrappers/HiGHS_jll.jl/releases/download/HiGHS-v0.3.3%2B0/HiGHS.v0.3.3.x86_64-w64-mingw32-cxx11.tar.gz
has `bin/libhighs.dll`. Is that not sufficient?

jkiviluo:
I just experienced a similar issue. When I try to execute highs from the command prompt, it cannot proceed because its missing `libgcc_s_sjlj-1.dll` and `libstdc++-6.dll`. I'm sure these can be acquired from e.g. mingw-w64-gcc, but it would be nice if the binary just worked out of the box.

jajhall:
> I just experienced a similar issue. When I try to execute highs from the command prompt on Windows10 after just downloading the HiGHS binary (x86_64-w64-mingw32-cxx11.tar.gz), it cannot proceed because its missing `libgcc_s_sjlj-1.dll` and `libstdc++-6.dll`. I'm sure these can be acquired from e.g. mingw-w64-gcc, but it would be nice if the binary just worked out of the box.

Noted, and we will get around to it at some point.

odow:
Oops. I should update the instructions. (I don't have a windows machine so I can't test.)

You can find the missing DLLs here:
https://github.com/JuliaBinaryWrappers/CompilerSupportLibraries_jll.jl/releases/tag/CompilerSupportLibraries-v0.5.1%2B0

You probably want to download this one
https://github.com/JuliaBinaryWrappers/CompilerSupportLibraries_jll.jl/releases/download/CompilerSupportLibraries-v0.5.1%2B0/CompilerSupportLibraries.v0.5.1.x86_64-w64-mingw32-libgfortran5.tar.gz

jkiviluo:
Thanks! Got it working standalone. I am using it to run a MathProg model.

[...]

PR #689, opened 2022-01-25
Update download instructions in README

odow:
x-ref: https://github.com/ERGO-Code/HiGHS/issues/680
x-ref: https://github.com/ERGO-Code/HiGHS/issues/590


jajhall (approved):
Thanks!

PR #690, opened 2022-01-25
Improve cut aging

lgottwald:
Finished performance runs for this now and looks good. Cleans up and improves how cutting planes are aged out of the LP. Before there where some inconsistencies and cutting plane ages where not allowed to increase for every node. Now cuts may age throughout every node during a dive down the tree but will have there age set to zero when they are useful at one of the nodes. Useful means that they contribute with a nonzero dual multiplier with some new special handling for cuts that have integer columns which are not sitting at a global bound in an LP solution. For those columns all cuts that have no slack have their ages reset to zero.

So with the changes some that do not seem useful may be removed from the LP much faster leading to a reduced LP size but also some cuts that where aged out before may now be kept due to also considering the LP solutions for aging when the LP exceeds the upper bound and by the additional rule for general (non-binary) integer columns.


lgottwald:
In the last two commits I revised slightly the printing in the solving report regarding the gap. The accuracy was fixed to 2 digits so a gap below 0.01% or a gap tolerance below 0.01% would be printed as 0.00% which I though might be misleading. Now the gap and its tolerance is printed up to precision 0.000001% but at most 3 significant digits precision. I feel that is an important  change because the default value for mip_rel_gap is 1e-4 (=0.01%) so with the default value all instances where HiGHS terminates with a gap below 0.005% where rounded down in the solving report and printed as 0.00%.

jajhall:
Indeed, I'd spotted that the gap was always coming out to be 0.00. I guessed that it can end up being exactly zero even with a tolerance of 1e-4, but not so uniformly.

jajhall (approved):
Fine

lgottwald:
Also now added an option to control whether use of parallism in the cliquetable is allowed. The parallelism there helps on several instances in the miplib2017 benchmark set. It is not an improvement by an order of magnitude but the small part where parallism is applied scales very well. Affects instances where significant amounts of time is spent on separating cliques, e.g., rail01, rail02, eilA101-2, nursesched, irp, nw04. The code is already in master but had no option attached to adjust the threshold. Now it can be disabled by setting to kHighsIInf or always used by setting to 0. Default value of 100000 entries seems to be sensible based on the affected instances. Also, when it does not help, overhead of this is negligible. When the number of available threads is set to 1 the parallel code is never executed to avoid any overhead when no more than 1 thread is available.

The parallel code produces exactly the same result as the serial code and is totally independent of the thread number. So any instance solves exactly the same regardless of the option value I introduced or the option value for the number of threads.

Solving time on rail01 on my machine is 289s with 8 threads and 343s with 1 thread. This instance is one of the more affected ones, because it is solved in the root node so the separation rounds are a larger fraction of the runtime compared to instances where branch&bound nodes need to be explored. Considering that the LP solves are still the majority of the time the scaling of just the clique separation is quite good. Just subtracting time for presolve and solving the first LP makes it 200s vs 255s and the LP resolves will be another >100s of that time.

PR #691, opened 2022-01-25
Parallel mip

lgottwald:
This contains some important ground work for parallelism in the MIP solver and changes the datastructures for the clique table. The changes in here are well tested by now and lying around long. This has been my working branch from which most of the other recent pull requests diverged as too many changes had gathered in here. Now it only contains the single changes I started the branch for and I think also be merged with master to not have more diverging behaviors of the MIP solver.


jajhall (approved):
Needs merging and, as you say, well tested

PR #693, opened 2022-01-25
fix performance bottleneck with moving around all implications in memory

lgottwald:
Discovered this randomly when testing things. E.g. occurs on neos-5104907-jarama where I noticed that there is a 50% slowdown just due to the presence of multiple threads. As it turned out when looking in the profiler most time was spent in mapping and unmapping memory, as well as, moving it around, due to many implications of binary variables being found. The slowdown seemed to just be due to a switch in the memory allocator to support multithreading or something like that.

The change is functionally 100% equivalent but stores implications of binary variables in individual vectors instead of one big vector so that adding new implications does not require to touch memory of previous ones. Presolve time on neos-5104907-jarama went down from 33 seconds (threads=1) and 45 seconds(threads=8) to 3 seconds for both values. The thread number is expected to not have an effect as the threads just exist but do nothing but sleeping until they get work.

### DIFF
--- a/src/mip/HighsImplications.cpp
+++ b/src/mip/HighsImplications.cpp
@@ -128,6 +128,7 @@ bool HighsImplications::computeImplications(HighsInt col, bool val) {
   HighsInt loc = 2 * col + val;
   implications[loc].computed = true;
   implications[loc].implics = std::move(implics);
+  this->numImplications += implications[loc].implics.size();
 
   return false;
 }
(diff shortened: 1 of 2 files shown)


lgottwald:
As a side effect this change will also get rid of the false positive "may be used uninitialized" warning that we had an issue about

lgottwald:
The build failure was due to a bug that was revealed by the new memory layout. Before the access was very likely in the region of another implication so that the column value read was in the expected range.

jajhall (approved):
Well spotted!

PR #695, opened 2022-01-28
Use mip start solution

lgottwald:
includes set-get-iterate branch


jajhall (approved):
I'm happy with this if you are!

lgottwald:
I also changes a few other minor things. I.e. for the `HighsBasis` struct the `HighsBasisStatus` was unnecessarily large using the default enum size, usually 4 bytes. I changed it to be 1 byte. Also added some missing `const` to the postsolve stack and fixed one situation that is technically undefined behavior where memcpy with zero size and `NULL` pointer was called.

lgottwald:
I think I am happy. Running little more tests, but would probably merge once the CI tests run through

PR #696, opened 2022-01-28
fix unintended behavior of RINS heuristic

lgottwald:
Just discovered this. The RINS heuristic implementation extends the original scheme by fixing additional variables, but the implemented behavior fixed additional variables before the RINS fixings where all applied.

### DIFF
--- a/src/mip/HighsMipSolver.cpp
+++ b/src/mip/HighsMipSolver.cpp
@@ -27,6 +27,7 @@
 #include "presolve/HighsPostsolveStack.h"
 #include "presolve/PresolveComponent.h"
 #include "util/HighsCDouble.h"
+#include "util/HighsIntegers.h"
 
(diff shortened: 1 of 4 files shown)


lgottwald:
Shortly testing the fix on a few instances, but heuristic success seems to be improved

lgottwald:
The changes for the heuristic are ok, but there is one other (unrelated) bug that I need to investigate where the symmetry detection is not terminating

lgottwald:
I would also want to merge this as it now fixes several bugs. One in the symmetry detection that already appeared several times before, but very rarely, and only on a long running instance which made it very hard to debug. I am glad I finally fixed it now. The bug in the primal heuristic seemed to decrease the success of the heuristic considerable and one other issue leading to bad restart decisions that appeared during my tests of the other changes.
Any objections @jajhall?

jajhall (approved):
Fine by me!

PR #698, opened 2022-01-29
Bug fixes

lgottwald:
I discovered some further bugs when I wanted to start to look into some prerequisites for user cut callbacks. I believe the callback likely is supposed to operate in the original model's space. I.e. it must be possible to transform the LP relaxation solution in to the original space, which already is the case, but then it must be possible to transform a cutting plane back to the presolved space. Regarding the duplicate column reduction this is not generally possible, except when both columns that are duplicate with the same scale. For now I added a flag that is set for columns that are involved in such a reduction to mark them as not linearly transformable in the postsolve information, i.e. a linear expression involving these columns cannot be transformed to the reduced problem. In that case clique and implication information involving such columns is discarded now. Clique information was not much of an issue before because the only relevant case where bad things could happen was when two binary variables where merged into an integer which then got its bounds tightened to have ub-lb == 1 so that it got converted back into a binary variable. For continuous columns, however, some wrong variable bound information might have been there in rare circumstances.

### DIFF
--- a/src/lp_data/HighsOptions.h
+++ b/src/lp_data/HighsOptions.h
@@ -900,10 +900,10 @@ class HighsOptions : public HighsOptionsStruct {
         kMaxPivotThreshold);
     records.push_back(record_double);
 
-    record_int = new OptionRecordInt("presolve_substitution_maxfillin",
-                                     "Strategy for CHUZC sort in dual simplex",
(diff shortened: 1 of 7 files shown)


lgottwald:
This should be ready for merging from my side as it contains only bugfixes

jajhall (approved):
Well done on handling this edge case.

PR #699, opened 2022-01-29
skip implicit integer columns in primal heuristic

lgottwald:
This was the intention to begin with but was done inconsistently. I will keep this as a pull request for now and wait for the current benchmark runs to finish. Once they have finished I will compare this change only on those models again which have any implicit integer variable as this is a much smaller subset.

### DIFF
--- a/src/mip/HighsPrimalHeuristics.cpp
+++ b/src/mip/HighsPrimalHeuristics.cpp
@@ -351,6 +351,8 @@ retry:
     for (HighsInt i : intcols) {
       if (localdom.col_lower_[i] == localdom.col_upper_[i]) continue;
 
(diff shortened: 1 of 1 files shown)


jajhall (approved):
Happy for this to be merged once the benchmarking is finished

lgottwald:
Benchmarks will finish today. I will then run the implicit integer instances again with this changes but without the other changes in master and will look into one more crash that was revealed on an instance. Once that is fixed I will do one full run of the benchmarks again with all the changes but will use 1 thread and 4 concurrent instances to speed things up. If everything goes well the release would then be ready from my side.

lgottwald:
Closing this as performance is slightly worse (although there were too few instances for this to be conclusive)

PR #704, opened 2022-02-01
Fix osi gams compile

svigerske:
Fixes compilation/link of Osi and GAMS interfaces.

### DIFF
--- a/check/TestOsi.cpp
+++ b/check/TestOsi.cpp
@@ -16,6 +16,7 @@
 #include "CoinHelperFunctions.hpp"
 #include "CoinPragma.hpp"
 #include "HCheckConfig.h"
+#include "HighsInt.h"
 #include "OsiHiGHSSolverInterface.hpp"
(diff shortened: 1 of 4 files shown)


jajhall (approved):
Thanks Stefan. I guess this means that the HiGHS OSI interface works for you. I can see that you've fixed in src/interfaces/OsiHiGHSSolverInterface.cpp all the things that I fixed in that file during my abortive attempt to fix the OSI interface.

All the CI tests pass, and it compiles fine for me - I've not got OSI installed currently, due to the problems I was having with the broken interface. So I'll merge it into master.

svigerske:
The Osi interface and its test build again (at least for me) and the test does not abort immediately with a segfault, but it still fails.

I was just procrastinating by fixing some compilation issues that I ran into :).

jajhall:
Thanks, I'll reinstall OSI at some point and fix the interface so that the tests pass

PR #712, opened 2022-02-03
Prevented indentation without prefix for callback and deleted redundant printmsgcb

jajhall:


lgottwald:
So I removed the msgbuffer now as static, so only the log callbacks are left

lgottwald:
Also removed the static callback variables and moved them to the log options. The only thing left is, assuming I have not overlooked where this happens, a possibility to set the log callback. Once the callback is set we also need to make sure to only use the callback and not print anything through printf anymore. I would suggest to maybe remove the callbacks from the HighsOptions Struct completely and only have them within the log options. I don't see why we would need two copies of them. Then I guess the Highs class should get methods to set the log callback with its data directly within the HighsLogOptions struct.

lgottwald:
I also made the default NULL arguments consistently use nullptr in the Highs.h header.

@jajhall Could you check whether the logging behaves as you would expect and I have not broken anything?

jajhall:
I've looked through what you've done - mainly to try to understand it rather than check for errors. I've improved the naming and checked that callbacks work in Highs::run() as intended, and updated the OSI interface now that callbacks can be set up via the method (now named) setLogCallback.

I had a quick go at saving the last error log, but adding a char buffer to `HighsLogOptions` leads to a violation of `log_options_` being const in `highsLogUser`. I guess that `HighsLogOptions` needs to know the pointer to where the buffer is, but making it work is beyond my C++. So, unless you've got a spare moment, @odow will have to wait.

lgottwald:
I can have a look tomorrow and see if I can come up with a satisfying solution

lgottwald:
in principle what could work is simply using mutable in front of the char buffer, then it can be written to regardless of const and the the last printed line is always there. One thing to have in mind though is that printing through the same log options with multiple threads is then not possible.

jajhall:
Let's leave this for now and get v1.2 released

lgottwald:
Ok, so can we merge this?

jajhall:
> Ok, so can we merge this?

Yes

PR #714, opened 2022-02-04
setLogCallback takes only 2 args now

svigerske:


jajhall (approved):
Sorry, I missed this one!

PR #715, opened 2022-02-06
Fix infeasible solution

lgottwald:
@jajhall there is one warning where HiGHS terminates with an infeasible solution in the most recent MIP runs. I remember that the check for this instances showed an infeasibility in the original space that I suspected to be due to rounding errors as the computed infeasibility was for row bounds and was just one smallest possible deviation that a double precision value could be above the feasibility tolerance. As the computed row values are subject to rounding errors I at that time did not want to discard the solution since it was feasible in the presolved space and was checked with quad precision there.
I now removed the hacky allowance of a tiny extra tolerance and compute the row values with quad precision instead. For this I simply added one additional function at the place where the function for computing the row values was used before. This should avoid to ever terminate with an infeasible solution, even very slightly infeasible, and still accept solutions when the exceeding of the feasibility tolerance is due to rounding errors as quad precision is used now.

Also the change to the default max scaling factor last minute might not have been a great idea as it seems to have some side effects leading some MIP instances to timeout that could otherwise be solved rather reliably. For the release I'd suggest to rollback to the previous value for now and do some more extensive testing whether to increase it for the next release where I have time to investigate. The effect for timing out must be some unexpected side effect since the LP performance is not altered.

### DIFF
--- a/src/lp_data/HConst.h
+++ b/src/lp_data/HConst.h
@@ -183,7 +183,7 @@ enum class HighsBasisStatus : uint8_t {
 };
 
 // Default and max allowed power-of-two matrix scale factor
-const HighsInt kDefaultAllowedMatrixPow2Scale = 10;
+const HighsInt kDefaultAllowedMatrixPow2Scale = 20;
 const HighsInt kMaxAllowedMatrixPow2Scale = 30;
 
 // Illegal values of num/max/sum infeasibility - used to indicate that true
--- a/src/lp_data/HConst.h
+++ b/src/lp_data/HConst.h
@@ -183,7 +183,7 @@ enum class HighsBasisStatus : uint8_t {
 };
 
 // Default and max allowed power-of-two matrix scale factor
-const HighsInt kDefaultAllowedMatrixPow2Scale = 20;
+const HighsInt kDefaultAllowedMatrixPow2Scale = 10;
 const HighsInt kMaxAllowedMatrixPow2Scale = 30;
 
 // Illegal values of num/max/sum infeasibility - used to indicate that true
--- a/src/lp_data/Highs.cpp
+++ b/src/lp_data/Highs.cpp
@@ -597,6 +597,15 @@ HighsStatus Highs::run() {
   highs::parallel::initialize_scheduler(options_.threads);
 
   max_threads = highs::parallel::num_threads();
+  if (options_.threads != 0 && max_threads != options_.threads) {
+    highsLogUser(
+        options_.log_options, HighsLogType::kError,
+        "Option 'threads' is set to %d but global scheduler has already been "
+        "initialized to use %d threads. The previous scheduler instance can be "
+        "destroyed by calling HighsTaskExecutor::shutdown().\n",
+        (int)options_.threads, max_threads);
+    return HighsStatus::kError;
+  }
   assert(max_threads > 0);
   if (max_threads <= 0)
     highsLogDev(options_.log_options, HighsLogType::kWarning,
--- a/src/lp_data/HighsLpUtils.cpp
+++ b/src/lp_data/HighsLpUtils.cpp
@@ -2242,6 +2242,34 @@ HighsStatus calculateRowValues(const HighsLp& lp, HighsSolution& solution) {
   return HighsStatus::kOk;
 }
 
+HighsStatus calculateRowValuesQuad(const HighsLp& lp, HighsSolution& solution) {
+  // assert(solution.col_value.size() > 0);
+  if (int(solution.col_value.size()) != lp.num_col_) return HighsStatus::kError;
+
+  std::vector<HighsCDouble> row_value;
+  row_value.assign(lp.num_row_, HighsCDouble{0.0});
+
+  solution.row_value.assign(lp.num_row_, 0);
+
+  for (HighsInt col = 0; col < lp.num_col_; col++) {
+    for (HighsInt i = lp.a_matrix_.start_[col];
+         i < lp.a_matrix_.start_[col + 1]; i++) {
+      const HighsInt row = lp.a_matrix_.index_[i];
+      assert(row >= 0);
+      assert(row < lp.num_row_);
+
+      row_value[row] += solution.col_value[col] * lp.a_matrix_.value_[i];
+    }
+  }
+
+  // assign quad values to double vector
+  solution.row_value.resize(lp.num_row_);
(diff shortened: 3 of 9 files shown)


jajhall:
I suspect that this inconsistency is due to you not computing primal infeasibilities the same way as I do in the simplex solver (and elsewhere in HiGHS). Namely, in HighsMipSolverData::transformNewIncumbent, you compute 

    row_violation_ =
        std::max(row_violation_,
                 mipsolver.orig_model_->row_lower_[i] - solution.row_value[i]);
    row_violation_ =
        std::max(row_violation_,
                 solution.row_value[i] - mipsolver.orig_model_->row_upper_[i]);

which, ultimately, gets compared with

      row_violation_ <= mipsolver.options_mip_->mip_feasibility_tolerance;

to determine feasibility. However, I compute

      double primal_infeasibility = 0;
      if (value < lower - scaled_primal_feasibility_tolerance) {
        primal_infeasibility = lower - value;
      } else if (value > upper + scaled_primal_feasibility_tolerance) {
        primal_infeasibility = value - upper;
      }
and count the number of times that primal_infeasibility exceeds the tolerance, as well as finding the max and sum of them.

In marginal cases, you'll get feasible = true when I get a zero count. Computing the row activities in quad precision may have fixed the issue you found, but the only sure fix is for us both to compute infeasibilities the same way. Since you compute row_violation_ and col_violation_ in only a few places, I'd like you to switch.

jajhall:
Naturally I'm fine with you changing kDefaultAllowedMatrixPow2Scale

lgottwald:
I can also switch the way it is computed but the issue for the instance is cleary that I allowed an extra tolerance of kHighsTiny here. The MIP solver marks the solution as infeasible too.

lgottwald:
I'll change the way infeasibility is computed definitely too though as inconsistent reports would be unpleasant too.
I let the runs also finish now as they are far progressed and make a final decision on the scaling factor by tomorrow.

jajhall:
Thanks, and you're right that adding kHighsTiny is another source of inconsistency that should be removed.

lgottwald:
Also caught an issue with the initialization of the global scheduler, which is currently not thread-safe. I don't know whether this is the source of the segfault reported in https://github.com/ERGO-Code/HiGHS/issues/625 but it should still be fixed.

Also it occurred to me that we may want to document that HiGHS will keep global data for the scheduler around between calls to run() and when the user wants HiGHS to relinquish all resources and shutdown all threads a call to HighsTaskExecutor::shutdown() is necessary. We might want to make this call available via the C API too. Note that the call to shutdown is not threadsafe and the user is expected to only make this call when no application thread might call run() on any HiGHS instance during the call.

I now added the use of a mutex to the initialization of the global executor. Hence the scheduler is now guaranteed to be initialized exactly once even if a user creates Highs instances in multiple threads and calls run on them. The result will always be correct and work properly and deterministically. What is still subject to a race is when the user does this with different values of the `threads` option in the different HiGHS condition. The HiGHS instance that will be the first to call initialize_scheduler within run will be the HiGHS instance that determines the number of threads. To avoid this I suggest we simply discourage setting the number of threads to different values for concurrent calls to run() on different HiGHS instances. We could also expose and document the highs::parallel::initialize_scheduler() function for users which would allow initializing the scheduler indepently of run(), but I feel that is not really needed. Waiting until run() is fine and I see no downside of telling users to not use different values for the number of threads when using HiGHS instances concurrently.

jajhall:
Thanks. I'd come the same conclusions before I'd finished reading your comment. I suggest that when an instance of HighsOptions is created a check is made to see whether something other than the default value for `threads` has been used in the first call to initialize_scheduler, in which case that will be used to initialise `threads`. Similarly, if a user attempts to change the `threads` option, a check is made to see whether it matches the value associated with the scheduler, and an error is reported if the user's value is rejected. Thus "discouragement" is hard-coded.

lgottwald:
I think this would not work. The user could change the option values independently before calling run, in which case the call to initialize_scheduler has not yet happened. So both changes would be accepted and the winner of the race gets to determine the value that will be used.

lgottwald:
but we could return an error in run when after the call to initialize the scheduler it does not have the expected number of threads

lgottwald:
And the error message can then state the changing the global thread number is not possible once the scheduler is initialized. Also we could reference the shutdown() call to reinitialize the scheduler with a new number of threads.

jajhall:
Sure, but we must still return an error if the user attempts to set a value of `threads` that we know to be illegal at the time.

lgottwald:
But we don't know at the time. The value is legal when it matches the thread number after the initialization in the run call. Whether that happens we cannot detect at the time when the option is set.

jajhall:
> I think this would not work. The user could change the option values independently before calling run, in which case the call to initialize_scheduler has not yet happened. So both changes would be accepted and the winner of the race gets to determine the value that will be used.

Yes, we can't avoid this, and would have to return an error in run(). To be honest, very few users will think of setting different values of threads in different Highs instances, and should read up on whether it's possible.

lgottwald:
A different possibility could work. I could make the thread option point to a global variable instead of into the HighsOptions instance. Then changing it will cause the value to change for all Highs instances at the same time automatically. For changing the threads option the initialization mutex should be acquired. For that I would need to make a new option record type that does this when altered.

lgottwald:
Then it would actually also be possible to check that the scheduler is uninitialized when there is an attempt to alter the number of threads and other wise let the changing of the option value fail.

lgottwald:
I also am not sure if it is worth the effort. Returning an error in run() might be enough and is the simplest thing we could do. Users will likely not notice and for the slight probability that a user tries to do this a meaningful error message should be enough to catch this as a user error.

jajhall:
Sounds over-complex, and changes the number of threads for a Highs instance that's already been chosen by the user. No, we make it clear that all Highs instances must have the same value of `threads`, allow the race to run(), and throw whatever error messages we like if it's violated. The race might make the code non-deterministic, but the user is doing something illegal if errors are returned.

jajhall:
Agreed

lgottwald:
The race would still produce correct and deterministic results. It would just make the property of how many threads participated in producing the result non-deterministic.

jajhall:
Yes, that's what I mean by non-deterministic

lgottwald:
This is ready to be merged and tagged for the release from my side

jajhall:
Did a few LP sanity checks, and silenced a compiler warning

PR #716, opened 2022-02-08
Add dense trans

jajhall:
Tidied up work that will be useful in future when using HFactor, rather than basiclu for IPX. 

For simplex, change the breaking of ties in non-hyper-sparse LPs, and hopefully improve performance since indices of *TRAN solutions will be ordered, so better cache-wise when accessing double values.


lgottwald:
This could now be merged I think?

Issue #478, opened 2021-02-26
constants are not exported in the C API


lovasoa:
It would be nice to have access to the various constants used as status codes in the C API.

Currently, one has to hardcode the constants values in code such as in  https://github.com/ERGO-Code/HiGHS/blob/master/examples/call_highs_from_c.c#L226

It would be great to be able to write 

```c
if (modelstatus == OPTIMAL) { ... }
```

with `OPTIMAL` a constant available in `highs_c_api.h`

jajhall:
Sure, and this is something that we'll do for the first proper release. Unfortunately OPTIMAL is part of an enum class in the C++, and these don't export (readily) to C

Nice to have the feedback, though! Hope you're happy with HiGHS otherwise.

lovasoa:
> OPTIMAL is part of an enum class in the C++, and these don't export (readily) to C

Yes, I meant manually exporting the constant values of the c++ enum. I did that in my rust binding: https://github.com/rust-or/highs-sys/blob/v0.0.5/src/lib.rs#L9-L27

> Hope you're happy with HiGHS otherwise.

Yes, I like working with it !

jajhall:
It's true that this is something that we could do in the first formal release

lovasoa:
@jajhall Why closing this issue now, then ? Maybe we can leave it open until this is actually done ?

jajhall:
This issue has been addressed to the fullest reasonable extent

PR #718, opened 2022-02-09
Document the C API

odow:
The new version of HiGHS.jl will let us pull in the documentation for the C API: 
https://github.com/jump-dev/HiGHS.jl/blob/35e1153a819f4110cf503630b8f03d0b51465b33/src/gen/libhighs.jl#L169-L174

So it's in my interest to tidy up some of the docstrings. (At the moment, the only way to know some things when using HiGHS from Julia is to look at the source code.)

I started a couple, but I thought I'd ask some questions before I continue:
 - [x] Are you okay if I do this? Or did you have a different plan to document things?
 - [x] Is there any particular format I should follow? (I've never used Doxygen. What fields should I add?)

Here's how the resulting docstrings look in the Julia help:
```julia
help?> Highs_clearModel
search: Highs_clearModel Highs_changeColsIntegralityByRange

  Highs_clearModel(highs)

  Empties the current model but does not invalidate the highs pointer.

  This function removes all variables and constraints from the model highs, but does not invalidate
  the pointer highs. Future calls (for example, adding new variables and constraints) are allowed.

  See Highs_destory to clear the model and free all associated memory.

  Parameters
  ––––––––––––

    •  highs: a pointer to the HiGHS model object
```


jajhall:
We've got plans for formal documentation for HiGHS, with @galabovaa deciding that the Julia documentation system is the way to go. A "proof of concept" site has been developed
image
It would be good if the same documentation data from the source code can be pulled into our documentation. As with a whole bunch of things, there will be no progress until Ivet has submitted her thesis.

odow:
I will admit, I find the Julia documentation system easy to use (but probably because I have more experience). I'm not sure how well it integrates C and C++ code though. You're probably better off using something that understands doxygen.

The new documentation of SCS used readthedocs:
 * https://www.cvxgrp.org/scs/index.html
 * https://www.cvxgrp.org/scs/api/c.htm

I think the docs look nice, and they have a good mix of automated API details, and written exposition.

> there will be no progress until Ivet has submitted her thesis.

Okay. In the mean-time I will slowly chip away at this. 

I'll follow SCS as a guide for the docstrings, since that seems to work:
https://github.com/cvxgrp/scs/blob/65a5451f4427477a332542438f89cee9ded570c7/include/scs.h#L201-L272
And I'll leave Ivet to make a decision of the final format.

odow:
Okay, this is now in a good state. There are a couple of TODOs about deprecating methods and one about the structure of the Hessian, but I can fix them in other PRs.

Julia manages to parse the Doxygen okay, so I think we're good on that front:
```julia
help?> Highs_deleteColsByRange
search: Highs_deleteColsByRange Highs_deleteColsByMask Highs_deleteColsBySet

  Highs_deleteColsByRange(highs, from_col, to_col)

  Delete multiple adjacent columns.

  Parameters
  ––––––––––––

    •  highs: a pointer to the HiGHS model object

    •  from_col: the index of the first column to delete

    •  to_col: the index of the last column to delete

  Returns
  –––––––––

  A non-zero return value indicates that a problem occured
```

Once this is merged, I can have a go at: https://github.com/ERGO-Code/HiGHS/issues/478, which should allow us to tidy up some of the "What values does Highs_getModelStatus return?" type questions.

There's also a few annoying inconsistencies in argument naming, but they're pretty minor and can be adjusted in a different PR. (Why sometimes `numnz` and others `num_nz`? Why `start`, `index` and `value`, but then in other cases `solution_start`, `solution_index` and `solution_value`? Why `q_format` but `qindex`? These don't have any bearing on the functionality, just the pedantic side of me reading.)

odow:
> here's no solution_start, and the use of solution_index and solution_value

Ah, yeah. I guess I mis-remembered. It was `matrix_start` etc in `get{Cols,Rows}By{Range,Set,Mask}`.

I'll fix the conflict.

odow:
Okay, conflicts are fixed. 

This PR is pretty hard to review, but since it's a strict improvement on what we had before (I think I fixed the worst offenders that had some off-by-one descriptions, but there might still be some typos etc.), I'd suggest we merge it as-is. 

I can take another editing pass when I go through and re-name some of the arguments, and another when I export the enums.

jajhall:
Thanks. Tomorrow I was intending to correct naming inconsistencies, as well as look for typos. I'll also make sure that the C++ Highs class is consistent with the C API in terms of names and documentation.

odow:
> Tomorrow I was intending to correct naming inconsistencies

Perhaps we can merge this first then? This is already a big PR, and it'd be easier to review if only the names were changing, not a whole lot of other things.

jajhall:
I've merged this into a branch of HiGHS and brought it up-to-date with master so that I can review and check it.
Maybe there's a more efficient way of doing this...

PR #727, opened 2022-02-11
Fix 723

jajhall:
Deleted `Highs_lpDimMpsRead` and `Highs_lpDataMpsRead` from C and Python APIs, modifying `minimal_api_mps()` in `examples/call_highs_from_c.c` accordingly


odow (approved):
Nice.

PR #729, opened 2022-02-12
fix possible memory leak in HighsHashTable::clear()

lgottwald:
Should go into 1.2.1. It fixes a memory leak when clear() is called in the hashtable which is the case in the Modk separator.


jajhall (approved):
Fine

PR #730, opened 2022-02-12
Ensure run() returns error if model constraint matrix contains a large value

### LINKED ISSUE #728: Unspecific "error loading file"
I am using PuLP to formulate an ILP problem, which works fine with gurobi. When using HiGHS, I get the pretty unspecific error (I am using the direct call that PuLP does to ensure that it's not a PuLP error):
```
$ highs --model_file /var/folders/pf/p_4b5xx54_n00pf2phf_1x980000gn/T/76a17ff536d847a3a5934991380c93d0-pulp.mps
Running HiGHS 1.2.0 [date: 2022-02-11, git hash: f11b7936]
Copyright (c) 2022 ERGO-Code under MIT licence terms
Number of BV entries in BOUNDS section is 178160
WARNING: LP matrix packed vector contains 52683 |values| in [1.79769e+308, 1.79769e+308] greater than 1e+15
Error loading file
```

Is there a way to better debug this error?

PS: I can share the `.mps` file any time, if that helps.

jajhall:
Highs::run() returns error if model constraint matrix contains a large value; 
read/passModel return kWarning if matrix contains a small value
closes #728

### DIFF
--- a/check/TestLpValidation.cpp
+++ b/check/TestLpValidation.cpp
@@ -4,6 +4,7 @@
 #include "catch.hpp"
 
 const bool dev_run = false;
+const double inf = kHighsInf;
 
 // No commas in test case name.
 TEST_CASE("LP-dimension-validation", "[highs_data]") {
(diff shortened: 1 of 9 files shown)


odow (approved):
I don't know if I can comment on the implementation specifics, but I agree that encountering a large value should be an error. I don't know why Gurobi would allow `inf * x` to be anything; it just leads to modeling errors.

Issue #721, opened 2022-02-09
DLLs not installed


svigerske:
I build HiGHS v1.2 under MSys2 with
```
cmake .. \
  -G "NMake Makefiles" \
  -DCMAKE_BUILD_TYPE=Release \
  -DCMAKE_INSTALL_PREFIX=/path/to/install \
  -DCMAKE_C_COMPILER=icl -DCMAKE_CXX_COMPILER=icl
nmake
nmake install
```
The last step misses installing the HiGHS DLLs into `$prefix/bin`.

It only installs the `.lib` files into `$prefix/lib` and `highs.exe` into `$prefix/bin` (and some non-binary files).

Also without `-DCMAKE_BUILD_TYPE=Release` I seem to get a debug build, while on other platforms `Release` was the default (which seems correct).

lgottwald:
Could you try to use `-DFAST_BUILD=on`?

svigerske:
If I only add `FAST_BUILD=on`, then I get only static libraries. If I do `-DFAST_BUILD=on -DBUILD_SHARED_LIBS=on`, then I get a DLL and it is installed by `nmake install`.

lgottwald:
Thanks, I see that this will be fixed

lgottwald:
The issue with the debug build I do not understand though. Could you maybe check if the problem occurs with a fresh build directory? And if it does could you check whether it makes a difference to move the if block on line 19 of the top level CMakeLists.txt (`if(NOT CMAKE_BUILD_TYPE)`), which sets the default build type to release, a few lines up before the `project()` call at line 15?

svigerske:
> The issue with the debug build I do not understand though. Could you maybe check if the problem occurs with a fresh build directory? And if it does could you check whether it makes a difference to move the if block on line 19 of the top level CMakeLists.txt (`if(NOT CMAKE_BUILD_TYPE)`), which sets the default build type to release, a few lines up before the `project()` call at line 15?

Yes, moving the `CMAKE_BUILD_TYPE` before `project()` fixes this problem.

lgottwald:
@galabovaa I don't understand the difference between the FAST_BUILD=on/off and installation of the library here. Do you have an idea?

mmuetzel:
Didn't notice this report before. But is likely fixed now that PR #823 is merged.

svigerske:
Yes, seems to work now.

Issue #720, opened 2022-02-09
task not trivially copyable


svigerske:
For HiGHS 1.2, compiling with Intel Compilers 2021.2.0.20210228 on Windows fails for me with
```
\path\to\HiGHS\src\parallel/HighsTask.h(94): error: static assertion failed with "given task type must be trivially copyable
"
      static_assert(IS_TRIVIALLY_COPYABLE(F),
      ^
          detected during:
            instantiation of "void HighsTask::setTaskData(F &&) [with F=lambda []()->void]" at line 282 of "\path\to\HiGHS\src\parallel/HighsSplitDeque.h"
            instantiation of "void HighsSplitDeque::push(F &&) [with F=lambda []()->void]" at line 39 of "\path\to\HiGHS\src\parallel/HighsParallel.h"
            instantiation of "void highs::parallel::spawn(HighsSplitDeque *, F &&) [with F=lambda []()->void]" at line 81 of "\path\to\HiGHS\src\parallel/HighsParallel.h"
            instantiation of "void highs::parallel::TaskGroup::spawn(F &&) const [with F=lambda []()->void]" at line 85 of "\path\to\HiGHS\src\mip\HighsMipSolverData.cpp"

compilation aborted for \path\to\HiGHS\src\mip\HighsMipSolverData.cpp (code 2)
```
Is there some workaround?

lgottwald:
Probably the static assert can be commented out. Likely the Intel compiler does simply not set the property of being trivially copyavle for the compiler generated type, but in reality nothing prevents it from being copied with memcpy.
Would recommend to check whether the unit tests for parallelism pass with several invocations.

svigerske:
Thanks. After taking the assert out, two invocations of `nmake tests` pass.  I hope that included tests for parallelism.

lgottwald:
Assuming it does the ctest of cmake then yes it does. It will invoke unit_test which contains a parallel Fibonacci Evaluation and matrix multiplication and will use some parallel invocations for the instance tests.
It is unfortunate that there is no compiler guaranteed way but the current solution is closest to what I could get to standard compliant and does not violate strict aliasing rules.

lgottwald:
And when the static assert passes it is guaranteed which it does for most compilers.

svigerske:
There were 148 tests, starting with
```
        Start   1: capi_unit_tests
  1/148 Test   #1: capi_unit_tests ..................   Passed    0.04 sec
        Start   2: unit-test-build
[...]

PR #731, opened 2022-02-13
revert to previous solution but use union type punning to avoid warning

lgottwald:
I reverted to a previous solution for calling a task. This solution has the advantage that it does not require the compiler generated type of the lambda expression to be trivially copyable, for which the static assert failed for some compilers. The disadvantage is that the pointer to the CallableBase class generated by a cast is not standard compliant. The base class could in principle have an offset and hence a pointer obtained by a simple cast reinterpret_cast from a derived class might not retrieve the correct pointer to the base class.
I avoided any warnings about strict aliasing violations by doing "type punning" of the pointer via a union. Also I added an assert that will fail at runtime if there should ever be an offset generated for the classes. In reality compilers do not do ever this for such cases with trivial single inheritance of one level so this should be safe in practice.

I would propose to merge into master for v1.2.1 to address issue #720. Runtime behavior is not affected.

Also contains (I think) a fix for #721, by installing libraries into bin on windows. This split between `FAST_BUILD=on/off` becomes more and more messy and I propose that we remove this soon. There should not be an option for FAST_BUILD but all builds should use the same path in how the targets are set up. To speed things up when building highs there are different things we could do. E.g. we can exclude some interfaces from the default compilation by using `EXCLUDE_FROM_ALL` or it is also easily possible to simply type `make highs` instead of `make` to only build the highs target when developing code without changing anything. We could also, e.g., exclude the unit_test build from the all target and only build the unit_tests as part of running ctest, which I think happens already.
But having these two build options causes these undesirable problems as in #721 where only one of the options works but behaves still differently regarding whether shared libraries are built or not. Installation differs between `FAST_BUILD=on/off` and regarding this particular issue what `FAST_BUILD=on` does is correct and the default for whether shared libraries are built also differs.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -427,37 +427,51 @@ install(FILES ${HIGHS_BINARY_DIR}/libhighs_export.h DESTINATION include)
 
 if (IPX_ON)
     if (UNIX)
-    #target_compile_options(libhighs PRIVATE "-Wno-defaulted-function-deleted")
-    #target_compile_options(libhighs PRIVATE "-Wno-return-type-c-linkage")
-    target_compile_options(libhighs PRIVATE "-Wno-return-type" "-Wno-switch")
+        #target_compile_options(libhighs PRIVATE "-Wno-defaulted-function-deleted")
+        #target_compile_options(libhighs PRIVATE "-Wno-return-type-c-linkage")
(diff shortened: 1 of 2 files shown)


jajhall (approved):
This is rather too technical for me to comment, although the case you make sounds good!

PR #732, opened 2022-02-13
move setting of default build type before project() call

lgottwald:
@svigerske  confirmed that this fixes the issue that the build type is not set to Release by default when using cmake with nmake makefile under Windows with MSys2.


svigerske (approved):

jajhall (approved):
Thanks @svigerske
Happy to approve

PR #737, opened 2022-02-15
Fix objective sense input to assessHessian

### LINKED ISSUE #736: Bug: PSD check for maximization quadratic programs fails
When calling `passHessian`, HiGHS does not take the objective sense into account when computing semi definiteness:
```
julia> highs = Highs_create()
Ptr{Nothing} @0x00007fea23874400

julia> Highs_addCol(highs, 0.0, 0.0, Inf, 0, C_NULL, C_NULL)
0

julia> Highs_changeObjectiveSense(highs, -1)
0

julia> Highs_passHessian(
           highs,
           1,
           1,
           1, # kTriangular
           Cint[0],
           Cint[0],
           Cdouble[-2.0],
       )
ERROR:   Hessian has 1 diagonal entries in [-2, 0) so is not positive semidefinite
-1
```

odow:
Closes #736.

I think this is sufficient. But there might be other implications of solving maximization QPs? Doesn't look like they are well tested.

### DIFF
--- a/check/TestCAPI.c
+++ b/check/TestCAPI.c
@@ -811,7 +811,7 @@ void test_getColsByRange() {
 
 void test_passHessian() {
   void* highs = Highs_create();
-  Highs_addCol(highs, 0.0, 0.0, 2.0, 0, NULL, NULL);
+  Highs_addCol(highs, 2.0, 0.0, 2.0, 0, NULL, NULL);
   Highs_changeObjectiveSense(highs, -1);
   HighsInt start[1] = {0};
   HighsInt indices[1] = {0};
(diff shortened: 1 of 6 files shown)


jajhall:
Thanks for this. I'll fix the 64-bit bug

jajhall:
Indeed, QP maximization not tested well! There was none in the C++ unit tests, and the only other was the one @odow introduced in `TestCAPI.c`. When maximizing, `Highs::callSolveQp()` only negates the cost vector, not the Hessian. Bug not exposed in the @odow example since it's max -x^2, so the turning point is the same whether or not the Hessian is negated!

jajhall (approved):
For me there is now sufficient testing of max QPs, and the process exposed several bugs that have now been fixed

feldmeier:
looks good to me - thanks for fixing this!

PR #739, opened 2022-02-15
Add default semi variable upper bound

jajhall:
Large upper bounds on semi-variables are now replaced when possible, and only within run(), by values that give acceptable matrix coefficients when a binary is introduced to transform semi-variable problems into pure MIPs. If the new bounds are active in the optimal solution, run() returns kError.

### DIFF
--- a/check/TestCAPI.c
+++ b/check/TestCAPI.c
@@ -245,24 +245,10 @@ void minimal_api_mip() {
 					 integrality,
 					 colvalue, rowvalue,
 					 &modelstatus);
-
   // Should return error
   assert( return_status == -1 );
 
-  // Non-empty integrality for a continuous problem yields a warning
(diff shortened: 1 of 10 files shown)


lgottwald:
I think we should not be very strict with the upper bound values. As I told you they will be reduced by presolve anyways whenever tighter upper bounds are implied on the semicontinuous columns by other columns bounds.
So rather lets aim for a largest value that can be reasonably handled, e.g. `1 / (k*feastol)`. A problem with very large coefficients is only that they would allow for large "trickle flows". The introduced binary may sit within (MIP) feasibility tolerance at zero but the large coefficient blows up the allowed value for the semicontinuous variable. Using the coefficient above the trickle flow would be bounded by 1/k. This will rarely be a problem at all due to presolve. Maybe k=100 or even k=10 would be ok.

PR #740, opened 2022-02-16
FilereaderLp.cpp now returns FilereaderRetcode::kFileNotFound when this happens, and not FilereaderRetcode::kParserError!

jajhall:
This caused confusion for a user and I. He'd had a .lp model that he's wanting to solve from the command line.

He had caused parser errors in the past - because he uses semi-continuous variables - and recently was generating models could be read, but not solved - because the semi-continuous variables had infinite upper bounds. I fixed the solver issue, and he came up with finite upper bounds. But he was still getting "parser errors". 

We were baffled until I spotted that "parser errors" were reported if the .lp file wasn't found. 

He'd got his file name wrong in the command line!

My fix may not be the most elegant, @feldmeier, but it works!

### DIFF
--- a/src/io/FilereaderLp.cpp
+++ b/src/io/FilereaderLp.cpp
@@ -141,6 +141,15 @@ FilereaderRetcode FilereaderLp::readModelFromFile(const HighsOptions& options,
     lp.sense_ = m.sense == ObjectiveSense::MIN ? ObjSense::kMinimize
                                                : ObjSense::kMaximize;
   } catch (std::invalid_argument& ex) {
+    // lpassert in extern/filereaderlp/def.hpp throws
+    // std::invalid_argument whatever the error. Hence, unless
+    // something is done specially - here or elsewhere -
+    // FilereaderRetcode::kParserError will be returned.
+    //
+    // This is misleading when the file isn't found, as it's not a
+    // parser error
+    FILE* file = fopen(filename.c_str(), "r");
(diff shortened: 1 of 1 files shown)


src/io/FilereaderLp.cpp
```diff
@@ -141,6 +141,15 @@ FilereaderRetcode FilereaderLp::readModelFromFile(const HighsOptions& options,
...
   } catch (std::invalid_argument& ex) {
+    // lpassert in extern/filereaderlp/def.hpp throws
+    // std::invalid_argument whatever the error. Hence, unless
+    // something is done specially - here or elsewhere -
+    // FilereaderRetcode::kParserError will be returned.
+    //
+    // This is misleading when the file isn't found, as it's not a
+    // parser error
+    FILE* file = fopen(filename.c_str(), "r");
+    if (file == nullptr) return FilereaderRetcode::kFileNotFound;
```
odow, line 152:
```suggestion
    if (file == nullptr) return FilereaderRetcode::kFileNotFound;
    fclose(file);
```
Unless I'm missing something obvious?

jajhall: You're right, it should be closed if opened successfully.

feldmeier:
thanks - Ill take a look at this today, there might be a more elegant solution. I discovered another bug in the LpFilereader as well that needs to be fixed.

PR #744, opened 2022-02-17
[README] Update download instructions

odow:
We're formulating a plan to improve Julia's ability to distribute shared libraries to other platforms, but until that happens, we have this new set of binaries which statically link the required libraries.


jajhall:
Thanks for the rewording

Issue #742, opened 2022-02-17
Meson build system


lukas-weber:
Hello and thanks for writing this library! I was able to get more than a threefold speedup over my own naive implementation of the revised simplex method when using it for some small problems.

Now as you know, adding library dependencies like HiGHS that are usually not installed on systems can be a chore, especially in an academic context where you run in environments you do not have complete control of. That is why for my projects I like to use the [meson build system](https://mesonbuild.com/). The biggest benefit of meson me is that it makes vendoring dependencies completely trivial. If a library is not installed system-wide, it can download and build and use a fallback, all completely behind the scenes. This means you do not have to include a copy of the dependency code in your repository.

Another advantage of meson is its solid design: Build definitions get much shorter and you do not get into the same technicalities or legacy practices that you run into with CMake. The availability is also excellent because it can be simply installed on virtually any system not having it via pip. That is why an increasing number of open source projects are switching to meson, including for example [systemd](https://www.freedesktop.org/wiki/Software/systemd/).

Now why am I writing this? The one thing required for the fallback mechanism mentioned above is that the dependency also uses meson. For this reason, I wrote a simple meson port of part of the HiGHS CMake build definitions. I am currently trying to get it accepted into mesonbuild/wrapdb#309 so that HiGHS becomes easy to use for all projects using meson.

Concerning this, I have two questions.

1. The most difficult part for the port is your usage of `generate_export_header`, which generates `libhighs_export.h`. In principle this header file does not depend much on the things CMake knows and the definition of different versions of the export macros for different compilers could also be done entirely in the preprocessor using a non-generated `libhighs_export.h`, arguably giving you more control over it, too. That is why meson does not have a similar function. My workaround is to supply a non-generated `libhighs_export.h` in my build system port, but including new source code files in such ports is against the rules of the meson wrapdb so that this is currently blocking the adoption of mesonbuild/wrapdb#309. Would you be okay with switching from `generate_export_header` to a non-generated export header?
2. In your CMakeLists.txt, I saw comments showing that you are not very happy with the old-style CMake practices. Instead of modernizing those definitions, would you be interested in giving meson a try as an alternative build system? Since I already wrote a functional meson port, this would be an easy thing to do. In particular, using meson fallbacks, you could easily decouple basiclu and ipx into their own dedicated libraries and repositories without making HiGHS development and usage more complicated.

jajhall:
[...]

PR #745, opened 2022-02-18
Ditch CMake generate_export_header in favor of a short self-written one

lukas-weber:
This is what removing the dependency on `generate_export_header` would look like (see #742 ). I do not have MSVC so it might be necessary to test on different systems.

On that note, it seems the export macros from this header are actually used in only a single file throughout the whole code base:

```
src/parallel/HighsTaskExecutor.h:46:  static LIBHIGHS_EXPORT cache_aligned::shared_ptr<HighsTaskExecutor>
src/parallel/HighsTaskExecutor.h:48:  static LIBHIGHS_EXPORT HighsSpinMutex initMutex;
```

So I guess the shared libraries export all symbols at the moment anyway and an alternative solution would be to remove the export macro entirely until more fine-grained control of exported symbols is needed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -525,9 +525,6 @@ endif()
 
 endif()
 
-include(GenerateExportHeader)
-generate_export_header(libhighs)
-
 # # Comment out for scaffold/ tests
 # add_subdirectory(scaffold)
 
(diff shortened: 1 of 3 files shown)


lgottwald:
The export is needed at that place because despite exporting all symbols the symbols of the above global variables are not exported automatically. They are indeed the only reason the header is generated and I am fine to drop the header if this is confirmed to work on all supported platforms as expected

lgottwald:
This will not be needed once #818 is merged

PR #750, opened 2022-02-21
[Depends on #749] Python Interface Updates

michaelbynum:
This PR adds a few more methods to the highspy python interface.

It will look like there are a ton of changes until #749 is merged.


michaelbynum:
The conflicts should be resolved.

PR #752, opened 2022-02-23
fix gap calculation for info_.mip_gap in callSolveMip

Baggins800:
It seems info_.objective_function_value accounts for objective sense and info_.mip_dual_bound is for minimization which leads to incorrect gap calculation for maximization problems see https://github.com/ERGO-Code/HiGHS/blob/8585b91291062b2fd95f13bcf9b488ce9b203834/src/lp_data/Highs.cpp#L2602-L2604.

### DIFF
--- a/src/lp_data/Highs.cpp
+++ b/src/lp_data/Highs.cpp
@@ -2600,8 +2600,8 @@ HighsStatus Highs::callSolveMip() {
   info_.mip_node_count = solver.node_count_;
   info_.mip_dual_bound = solver.dual_bound_;
   info_.mip_gap =
-      100 * std::abs(info_.objective_function_value - info_.mip_dual_bound) /
-      std::max(1.0, std::abs(info_.objective_function_value));
+      100 * std::abs(solver.primal_bound_ - info_.mip_dual_bound) /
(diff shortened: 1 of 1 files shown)


jajhall:
@Baggins800 Your fix corrects the MIP gap value, but it's not the way I'd like to see this corrected

jajhall:
The issue @lgottwald is that the primal and dual bounds relate to the equivalent minimization. 

For me the question is the point at which the sense of the objective should be reflected in MIP objective values

I can understand why it's a pain to modify all the MIP solver's reporting to reflect the objective sense, but
`solver.solution_objective_` incorporates the objective sense, as I copy this directly into
`info_.objective_function_value`
So
`info_.mip_dual_bound`
should incorporate the objective sense, otherwise it's inconsistent with `info_.objective_function_value`

lgottwald:
solver.solution_objective_ uses the objective value recalculated for the original model. for computing the gap solver.primal_bound_ should be used, but also, the gap is already computed before in cleanupSolve() so I suggest we simply retain the value there as the gap value that is recomputed there is also not using the correct formula. Also in cleanuSolve the bounds should be multiplied be adjusted for the objective sense.

PR #754, opened 2022-02-23
Check 752

jajhall:
I've extended MIP-maximize to expose errors in `info.mip_gap` and `info.mip_dual_bound` when solving maximization MIPs

@lgottwald, could you modify `cleanupSolve()` 
- to retain the value of `gap` so that it can be copied into `info.mip_gap` in `Highs::callSolveMip`
- to adjust the bounds for the objective sense so that `solver.dual_bound_` is correct when copied into `info.mip_dual_bound`  in `Highs::callSolveMip`
Thanks!

This will close the issue raised via #752

### DIFF
--- a/check/TestMipSolver.cpp
+++ b/check/TestMipSolver.cpp
@@ -153,10 +153,11 @@ TEST_CASE("MIP-maximize", "[highs_test_mip_solver]") {
   double optimal_objective;
   special_lps.distillationMip(lp, require_model_status, optimal_objective);
   // Add an offset to make sure this is handled correctly
-  const double offset = -20;
+  double offset = -20;
   lp.offset_ = offset;
   optimal_objective += offset;
   Highs highs;
+  if (!dev_run) highs.setOptionValue("output_flag", false);
   const HighsInfo& info = highs.getInfo();
   const HighsOptions& options = highs.getOptions();
   REQUIRE(highs.passModel(lp) == HighsStatus::kOk);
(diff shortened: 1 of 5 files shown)


lgottwald:
I adjusted the things requested. Additionally I changed the reporting to display the BestBound and BestSol column with respect to the original objective sense. I have no maximization problem at hand, but could you verify that the values reported during the solve in those columns matches the final printed value in the report, and that all have the correct sign?

jajhall:
The output looks OK, but the following isn't satisfied (in `TEST_CASE("MIP-maximize", "[highs_test_mip_solver]")`

`std::abs(info.objective_function_value - info.mip_dual_bound) <= options.mip_abs_gap`

Should it?

lgottwald:
I have a look what causes the difference

lgottwald:
Ah I see. The issue was the default mip_rel_gap of 1e-4 together with the rather large magnitudes of the objective function values of dcmulti

jajhall:
Ah, I assumed (as users might) that setting `options.mip_abs_gap` would guarantee that it's satisfied by the MIP solutio. IN other words, the MIP solver would satisfy both the relative and absolute gap values in `options`

lgottwald:
Hm. I assumed that when either tolerance is satisfied then optimality may be assumed

lgottwald:
I think it should be how it is now. Because otherwise you cannot set a mip_rel_gap without changing mip_abs_gap. The parameter description on gurobi also reads to me like that. The description says will terminate with optimal status if <criterion>, so if the criterion is reaches for either parameter, then HiGHS may terminate.

jajhall:
OK, that's fine. We just need to document it as Gurobi

jajhall:
I'm OK to merge this, now

lgottwald:
Me too

lgottwald (approved):

PR #761, opened 2022-02-23
C API: manually export the enums and document

### LINKED ISSUE #478: constants are not exported in the C API

odow:
Closes #478 

I don't know if there are any automated tools to keep these in sync. But it's probably not hard to manually update them if more enums are added.

### DIFF
--- a/src/interfaces/highs_c_api.h
+++ b/src/interfaces/highs_c_api.h
@@ -14,9 +14,58 @@
 #define HIGHS_C_API
 
 #include "util/HighsInt.h"
+
 const HighsInt HighsStatuskError = -1;
 const HighsInt HighsStatuskOk = 0;
 const HighsInt HighsStatuskWarning = 1;
+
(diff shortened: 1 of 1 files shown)


odow:
Okay I'll leave this branch as-is for you to finish how you like. I think, with these changes, the C API should be in a good state for HiGHS.jl.

jajhall:
Thanks: I'll merge this into od/document-c-api and check through the documentation.
Note that HiGHS constants always begin with k, so I'll change Highs*k$ to kHighs*$

jajhall:
Merged into od/document-c-api

PR #762, opened 2022-02-24
Od/document c api

jajhall:
Added some more `enums` as `const HighsInt` in `highs_c_api.h`, and using them in `TestCAPI.c` and `call_highs_from_c.c`. Worked through the documentation in `highs_c_api.h`.

### DIFF
--- a/examples/call_highs_from_c.c
+++ b/examples/call_highs_from_c.c
@@ -41,14 +41,14 @@ void minimal_api() {
   // After a successful call to Highs_lpCall, the primal and dual
   // solution, and the simplex basis are returned as follows
   //
-  // The vector x is colvalue
+  // The vector x is col_value
   // The vector Ax is row_value
   // The vector of dual values for the variables x is col_dual
   // The vector of dual values for the variables Ax is row_dual
   // The basic/nonbasic status of the variables x is col_basis_status
   // The basic/nonbasic status of the variables Ax is row_basis_status
   //
-  // The status of the solution obtained is modelstatus
+  // The status of the solution obtained is model_status
   //
   // The use of Highs_lpCall is illustrated for the LP
   //
--- a/examples/call_highs_from_c.c
+++ b/examples/call_highs_from_c.c
@@ -62,9 +62,6 @@ void minimal_api() {
   // bound on x_1, it serves to illustrate a non-trivial packed
   // column-wise matrix.
   //
-  const int kHighsObjSenseMinimize = 1;
-  const int kHighsObjSenseMaximize = -1;
-
   const int num_col = 2;
   const int num_row = 3;
   const int num_nz = 5;
--- a/examples/call_highs_from_c.c
+++ b/examples/call_highs_from_c.c
@@ -62,14 +62,14 @@ void minimal_api() {
   // bound on x_1, it serves to illustrate a non-trivial packed
   // column-wise matrix.
   //
-  const int sense_minimization = 1;
-  const int sense_maximization = -1;
+  const int kHighsObjSenseMinimize = 1;
+  const int kHighsObjSenseMaximize = -1;
 
   const int num_col = 2;
   const int num_row = 3;
   const int num_nz = 5;
   // Define the optimization sense and objective offset
-  int sense = sense_minimization;
+  int sense = kHighsObjSenseMinimize;
   const double offset = 3;
 
   // Define the column costs, lower bounds and upper bounds
@@ -104,8 +104,8 @@ void minimal_api() {
 			   col_basis_status, row_basis_status,
 			   &model_status);
   // The run must be successful, and the model status optimal
-  assert(run_status == 0);
-  assert(model_status == 7);
+  assert(run_status == kHighsStatusOk);
+  assert(model_status == kHighsModelStatusOptimal);
 
   printf("\nRun status = %d; Model status = %d\n", run_status, model_status);
 
@@ -122,7 +122,7 @@ void minimal_api() {
   printf("Optimal objective value = %g\n", objective_value);
 
   // Switch the sense to maximization and solve the LP again
-  sense = sense_maximization;
+  sense = kHighsObjSenseMaximize;
   run_status = Highs_lpCall(num_col, num_row, num_nz, a_format,
 			   sense, offset, col_cost, col_lower, col_upper, row_lower, row_upper,
 			   a_start, a_index, a_value,
@@ -130,8 +130,8 @@ void minimal_api() {
 			   col_basis_status, row_basis_status,
 			   &model_status);
   // The run must be successful, and the model status optimal
-  assert(run_status == 0);
-  assert(model_status == 7);
+  assert(run_status == kHighsStatusOk);
+  assert(model_status == kHighsModelStatusOptimal);
 
   printf("\nRun status = %d; Model status = %d\n", run_status, model_status);
 
@@ -158,8 +158,8 @@ void minimal_api() {
 			     col_value, row_value, 
 			     &model_status);
   // The run must be successful, and the model status optimal
-  assert(run_status == 0);
-  assert(model_status == 7);
+  assert(run_status == kHighsStatusOk);
+  assert(model_status == kHighsModelStatusOptimal);
 
   printf("\nRun status = %d; Model status = %d\n", run_status, model_status);
 
@@ -195,7 +195,7 @@ void minimal_api_qp() {
   const int num_row = 1;
   const int num_nz = 3;
   // Define the optimization sense and objective offset
-  int sense = 1;
+  int sense = kHighsObjSenseMinimize;
   const double offset = 0;
 
   // Define the column costs, lower bounds and upper bounds
@@ -206,12 +206,12 @@ void minimal_api_qp() {
   const double row_lower[1] = {1};
   const double row_upper[1] = {1.0e30};
   // Define the constraint matrix row-wise
-  const int a_format = 2;
+  const int a_format = kHighsMatrixFormatRowwise;
   const int a_start[2] = {0, 3};
   const int a_index[3] = {0, 1, 2};
   const double a_value[3] = {1.0, 1.0, 1.0};
 
-  const int q_format = 1;
+  const int q_format = kHighsHessianFormatTriangular;
   const int q_num_nz = 5;
   const int q_start[3] = {0, 2, 3};
   const int q_index[5] = {0, 2, 1, 0, 2};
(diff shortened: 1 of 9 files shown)


examples/call_highs_from_c.c
```diff
@@ -62,14 +62,14 @@ void minimal_api() {
   // bound on x_1, it serves to illustrate a non-trivial packed
   // column-wise matrix.
   //
-  const int sense_minimization = 1;
-  const int sense_maximization = -1;
+  const int kHighsObjSenseMinimize = 1;
+  const int kHighsObjSenseMaximize = -1;
```
odow: These should be removed

src/interfaces/highs_c_api.h
```diff
@@ -14,9 +14,74 @@
 #define HIGHS_C_API
 
 #include "util/HighsInt.h"
-const HighsInt HighsStatuskError = -1;
-const HighsInt HighsStatuskOk = 0;
-const HighsInt HighsStatuskWarning = 1;
+
+const HighsInt kHighsStatusError = -1;
```
odow, line 18: Two comments:
 * This is technically breaking for anyone who was using this in the C API
 * Why the `k` prefix? Is that standard somewhere? 

I had it as `[enum class][enum name]` which seemed to make sense to me. The class is `HighsStatus` and the enum is `kError`. It also makes all of the symbols exported by HiGHS in the C interface start with `Highs` which makes discoverability easier. But I'm not super attached to it, so `kHighs` is okay if you prefer.

src/interfaces/highs_c_api.h
```diff
@@ -25,110 +90,115 @@ extern "C" {
...
- * @param aindex    array of length [numnz] with indices of matrix entries
- * @param avalue    array of length [numnz] with values of matrix entries
- *
- * @param colvalue       array of length [numcol], filled with the primal column
- *                       solution
- * @param coldual        array of length [numcol], filled with the dual column
+ *                  three arrays, `a_start`, `a_index`, and `a_value`. `a_start`
+ *                  is an array of length [num_col] containing the starting
+ * index of each column in `a_index`. If `a_format` is `kRowwise` the array is
+ * of length [num_row] corresponding to each row.
```
odow: Indentation is wrong

src/interfaces/highs_c_api.h
```diff
@@ -25,110 +90,115 @@ extern "C" {
...
- *                       of the rows
- * @param modelstatus    termination status of the model after the solve
- *
- * @returns A non-zero return value indicates that a problem occured
- */
-HighsInt Highs_lpCall(const HighsInt numcol, const HighsInt numrow,
-                      const HighsInt numnz, const HighsInt a_format,
+ * @param col_basis_status  array of length [num_col], filled with the basis
+ *                          status of the columns in the form of a
+ * `HighsBasisStatus` enum
```
odow: Indentation is wrong

src/interfaces/highs_c_api.h
```diff
@@ -25,110 +90,115 @@ extern "C" {
...
+ * @param row_basis_status  array of length [num_row], filled with the basis
+ *                          status of the rows in the form of a
+ * `HighsBasisStatus` enum
```
odow: Indentation is wrong

src/interfaces/highs_c_api.h
```diff
@@ -25,110 +90,115 @@ extern "C" {
...
+ * @param model_status      termination status of the model after the solve in
+ * the form of a `HighsModel_Status` enum
```
odow: Indentation is wrong

src/interfaces/highs_c_api.h
```diff
@@ -236,83 +307,90 @@ HighsInt Highs_passLp(void* highs, const HighsInt numcol, const HighsInt numrow,
...
+ * @param num_col     the number of columns
+ * @param num_row     the number of rows
+ * @param num_nz      the number of elements in the constraint matrix
+ * @param q_num_nz    the number of elements in the Hessian matrix
+ * @param a_format    the format of the constraint matrix to use in th form of a
+ *                    `HighsMatrixFormat` enum
+ * @param q_format    the format of the Hessian matrix to use in the form of a
+ *                    `HighsHessianFormat` enum
+ * @param sense       the optimization sense in the form of a `HighsObjSense`
+ * enum
```
odow: Indentation is wrong

src/interfaces/highs_c_api.h
```diff
@@ -236,83 +307,90 @@ HighsInt Highs_passLp(void* highs, const HighsInt numcol, const HighsInt numrow,
...
+ * @param row_lower   array of length [num_row] with the upper row bounds
+ * @param row_upper   array of length [num_row] with the upper row bounds
+ * @param a_start     the constraint matrix is provided to HiGHS in compressed
+ *                    sparse column form (if `a_format` is `kColwise`, otherwise
+ *                    compressed sparse row form). The sparse matrix consists of
+ *                    three arrays, `a_start`, `a_index`, and `a_value`.
+ *                    `a_start` is an array of length [num_col] containing the
+ *                    starting index of each column in `a_index`. If `a_format`
+ *                    is `kRowwise` the array is of length [num_row]
+ * corresponding to each row.
```
odow: Indentation is wrong

src/interfaces/highs_c_api.h
```diff
@@ -236,83 +307,90 @@ HighsInt Highs_passLp(void* highs, const HighsInt numcol, const HighsInt numrow,
...
+ * @param a_index     array of length [num_nz] with indices of matrix entries
+ * @param a_value     array of length [num_nz] with values of matrix entries
+ * @param q_start     the Hessian matrix is provided in the same format as the
+ *                    constraint matrix, using `q_start`, `q_index`, and
+ *                    `q_value` in the place of `a_start`, `a_index`, and
+ * `a_value`. If the model is linear, pass NULL.
```
odow: Indentation is wrong

src/interfaces/highs_c_api.h
```diff
@@ -799,106 +878,107 @@ HighsInt Highs_addRow(void* highs, const double lower, const double upper,
...
 
 /**
  * Change the objective sense of the model.
  *
- * @param highs a pointer to the HiGHS model object
- * @param sense the new optimization sense. Use -1 for maximization and 1 for
- *              minimization.
+ * @param highs     a pointer to the Highs instance
+ * @param sense     the new optimization sense in the form of a `HighsObjSense`
+ * enum
```
odow: Indentation is wrong

src/interfaces/highs_c_api.h
```diff
@@ -1441,85 +1524,62 @@ HighsInt Highs_getHessianNumNz(const void* highs);
...
- * @param col_value array of length [m] with optimal primal solution for each
- *                  column
- * @param col_dual  array of length [n] with optimal dual solution for each
- *                  column
- * @param row_dual  array of length [m] with optimal dual solution for each row
+ * @param highs      a pointer to the Highs instance
+ * @param num_col    the number of variables
+ * @param num_row    the number of rows
+ * @param col_value  array of length [num_col] with optimal primal solution for
+ * each column
```
odow: Indentation is wrong

src/interfaces/highs_c_api.h
```diff
@@ -1441,85 +1524,62 @@ HighsInt Highs_getHessianNumNz(const void* highs);
...
+ * @param col_dual   array of length [num_col] with optimal dual solution for
+ * each column
```
odow: Indentation is wrong

src/interfaces/highs_c_api.h
```diff
@@ -1441,85 +1524,62 @@ HighsInt Highs_getHessianNumNz(const void* highs);
...
+ * @param row_dual   array of length [num_row] with optimal dual solution for
+ * each row
```
odow: Indentation is wrong

src/interfaces/highs_c_api.h
```diff
@@ -25,110 +90,115 @@ extern "C" {
...
  *
- * @param numcol    the number of columns
- * @param numrow    the number of rows
- * @param numnz     the number of nonzeros in the constraint matrix
- * @param a_format  the format of the constraint matrix
- * @param sense     the optimization sense. Use -1 for maximization, 1 otherwise
+ * @param num_col   the number of columns
+ * @param num_row   the number of rows
+ * @param num_nz    the number of nonzeros in the constraint matrix
+ * @param a_format  the format of the constraint matrix as a `HighsMatrixFormat`
```
odow:
```suggestion
 * @param a_format  the format of the constraint matrix as a `kHighsMatrixFormat`
```

We should use however we chose to name this. There are lots of other instances of enums as well.

odow (commented):
Name changes look good. 

* I think I prefer `HighsStatuskError` as it translates more directly to the C++ API, but if we change to `kHighsStatusError` then there are some updates to be made through the docstrings.
* I tried to flag the indentation issues, but I might have missed some

jajhall:
> Name changes look good.
> 
>     * I think I prefer `HighsStatuskError` as it translates more directly to the C++ API, but if we change to `kHighsStatusError` then there are some updates to be made through the docstrings.
> 
>     * I tried to flag the indentation issues, but I might have missed some
> Two comments:
>
>    * Why the k prefix? Is that standard somewhere?
>
> I had it as [enum class][enum name] which seemed to make sense to me. The class is HighsStatus and the enum is kError. It also makes all of the symbols exported by HiGHS in the C interface start with Highs which makes discoverability easier. But I'm not super attached to it, so kHighs is okay if you prefer.

I can see your logic, but the "k" still corrupts names, and HiGHS uses

https://google.github.io/styleguide/cppguide.html#Constant_Names

The "enums" in the C++ header files are named correctly in the docstrings. In C they aren't enums. Why don't the docstrings refer to the `const HighsInt` "families?" - or are these `const HighsInt` values considered to be enums in C?


>    * This is technically breaking for anyone who was using this in the C API

Indeed, and I see that they were in v1.1.1. We can keep these three, but document why they are different and indicate that they are deprecated

Sorry for all the indentation errors. Everything was in order until I ran `clang` formatting, which must have objected to some long lines. I'll fix these

odow:
Ah I didn't know that about the Google style guide. Happy to go with that then.

> or are these const HighsInt values considered to be enums in C?

Perhaps `enum` was the wrong word in C. 

> I'll fix these tomorrow, since it's 22:00 here, now.

I'll take a pass at sorting the issues. I should have permission to push to this branch.

jajhall:
Yes, @galabovaa championed the Google style guide. 
I've made a few changes just now, so I'll push them and leave you to it

jajhall:
So, Windows doesn't like 

const HighsInt foo = 0;
const HighsInt bar = foo;

Live and learn!

odow (approved):
I think this is in a good state.

jajhall:
Thanks. I'll merge once the tests go through. I've run `clang` formatting on my machine and it changes nothing

Issue #746, opened 2022-02-20
Experiment: new release binaries


odow:
We're experimenting with a new way of distributing binaries compiled for Julia. Here are three generic binaries for x86_64 machines:

**v1.2.2**

 * [highs_linux.tar.gz](https://github.com/ERGO-Code/HiGHS/files/8551106/highs_linux.tar.gz)
 * [highs_macOS.tar.gz](https://github.com/ERGO-Code/HiGHS/files/8551112/highs_macOS.tar.gz)
 * [highs_windows.tar.gz](https://github.com/ERGO-Code/HiGHS/files/8551113/highs_windows.tar.gz)

**v1.2.0**

 * [highs_linux.tar.gz](https://github.com/ERGO-Code/HiGHS/files/8104846/highs_linux.tar.gz)
 * [highs_windows.tar.gz](https://github.com/ERGO-Code/HiGHS/files/8104847/highs_windows.tar.gz)
 * [highs_macOS.tar.gz](https://github.com/ERGO-Code/HiGHS/files/8104848/highs_macOS.tar.gz)

If people, particularly on Windows machines, can try out these releases that'd be very helpful.

Everything is dynamically linked, but the tarball contains all of the linked files, so it should be relocatable with no issues. People using the executable can download the release, extract the contents, and then point their software to `/bin/highs`.

This should let people download and extract the binary without having to recompile: https://github.com/JuliaPackaging/Yggdrasil/pull/4456#issuecomment-1046218509
And we can host the releases on the HiGHS repository so we don't have people digging through weirdly named HiGHS_jll releases.

_As a reminder to myself, here's the code to construct releases_
```Julia
import JLLPrefixes
import Tar

for (name, platform) in Dict(
    "windows" => Base.BinaryPlatforms.Platform("x86_64", "windows"),
    "macOS" => Base.BinaryPlatforms.Platform("x86_64", "macOS"),
    "linux" => Base.BinaryPlatforms.Platform("x86_64", "linux"),
)
    prefix = JLLPrefixes.expanduser("/tmp/$name")
    artifact_paths = JLLPrefixes.collect_artifact_paths(
        ["HiGHS_jll"];
        platform = platform,
    )
    JLLPrefixes.copy_artifact_paths(prefix, artifact_paths)
    tarball = "highs_$(name).tar.gz"
    Tar.create(prefix, pipeline(`gzip -9`, tarball))
end
```

NPC:
[...]

PR #764, opened 2022-02-26
Improve C# API

tosttost:
* Implement IDisposable / dispose pattern. ~HighsLpSolver is only run if the GC is collecting the instance. IDisposable is the standard way for deterministic destruction in .NET and supported by 'using blocks' in C#. See https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose for details.
* Add the new getter/setter methods for options
* replace void* with IntPtr. Therefore it is no longer necessary to use "unsafe" code.
* discard the file extension from highslibname. Without the ".dll" it is possible to use the C# API on Linux as well, as the .NET runtime is clever enough to search for "libhighs.so" - see https://docs.microsoft.com/en-us/dotnet/standard/native-interop/cross-platform for details.
   * This does not work for "libhighs.so.1.2" -> symlink or similar required.
   * This does not work for "libhighs.dll" as produced by mingw. We could switch to "libhighs" - the binaries of @odow at https://github.com/ERGO-Code/HiGHS/issues/746 seem to be compiled with mingw. Visual C++ produces a "highs.dll" which would no longer be supported after prepending "lib"...

### DIFF
--- a/src/interfaces/highs_csharp_api.cs
+++ b/src/interfaces/highs_csharp_api.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Linq;
 using System.Runtime.InteropServices;
+using System.Text;
 
 // mcs -out:highscslib.dll -t:library highs_csharp_api.cs -unsafe
 
@@ -130,11 +131,12 @@ public class HighsBasis
    }
 }
 
-public unsafe class HighsLpSolver
+public class HighsLpSolver : IDisposable
 {
-   private void* highs;
+   private IntPtr highs;
+   private bool _disposed;
 
-   private const string highslibname = "highs.dll";
+   private const string highslibname = "highs";
 
    [DllImport(highslibname)]
    private static extern int Highs_call(Int32 numcol, Int32 numrow, Int32 numnz, double[] colcost,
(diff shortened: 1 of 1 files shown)


src/interfaces/highs_csharp_api.cs
```diff
@@ -130,173 +131,198 @@ public HighsBasis(HighsBasisStatus[] colbasisstatus, HighsBasisStatus[] rowbasis
...
-   private static extern double Highs_getObjectiveValue(void* highs);
+   private static extern int Highs_getIntOptionValue(IntPtr highs, string option, out int value);
 
    [DllImport(highslibname)]
-   private static extern int Highs_getIterationCount(void* highs);
+   private static extern int Highs_getDoubleOptionValue(IntPtr highs, string option, out double value);
 
    [DllImport(highslibname)]
-   private static extern int Highs_getModelStatus(void* highs);
+   private static extern int Highs_getStringOptionValue(IntPtr highs, string option, [Out] StringBuilder value);
```
tosttost, line 190: can't use string here because HiGHS is writing to the parameter.

jajhall: Sorry, but I understand little of this. I see that we've no getters and setters for the options, and maybe you indicate that we should have `Highs_getStringOptionValue`. But why delete `Highs_getModelStatus`? Also, are you saying that `string` can't be used in the parameter list of `Highs_getStringOptionValue` - that you created?

Unfortunately, no-one in the HiGHS team is a C# user, and we have little Windows expertise.

tosttost: Highs_getModelStatus is still there. Only "void*" has been replaced with IntPtr causing git to show the line as deleted and newly added (scroll down...).

The problem is that the internal representation of string in C# is incompatible with C-style 0-terminated strings. The .NET Runtime therefore has to 'marshall' string parameters. If a string is passed into a function this works by allocating a temporary char* buffer of suitable size, copying the content into that and then pass the buffer to C. All this happens automatically.

Furthermore strings in C# are immutable. Therefore it is impossible to let a C function manipulate the string without risking memory corruption. 
The StringBuilder is mutable and therefore usable if the C function is manipulating the string - the .NET Runtime can simply copy the contents of the char* buffer back into the StringBuilder after the function has returned.
It is not visible to the user anyway.

jajhall: > Highs_getModelStatus is still there. Only "void*" has been replaced with IntPtr causing git to show the line as deleted and newly added (scroll down...).

OK, that's clear enough. I now see [478d3c6](https://github.com/ERGO-Code/HiGHS/pull/764/commits/478d3c66de93155c864d405b7b268ad793c9d40d) and realise that you've made all the necessary changes. 

I'm happy to approve the changes to the C# API

As for the discussion on libraries, again it means nothing to me as I'm a very limited Windows user. However, since you're only changing the C#API I guess that these issues are still to be addressed.

jajhall (approved):
Only changes the C# interface, and for apparently sound reasons

PR #766, opened 2022-03-01
QP solver improvements

feldmeier:
Improvements for QP solver, e.g. renaming (QUadratic Active Set Solver -> Quass), now exiting gracefully when errors are encountered (negative Eigenvalues in reduced Hessian, degeneracy, ...), pivoting strategy can be chosen via settings rather than constructing some object, memory leaks fixed, two bugs in .lp reader fixed (offset, hessian dimension for semidefinite instances)


jajhall (approved):
Looks good!

Issue #765, opened 2022-02-28
Fail on (numerically challenging) binary expansion problems


joaquimg:
I have been trying to use HiGHS as the testing solver and readme solver in https://github.com/joaquimg/QuadraticToBinary.jl (current is Cbc).

This package generates binary expansions to solver QCQP as a MIP. Because of binary expansion, the problem does have numerical issues.

I gathered a few examples to showcase some fail of HiGHS. All of them come from binary expansions of the qp:
```
     minimize 2 x^2 + y^2 + xy + x + y + 1
        s.t.  x, y >= 0
              x + y = 1
```

They are named: `moiqp3qtbX.mps`, where `X` is a proxy on the precision, so higher `X` more terms in the binary expansion and less numerical stability, `X` varies from 3 to 8.

HiGHS seems to do well for `X=3`, `X=4` and `X=5`.
HiGHS seems to return wrong results for `X=6` (root node solution seems off) and `X=7` (this stops at presolve).
HiGHS returns Infeasible for `X=8`

Here are the files:
https://gist.github.com/joaquimg/56abb500c315d3d1251ddc939dd5529b

I tested those mps files on Xpress, which returns reasonable results for all, but prints that `X=8` might be numerically unstable (result precisely matches the QP solution though).

I have been using https://github.com/jump-dev/HiGHS.jl 1.0.0 and other instances seem fine (cc @odow)

jajhall:
Thanks, We'll have a look at these

lgottwald:
I had a look and the matrix values are simply too small for default feasibility tolerances starting from the X=6. From then they still solve fine when setting tightened feasibility tolerances by setting:
```
mip_feasibility_tolerance = 1e-9
primal_feasibility_tolerance = 1e-9
```
I need to have a closer look if I can be more careful in presolve to avoid these issues, but many things can happen with a default tolerance of 1e-6 and constraints like this:
```
x56 -5.960464477539063e-8 * x1 <= 0
```
Or bounds like this:
```
LO bounds    x3        0
UP bounds    x3        5.960464477539063e-8 
```

The variable x3 with above bounds is allowed to take values v with -1e-6 <= v <= 1.06e-6 in default feasibility tolerance. This changes the meaning of the column considerably as the range increases on the order of 1e2.
Is it possible to scale above columns differently when they need to have such small bound values?

jajhall:
[...]

PR #767, opened 2022-03-02
avoid using primal_feasibility_tolerance for MIP presolve, use mip_fe…

lgottwald:
…asibility_tolerance instead.

Realized this undesired behavior due to #765. For MIP behavior it should be sufficient to alter the mip_feasibility_tolerance option.

### DIFF
--- a/src/presolve/HPresolve.cpp
+++ b/src/presolve/HPresolve.cpp
@@ -94,10 +94,10 @@ void HPresolve::setInput(HighsLp& model_, const HighsOptions& options_,
   }
 
   if (mipsolver == nullptr) {
-    feastol = options->primal_feasibility_tolerance;
+    primal_feastol = options->primal_feasibility_tolerance;
     model->integrality_.assign(model->num_col_, HighsVarType::kContinuous);
   } else
-    feastol = options->mip_feasibility_tolerance;
+    primal_feastol = options->mip_feasibility_tolerance;
 
   if (model_.a_matrix_.isRowwise())
     fromCSR(model->a_matrix_.value_, model->a_matrix_.index_,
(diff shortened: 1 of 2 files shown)


jajhall (approved):
Fine by me, and consistent with the setting of `options.primal_feasibility_tolerance` in `HighsLpRelaxation.cpp`. 

I find `feastol` rather opaque, since `options.dual_feasibility_tolerance` is also used in presolve, so I suggest that `primal_feasibility_tol(erance)` is used. Indeed, since the word "primal" is only associated with feasibility (and the setting `kPrimalInfeasible`) in `HPresolve.cpp`, if you want something short, use `primal_tol`. For consistency, I suggest that you also introduce `dual_tol` for `options.dual_feasibility_tolerance`.

I also see some "magic numbers" in calculations like
`impliedBound > implRowDualLower[row] + 1000 * feastol` and ``
For security and readability, it would be better if these were defined constants

lgottwald:
The numbers are a bit magic, but also not entirely, as they are a relative threshold over the feasibility tolerance. It is not only used in HPresolve this way so I would replace this separately.

I renamed to primal_feastol now, to be also consistent with the naming of the field in the MIP solver class (feastol).

I did not introduce a similar field for the dual feasibility tolerance because I am not entirely sure about the proper intended behavior. The dual feasibility tolerance would be used for some dual reductions during presolve and it might be valuable, even for MIP to allow to adjust it differently via options.

For the primal_feasibility_tolerance, however, it overlaps with the mip_feasibility_tolerance, and for MIP clearly the latter should be used also presolve.

Do you think the mip_feasibility_tolerance should also be used for dual reductions when presolving a MIP?

jajhall:
Sorry, by `magic`, I don't mean that they aren't well considered, rather that they are hard-coded as specific values in multiple places. If you ever wanted to change this value, all occurrences of 1000 would have to be edited, and not all 1000's might relate to the same usage

> Do you think the mip_feasibility_tolerance should also be used for dual reductions when presolving a MIP?

No, I was just thinking that since you are setting `primal_feastol ` to `options.mip_feasibility_tolerance` or `options.primal_feasibility_tolerance` according to whether the problem being presolved is a MIP or LP, for clarity it would make sense to set a local value of `dual_feastol` to `options.dual_feasibility_tolerance`. However, it may well be that the dual tolerance is used in very different places, so there's no comparison to be made.

lgottwald:
Currently there is no analogy because MIP does not have a counterpart for the dual feasibility tolerance, hence there is no overlap of use that needs to be streamlined by a common variable dual_feastol. The intent of the field primal_feastol is to allow usage of the same primal tolerance for MIP and LP presolve and streamline the difference in how they are represented in the options. So for the dual feasibility we could do either of the following:

1. Define dual_feasibility_tolerance to affect LP and reductions based on LP-duality in LP and MIP presolve
2. Define dual_feasibility_tolerance to only affect LP and hard-code a dual_feasibility_tolerance value as relative factor to the MIP feasibility tolerance so that setting the option dual_feasibility_tolerance does not affect MIP presolve and mip_feasibility_tolerance is the only option affecting MIP.

I would currently favor option 1 because it seems more logical to me. It allows more control and also avoids arbitrarily using a dual_feasibility_tolerance value that is derived from the mip_feasibility_tolerance, as I feel there is no good argument of why that might be a good idea. There is only limited duality in MIP and implied bounds on the dual multipliers in MIP can only be derived from non-boxed continuous columns.

jajhall:
Thanks, let's go with option 1

PR #770, opened 2022-03-11
some changes in QUASS for benchmarking

feldmeier:
some default settings for QP solver, no printf output, correct passing of time limit


jajhall (approved):
Thanks for these, I'll merge and create v1.2.2

PR #774, opened 2022-03-18
Fix 771

jajhall:
Fixed erroneous reporting of dual infeasibilities after MIP optimality failure


jajhall:
clang formatting failure probably due to using newly installed clang format on markov

PR #779, opened 2022-03-24
Extend MPS free-format reader

svigerske:
I needed an MPS reader and decided to use the one from HiGHS. But I was missing some features and added them. Here they are.

- support `QCMATRIX` and proper support for `QSECTION`: quadratic entries for non-objective rows are stored in new `qrows_entries` vector
- add support for `CSECTION`: cones are stored in new `cone_*` members
- add support for SOS (`SOS` and `SETS` sections): SOS are stored in new `sos_*` members
- add flag to indicate whether variables that appear between `INTORG`/`INTEND` markers of `COLUMNS` section should be binary (most modern solvers and still the default) or integer (original IBM interpretation)
- fixed some compiler warnings in fixed-format reader


jajhall (approved):
Thanks for this, Stefan. I can't imagine that you've compromised anything by adding these. I'm a little puzzled that no CI tests have run, but I'll take the risk

PR #783, opened 2022-03-25
MPS free-format reader: create columns that appear after COLS section

svigerske:
This changes behavior such that when a column first appears in a `BOUNDS` or any of the quadratic matrix sections, it is created instead of ignoring the entry.
Otherwise problems where variables appear only in quadratic parts were probably read incorrectly (though there are warnings).

There was code for `BOUNDS` that was probably meant to create such columns already, but due to a `continue;` after the warning, I don't think it did what it meant to do.

Added some newlines that got lost in previous pull request.


jajhall (approved):
Well spotted!

jajhall:
@feldmeier This change looks as if it's most likely to affect QP. Could this have been occurring in any QP test problems that solved, but with the wrong objective?

Issue #786, opened 2022-03-29
read compressed MPS/LP files


svigerske:
It would be nice if HiGHS could read compressed input files.
I guess something like `zless prob.mps.gz | highs /dev/stdin` would already work (?).
But allowing `highs prob.mps.gz` may also be nice.

The header-only library from https://github.com/tmaklin/bxzstr, which makes several popular compression libraries available as C++ I/O  streams, looks promising. I could try to prepare a pull-request which adds this lib in `extern/`, adds checks to `CMakeLists.txt`, and uses it in the file readers, if that is of interest. bxzstr is under [Mozilla Public License](https://www.mozilla.org/en-US/MPL/2.0/FAQ/).

jajhall:
We've considered this, but I don't want to do things that make HiGHS complicated license-wise. In particular, we don't want to compromise our MIT status. 

However, I see that  `https://github.com/tmaklin/bxzstr` "is a fork of the zstr library by Matei David, and the core functionality of this library remains largely the same.", and `https://github.com/mateidavid/zstr` is MIT. So, if you can build on `zstr` - or are sure that using `bxzstr` is OK - please go ahead.

svigerske:
I preferred `bxzstr` because it seemed to support more compressions formats than `zstr`, i.e., bzlib and lzma in addition to zlib, and the MPL FAQ that I linked above, in particular Q7, seemed to require only that one include a link to the bxzstr code repo.   
But for my purposes, the zlib part is actually sufficient.

jajhall:
Thanks. It sounds OK to me to use bxzstr, then

mdhaber:
@jajhall would you like us to check on this? HiGHS is an important dependency of SciPy now, and MPL is a copyleft license, so I would rather be safe than sorry. 

Would the source code of `bxzstr` be distributed with HiGHS or just binaries?

svigerske:
I was just going start working on this. But I'll just be going with `zstr` for now. Even though `bxzstr` may be ok, the "we don't want to compromise our MIT status" argument above is convincing.

sschnug:
The file [extern/zstr/zstr_make_unique_polyfill.h](https://github.com/ERGO-Code/HiGHS/blob/eec5ec9bb1419ba0e097e5cf920f06ebda631be8/extern/zstr/zstr_make_unique_polyfill.h) in the current pull-request looks like it's GPL licensed, despite the source being MIT licensed in aggregate.

[...]

PR #787, opened 2022-03-31
read zlib-compressed (.gz) mps and lp files

svigerske:
This adds reading of `.lp.gz` and `.mps.gz` if zlib is available.

- Adds source files from [zstr](https://github.com/mateidavid/zstr) to `extern/zstr`
- Add check for zlib to cmake system. It is meant to be optional.
- Changed fixed-format MPS and LP reader to use C++ iostreams for reading. For the LP reader, I also changed the linebuffer to a `std::string` that can always hold a complete line (unless out-of-memory), because `fgets(FILE*)` does not match exactly to `std::istream::get()` and I gave up on trying to follow the logic on partially read lines.
- Switch from `std::ifstream` to `zstr::ifstream` if zlib is found in both MPS readers and LP reader. It would have been an even smaller modification if `zstr::ifstream::open()` would not throw an exception if opening a file fails.
- Use compressed version of `afiro.mps` in ctest if zlib is available.


jajhall:
Of course, we trust @svigerske to have written correct code for what he's introduced, so the request to review is more to highlight what he's done, and confirm that we're all happy with software engineering issues that I can't assess.

jajhall (commented):
This is a welcome extension of file-reading capability in HiGHS, and discussions in #786 satisfy me that our MIT status isn't compromised.

svigerske:
> Of course, we trust @svigerske to have written correct code for what he's introduced, so the request to review is more to highlight what he's done, and confirm that we're all happy with software engineering issues that I can't assess.

It works for me. What more do we want? :)

jajhall:
> It works for me. What more do we want? :)

I'm sure it works, but I'm wary of approving a PR that makes changes to code written by others, as they need to see what's changed.

lgottwald:
The changes all look reasonable. One thing to check would be the reading performance in case of large models stored in `.lp` format, to rule out bad effects on the reading performance due to the line buffering changes.

PR #789, opened 2022-04-05
fix handling of quadratics in constraints of .lp file

svigerske:
Adds a missing break to avoid an endless loop in case of a wrong syntax in an .lp file. Parsing now fails instead.

Handle that there is no "/2" following quadratic term in constraints, at least for Gurobi: https://www.gurobi.com/documentation/9.5/refman/lp_format.html


jajhall (approved):
Fine - and thanks for the spelling correction!

PR #790, opened 2022-04-05
finish SOS support in lp reader

svigerske:
This finishes the implementation on reading SOS in the LP reader.

Also make some local functions `static` to fix compiler warnings on my side.


jajhall (approved):
Fine

PR #793, opened 2022-04-06
open ifstream in zstr as binary always

svigerske:
Sorry that things come in small pieces.
I had problems reading compressed input on Windows with zstr. (There was an exception from `std::string` thrown.)

Turned out that this is due to mateidavid/zstr#15, i.e., not opening the input stream as binary. On Windows, there are some conversions of carriage-return (`\r`) happening in that case that cause trouble.

This PR changes the zstr library to always open the input stream in binary form.
As far as I understand zstr/issues#15, this means that `\r` in the input are not converted away.
Fortunately, your LP and MPS readers seem to handle this case:
- In the fixed-form MPS reader, deleting trailing whitespace includes removing `\r`, because `isspace('\r')` is true:  https://github.com/ERGO-Code/HiGHS/blob/master/src/io/HMPSIO.cpp#L474-L484
- In the free-form MPS reader, you use your own `trim()` function to remove white-space, which includes `\r`: https://github.com/ERGO-Code/HiGHS/blob/master/src/util/stringutil.h#L27
- In the LP reader, there is an explicit check for `\r`: https://github.com/ERGO-Code/HiGHS/blob/master/extern/filereaderlp/reader.cpp#L898-L900

### DIFF
--- a/extern/zstr/zstr.hpp
+++ b/extern/zstr/zstr.hpp
@@ -414,7 +414,11 @@ class ifstream
 {
 public:
     explicit ifstream(const std::string filename, std::ios_base::openmode mode = std::ios_base::in, size_t buff_size = default_buff_size)
-        : detail::strict_fstream_holder< strict_fstream::ifstream >(filename, mode),
+        : detail::strict_fstream_holder< strict_fstream::ifstream >(filename, mode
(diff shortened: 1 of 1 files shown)


svigerske:
Of course it had to fail on mac now.

jajhall (approved):
Thanks

PR #796, opened 2022-04-08
be more careful with numerical tolerances in singleton row presolve

### LINKED ISSUE #771: MIP solver reports feasible MIP as infeasible
Found by https://github.com/jump-dev/HiGHS.jl/issues/104

MPS file (note the big-M constraint):
```
NAME        
ROWS
 N  COST
 E  R0      
 L  R1      
COLUMNS
    MARK0000  'MARKER'                 'INTORG'
    C0        R1        -1000000
    MARK0001  'MARKER'                 'INTEND'
    C1        R0        1
    C1        R1        1
RHS
    RHS_V     R0        1
BOUNDS
 BV BOUND     C0      
 FR BOUND     C1      
ENDATA
```

Output
```
julia> using HiGHS_jll

julia> HiGHS_jll.highs() do exe
           run(`$(exe) bug.mps`)
       end
Running HiGHS 1.2.1 [date: , git hash: ]
Copyright (c) 2022 ERGO-Code under MIT licence terms
Number of BV entries in BOUNDS section is 1
MIP  bug has 2 rows; 2 cols; 3 nonzeros; 1 integer variables
Presolving model
0 rows, 0 cols, 0 nonzeros
0 rows, 0 cols, 0 nonzeros
Presolve: Optimal
WARNING: Untransformed solution with objective 0 is violated by 1 for the original model

Solving report
  Status            Optimal
  Primal bound      0
  Dual bound        0
  Solution status   infeasible
                    0 (objective)
                    0 (bound viol.)
                    0 (int. viol.)
                    1 (row viol.)
  Timing            0.00 (total)
                    0.00 (presolve)
                    0.00 (postsolve)
  Nodes             0
  LP iterations     0 (total)
                    0 (strong br.)
                    0 (separation)
                    0 (heuristics)
ERROR:   MIP solver claims optimality, but with num/sum/max primal(1/1/1) and dual(-1/inf/inf) infeasibilities
```

lgottwald:
This fixes #771

The issue was that the bigM constraint becomes a singleton row and then singleton row presolve computes the bounds [1e-6,1] on the integral bigM column. Since 1e-6 is within the feasibility tolerance it then decides to not tighten the bound.

To fix the issue I determine whether the tightened bound should be used by using the tolerance primal_feasibility_tolerance / abs(coefficient) when coefficient > 1 since this tighter tolerance must hold for the unscaled row. As a result the lower bound 1e-6 is still used and rounded to 1 for the integral column and the violation in the original space is avoided. Previously the singleton row was essentially scaled down by factor 1e-6 and then determined to be redundant.

### DIFF
--- a/src/presolve/HPresolve.cpp
+++ b/src/presolve/HPresolve.cpp
@@ -2572,7 +2572,7 @@ HPresolve::Result HPresolve::singletonRow(HighsPostsolveStack& postSolveStack,
   }
 
   // zeros should not be linked in the matrix
-  assert(std::abs(val) > options->small_matrix_value);
+  assert(std::fabs(val) > options->small_matrix_value);
 
   double newColUpper = kHighsInf;
   double newColLower = -kHighsInf;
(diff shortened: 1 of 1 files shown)


jajhall (approved):
Thanks for this. I'm puzzled on the use of abs vs fabs. I thought that abs was to be avoided, but then thought I'd learned that abs and fabs are the same when given a double argument.

Yet you used abs initially, and have now changed to fabs - suggesting that there's a difference

lgottwald:
In C++ the function `std::abs` has overloads for different argument types, so when the argument is double then it will be the same as `std::fabs`. In C code a source of errors can be that `abs` is used for a double argument, because in C there are no overloads and the double argument would be cast to an integer automatically and hence truncated. In C++ this might happen when including `<math.h>`, the C version of the header, instead of <cmath> and then using abs without the `std::` prefix.

Since most other code uses std::fabs and fabs states the intent clearly and avoids such errors I decided to switch to fabs for my code too. I do so for new code and whenever I touch a function or part of the code.

Issue #784, opened 2022-03-25
Add HiGHS version number to the log


NPC:
Hello again,

Hopefully a minor thing, it would be useful to have the library version number included in the library log output. If I run highs.exe it prints `Running HiGHS 1.2.0` at the top, but this information is missing if the library is called from, in my case, C#, where the first line in the log is “Presolving model”.

jajhall:
I see what you mean: it would be printed before the first write of logging information.

jajhall:
closed by #798

PR #798, opened 2022-04-09
logHeader reported when first non-const public Highs method is called

jajhall:
Will clear #784

### DIFF
--- a/app/RunHighs.cpp
+++ b/app/RunHighs.cpp
@@ -14,9 +14,9 @@
  * @brief HiGHS main
  */
 #include "Highs.h"
+//#include "io/HighsIO.h"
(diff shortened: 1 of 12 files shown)


lgottwald:
This breaks the MIP solver output. The log header should not be printed for every public method but only for a call to run(). Otherwise there is no possibilty to disable output or change any options before the header is printed.

lgottwald:
Maybe also for a call to presolve

jajhall:
The log header isn't printed when calling methods to set options, so it should still be possible to prevent any output. If it's only printed before `run()` then it'll appear in the middle of log output - which looks poor.

I'll see what happens with MIP output, though

Issue #799, opened 2022-04-10
Unreduced LP with model_presolve_status_ = HighsPresolveStatus::kReduced


jajhall:
For #797 I resurrected code written a long time ago, and found that a unit test failed. One of the LPs used in `TEST_CASE("Presolve", "[highs_test_presolve]")` (see ToDo @leona in `TestPresolve.cpp`) is not reduced by presolve, but `HighsPresolveStatus PresolveComponent::run()` returns `HighsPresolveStatus::kReduced` rather than `HighsPresolveStatus::kNotReduced`. 

For LPs where no reduction takes place, there is a slight inefficiency internally, as the simplex solver is called after applying the null postsolve stack, requiring a reinversion of the basis matrix. However, Highs::presolve() is designed for folk who just want to use the HiGHS presolve, and I feel that we should communicate whether no reductions were possible.

lgottwald:
fixed now

PR #800, opened 2022-04-10
Fix 797

### LINKED ISSUE #797: Write Highs::presolve() and Highs::postsolve()
Highs.h contains `HighsStatus presolve();` and `HighsStatus postsolve(const HighsSolution& solution, const HighsBasis& basis);` but the code for these methods isn't defined. They should exist.

jajhall:
This still requires #799 to be addressed, but will close #797


lgottwald (approved):
Looks fine

PR #801, opened 2022-04-11
Cmake options

galabovaa:
Fast build ON by default and disabled looking for a fortran or C sharp compiler by default.


jajhall (approved):
Fine by me: merge once CI tests have passed

PR #803, opened 2022-04-11
Highspy: Enable construction of HighsLp

michaelbynum:
This PR depends on #802.

This PR enables the construction of an empty `HighsLp` from Python.


jajhall (approved):
Very helpful!

PR #804, opened 2022-04-12
fixes for cmake after switching default to FAST_BUILD

lgottwald:
Fixes to use proper flags and to use zlib if available after changes in #801

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -105,9 +105,10 @@ else()
         HIGHS_HAVE_BUILTIN_CLZ)
 endif()
 
+include(CheckCXXCompilerFlag)
+
 if (NOT FAST_BUILD)
 # Function to set compiler flags on and off easily.
-include(CheckCXXCompilerFlag)
 function(enable_cxx_compiler_flag_if_supported flag)
     string(FIND "${CMAKE_CXX_FLAGS}" "${flag}" flag_already_set)
     if(flag_already_set EQUAL -1)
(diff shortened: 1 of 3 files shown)


lgottwald:
Also FAST_BUILD still has some issues. It does not set the githash or date and there is still #721 with difference in install behavior between FAST_BUILD and non FAST_BUILD. The double maintenance effort of these two options does not seem worthwhile, I propose we switch to use one of the two builds and improve it to cover all functionality.

lgottwald:
Also we may want to change this message:
```
Note: The HiGHS team is preparing for our first official release. If you
       experience any issues please let us know via email or on GitHub.
```

Users will also not usually change the default to this FAST_BUILD options. So all features users expect regarding interfaces should be enabled in FAST_BUILD or we'll get many more support requests why a feature stopped working.

lgottwald:
I think for v1.2.2 we should not make such a big switch of the default build which is not as much tested by every user setup. This fixes should go into the FAST_BUILD CMakeLists.txt nevertheless, but I can also base them on the version after revert.

lgottwald:
So I would propose to revert in master, reopen a PR which reverts the revert but don't merge that PR until v1.2.2 is released. I fear we would get new follow up issues from many users if we do such a switch of the build system in a minor release.

lgottwald:
I set default for FAST_BUILD to off for now and also added a fix for the extra print of the header in the MIP solver. We can switch to FAST_BUILD again of course but I think such a switch should be done with more testing as it changes a lot of things in the default build that users are currently relying on.

lgottwald:
Its possible to exclude targets from the all target without separating builds. E.g. we could create an interfaces target and exclude interfaces from the all target. Then typing "make" would not build interfaces, but typing "make interfaces" will.

I think the core part of building, linking, and testing should not be duplicated. Features that we do not want to build by default can get either excluded from the all target or be hidden behind an option that is more descriptive than FAST_BUILD with a default value off.

Not even I as an advanced user could list the full difference between FAST_BUILD on/off right now.

jajhall:
> Its possible to exclude targets from the all target without separating builds. E.g. we could create an interfaces target and exclude interfaces from the all target. Then typing "make" would not build interfaces, but typing "make interfaces" will.

This might well be something that `FAST_BUILD` could work towards, but it's for @galabovaa to manage

PR #805, opened 2022-04-12
Revert "Cmake options"

jajhall:
Reverts ERGO-Code/HiGHS#801


lgottwald:
Reverting to use the current build for v1.2.2 may be better to avoid unforeseen issues. We can do more thorough testing for a later release and switch to fast build there.

lgottwald:
I think revert is not necessary so I'll close

PR #806, opened 2022-04-12
only print log header for functions that produce info output

lgottwald:
printing the log header for all public functions is a bit overkill and produces unnessary overhead by adding an additional conditional branch for mundane things like changing a bound of a column. The intention of the change was to print the log header in the library and in the binary in the same way. Hence it should appear when a function is called that produces info output. There is info output for presolve(), run(), and a report on the model size for readModel(). For all other functions a `logHeader()` call does not seem necessary. This way the header should now be printed whenever a HiGHS instance is used to do something.

One thing that popped my eye is the crossover function. It uses cout where it should probably use `highsLogUser()`? Maybe there should be a call for the crossover function that accepts a solution too. The runCrossover call also contains uses of cout.

### DIFF
--- a/src/lp_data/Highs.cpp
+++ b/src/lp_data/Highs.cpp
@@ -39,19 +39,16 @@
 Highs::Highs() {}
 
 HighsStatus Highs::clear() {
-  this->logHeader();
   resetOptions();
   return clearModel();
 }
 
 HighsStatus Highs::clearModel() {
-  this->logHeader();
   model_.clear();
   return clearSolver();
 }
 
 HighsStatus Highs::clearSolver() {
-  this->logHeader();
   HighsStatus return_status = HighsStatus::kOk;
   clearPresolve();
   clearUserSolverData();
@@ -159,7 +156,6 @@ HighsStatus Highs::getOptionType(const std::string& option,
 }
 
 HighsStatus Highs::resetOptions() {
-  this->logHeader();
   resetLocalOptions(options_.records);
   return HighsStatus::kOk;
 }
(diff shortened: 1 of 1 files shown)


jajhall (changes requested):
The header must appear before any other output of any kind from HiGHS - except for error messages when setting options. Thus most of your changes need to be reverted

jajhall:
> One thing that popped my eye is the crossover function. It uses cout where it should probably use `highsLogUser()`? Maybe there should be a call for the crossover function that accepts a solution too. The runCrossover call also contains uses of cout

Indeed, this must use `highsLogUser()`

There is Highs::crossover(HighsSolution& solution);

lgottwald:
I think this is overkill because many methods can produce warnings but they usually indicate a coding error. In that case I'd want an internal interface for the MIP solver which does not need those extra checks for every little call. For passModel, its arguably ok. But I think the header should be printed before HiGHS does something for the first time. If there is an error before that it may look a bit ugly, but it is an error anyways. The MIP solver creates and uses many HiGHS instances and within the LP relaxation modifies it all the time. The logHeader function creates an additional conditional branch on EVERY modification of the LP relaxation and due to inlining the short logHeader function might be inlined in many of the function within the Highs class. This creates an additional conditional branch which can all clutter the branch-target buffer as individual branches. For bigger function  calls like calls to run(), presolve(), passModel(), or IO functions that might not be something significant.

For interface functions that are used a lot in the MIP solver, like changing column bounds, retrieving dual information, doing basis solves, these are undesirable for something as unimportant as an output header. The MIP solver uses many of these functions and since they occur within loops of loops calls them a lot.

The simplest solution to the original issue would be to not print the header automatically, but simply expose an API function to print the header for library users instead of blowing up every function of the Highs class and increasing its size further. Then any user of HiGHS as a library can call the API function first if such output is desired and every other user, including the HiGHS MIP solver, will not have more printed headers and checks per HiGHS instance.

lgottwald:
That makes sense. I fixed the MIP output in the other branch by first setting the output_flag in the temporary Highs instance that produced the second header. If the calls are avoided for modification functions and get functions things should be fine.

lgottwald:
Yes, starting over seems better. But the call in the executable would happen automatically now that it is done by read model, so I thought the explicit call could go

PR #808, opened 2022-04-13
Api function to release global resources

lgottwald:
I added an API function in the Highs class and the C API to allow resetting the global scheduler.

This is needed if the user wants to release all memory held by the highs library and is also needed for changing the number of threads after the scheduler has been initialized by a previous Highs call.


jajhall (approved):
Great, thanks. I'll add it to the Python interface in due course

PR #809, opened 2022-04-13
Fix 784 and introduce highspy

### LINKED ISSUE #784: Add HiGHS version number to the log

jajhall:
This closes #784, but would also merge highspy into master. Although this Python interface is not complete, it works, and there are more than enough methods to make it useful. The names of some methods will change. These are the ones used in cases where names are overloaded in the C++ class, but @michaelbynum has a solution for this. 

I know that an issue fix isn't the way to merge something major like highspy into master, but I merged highspy into fix-784 to check for conflicts without thinking


lgottwald:
I think merging highspy should be fine as it is a separate interface and any python functionality is better than none.

lgottwald (approved):

PR #811, opened 2022-04-14
highspy method overloading

michaelbynum:
This PR changes the names of `passLP`, `passModelPointers`, and `passLpPointers` to `PassModel` for consistency with the C++ interface. Additionally, it changes the name of `passHessianPointers` to `passHessian`. It is also possible to overload methods with a Pybind11 syntax like `.def("passModel", py::overload_cast<HighsModel>(&Highs::passModel))`, but we need the `highs_passModel` functions here anyway to convert the NumPy arrays.


michaelbynum:
It looks like the Python tests are not running for some reason...

michaelbynum:
Oh, they are only set to run on `push`. I'll update that.

jajhall (approved):
Thanks: looks fine to me

PR #814, opened 2022-04-16
No scaled model status 

jajhall:
The `scaled_model_status_` data member of the Highs class has been unnecessary since the simplex solver resolved primal/dual infeasibilities in the unscaled LP. Indeed, HiGHS ensured that `scaled_model_status_` was identical to `model_status_`. 

In removing `scaled_model_status_`, `const HighsModelStatus& getModelStatus(const bool scaled_model = false)` no longer needs a parameter, so this method is now deprecated (it returns `model_status_`) and is replaced by `const HighsModelStatus& getModelStatus()`.

Occurrences of `getModelStatus(true/false)` across the codebase have been replaced by calls to `getModelStatus()`, including some in presolve and the MIP solver, @lgottwald > However, since `scaled_model_status_` was identical to `model_status_`, this should have no effect.


lgottwald (approved):

lgottwald:
looks fine

PR #817, opened 2022-04-19
FreeBSD fix

yurivict:


svigerske (approved):
Should be here: Not that I can test this, but code that is good on `__APPLE__` has a good chance to be good for `__FreeBSD__` as well.

jajhall (approved):
Fine by me is @svigerske and @lgottwald are OK with it

lgottwald:
I am also fine with it

Issue #810, opened 2022-04-14
Thread Safety


Downsite:
I am on commit  "71451792f302a4d38dfa90d1071491116306d797" (the last time I tried to update, my CMake script was unable to find the HiGHS headers suddenly).

When I try to run different instances of HiGHS inside of std::async, I receive a SEGVAULT inside the file

```
Thread 3 "X" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffeffff640 (LWP 48347)]
0x0000555555c12b6a in HighsSplitDeque::getCurrentHead (this=0x0) at /home/X/dep/HiGHS/src/./parallel/HighsSplitDeque.h:481
481	  int getCurrentHead() const { return ownerData.head; }
(gdb) bt
#0  0x0000555555c12b6a in HighsSplitDeque::getCurrentHead (this=0x0) at /home/X/dep/HiGHS/src/./parallel/HighsSplitDeque.h:481
#1  0x0000555555c12d99 in highs::parallel::TaskGroup::TaskGroup (this=0x7fffefff85e0) at /home/X/dep/HiGHS/src/./parallel/HighsParallel.h:75
#2  0x0000555555c0be61 in HighsMipSolverData::evaluateRootNode (this=0x7fffe809b140) at /home/X/dep/HiGHS/src/mip/HighsMipSolverData.cpp:1198
#3  0x0000555555bf0ae7 in HighsMipSolver::run (this=0x7fffefff90d0) at /home/X/dep/HiGHS/src/mip/HighsMipSolver.cpp:122
#4  0x0000555555b8f499 in Highs::callSolveMip (this=0x7fffefff9d90) at /home/X/dep/HiGHS/src/lp_data/Highs.cpp:2761
#5  0x0000555555b84afa in Highs::run (this=0x7fffefff9d90) at /home/X/dep/HiGHS/src/lp_data/Highs.cpp:804
```

Interestingly, I also encounter the same problem if I encapsulate all calls to HiGHS functions in a single function and use std::lock_guard to avoid it being called by multiple threads at the same time.

Outside of the std::async, I got the following error trying to call the HiGHS-containting-function (still with barriers around) from different threads with different settings for threads.

`Option 'threads' is set to 2 but global scheduler has already been initialized to use 1 threads. The previous scheduler instance can be destroyed by calling HighsTaskExecutor::shutdown().`

Is there something about using HiGHS in parallel that I should be aware of?

Downsite:
[...]

PR #818, opened 2022-04-20
create a distinct executor for each user thread that initializes the …

lgottwald:
…HiGHS scheduler

This should fix the issue in #810 although I need more testing. It means when `Highs::run()` is called from a new thread where it never has been called before, then the scheduler will be initialized for the thread number in that Highs instance and subsequent calls to Highs::run() in that same thread will reuse the scheduler instance. When `Highs::run()` is now called from a different user thread then it will not anymore interact with scheduler instances used on other user threads.

When a user thread exits for which a Highs scheduler was initialized then it should automatically call the shutdown() function causing all background threads attached to that scheduler instance to also exit and all memory to be freed once the last of that scheduler instance exits.

Also this removed the need for the generated export header which I removed from cmake in this branch.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -533,9 +533,6 @@ endif()
 
 endif()
 
-include(GenerateExportHeader)
-generate_export_header(libhighs)
-
 # # Comment out for scaffold/ tests
 # add_subdirectory(scaffold)
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -401,7 +401,6 @@ foreach ( file ${headers} )
     install( FILES ${file} DESTINATION include/${dir} )
 endforeach()
 install(FILES ${HIGHS_BINARY_DIR}/HConfig.h DESTINATION include)
-install(FILES ${HIGHS_BINARY_DIR}/libhighs_export.h DESTINATION include)
 
 if (IPX_ON)
     if (UNIX)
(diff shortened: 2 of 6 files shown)


jajhall (approved):
I can't comment on the technical correctness, but it sounds as if @lukas-weber and @lgottwald are OK with it. Since it touches CMake, what's your view @galabovaa ?

lgottwald:
For cmake it only removes the export header that I added before, no changes otherwise.

For my local tests things seem to work fine. Although I don't know for sure whether all the problems in the issue are resolved I would be fine to already merge this, as it is definitely an improvement.

One other solution I considered is to register different user threads inside the scheduler which would allow work balancing between Highs instances running on different user threads. This would, however, be a much bigger change and seemed to get too messy. It would also mean that a Highs instance set to use 1 thread which runs on a user thread within a users multi-threaded application would possibly use additional threads created by a different Highs instance running on another user thread in the same application or join in to complete tasks for solving the instance running on a different thread while waiting for the completion of a stolen sub-task.

Since there is no control on our side about how much threads a user may create I believe the way it is done here is best. It will now create n-1 background threads for each user thread when `n` is the value of the `threads` option. Using different values of the threads option for different user threads is now allowed and each user thread will only use those background threads that are created in its own private scheduler instance.

For sharing work between different Highs instances users can still opt in by explicitly using the parallel facilities of the `parallel/HighsParallel.h` header without creating threads themselves, e.g. for solving a set of problems in a batch which may be unbalanced in size and each use multi-threading for their individual solve.

jajhall:
Thanks for giving this such careful consideration. Ultimately, some combination of user, architecture and OS will lead to a performance hit, so I think it's good to stick with something simple - with warnings in the documentation - so that when things go wrong we have a good chance of understanding why

PR #825, opened 2022-04-25
Task interrupt

lgottwald:
This fails right now on mac os x, because the thread_local variables in HighsTaskExectuor.cpp are missing from IPX and on OS X the creation of the IPX library without the C++ file that contains the declarations is forbidden.

This could be fixed in several ways, and I don't know what is preferred. One way would be to get rid of the separate IPX library and instead add the IPX sources to the HiGHS sources and into the HiGHS library. Maybe the issue also goes away when IPX is always compiled as static library, in which case I think OS X will not complain. Another possibility is to compile the single parallel source file HighsTaskExecutor.cpp into a separate library and then link that library to libipx and to libhighs.

For simplicity I'd prefer anything but the solution with another extra library. Besides I think IPX can only be used via the Highs class because the call that I added that checks for an interrupt requires the scheduler to be initialized, which only happens in the Highs class.


jajhall:
Thanks for this. The separate IPX and basiclu libraries are a throwback to when Lukas had a non-MIT license for them and we wanted to be able to issue HiGHS without IPX. However, not only did Lukas agree to removing this license, but IPX is now fully integrated into HiGHS due to its use in Highs::crossover() and to solve nodes in the MIP solver.

Having IPX compile like the rest of HiGHS is handy because this may solve an issue I've got in a development branch where I'm using HFactor, and it fails to build in the MacOS CI tests due to not finding `HFactor.h`

jajhall (approved):
This is OK with me

lgottwald:
I now removed all IPX_ON defines and the IPX option from CMake in this branch and also fixed the issue of HiGHS running past the time limit on supportcase19

Issue #807, opened 2022-04-12
Quadratic Problem to .lp file bug?


Downsite:
Hi,

Testing my implementation of the QP in thhe C++ API, I tried to solve the problem
```
variables
 x within [-2,2];
 y within [-2,2];

minimize:
x*y+x*x;

constraints:
x>=-1;
```
when I write the Hessian correctly(?) as the triangular and sparse version of [2 1;1 0]
the output written by the file writer is incorrect (the returned solution is correct).

```
\ File written by Highs .lp filereader
min
 obj: +0 x1 +0 x2 + [ +2 x0 * x0 +1 x0 * x1 +0 x1 * x1  ]/2 
st
 con1lo: +1 x1 >= -1
bounds
 -2 <= x1 <= +2
 -2 <= x2 <= +2
bin
gen
semi
end
```

feldmeier:
Thanks - this will be fixed in #828, it make take a bit to be merged thoguh

PR #828, opened 2022-04-27
QP updates

feldmeier:


jajhall:
This will close #807

jajhall (approved):
This is fine now that the unassigned read is avoided

PR #834, opened 2022-05-01
Enable concurrent compilation under MSVC

metab0t:
https://docs.microsoft.com/en-us/cpp/build/reference/mp-build-with-multiple-processes?view=msvc-170


jajhall:
This looks uncontroversial. What do you think @galabovaa ?

galabovaa (approved):

PR #835, opened 2022-05-01
Minor fixes

lgottwald:


lgottwald:
fixes a memory leak in the zstr header and adds a minor followup change which I held back because I wanted to test it and it does not affect many instances.

jajhall (approved):
Fine - and CI pass!

PR #839, opened 2022-05-03
Allow LP file parse to handle var*var in expressions

arcondello:
Previously LP files with the format
```
Minimize
 10 x1 + [ 6 x0*x1 ] / 2
End
```
would have `x0*x1` treated as a variable, rather than as a multiplication of two variables.

I used the LP reader code as a basis for another project (https://github.com/dwavesystems/dimod/pull/1182). Would have loved to just use y'all as a submodule, but we needed to make pretty invasive changes. Nonetheless, while doing so discovered this bug, so thought I would make the change upstream as well.

### DIFF
--- a/check/CMakeLists.txt
+++ b/check/CMakeLists.txt
@@ -43,6 +43,7 @@ set(TEST_SOURCES
     TestHighsModel.cpp
     TestHSet.cpp
     TestLogging.cpp
+    TestLPFileFormat.cpp
     TestLpValidation.cpp
     TestLpModification.cpp
     TestLpOrientation.cpp
--- a/check/TestLPFileFormat.cpp
+++ b/check/TestLPFileFormat.cpp
@@ -1,0 +1,13 @@
+
+#include "catch.hpp"
+#include "Highs.h"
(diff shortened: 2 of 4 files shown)


arcondello:
I would be happy to add better testing, pointers to an example of how to read the coefficients off the loaded model would be appreciated!

jajhall:
This looks uncontroversial to me, @feldmeier, but you're the `.lp` file expert so I'd want you to give it your blessing

jajhall:
> I used the LP reader code as a basis for another project ([dwavesystems/dimod#1182](https://github.com/dwavesystems/dimod/pull/1182)). Would have loved to just use y'all as a submodule, but we needed to make pretty invasive changes. Nonetheless, while doing so discovered this bug, so thought I would make the change upstream as well.

> I would be happy to add better testing, pointers to an example of how to read the coefficients off the loaded model would be appreciated!

Many thanks!

feldmeier:
Thanks - I'm a bit surprised this error didn't show up earlier. Nice to see my code being used in quantum computing - even if its just adibiatic QC!

feldmeier (approved):
Looks good to me

Issue #840, opened 2022-05-03
Trying to load an LP file without a newline at the end of the file leads to out-of-memory errors


arcondello:
I built the solver from the current [master](https://github.com/ERGO-Code/HiGHS/tree/d1d9839b83800ac04eb4a0bb50d64e2c3efed63c). When I used `highs pass.lp` on [pass.lp](https://gist.github.com/arcondello/3f4a71bd5742952b5ede0546edb1e423) it works fine, but running `highs fail.lp` on [fail.lp](https://gist.github.com/arcondello/a37cb4f4eb6292fb1e8f3f1ee1d4e0e3) results in running out of memory. The difference between the two is the newline at the end of the file.

Reverting to [v1.2.1](https://github.com/ERGO-Code/HiGHS/releases/tag/v1.2.1) resolves the issue

jajhall:
Strange. Please would you look at this @feldmeier

PR #841, opened 2022-05-04
Fix EOF detection in LP file reader

fuglede:
The condition `this->linebufferpos == this->linebuffer.size()` would never be `true` except potentially (but not necessarily) at the end of the file, since whenever an `'\n'` is encountered, `this->linebufferpos` is manually set to correspond to the end of the line (on line 998).

Moreover, the condition would mean that an actual EOF would not be properly handled when the final line of the file is not empty, which would instead cause an infinite loop as the final line is read over and over. In particular, this infinite loop is encountered for the example given in gh-840:

```
4d 69 6e 69 6d 69 7a 65 0a 20 78 30 20 2b 20 78 31 0a 73 75 62 6a 65 63 74 20 74 6f 0a 20 63 31 3a 20 78 30 20 2b 20 78 31 20 3c 3d 20 31 0a 45 6e 64
```

Here we just move the EOF check to before reading a new line.

This closes gh-840.

Now at the same time, it certainly seems like the condition wouldn't be put there for no reason, so maybe I'm missing something.

### DIFF
--- a/extern/filereaderlp/reader.cpp
+++ b/extern/filereaderlp/reader.cpp
@@ -903,8 +903,8 @@ void Reader::readnexttoken(bool& done) {
      this->linebufferpos = 0;
    }
 
-   // if all line has been read and we are at end of file, then stop
-   if (this->linebufferpos == this->linebuffer.size() && this->file.eof()) {
+   // if we are at end of file, then stop
+   if (this->file.eof()) {
      this->rawtokens.push_back(std::unique_ptr<RawToken>(new RawToken(RawTokenType::FLEND)));
      done = true;
      return;
--- a/extern/filereaderlp/reader.cpp
+++ b/extern/filereaderlp/reader.cpp
@@ -892,7 +892,12 @@ void Reader::readnexttoken(bool& done) {
    done = false;
 
    if (this->linebufferpos == this->linebuffer.size()) {
-     // read next line
+     // read next line if any are left. 
+     if (this->file.eof()) {
+         this->rawtokens.push_back(std::unique_ptr<RawToken>(new RawToken(RawTokenType::FLEND)));
+         done = true;
+         return;
+     }
      std::getline(this->file, linebuffer);
 
      // drop \r
@@ -901,13 +906,6 @@ void Reader::readnexttoken(bool& done) {
 
      // reset linebufferpos
      this->linebufferpos = 0;
-   }
-
-   // if we are at end of file, then stop
-   if (this->file.eof()) {
-     this->rawtokens.push_back(std::unique_ptr<RawToken>(new RawToken(RawTokenType::FLEND)));
-     done = true;
-     return;
    }
 
    // check single character tokens
--- a/check/TestFilereader.cpp
+++ b/check/TestFilereader.cpp
@@ -50,6 +50,15 @@ TEST_CASE("filereader-edge-cases", "[highs_filereader]") {
   run_status = highs.run();
   REQUIRE(run_status == HighsStatus::kOk);
 
+  // Load an LP file that does not end with a newline
+  model = "no-newline-eof";
+  model_file = std::string(HIGHS_DIR) + "/check/instances/" + model + ".lp";
+  return_status = highs.readModel(model_file);
+  REQUIRE(return_status == HighsStatus::kOk);
+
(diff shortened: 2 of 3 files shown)


extern/filereaderlp/reader.cpp
```diff
@@ -903,8 +903,8 @@ void Reader::readnexttoken(bool& done) {
      this->linebufferpos = 0;
    }
 
-   // if all line has been read and we are at end of file, then stop
-   if (this->linebufferpos == this->linebuffer.size() && this->file.eof()) {
```
fuglede, line 907: So, one case this doesn't handle is what happens if "End" is not on its own line. I guess that's no good.

Edit: Fixed by simply making the check before reading another line.

extern/filereaderlp/reader.cpp
```diff
@@ -892,7 +892,12 @@ void Reader::readnexttoken(bool& done) {
    done = false;
 
    if (this->linebufferpos == this->linebuffer.size()) {
-     // read next line
+     // read next line if any are left. 
+     if (this->file.eof()) {
+         this->rawtokens.push_back(std::unique_ptr<RawToken>(new RawToken(RawTokenType::FLEND)));
+         done = true;
```
fuglede, line 898: The value of `done` is never actually used for anything; I can remove that in a separate PR if relevant.

jajhall: Thanks, but let's leave that to @feldmeier , who might want it left in for future use

svigerske (approved):
Seems like I messed this up.
I didn't think that `std::getline(this->file, linebuffer);` would leave `linebuffer` untouched when at eof.    
So an alternative fix would be
```diff
--- a/extern/filereaderlp/reader.cpp
+++ b/extern/filereaderlp/reader.cpp
@@ -893,6 +893,7 @@ void Reader::readnexttoken(bool& done) {
 
    if (this->linebufferpos == this->linebuffer.size()) {
      // read next line
+     linebuffer.clear();
      std::getline(this->file, linebuffer);
 
      // drop \r
```
But moving the eof check is also good (maybe also better).

Adding `fail.lp` from #840 to the unittests would be nice.

fuglede:
> Adding fail.lp from https://github.com/ERGO-Code/HiGHS/issues/840 to the unittests would be nice.

I've added such a test to `TestFilereader.cpp`.

jajhall:
> So an alternative fix would be
> 
> ```diff
> --- a/extern/filereaderlp/reader.cpp
> +++ b/extern/filereaderlp/reader.cpp
> @@ -893,6 +893,7 @@ void Reader::readnexttoken(bool& done) {
>  
>     if (this->linebufferpos == this->linebuffer.size()) {
>       // read next line
> +     linebuffer.clear();
>       std::getline(this->file, linebuffer);
>  
>       // drop \r
> ```
> 
> But moving the eof check is also good (maybe also better).

What do you reckon @fuglede ?

fuglede:
> What do you reckon @fuglede ?

I'm probably biased since this solution is the one that happened to be what I found natural first. Perhaps one can argue that sticking to `eof` makes semantics clearer?

Another thing that could also work is

```cpp
     std::getline(this->file, linebuffer);
     if (!this->file)
     {
         this->rawtokens.push_back(std::unique_ptr<RawToken>(new RawToken(RawTokenType::FLEND)));
         done = true;
         return;
     }
```

extern/filereaderlp/reader.cpp
```diff
@@ -892,9 +892,14 @@ void Reader::readnexttoken(bool& done) {
    done = false;
 
    if (this->linebufferpos == this->linebuffer.size()) {
-     // read next line
+     // read next line if any are left. 
      std::getline(this->file, linebuffer);
-
+     if (this->file.eof())
```
svigerske: Doesn't this now skip parsing the content of the last line, if not empty?

I was also ok with the previous solution.

fuglede: You're right; `eof` triggers on the last line, even if it has contents. I guess it's safe anyway given that the end marker appears on a line of it's own, but relying on that seems iffy. Let me just revert the last commit.

feldmeier:
thanks for fixing this, changes look fine to me

PR #845, opened 2022-05-06
Remove unused parameter from Reader::readnexttoken

fuglede:
This removes the unused parameter `done` from the LP file reader, but cf. [this comment](https://github.com/ERGO-Code/HiGHS/pull/841/files#r864694671), maybe it's there for future use? Cc @feldmeier


jajhall (approved):
Uncontroversial

PR #849, opened 2022-05-10
Fix header in "Get started" section of documentation

fuglede:
Currently, the page looks as follows:

image

This change removes the duplicated header and fixes the formatting of the second one.


jajhall (approved):
This seems uncontroversial

PR #862, opened 2022-05-16
MPSFF: create colums that appear first in BOUNDS section

svigerske:
Changes the MPS reader to create again columns when they are mentioned in the BOUNDS section first.
Fixes reading of
```
NAME Latecol
OBJSENSE
 MAX
ROWS
 N obj
COLUMNS
RHS
BOUNDS
 LO bnd x1 0
 UP bnd x1 2
 LO bnd x2 0
 UP bnd x2 4
QMATRIX
 x1 x1 2
 x1 x2 2
 x2 x1 2
 x2 x2 2
ENDATA
```

### DIFF
--- a/src/io/HMpsFF.cpp
+++ b/src/io/HMpsFF.cpp
@@ -1251,11 +1251,11 @@ HMpsFF::Parsekey HMpsFF::parseBounds(const HighsLogOptions& log_options,
     // if not existing yet
     HighsInt colidx = getColIdx(marker, false);
     if (colidx < 0) {
-      highsLogUser(
-          log_options, HighsLogType::kWarning,
-          "Column name \"%s\" in BOUNDS section is not defined: ignored\n",
-          marker.c_str());
-      continue;
+      // add new column if did not exist yet
+      colidx = getColIdx(marker, true);
+      assert(colidx == num_col-1);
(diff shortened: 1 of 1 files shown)


jajhall (approved):
Thanks. I'd never thought of a model having an empty columns section, but this example illustrates why. The N-row isn't used: is it necessary to have one?

svigerske:
Looks like nothing is changing if the N-row is omitted.

jajhall:
I note that gurobi doesn't read this MPS file, since it ignores the column definition in the BOUNDS section, and then complains 

Error 10012: MPS read error at line 14:
 x1 x1 2

I guess they are expecting folk to define the columns with zero cost in the COLUMNS seciton - which is inefficient.

Not saying that we need to do the same, and I can't see any downside as the MPS file is a reasonable representation of a box constrained pure QP

jajhall:
Of course, it's a non-convex problem, so HiGHS can't solve it :D

I also see that gurobi demands a ROWS and COLUMNS section. Again, I don't see why it's necessary

svigerske:
I actually never thought about solving this instance :).

For LP/MIP people, it sometimes seems to be difficult to imagine that there is problem without A. :stuck_out_tongue:

PR #880, opened 2022-05-22
Added fix to omit cost row in Glpsol solution file only if empty and artificial

jajhall:
Gets around another crazy quirk of Glpsol


jajhall (commented):
Minor change in code orthogonal to main HiGHS usage

PR #887, opened 2022-05-25
Added some comments and another test in TestAlienBasis.cpp - for a client

jajhall:
Should be utterly uncontroversial


jajhall (commented):
Passes CI tests

PR #890, opened 2022-06-03
Presolve cleanup

jajhall:
Much redundant code stripped out of presolve


lgottwald:
So I finally looked through this and it looks fine. One last thing that can be removed is HAggregater.{h,cpp}. Will do some more tests.

jajhall:
Thanks. I think I remember identifying that this could be done, but clearly I forgot to do it.

PR #892, opened 2022-06-09
Wheel

galabovaa:
Python wheel with C++ lib included builds and loads on macOS Big Sur, next check other platforms and CI


jajhall (approved):
Looks good to me

PR #898, opened 2022-06-26
Use CMake ZLib target for CMake config interoperability

Dekker1:
This is a rather simple suggested change that uses the `ZLIB::ZLIB` target provided by the [FindZLIB](https://cmake.org/cmake/help/latest/module/FindZLIB.html) module, rather than the location of the include directories and libraries directly.

While setting up an interface between MiniZinc and HiGHS, I ran into some problems that I could not use the HiGHS version compiled by our CI on my local computers. Looking through the exported CMake configuration I found that it was linking explicitly against the ZLib location on the CI machine. I believe this would be the suggested way to resolve this issue. The machine using the exported CMake configuration can just again use `find_package(ZLIB)` for the `ZLIB::ZLIB` target to be defined within their usage of highs.

Something that we have done for some MiniZinc dependencies that could maybe even improve on this, but that is a bit more work, is to have a custom `*Config.cmake`. Then you can include a `find_dependecy(ZLIB)` call (when HiGHS is compiled with ZLib) using [CMakeFindDependencyMacro](https://cmake.org/cmake/help/latest/module/CMakeFindDependencyMacro.html). Then users of the exported CMake package do not have include ZLib themselves.


Dekker1:
It seems I overlooked the fact that there is already a `highs-config.cmake.in` template in which the `find_depency` call could be placed. I'll adjust this PR accordingly

galabovaa (approved):

PR #900, opened 2022-06-27
fix some compiler warnings

svigerske:
I compile HiGHS 1.2.2 with GCC 12 and ` -Wall -Wreturn-type -Wmissing-declarations -Wno-sign-compare -Wno-unused-variable -Wno-format-truncation`. This PR fixes some warnings that still come up.
In particular, functions are declared static if in cpp and local, or the signature in the corresponding declaration is fixed, or is declared inline if in header. Some include paths are fixed. The pretty annoying `-Wnon-null` warning from the hashing code is avoided. All cases are handled in switch statements.

Regarding the last, there seems to be another issue. In `HighsMipSolverData::runSetup()` there was a switch on variable types which doesn't handle the cases of semicontinuous and semiinteger variables. I added that there are handles as if continuous/integer to squash the warning, but, not surprisingly, I get wrong results on a MIP with these variables. It somehow looks as if HiGHS allows to input semicontinuous/integer variables, but does not handle them appropriately further on. Shouldn't it reject such problem instances at least?

### DIFF
--- a/src/qpsolver/ratiotest.cpp
+++ b/src/qpsolver/ratiotest.cpp
@@ -98,10 +98,10 @@ RatiotestResult ratiotest(Runtime& runtime, const Vector& p,
                           const Vector& rowmove, double alphastart) {
   switch (runtime.settings.ratiotest) {
     case RatiotestStrategy::Textbook:
-    default:  // to fix -Wreturn-type warning
       return ratiotest_textbook(runtime, p, rowmove, runtime.instance,
                                 alphastart);
     case RatiotestStrategy::TwoPass:
+    default:  // to fix -Wreturn-type warning
       Instance relaxed_instance = runtime.instance;
       for (double& bound : relaxed_instance.con_lo) {
         if (bound != -std::numeric_limits<double>::infinity()) {
--- a/src/qpsolver/ratiotest.cpp
+++ b/src/qpsolver/ratiotest.cpp
@@ -1,6 +1,6 @@
 #include "ratiotest.hpp"
 
-double step(double x, double p, double l, double u, double t) {
+static double step(double x, double p, double l, double u, double t) {
   if (p < -t && l > -std::numeric_limits<double>::infinity()) {
     return (l - x) / p;
   } else if (p > t && u < std::numeric_limits<double>::infinity()) {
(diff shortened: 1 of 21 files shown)


jajhall:
Thanks for this. I'm puzzled by your comments about getting wrong results on a MIP with semi-continuous and semi-integer variables. HiGHS can handle semi-continuous and semi-integer variables so long as the upper bound does not exceed `kMaxSemiVariableUpper`, which is `1e5` - by modelling them as a binary and continuous variable. There's a unit test for this that passes for your branch.

What is your example that gives wrong results?

The `case` that you've modified should trigger an assert if semi-continuous and semi-integer variables reach `HighsMipSolverData::runSetup()`.

jajhall (approved):
I trust you on the insertion of `static`, as I don't know the effect. Otherwise these look fine in the context of the code that I wrote. Again, because I don't understand the code, I have to trust your replacement of `std::vector<OptionRecord*>&` by `OptionRecordInt&` etc. @galabovaa wrote this so she should approve the change. 

The change to `src/util/HighsHash.h` should be approved by @lgottwald

src/qpsolver/ratiotest.cpp
```diff
@@ -98,6 +98,7 @@ RatiotestResult ratiotest(Runtime& runtime, const Vector& p,
                           const Vector& rowmove, double alphastart) {
   switch (runtime.settings.ratiotest) {
     case RatiotestStrategy::Textbook:
+    default:  // to fix -Wreturn-type warning
```
mmuetzel: IIUC, `switch` statements don't look ahead. So, this case will be executed now for all subsequent cases as well. If it doesn't cause an error, it probably still causes unexpected behavior.

svigerske:
Regarding the semicontinuous vars, I was confused that this case wasn't handled in `HighsMipSolverData::runSetup()`.

I will need to work on a reproducible example, but there are two issues I get when having a semicontinuous variable. I setup the model via the usual C++ API of HiGHS, specifying semicontinuous and semiinteger variables. If I do not set a start solution, I get into the loop in `HighsMipSolverData::runSetup()` that does not handle this variable type. If I set an infeasible start solution, then HiGHS accepts this as feasible (`MIP start solution is feasible, objective value is 0`), but then terminates with `ERROR:   MIP solver claims optimality, but with num/sum/max primal(1/0.4/0.4) infeasibilities`.    
If you say that this should have worked, then I'll create a separate issue with some code to reproduce that.

lgottwald (approved):

PR #918, opened 2022-07-18
Presolve logging

jajhall:
This branch inserts logging of rows and columns removed during LP presolve, associating them with 13 presolve rules. Also allows more advanced rules to be switched off.

Logging is off by default, and only overhead is the test of a bool whenever a rule would be successful or, in the case of dependent equations and aggregator, when it is considered.

`struct HighsPresolveLog` has been introduced to accumulate logging data in `Highs`, with `Highs::getPresolveLog` introduced to extract it.


jajhall:
I've run it through presolve for all the LP problems I have - in debug with logging on. I can't say that the reductions are identical - because I didn't check - but I think that if I'd broken anything it would have shown up.

Issue #886, opened 2022-05-24
.lp file reader has unusably large RAM requirements on large .lp files


kkofler:
I have tried running the HiGHS CLI on a 1.9 MiB .lp file. The .lp file reader fails spectacularly on it, thrashing swap despite this machine having 16 GiB of RAM! That file can be read just fine by other solvers.

Valgrind Massif points to a single function as the culprit for the memory allocations: `Reader::readNextToken` from `extern/filereaderlp/reader.cpp`. 77% of the allocations come directly from `Reader::readNextToken`, 23% from the `std::vector::_M_realloc_insert` method that it calls, all other allocations are negligible (at least at the time point where I interrupted the measurement due to RAM exhaustion, which was still during input file parsing).

As far as I can tell, this method heap-allocates a C++ object for every single read token (which by itself is already a no-no), wraps it in a `std::unique_ptr`, and then appends it to the buffer vector `this->rawtokens`. This buffer is only appended to and never removed from. It is impossible to parse large files using this algorithm.

For memory efficiency, the parser needs to stream tokens from the lexer (as opposed to the lexer appending them to a buffer – instead, streaming means the lexer returns to the parser after each token, with the token as the return value, and the parser directly processes that return value), any already parsed tokens need to be discarded (not buffered), and the token representation needs to be as compact and as devoid of indirections (requiring heap allocations) as possible (best use a POD `struct RawToken { RawTokenType type; union {double d; const char *s; } data; }` without inheritance and pass that around by value rather than by `unique_ptr`, then only the string data needs one allocation, all other token types need none).

jajhall:
Sorry about this. The LP file reader hasn't been tested on files of this size - not that it's particularly large.

Thanks for your detailed observations: I'm sure that they will mean something to the author of the LP file reader, and hope that they can be addressed relatively easily.

feldmeier:
Thanks for reporting this issue.

Yes, the .lp filereader could be more efficient, but that would require a complete rewrite taking up amount of time I don't have at the moment. I would be happy to review any pull requests implementing these changes.

Right now, the .lp filereader should take linear (in model size) amount of memory - as you noted, it stores each token in a buffer. This is not horrible, though, since this should be in the same order of magnitude as storing the HighsModel afterwards (of which multiple copies exist at runtime anyways).

[...]

PR #921, opened 2022-07-21
first round on improving performance of .lp reader

svigerske:
As discussed in #886, the .lp reader seems to require too much memory and time when reading larger input files.
The huge memory requirement may be caused by having too much data stored simultaneously, in particular, the vector of all raw tokens, the vector of all processed tokens, the vectors of processed tokens for each section, the LP stored in the readers Model class, and finally the LP stored in HiGHS.    
The large time requirement may be due to some inefficient string comparisons or lookups, memory reallocations, or unnecessary polymorphism.

This PR mainly tackles the raw tokens and speeds up some bottlenecks.
In particular,
- raw tokens are read and created while processed tokens are generated; at most 3 raw tokens are stored at a time
- when the vectors of processed tokens for each section are generated, only pointers to the begin and end of each section in the large vector of processed tokens are stored
-  processed tokens are freed after the Model has been generated
- speed up `getvarbyname()` by using a hashmap instead of a tree-based map
- speed up `parsesectionkeyword()` by using a hashmap instead of comparing with each allowed section keyword; convert string that is looked up to lower-case once; as this function is called for every string (e.g., variable and constraint names), it better be fast
- remove polymorphism for RawToken and ProcessedToken; use union for ProcessedToken to save a bit of memory
- simplify handing of objective keyword by distinguishing into min and max early

I've run this on .lp files generated from the MIPLIB2010 MPS files and other .lp files that I found on my disk (quadratics, sos, semi, etc). I've run master and this PR with `bin/highs --write_model_file out/$f.mps --model_file $i --time_limit 0 --presolve off > out/$f.log` and confirmed that the log and MPS files are the same.

This scatter plot compares the runtime for the above command per instance:    
time

This scatter plot compares the peak memory usage (KB) for the above command per instance:    
mem

There is more to do, but this is now a good place for a first cut.

### DIFF
--- a/extern/filereaderlp/builder.hpp
+++ b/extern/filereaderlp/builder.hpp
@@ -1,14 +1,14 @@
 #ifndef __READERLP_BUILDER_HPP__
 #define __READERLP_BUILDER_HPP__
 
-#include <map>
+#include <unordered_map>
 #include <memory>
 #include <string>
 
 #include "model.hpp"
 
 struct Builder { 
-   std::map<std::string, std::shared_ptr<Variable>> variables; 
+   std::unordered_map<std::string, std::shared_ptr<Variable>> variables;
 
    Model model;
 
--- a/extern/filereaderlp/builder.hpp
+++ b/extern/filereaderlp/builder.hpp
@@ -12,12 +12,14 @@ struct Builder {
 
    Model model;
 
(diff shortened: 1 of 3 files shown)


jajhall (approved):
I trust you @svigerske

PR #922, opened 2022-07-21
Fix 917

### LINKED ISSUE #917: addRow has bottleneck in assessMatrix
I have the following model (really, something more complicated, but this is the essential constraint):
```Julia
julia> using HiGHS

julia> function create_model(N)
           highs = Highs_create()
           for _ in 1:N^2
               Highs_addCol(highs, 0.0, -Inf, Inf, 0, C_NULL, C_NULL)
           end
           for i in 1:N
               for j in 2:N
                   b = N * (i - 1) + j
                   a = b - 1
                   Highs_addRow(highs, -Inf, 0.0, 2, [a, b], [1.0, -1.0])
               end
           end
           Highs_destroy(highs)
           return 
       end
```
and it's slow to build. (On the order of 1-5 minutes to build the full model for moderate instances, and yet the LP solves in 5 seconds.)

I know that I could create the full matrix up front and call `passModel`, but ideally addRow wouldn't be too much slower. Here's a profile for `create_model(200)`:

image

which shows that the bottleneck is in the `assessMatrix` method used to check that the constraint doesn't contain small/large/duplicate coefficients. In particular, it's this `assign` call:
https://github.com/ERGO-Code/HiGHS/blob/dcdb627398f0748273c0426d71c7834438bc0d31/src/util/HighsMatrixUtils.cpp#L146-L151

Here `vec_dim` is the number of columns, `N^2`, and we create a new vector for each of the `N * (N - 1)` constraints, even though each row only has two coefficients!

Some potential solutions would be:

 * If there are few coefficients, just do a loop through the coefficients to check for duplicates.
 * use a different data-structure, like a map
 * cache the vector so we can re-use it between calls, avoiding the need to malloc a new vector for each constraint
 * add an option to disable assessMatrix, assuming the user knows they are providing something good.

jajhall:
Using a map rather than a full-length vector to check for duplicate entries in `assessMatrix` will close #917

### DIFF
--- a/check/TestLpModification.cpp
+++ b/check/TestLpModification.cpp
@@ -5,7 +5,7 @@
 #include "util/HighsRandom.h"
 #include "util/HighsUtils.h"
 
-const bool dev_run = true;
+const bool dev_run = false;
 const double inf = kHighsInf;
 const double double_equal_tolerance = 1e-5;
 void HighsStatusReport(const HighsLogOptions& log_options, std::string message,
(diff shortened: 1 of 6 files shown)


lgottwald:
I would propose to use HighsHashTable<HighsInt, HighsInt> instead of map, unless you need the ordered storage of map?
That should be faster and the interface is almost the same.
Also I wonder why you use a map? Do you use the value for a key? I only see a lookup for membership so std::set would also be better in that case provided you use the ordering, or HashHashTable<HighsInt>, which then works like a set.

jajhall:
I used a map because it was one of @odow 's suggestions. I simply need to check for duplicate entries in a list of integers. Feel free to replace it with something more efficient.

lgottwald:
Ok will change it

odow:
I guess I meant map in a generic sense, rather than a specific C++ object. Sorry for the wrong steer!

lgottwald (approved):

Issue #780, opened 2022-03-24
LP failing with parallel=on (EKK parallel dual simplex)


NPC:
Hello,

I'm using Windows binaries from #746, but also calling v1.2.1 from my C# code. If running highs.exe with my simple MPS file, with parallel=on it fails silently without finding a solution:  
```
highs.exe --parallel on my.mps
Running HiGHS 1.2.0 [date: , git hash: ]
Copyright (c) 2022 ERGO-Code under MIT licence terms
LP   mySno has 3 rows; 4 cols; 6 nonzeros
Presolving model
1 rows, 2 cols, 2 nonzeros
1 rows, 2 cols, 2 nonzeros                                                                                                                                      
Presolve : Reductions: rows 1(-2); columns 2(-2); elements 2(-4)                                                                                                
Solving the presolved LP                                                                                                                                        
Using EKK parallel dual simplex solver - PAMI with concurrency of 8                                                                                               Iteration        Objective     Infeasibilities num(sum)                                                                                                                 
0     0.0000000000e+00 Ph1: 0(0) 0s
```

Without parallel I get the following output, a solution is found:  
```
highs.exe --parallel off my.mps                                               
Running HiGHS 1.2.0 [date: , git hash: ]                                                                                                                        
Copyright (c) 2022 ERGO-Code under MIT licence terms                                                                                                            
LP   mySno has 3 rows; 4 cols; 6 nonzeros                                                                                                                       
Presolving model                                                                                                                                                
1 rows, 2 cols, 2 nonzeros                                                                                                                                      
1 rows, 2 cols, 2 nonzeros
Presolve : Reductions: rows 1(-2); columns 2(-2); elements 2(-4)                                                                                                
Solving the presolved LP                                                                                                                                        
Using EKK dual simplex solver - serial                                                                                                                            
Iteration        Objective     Infeasibilities num(sum)                                                                                                                 
0     0.0000000000e+00 Ph1: 0(0) 0s                                                                                                                             
1    -1.4952000000e+02 Pr: 0(0) 0s                                                                                                                    
Solving the original LP from the solution after postsolve                                                                                                       
Model   status      : Optimal                                                                                                                                   
Simplex   iterations: 1                                                                                                                                         
Objective value     : -1.4952000000e+02                                                                                                                         
HiGHS run time      :          0.02
```

[...]

PR #924, opened 2022-07-27
Fixed bug due to not initialising multi_choice[*].row_out and immediate optimality in dual Phase 1

jajhall:
Trivial one-line fix that only affects parallel dual simplex


jajhall:
This will fix #780

PR #930, opened 2022-08-08
several improvements and fixes for MIP

lgottwald:


lgottwald:
I don't know what happens with the one failed test. There is an assertion failure in solveLp after running IPX only on mac os x:
```
 } else if (options.solver == kIpmString) {
    // Use IPM
    bool imprecise_solution;
    // Use IPX to solve the LP
    try {
      call_status = solveLpIpx(solver_object);
    } catch (const std::exception& exception) {
      highsLogDev(options.log_options, HighsLogType::kError,
                  "Exception %s in solveLpIpx\n", exception.what());
      call_status = HighsStatus::kError;
    }
    // Non-error return requires a primal solution
    assert(solver_object.solution_.value_valid);
```

The assert on the last line fails. All other tests in the branch pass. Also I added the HighsRaceTimer class. It gets one template argument so you can decide what type for the underlying deterministic clock value you want to use.

jajhall:
> I don't know what happens with the one failed test. There is an assertion failure in solveLp after running IPX only on mac os x:

Very odd. I'll see if @galabovaa can reproduce this on her Mac

PR #932, opened 2022-08-11
avoid warning on empty loop body from Clang 11

svigerske:
The warning suggests to put the semicolon onto a separate line, so do this.
Also there is no need for a second semicolon.

The empty body for the loop is intentional. We loop here until readnexttoken() succeeds (returns true). It's a bit unfortunate that the readnexttoken() function does not always actually read a token, but may just skip over whitespace.
But as the end of the file is a token, these loops will terminate eventually (if you read from files with finite length, don't try `/dev/random` or so).

### DIFF
--- a/extern/filereaderlp/reader.cpp
+++ b/extern/filereaderlp/reader.cpp
@@ -281,7 +281,8 @@ Model Reader::read() {
    // read first NRAWTOKEN token
    // if file ends early, then all remaining tokens are set to FLEND
    for(size_t i = 0; i < NRAWTOKEN; ++i )
-      while( !readnexttoken(rawtokens[i]) ) ;;
+      while( !readnexttoken(rawtokens[i]) )
+      ;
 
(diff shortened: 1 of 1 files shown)


jajhall (approved):
Thanks. Once the CI tests have passed I'll complete the PR

svigerske:
So many checks. If https://github.com/ERGO-Code/HiGHS/runs/7782976378?check_suite_focus=true is the right one to look at, then there seems to be no warning (anymore, I hope).

mmuetzel:
There are still a few warnings when compiling with clang 14 (in case you are interested in silencing those, too):
https://github.com/ERGO-Code/HiGHS/runs/7782940311?check_suite_focus=true#step:5:7

jajhall:
Thanks. I see from some of our CI tests that some compiler+architecture settings yield warnings. I'll set aside some time to investigate them

PR #965, opened 2022-10-03
Bazel

galabovaa:


sschnug:
Interesting. I didn't expect a bazel-based build here.

*What's the plan or goal here?*

Do you intent to maintain two build-systems in the future (cmake and bazel)? For what purpose?

**Imho**, this PR is rather *non bazel-like* (coarse-granular / monolithic) and looks like a workaround to offer people to use *HiGHS* in bazel-based projects without introducing too many *noise* into the original project.

In this case howewer, i don't see much gain as https://github.com/bazelbuild/rules_foreign_cc offers *basically* the same without any bazel-induced changes to the repository. It's also trivial on the consumer-side and automatically evolves with the original cmake-based build. (Yes, it will show worst-case caching behaviour and some other quirks; but some of those inefficiencies are shared with this PRs style of introducing bazel)

sschnug:
### Problem
There are currently **issues in regards to pthread** or whatever other OSes need for their std::thread implementation.

### Observation on my side
I consumed this branch in a bazel-based project of mine which has, among others, two targets:

- grpc-based microservice where highs is used internally
- "call_highs_from_cpp.cpp" cc_binary as a "canary-like" check

The former builds okay. The latter does not.

Basically:

	Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
	ld.lld: error: undefined symbol: pthread_create
	>>> referenced by Highs.cpp
	>>>               bazel-out/k8-fastbuild/bin/external/highs/_objs/highs/Highs.pic.o:(std::thread::thread<HighsTaskExecutor::HighsTaskExecutor(int)::'lambda'(int), int&, void>(HighsTaskExecutor::HighsTaskExecutor(int)::'lambda'(int)&&, int&))
	collect2: error: ld returned 1 exit status
	INFO: Elapsed time: 17.344s, Critical Path: 8.32s
	INFO: 12 processes: 7 internal, 5 linux-sandbox.
	FAILED: Build did NOT complete successfully
	
Why is one working and not the other? Well... grpc is *probably* already linking in pthread and therefore we are in luck (i guess this project had something like this in the past with openmps implications on pthread linking)!

### Question: Why was this problem not observed in this branch

The example target introduced SHOULD be enough to replicate this error on linux-based systems. Why doesn't it?

That's because the target is not correctly defined and it's never build!

	sascha@venezia:~/Dev/highs_bazel_build_debug/bazel_branch/HiGHS$ bazelisk query ...
	DEBUG: Rule 'com_google_protobuf' indicated that a canonical reproducible form can be obtained by modifying arguments commit = "22d0e265de7d2b3d2e9a00d071313502e7d4cccf", shallow_since = "1643340956 -0800" and dropping ["tag"]
	DEBUG: Repository com_google_protobuf instantiated at:
	  /home/sascha/Dev/highs_bazel_build_debug/bazel_branch/HiGHS/WORKSPACE:6:15: in <toplevel>
	Repository rule git_repository defined at:
	  /home/sascha/.cache/bazel/_bazel_sascha/4109ed2b554a50b2b6ab5545bb5aa65c/external/bazel_tools/tools/build_defs/repo/git.bzl:199:33: in <toplevel>
	//:config
	//:highs
	//:highs-config
	
There is no example! This also means we cannot build it's target explicitly. Querying the target will fail as there is no valid package.

#### Resolution
It seems the name is wrong according to bazels expectations:

`build.BAZEL` -> `BUILD.bazel`

	sascha@venezia:~/Dev/highs_bazel_build_debug/bazel_branch/HiGHS$ bazelisk query ...
	DEBUG: Rule 'com_google_protobuf' indicated that a canonical reproducible form can be obtained by modifying arguments commit = "22d0e265de7d2b3d2e9a00d071313502e7d4cccf", shallow_since = "1643340956 -0800" and dropping ["tag"]
	DEBUG: Repository com_google_protobuf instantiated at:
	  /home/sascha/Dev/highs_bazel_build_debug/bazel_branch/HiGHS/WORKSPACE:6:15: in <toplevel>
	Repository rule git_repository defined at:
	  /home/sascha/.cache/bazel/_bazel_sascha/4109ed2b554a50b2b6ab5545bb5aa65c/external/bazel_tools/tools/build_defs/repo/git.bzl:199:33: in <toplevel>
	//:config
	//:highs
	//:highs-config
	//examples:call_highs_example

Now building `//examples:call_highs_example` or wildcard-based `//...` will 100% show the pthread issues!

### Remark

While in my case it's enough to introduce `linkopts = ["-lpthread"]` to `cc_binary` or `cc_libary` to fix this, this is probably platform-specific and a generic approach will probably need to introduce *branching* there for windows and mac.

sschnug:
Okay, but two remarks:

- This is a pull-request to be merged to master
  - This signaled to me, that it's ready to be merged to master when reviewed
  - If not, why open up a PR?
- In regards to or-tools:
  - This project has tons of non-bazel depenencies and they consume those without bazelifying those in their respective source-of-truth-repos (e.g. SCIP, glpk, ...)
  - Mutating [HiGHS](https://github.com/ERGO-Code/HiGHS) opposed to wrapping build-systems at or-tools side seems strange

Nonetheless... this is some interesting movement.

galabovaa:
Thank you also for your further remarks.

We are being advised by Google on the most suitable ways to do the integration. 

In our case not all PRs are always ready to merge: some are work in progress. Some are features we consider, but eventually decide to not include in the master branch. 

Some are opened by external developers and would only be included in HiGHS if we consider the changes necessary at the current stage. 

Some workflows can be set to only run on PR so perhaps we want to ensure something works. 

While what you describe may be the case for other projects, for HiGHS in particular, please refer to the master branch. 

If you have any questions or comments in relation to applications you are involved with, do get in touch with us on highsopt@gmail.com

galabovaa:
closing, WIP

PR #970, opened 2022-10-12
Added warning to the textual description for the "solver" option

### LINKED ISSUE #904: option "solver" on MIP
If I run HiGHS on a MIP and set option "solver", it seems to solve only its continuous relaxation.
How to set the LP solver that is used for the first LP relaxation at the root node of the MIP then?

jajhall:
This will close #904


jajhall:
Simply extends the text in the description of the "solver" option

Issue #972, opened 2022-10-12
Passing a start solution via HiGHS C# API


jannicklange:
Hi!

I'm trying to access the HiGHS solver through the [C# interface](https://github.com/ERGO-Code/HiGHS/blob/master/src/interfaces/highs_csharp_api.cs).

I found the [Highs_setSolution](https://github.com/ERGO-Code/HiGHS/blob/c0e1182bd350d4ecfc4600aee3e7ff71aacae6cd/src/interfaces/highs_c_api.h) in your C api, and tried calling it analogous to the other API calls in the C# interface:

```
[DllImport(highslibname)]
private static extern int Highs_setSolution(
        IntPtr highs,
        double[] col_value,
        double[] row_value,
        double[] col_dual,
        double[] row_dual);

public HighsStatus SetStartSolution(HighsSolution solution)
{
    return (HighsStatus)HighsLpSolver.Highs_setSolution(
        this.highs,
        solution.colvalue,
        solution.coldual,
        solution.rowvalue,
        solution.rowdual);
}
```
The extern call fails, stating that the `Highs_setSolution` method cannot be found on my version of the `libhighs.dll`. I'm using the DLL from the [HiGHS.v1.2.2.x86_64-w64-mingw32-cxx11.tar.gz](https://github.com/JuliaBinaryWrappers/HiGHS_jll.jl/releases) package.

According to this [commit](https://github.com/ERGO-Code/HiGHS/commit/0c5f3712f64c6a6ce799dbef7f039beef32f1928) and the `1.2.2` Tag, I'd expect the method to exist. The PR was merged on 2022-04-14, the [release tag](https://github.com/ERGO-Code/HiGHS/releases/tag/v1.2.2) was created on 2022-04-18. However, I noticed that you linked the zipped source code in archives that were created on 2022-04-08. 

Is it possible, that the PR #508 is not part of the actual deployed version 1.2.2? Or do I need to change my approach on extending the C# interface?
Note: I already added a `getInfo` method to the C# interface, which returns the `"mip_gap", "mip_dual_bound", ...`-HighsInfos, by using an equivalent approach as described above (i.e. relying on extern calls to the `highs_c_api`).

Thanks,
Jannick

jajhall:
[...]

PR #974, opened 2022-10-13
Extend highs_sharp_api: Add passMip, getInfo, setSolution

jannicklange:
refs #972

### DIFF
--- a/examples/call_highs_from_csharp.cs
+++ b/examples/call_highs_from_csharp.cs
@@ -16,7 +16,7 @@ class Program {
       double offset = 0;
       HighsMatrixFormat a_format = HighsMatrixFormat.kColwise;
 
-      HighsModel model = new HighsModel(cc, cl, cu, rl, ru, astart, aindex, avalue, offset, a_format, sense);
+      HighsModel model = new HighsModel(cc, cl, cu, rl, ru, astart, aindex, avalue, null, offset, a_format, sense);
 
       HighsLpSolver solver = new HighsLpSolver();
 
--- a/src/interfaces/highs_csharp_api.cs
+++ b/src/interfaces/highs_csharp_api.cs
@@ -148,7 +148,7 @@ public class HighsLpSolver : IDisposable
 
    private bool _disposed;
 
-   private const string highslibname = "libhighs.dll";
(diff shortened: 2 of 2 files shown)


jajhall:
Thanks. It looks as if you've made modifications to the C# API that I presume are correct, but conflict with `https://github.com/ERGO-Code/HiGHS/blob/master/examples/call_highs_from_csharp.cs`, resulting in some CI failures. If you've got time, could you look at the CI failures, work out how to modify `call_highs_from_csharp.cs` and add it to the PR?

jannicklange:
I added an argument to the HighsModel ctor. The example is now adjusted accordingly.

I am uncertain regarding the `private const string highslibname = "libhighs";`, though. In your initial C# API, it was set to `highs`.
When I compile the code under linux, `HiGHS\build\lib` contains `.so` files that are called `libhighs.so(.1.2.2)` The dll in your release package also is called `libhighs.dll`. However, when I compile under Windows, the binaries are called `HiGHS\build\bin\Debug\highs.(dll / exe)`.

We need to settle on one name. I'd suggest to go with libhighs, since this already is used by the official package, and the "linux world". Do you know, where the dll names are specified (and why the Win-Debug build produces a DLL with a different name than used in the release package)?

jajhall:
> I added an argument to the HighsModel ctor. The example is now adjusted accordingly.

Thanks: the CI tests pass

> I am uncertain regarding the `private const string highslibname = "libhighs";`, though. In your initial C# API, it was set to `highs`. When I compile the code under linux, `HiGHS\build\lib` contains `.so` files that are called `libhighs.so(.1.2.2)` The dll in your release package also is called `libhighs.dll`. However, when I compile under Windows, the binaries are called `HiGHS\build\bin\Debug\highs.(dll / exe)`.
> 
> We need to settle on one name. I'd suggest to go with libhighs, since this already is used by the official package, and the "linux world". Do you know, where the dll names are specified (and why the Win-Debug build produces a DLL with a different name than used in the release package)?

Thanks for raising this. I'll leave @galabovaa to look at it

jajhall (approved):
Great, thanks!

Issue #928, opened 2022-08-03
error: 'strdup' was not declared in this scope, Docker Ubuntu MinGW


TAz00:
I've been using a docker ubuntu image to build HiGHS using mingw. And it stopped working for the latest HiGHS version

Ubuntu 22.04 only has mingw gcc version 10.3.0

A couple of weeks ago, a single use of strdup was introduced 
https://github.com/ERGO-Code/HiGHS/commit/646679ff72438434e8f96b599bc8e6f836412543

Error

```
-- Build files have been written to: /code/HiGHS/build
[  2%] Building CXX object src/CMakeFiles/libhighs.dir/io/FilereaderLp.cpp.o
[  2%] Building CXX object src/CMakeFiles/libhighs.dir/__/extern/filereaderlp/reader.cpp.o
[  2%] Building CXX object src/CMakeFiles/libhighs.dir/io/Filereader.cpp.o
[  3%] Building CXX object src/CMakeFiles/libhighs.dir/io/FilereaderEms.cpp.o
[  3%] Building CXX object src/CMakeFiles/libhighs.dir/io/FilereaderMps.cpp.o
/code/HiGHS/extern/filereaderlp/reader.cpp: In constructor 'ProcessedToken::ProcessedToken(ProcessedTokenType, const string&)':
/code/HiGHS/extern/filereaderlp/reader.cpp:176:14: error: 'strdup' was not declared in this scope
  176 |       name = strdup(s.c_str());
      |              ^~~~~~
make[2]: *** [src/CMakeFiles/libhighs.dir/build.make:76: src/CMakeFiles/libhighs.dir/__/extern/filereaderlp/reader.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [CMakeFiles/Makefile2:116: src/CMakeFiles/libhighs.dir/all] Error 2
make: *** [Makefile:166: all] Error 2
```

According to some guy on stackoverflow (in 2016), strdup is not part of the standard c library
https://stackoverflow.com/a/40766163

It could very well be that my problem is purely of my own creation, apparently strdup was added into the standard in 2019.
https://en.cppreference.com/w/c/experimental/dynamic/strdup and its just a question of waiting for ubuntu to catch up c11.

```
#dockerfile
FROM ubuntu

RUN apt-get update
RUN apt-get install git -y
RUN apt-get install cmake -y
RUN apt-get install g++-mingw-w64-x86-64 -y				

[...]

PR #981, opened 2022-10-24
Add cstring to filereaderlp

odow:
This is needed because of the use of strdup.

x-ref https://github.com/ERGO-Code/HiGHS/issues/928

Not sure if it's the best fix. An alternative might be to remove the use of `strdup`.

Also needed to get the Julia binaries to compile: https://github.com/JuliaPackaging/Yggdrasil/pull/5713


jajhall:
Thanks for this @odow Once you're happy with the new release I'll update the tag

PR #982, opened 2022-10-24
Fix usage of popcnt on unsupported platforms

odow:
popcnt is not supported on a range of common platforms, including aarch64.

This patch was taken from the Julia binaries:
https://github.com/JuliaPackaging/Yggdrasil/pull/5713

This needs some checking by @lgottwald.  From what I could tell, `FLAG_MPOPCNT_SUPPORTED` wasn't actually used.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -127,13 +127,10 @@ endif()
 
 if(CMAKE_SYSTEM_PROCESSOR MATCHES "^(x86\_64|i686)")
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -mpopcnt")
-  set(FLAG_MPOPCNT_SUPPORTED ON)
 elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "^(ppc64|powerpc64)")
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -mpopcntd")
-  set(FLAG_MPOPCNT_SUPPORTED ON)
 else()
   message("FLAG_MPOPCNT_SUPPORTED is not available on this architecture")
-  set(FLAG_MPOPCNT_SUPPORTED OFF)
 endif()
 
 option(DEBUGSOL "check the debug solution" OFF)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -125,9 +125,16 @@ enable_cxx_compiler_flag_if_supported("-Wno-format-truncation")
 enable_cxx_compiler_flag_if_supported("-pedantic")
 endif()
 
-check_cxx_compiler_flag("-mpopcnt" FLAG_MPOPCNT_SUPPORTED)
-set(CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} -mpopcnt")
-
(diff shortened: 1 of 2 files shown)


jajhall:
> This needs some checking by @lgottwald. From what I could tell, `FLAG_MPOPCNT_SUPPORTED` wasn't actually used.

We can't expect checking by @lgottwald at short notice. I'll be seeing her next week. 

Do you have an insight into this @galabovaa ?

galabovaa:
I don't know of particular uses of FLAG_MPOPCNT_SUPPORTED. The changes look good to me. 

There is a refactoring of the CMake code coming up as well.

odow:
The binaries with this patch are running through our tests here: https://github.com/jump-dev/HiGHS.jl/pull/133

odow:
> We can't expect checking by `@lgottwald` at short notice

There's no rush. We've built with a patch, so this can wait a week or so for v1.3.1. It might stop scipy from building on other platforms through.

jajhall:
In passing, apart from the formatting failures that @galabovaa and @jajhall are working on, there's a persistent bug that flags up in the CI tests for MacOS that @jajhall thinks will be fixed by the branch he's working on

lgottwald:
The `__builtin_popcountll` is not an intrinsic for a plattform specific instruction and hence it can (and should) be used when available regardless of plattform (i.e. for all gcc and clang builds). Only the MSVC intrinsic is x86 specific and the `-mpopcnt` flag to enable the x86 instruction. I think the flag variable in cmake was used implicitly by the HConfig.h.in configuration.

jajhall:
> The `__builtin_popcountll` is not an intrinsic for a plattform specific instruction and hence it can (and should) be used when available regardless of plattform (i.e. for all gcc and clang builds). Only the MSVC intrinsic is x86 specific and the `-mpopcnt` flag to enable the x86 instruction. I think the flag variable in cmake was used implicitly by the HConfig.h.in configuration.

What does this mean for the PR?

lgottwald:
That this is too restrictive
```
#ifdef FLAG_MPOPCNT_SUPPORTED
  static int popcnt(uint64_t x) { return __builtin_popcountll(x); }
```
Because clang and gcc can emit efficient code for this depending on plattform.

The FLAG_MPOPCNT_SUPPORTED was meant to be used only for adding the flag `-mpopcnt` to the compiler flags, which enables gcc/clang to emit the x86 instruction for __builtin_popcntll. The fix that additionally checks for this flag to to be only added on x86 is okay I think.

The logic should be:
- On clang/gcc use `__builtin_popcntll` regardless of plattform. On x86 additionally add `-mpopcnt` to the compile flags to enable the instruction.( To achieve the first part I simply added the __builtin_popcnt to be used when the __builtin_clz is used, which should be fine I believe)
- On MSVC do not use the flag mpopcnt. Use the MSVC intrinsic for the x86 instruction on x86 and the fallback implementation otherwise.

jajhall:
> So if I understand correctly, the changes to CMake were okay, but I should revert the changes to HighsHash?

Is this correct @lgottwald ?

odow:
What's the status of this? I just rebased onto the latest changes from `master`.

Issue #983, opened 2022-10-25
Segfault with start values in C API


odow:
Is this how I should be using `Highs_setSolution`, or am I doing something dumb?

```c
#include "interfaces/highs_c_api.h"

#include <stdlib.h>
#include <math.h>

int main() {
    const int N = 100;
    int indices[N];
    double values[N], start[N];
    for (int i = 0; i < N; i++) {
        indices[i] = i;
        values[i] = sin((double)i + 1.0);
    }
    // Round 1
    void* highs = Highs_create();
    for (int i = 0; i < N; i++) {
        Highs_addCol(highs, -1.0, 0.0, 1.0, 0, NULL, NULL);
        Highs_changeColIntegrality(highs, i, kHighsVarTypeInteger);
    }
    Highs_addRow(highs, 0.0, 1.0, N, indices, values);
    Highs_run(highs);
    Highs_getSolution(highs, start, NULL, NULL, NULL);
    Highs_destroy(highs);
    // Round 2
    highs = Highs_create();
    for (int i = 0; i < N; i++) {
        Highs_addCol(highs, -1.0, 0.0, 1.0, 0, NULL, NULL);
        Highs_changeColIntegrality(highs, i, kHighsVarTypeInteger);
    }
    Highs_addRow(highs, 0.0, 1.0, N, indices, values);
    Highs_setSolution(highs, start, NULL, NULL, NULL);
    Highs_run(highs);
    Highs_destroy(highs);
    return 0;
}
```

```
(base) oscar@Oscars-MBP examples % clang main.c -o main -I$highs/include/highs -L$highs/lib -lhighs
export DYLD_LIBRARY_PATH=$highs/lib; ./main

Running HiGHS 1.3.0 [date: 2022-10-25, git hash: 0a85f4f84]
Copyright (c) 2022 ERGO-Code under MIT licence terms
Presolving model
1 rows, 100 cols, 100 nonzeros
1 rows, 100 cols, 100 nonzeros
Objective function is integral with scale 1

Solving MIP model with:
   1 rows
   100 cols (100 binary, 0 integer, 0 implied int., 0 continuous)
   100 nonzeros

        Nodes      |    B&B Tree     |            Objective Bounds              |  Dynamic Constraints |       Work      
     Proc. InQueue |  Leaves   Expl. | BestBound       BestSol              Gap |   Cuts   InLp Confl. | LpIters     Time

[...]

PR #986, opened 2022-10-25
Highs::setSolution ensures that the matrix is column-wise before calling calculateRowValues and calculateColDuals

jajhall:


src/lp_data/HighsLpUtils.cpp
```diff
@@ -2286,6 +2286,9 @@ HighsStatus readBasisStream(const HighsLogOptions& log_options,
 HighsStatus calculateColDuals(const HighsLp& lp, HighsSolution& solution) {
   //  assert(solution.row_dual.size() > 0);
   if (int(solution.row_dual.size()) < lp.num_row_) return HighsStatus::kError;
+  const bool is_colwise = lp.a_matrix_.isColwise();
+  assert(is_colwise);
```
odow, line 2290: I don't understand this combination of assert and the next line. If this throws an assert, then the `kError` is unreachable?

jajhall: The `kError` return is there to ensure that a row-wise matrix isn't used in `release`. 
It shouldn't be necessary, as `calculateRowValues` and `calculateColDuals` are only called internally, so the assert in development should be enough.

odow: D'oh, silly me. The asserts aren't compiled into the `Release`.

odow (commented):
Perhaps we can add the test case from #983?

jajhall:
Done!

> Perhaps we can add the test case from #983?

Done!

PR #990, opened 2022-10-28
Fix 987 revert

### LINKED ISSUE #987: Make fixed bounds in .lp files take precedece over presence in binary section
Also make sure that when HiGHS writes out the .lp and .mps file of a problem with fixed integer variable they are correct.

jajhall:
Tests the main issue

This will close #987


jajhall:
The correction of `fixed_binary.lp` shouldn't have affected anything, but eliminated the four MacOS segfaults. Spent too long on this already!

PR #1006, opened 2022-11-15
For hans simpler

jajhall:
Need to be able to run HiGHS without crossover for benchmarking, but this isn't so simple, as the status after IPX is more often imprecise than after crossover is run. Hence it's necessary to allow crossover in these cases. 

So, run-crossover is now a string rather than bool, with "choose" set when crossover is allowed to run in the case of an imprecise IPX solution. Since this breaks its definition in v1.3.2, we need v1.4. It's also possible to set run_crossover as a command-line option.


galabovaa:
Looks good to me, would you please update the version in src/interfaces/highspy/setup.py as well? 

And perhaps we could use the branch latest, not necessarily right now but in general :)

galabovaa (approved):

jajhall:
Sorry I hadn't spotted that latest had been created 🙂

Still some experiments to do before merging

PR #1009, opened 2022-11-20
Fix L2CTA3D

jajhall:
Testing on the 51 Mittlemann benchmark LPs, one (karted) takes an excessive time in presolve, and another (L2CTA3D) fails to return from presolve altogether. This is due to the dependent equations check taking too long. 

A timeout has been introduced. Currently it's min(1000, time_limit/100), making it (technically) non-deterministic. In due course it should be replaced by a limit on the existing pseudo-clock that's related to the size of the equation set.


jajhall:
Needs to be merged into master, as (a) Hans needs it soon and (b)

PR #1023, opened 2022-12-02
Add --version to command line options

jajhall:


mingodad:
It would be nice to also have an API to get the version because we can have an application that dynamically link to HiGHS and probably can work with more than one version and it's important to have a way to identify it for debugging/support purposes.

Something like:
```
const char* Highs_version(void);
```

jajhall:
> It would be nice to also have an API to get the version because we can have an application that dynamically link to HiGHS and probably can work with more than one version and it's important to have a way to identify it for debugging/support purposes.
> 
> Something like:
> 
> ```
> const char* Highs_version(void);
> ```

OK, that's soon done!

src/interfaces/highs_c_api.h
```diff
@@ -214,6 +214,13 @@ void* Highs_create(void);
  */
 void Highs_destroy(void* highs);
 
+/**
+ * Return the HiGHS version number vX.Y.Z
+ *
+ * @returns the HiGHS version as a char*
+ */
+const char* Highs_version(void);
```
odow, line 222: So if we're following semantic versioning, https://semver.org, this would be a new feature, requiring `v1.5.0` instead of `v1.4.1`.

I'm okay if you decide "nah, minor releases are for more substantial changes," but it would be good to have a policy.

jajhall: Thanks @odow. Where would we be without you 😍

mingodad: An application can have code to allow work with more than one version at runtime if it can identify the provided version of the supplied library.

So would be nice to have this as API calls:
```
const char *Highs_version_githash() {
  return HIGHS_GITHASH;
}
const char *Highs_version_compilation_date() {
  return HIGHS_COMPILATION_DATE;
}
int Highs_version_major() {
  return HIGHS_VERSION_MAJOR;
}
int Highs_version_minor() {
  return HIGHS_VERSION_MINOR;
}
int Highs_version_patch() {
  return HIGHS_VERSION_PATCH;
}
```

jajhall: OK. I wondered about doing this, but felt it was unnecessary since the version statement could be parsed. The compilation date and githash are sometimes not defined, but I'll still add them all.

mingodad: Also this one seems simpler:
```
#define HIGHS_GITHASH "d85660e3b-dirty"
#define HIGHS_COMPILATION_DATE "2022-12-01"
#define HIGHS_VERSION_MAJOR 1
#define HIGHS_VERSION_MINOR 4
#define HIGHS_VERSION_PATCH 0

#define TOSTR_DEF_HELPER(x) #x
#define TOSTR_DEF(x) TOSTR_DEF_HELPER(x)

const char Ĥighs_version() {
	return "HiGHS v." TOSTR_DEF(HIGHS_VERSION_MAJOR)
		"." TOSTR_DEF(HIGHS_VERSION_MINOR)
		"." TOSTR_DEF(HIGHS_VERSION_PATCH)
		" " HIGHS_COMPILATION_DATE " " HIGHS_GITHASH;
}
```
Output of preprocessor:
```
gcc -E highs-version.c
# 1 "highs-version.c"
# 1 "<built-in>"
# 1 "<command-line>"
# 31 "<command-line>"
# 1 "/usr/include/stdc-predef.h" 1 3 4
# 32 "<command-line>" 2
# 1 "highs-version.c"
# 10 "highs-version.c"
const char Ĥighs_version() {
 return "HiGHS v." "1"
  "." "4"
  "." "0"
  " " "2022-12-01" " " "d85660e3b-dirty";
}
```

jajhall: You've lost me in all this, but we're certainly only defining the version, compile date and githash via CMake

jajhall: Why is `void` an the argument of `const char* Highs_version(void);`, but not `const char *Highs_version_githash()` ?

PR #1036, opened 2022-12-07
Ipx fix

### LINKED ISSUE #280: BUG: IPX w/out presolve fail on easy problem
When solve the following problem the model gets unset (presumably from an error during the course of optimization).  Any insights into why it might be failing?

```
c [-1, 1] (MIN)
A = [[1, 0], [0, 1]]
lhs = [-inf, 2]
rhs = [1, 2]
lb = [1, 2]
ub = [1, 2]
```

The solution should be trivial, but `simplex` with `presolve` returns unset model as well as any variant of `ipm`.  The only thing that solves this is `simplex` with `presolve` turned off.

jajhall:
Applied first of IPX fixes from @lschork2. 

This will close #280

primal-dual-infeasible1 failure remains, and isn't fixed for me by second of IPX fixes from @lschork2. However, unit test is restored now that simplex clean-up masks IPX failure

### DIFF
--- a/check/SpecialLps.h
+++ b/check/SpecialLps.h
@@ -189,7 +189,7 @@ class SpecialLps {
     lp.model_name_ = "primalDualInfeasible1";
     lp.num_col_ = 2;
     lp.num_row_ = 2;
-    lp.col_cost_ = {-2, 1};
+    lp.col_cost_ = {2, -1};
     lp.col_lower_ = {0, 0};
     lp.col_upper_ = {inf, inf};
     lp.row_lower_ = {-inf, -inf};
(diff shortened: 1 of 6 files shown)


jajhall:
> primal-dual-infeasible1 failure remains, and isn't fixed for me by second of IPX fixes from @lschork2.

Second of IPX fixes from @lschork2 leads to correct primal residual value: IPX isn't designed to identify primal-dual-infeasible

jajhall:
Applied IPX mods from @lschork2 to add detection of (primal or dual) infeasibility based on divergence of the objective values.

galabovaa (approved):

PR #1043, opened 2022-12-11
Readme: Add badges with latest PyPI release and monthly downloads

EwoutH:
Add two PyPI badges to the readme, one with the latest version and one with the number of monthly downloads. This increases visibility and makes the HiGHS easier to find on PyPI (since both link to https://pypi.org/project/highspy/).


jajhall:
Thanks. I don't imagine that this controversial is in the least, but I've assigned @galabovaa to review it

galabovaa (approved):

PR #1073, opened 2023-01-01
fix processing of semi section in .lp reader

svigerske:
`Semi` section was ignored if there were no general integer variables due to a stupid copy-paste-bug.


jajhall:
Already fixed in `latest`

PR #1075, opened 2023-01-01
improve description of write_solution_style option

svigerske:
Added description for new option value -1 (which was 2 before).
Change some formatting.


jajhall:
Thanks for this and the other bits of "housekeeping" you've done. I'd spotted the semi-variable .LP read error before Christmas and it's fixed in "latest", which is where we're collecting modifications before pushing to master, after running performance tests. I'd had in mind to remove some compiler warnings, so thanks for this!

I'll be back at work on 3 January

jajhall:
Change incorporated into modified lines in `latest`

PR #1079, opened 2023-01-03
Fix readerlp semi

jajhall:
Checking that this is already fixed in latest


jajhall:
Already fixed in `latest`

PR #1097, opened 2023-01-08
Add integrated Webdemo

xor2k:
Hi everybody & happy New Year!

I've added code for compiling a webdemo for HiGHS directly, so it can be run in a browser serverlessly from a single HTML file of around 3.4 MB. This is done with Emscripten, a C/C++ to JavaScript/WebAssembly compiler. The integration with cmake is a little hacky (by design, Emscripten overrides a lot of parameters), so it needs a script to trigger the build. I've wrapped everything up in this pull request.

Basically a copy of the pull request I've added to SoPlex today.

The difference to the existing approach is that only Emscripten is required, which makes it very simple, a baseline so to say. This allows e.g. to integrate the build into automated tests to ensure JavaScript/Webassembly builds don't break and it will also always be up-to-date, as it is integrated with the actual code. Further, this option rather addresses C/C++ programmers that want to build larger applications that contain HiGHS and then optionally compile them to WebAssembly in contrast to using it primarily as a Node.js module. For example, Pyodide is a Python distribution that runs in a browser. If the maintainers wanted to add HiGHS to Pyodide, one needs to ensure it compiles with Emscripten alongside with its Python bindings.

Demo build see [highs.zip](https://github.com/ERGO-Code/HiGHS/files/10369637/highs.zip)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -266,3 +266,6 @@ qjh.mps
 
 *.whl
 bazel*
+
+# webdemo
+build_webdemo
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,6 +59,9 @@ option(PYTHON "Build Python interface" OFF)
 option(FORTRAN "Build Fortran interface" OFF)
 option(CSHARP "Build CSharp interface" OFF)
 
+# emscripten
+option(EMSCRIPTEN_HTML "Emscripten HTML output" OFF)
+
 # set the correct rpath for OS X
 set(CMAKE_MACOSX_RPATH ON)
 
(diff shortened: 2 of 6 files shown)


jajhall:
Thanks for this. It seems as if it's a useful new interface to HiGHS. The CI tests are now running

galabovaa:
Thank you for the contribution! Looks good to me. Could you possibly add a GitHub Action workflow to our .github/workflows directory which tests the build of the interface?

jajhall (approved):
Thanks, this looks fine

Issue #1040, opened 2022-12-10
Error when linking to HiGHS with clang15


FlorianSchwendinger:
In the new HiGHS version LTO is enabled by default for LINUX and clang.
The corresponding code chunk from the `CMakeList.txt` file can be seen below:

```cmake
if(LINUX AND NOT MSVC)
    include(CheckIPOSupported)
    check_ipo_supported(RESULT ipo_supported OUTPUT error)

    if(ipo_supported)
        message(STATUS "IPO / LTO enabled")
        set(CMAKE_INTERPROCEDURAL_OPTIMIZATION TRUE)
    else()
        message(STATUS "IPO / LTO not supported: <${error}>")
    endif()
endif()
```

When I try to link to the HiGHS library with clang15 on debian testing I get the following error message.
```sh
/usr/bin/ld: /tmp/Rtmpy6uejk/R.INSTALL3d922bc7e7f1/highs/src/highslib/lib/libhighs.a: error adding symbols: file format not recognized
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

I resolved it in my installation by changing 
```cmake
if(LINUX AND NOT MSVC)
```
to
```cmake
if(LINUX AND (NOT MSVC) AND (NOT CMAKE_CXX_COMPILER_ID STREQUAL "Clang"))
```

Not sure what is the best fix HiGHS, since I am not sure if this is a debian testing issue or a clang15 issue.

ttencate:
@FlorianSchwendinger You are compiling with clang but linking with the GNU linker, `ld`. LTO is compiler-specific, so the GNU linker can only perform LTO on the output of GCC, not on the output of clang. `cmake -DCMAKE_CXX_FLAGS="-fuse-ld=lld"` tells CMake to tell the compiler to use clang's `lld` linker instead.

The underlying problem in HiGHS is that LTO (IPO in CMake terminology) is forcibly enabled if supported by the compiler, even if the user turned it off explicitly using `-DCMAKE_INTERPROCEDURAL_OPTIMIZATION=OFF`. If `CMAKE_INTERPROCEDURAL_OPTIMIZATION` already has a value, I think `CMakeLists.txt` should not override it.

Also consider enabling LTO only if HiGHS is compiled as a dynamic library.  When building as a static library, HiGHS shouldn't assume that it will eventually get linked with a compatible linker, of a compatible version, with compatible settings.

ttencate:
[...]

PR #1103, opened 2023-01-17
don't force LTO when the user explicitly disabled it

lovasoa:
never override the value of CMAKE_INTERPROCEDURAL_OPTIMIZATION set by the user

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,7 +36,9 @@ endif()
 
 option(BUILD_TESTING "Build Tests" ON)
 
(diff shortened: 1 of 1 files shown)


jajhall:
I've no idea about the implications of this change. How about you, @galabovaa ?

galabovaa (approved):

lovasoa:
Currently I switched the rust binding to a fork of HiGHS that has this patch, this will allow me to switch back to the official HiGHS.

jajhall:
> No, this won't close #1040 ,

Noted, thanks

PR #1104, opened 2023-01-18
No IPO / LTO on static libs, and do not override user preference

### LINKED ISSUE #1040: Error when linking to HiGHS with clang15

ttencate:
The comments should describe what's going on here. Note that this removes the checks for Linux, ~~MSVC~~ and Clang in favour of relying on CheckIPOSupported to do its job.

Fixes #1040

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,16 +36,60 @@ endif()
 
 option(BUILD_TESTING "Build Tests" ON)
 
-if(LINUX AND (NOT MSVC) AND (NOT CMAKE_CXX_COMPILER_ID STREQUAL "Clang"))
-    include(CheckIPOSupported)
-    check_ipo_supported(RESULT ipo_supported OUTPUT error)
+# whether to use shared or static libraries
+option(SHARED "Build shared libraries" ON)
+set(BUILD_SHARED_LIBS ${SHARED})
+message(STATUS "Build shared libraries: " ${SHARED})
 
-    if(ipo_supported)
-        message(STATUS "IPO / LTO enabled")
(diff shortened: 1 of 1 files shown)


galabovaa (approved):

jajhall:
Some major Windows CI test failures. I've no idea why this should happen. Suggestions?

ttencate:
I have no clue either and I don't use Windows. For now, let me just disable LTO on MinGW as well.

PR #1115, opened 2023-01-27
Modified C API for Highs_version, Highs_githash and Highs_compilation_date

jajhall:
Simple change to these methods so that what's returned doesn't become invalid once the function returns and the string goes away.


jajhall:
This will close #1115

jajhall:
Simpler highsVersion, highsGithash and highsCompilationDate now return const char* simplifying C API

PR #1116, opened 2023-01-28
Cmake namespace

galabovaa:
CMake namespace, HiGHS version handling, library names in FAST_BUILD.


jajhall (approved):
So glad you can do all this. Have to assume it's OK!

Issue #1114, opened 2023-01-27
Enumerating `OptionRecord`s using C API


cyderize:
Hi, I'm in the process of converting MiniZinc's new HiGHS interface to use the C API rather than the C++ API (this way we can turn it into a plugin which dynamically loads the HiGHS library at runtime).

However, there are a few C++ APIs which are not available using the C interface. Most of these I've been able to easily add in my fork (I'll open a pull request soon).

We need to be able to get all of the available options, so that we can generate a GUI for setting the options from the MiniZinc IDE. Currently in C++, we loop through the `records` field of `HighsOptions`, but I'm not entirely sure what an API for accessing these records in C should look like (we need access to the default values and bounds of the options).

Gurobi provides `GRBgetnumparams` and `GRBgetparamname` to get the number of parameters, and then get the name from the index. Perhaps something similar could be done here, with additional functions to access the descriptions, default values, and bounds of options.

jajhall:
By "number of parameters", do you mean the number of options? I've searched for `GRBgetnumparams` in the Gurobi documentation, but nothing comes up.

Creating a method in the C API to achieve what you want won't be difficult. I'll get on with it this weekend

Note that there are some methods in the C++ API that are for internal use only, so shouldn't be in the C API. I'll see what you've put in your PR - please make this to `latest`, not `master` - and delete as appropriate.

cyderize:
Indeed those Gurobi functions seem to appear in the headers but apparently aren't documented anywhere. The functions we use for Gurobi have these signatures:

```c
int GRBgetnumparams(GRBenv *env);
int GRBgetparamname(GRBenv *env, int i, char **paramnameP);
int GRBgetparamtype(GRBenv *env, const char *paramname);
int GRBgetintparaminfo(GRBenv *env, const char *paramname, int *valueP, int *minP, int *maxP, int *defP);
int GRBgetdblparaminfo(GRBenv *env, const char *paramname, double *valueP, double *minP, double *maxP, double *defP);
int GRBgetstrparaminfo(GRBenv *env, const char *paramname, char *valueP, char *defP);
```

Thanks, I'll make sure I target the `latest` branch.

jajhall:
Thanks, I'll match these methods

jajhall:
[...]

PR #1118, opened 2023-01-30
Option info

jajhall:
Introduced new methods for accessing option information to support extension of C API, as per #1114


jajhall:
Needs consolidation of methods, as get*Option and getOptionType can call new, more general methods

jajhall:
Finally identified reason for CI failures on Windows. In `Highs_getOptionName`, space is mallocced for the name as a `char*`, and I used the length of the name as a string. But, (because the char* is null-terminated?) it needs one adding to the length on Windows!

There was also a `memset(current_value, 0, 7);` inherited from `Highs_getStringOptionValue`: I cannot see why this was ever useful. Must assume that `current_value` is of length at least 7, which isn't necessarily true!

PR #1121, opened 2023-02-01
relax asserts on indices in freelist: may be rows

ambros-gleixner:


jajhall:
Necessary changes now acieved by #1122

Issue #1129, opened 2023-02-03
Setting  D_ITERATOR_DEBUG_LEVEL=0 under Windows generate conflicts for using HiGHS as a dependency


Downsite:
The change that was added in a126bc8, is causing some issues for me.

I am trying to build HiGHS as part of a larger CMake project.

` LNK2038 : mismatch detected for '_ITERATOR_DEBUG_LEVEL'`

This is caused by the project being built in Debug Mode. This will set the flag to 2 by default, but HiGHs overwrites this setting locally.

I would like to be able to choose if this flag is set as a CMake option.

jajhall:
Am I correct in thinking that `_ITERATOR_DEBUG_LEVEL`has a default value of 0, and a fix for the issue is that if HiGHS finds that it's set to something else it should not set it to 2?

Downsite:
In Debug mode 2 is the default value. So I think there would not be any value defined during the CMake run.
See here (https://learn.microsoft.com/en-us/cpp/standard-library/iterator-debug-level?view=msvc-170).

jajhall:
So, you're saying that you want to use the default value, and the problem is that we're using a different value. @lgottwald will have done this for good reason - to speed up debug runs when checking for other issues.

For me, the solution is to comment out our local setting, and just remember that it might be useful to use it when we're developing HiGHS.

Downsite:
Yes, since other packages seem to be built with level 2 in mind. At least we have some issues with prebuilt libraries (specifically CPLEX) building their debug binaries with level 2.

However, I see the benefit of faster compilation.  
For me, it could be part of (or work similarly to) the FAST_BUILD option.

jajhall:
So, you're suggesting that the HiGHS setting of `_ITERATOR_DEBUG_LEVEL` is something that a user can define - like `FAST_BUILD`?

jajhall:
Fixed when #1135 is pushed to `latest`

jajhall:
Hmmm... it looks as if not setting `D_ITERATOR_DEBUG_LEVEL=0` causes problems on Windows.

I'll try to investigate

jajhall:
The continuous integration failures on Windows if `D_ITERATOR_DEBUG_LEVEL` is not set to zero can't be serious, since the tests pass on Linux/Mac. However, I wanted to get to the bottom of them.

I spent 2-3 hours yesterday failing to get Visual Studio to run in debug so that I could instigate.

[...]

PR #1135, opened 2023-02-06
fix-1129

jajhall:
Commented out `add_definitions(-D_ITERATOR_DEBUG_LEVEL=0)` in main `CMakeLists.txt`

This fouls up building HiGHS as part of a larger CMake project: see #1129 

Setting it will speed up run-time in debug (and compile-time?) so leave in `CMakeLists.txt`for local change by developer


jajhall:
Pushing this to `latest` will close #1129

jajhall:
Hmmm... it looks as if not setting `D_ITERATOR_DEBUG_LEVEL=0` causes problems on Windows.

I'll try to investigate

PR #1141, opened 2023-02-09
initialize Qdim in HMPSIO

svigerske:
Set it to some sensible value like for `numRow` or `numCol`, in case there are no quadratic terms.

### DIFF
--- a/src/io/HMPSIO.cpp
+++ b/src/io/HMPSIO.cpp
@@ -47,6 +47,7 @@ FilereaderRetcode readMps(
   // MPS file buffer
   numRow = 0;
   numCol = 0;
+  Qdim = 0;
   cost_row_location = -1;
   objOffset = 0;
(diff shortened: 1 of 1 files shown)


jajhall (approved):
Nothing wrong with doing this, of course. 

However, I'm puzzled that not setting `Qdim = 0` hasn't led to unassigned reads with valgrind

svigerske:
In the one place where HiGHS calls `readMps`, you pass in a `hessian.dim_` for `Qdim`. `hessian` is a `HighsHessian` object. And the constructor of that object sets `dim_ = 0`. So it is already initialized.

In my code, I passed in an uninitialized `HighsInt`.

jajhall:
Indeed, I've just seen this, so realised why no unassigned reads are triggered. 

Do note that `readMps` is the fixed format MPS reader, which isn't (currently) called via a unit test.

Issue #1140, opened 2023-02-08
C API: Highs_getRowsByRange incorrectly sets num_nz == 0


rschwarz:
In using the C API, I want to query the model properties (bounds, coefficients), using `Highs_getColsByRange` and `Highs_getRowsByRange`, which I apply to single columns and rows.

I found that `Highs_getColsByRange` works as advertised, but when I call `Highs_getRowsByRange`, the argument `int* num_nz` is always populated with 0, as if the row had no coefficients.

I added two blocks to the C API unit tests to demonstrate, see commit https://github.com/rschwarz/HiGHS/commit/0ba43511aa76db1376da16b6268cb000f22e2f94 on branch https://github.com/rschwarz/HiGHS/tree/rs/getrows_c_api.

jajhall:
Thanks: I've added this to the TestCAPI.c and reproduced the bug. Should be soon fixed

PR #1145, opened 2023-02-09
Highs::getRowsInterface now returns get_num_nz when row_matrix_start is nullptr

jajhall:
Fixed a bug in `Highs::getRowsInterface` which always returned `get_num_nz=0` when `row_matrix_start` was `nullptr`


jajhall:
This will close #1140

rschwarz:
Thanks, this also provides me with a work-around :-)

Issue #958, opened 2022-09-20
In CholeskyFactor.resize the index can exceed the vector length.


FlorianSchwendinger:
In [`src/qpsolver/factor.hpp`](https://github.com/ERGO-Code/HiGHS/blob/9c7b2cf2749e8717f54327df77d08b821ef51d0c/src/qpsolver/factor.hpp) in method `resize` (line 73), the indices used in the vectors `L` and `L_old` will exceed the vector length of the corresponding vector, for the case when `new_k_max < current_k_max`.

This case occurred e.g., in the simple example 
```
# Minimize:
#  -x_2 - 3x_3 + (1/2) * (2 x_1^2 - 2 x_1x_3 + 0.2 x_2^2 + 2 x_3^2)
# Subject to:
#  x_1 + x_3 <= 2
#  0 <= x
```
and causes an `container-overflow on address` the full message can be found below
```
==2813085==ERROR: AddressSanitizer: container-overflow on address 0x603000024378 at pc 0x7fe02221bccf bp 0x7fff59dc5630 sp 0x7fff59dc5628
WRITE of size 8 at 0x603000024378 thread T0
    #0 0x7fe02221bcce in CholeskyFactor::resize(int) /R/packages/incoming/highs.Rcheck/00_pkg_src/highs/inst/HiGHS/src/qpsolver/factor.hpp:79:32
...
0x603000024378 is located 8 bytes inside of 32-byte region [0x603000024370,0x603000024390)
allocated by thread T0 here:
    #0 0x4f078d in operator new(unsigned long) /Sources2/LLVM/14.0.6/llvm-project-14.0.6.src/compiler-rt/lib/asan/asan_new_delete.cpp:95:3
    #1 0x7fe021c431cf in void* std::__1::__libcpp_operator_new<unsigned long>(unsigned long) usr/local/bin../include/c++/v1/new:245:10
    #2 0x7fe021c431cf in std::__1::__libcpp_allocate(unsigned long, unsigned long) usr/local/bin../include/c++/v1/new:271:10
    #3 0x7fe021c431cf in std::__1::allocator<double>::allocate(unsigned long) usr/local/bin../include/c++/v1/__memory/allocator.h:105:38
    #4 0x7fe021c431cf in std::__1::allocator_traits<std::__1::allocator<double> >::allocate(std::__1::allocator<double>&, unsigned long) usr/local/bin../include/c++/v1/__memory/allocator_traits.h:262:20
[...]

PR #1158, opened 2023-02-14
Fix highs node queue clear

jajhall:
`HighsNodeQueue::clear()` was written originally as 

> HighsNodeQueue::clear() 
{
  HighsNodeQueue nodequeue;
  nodequeue.setNumCol(numCol);
  *this = std::move(nodequeue);
}

However, on Windows _without_

-D_ITERATOR_DEBUG_LEVEL=0

a segfault occurs in `std::move(nodequeue);`. This seems to be fixed by performing `std::move` operations in the right order!


jajhall:
Note that the branch has

-D_ITERATOR_DEBUG_LEVEL=0

since #958 is still exposed

PR #1160, opened 2023-02-14
Fix 1129

jajhall:
Contains the two successful fixes for #1129, and many replacements of `&vector[0]` by `vector.data()` that were inspired by an observation in #1129


galabovaa:
looks good to me

galabovaa (approved):

Issue #1098, opened 2023-01-09
Presolve emits infeasible for feasible MIP instance


h2oche:
HiGHS version: v1.4.1
OS: Ubuntu 20.04

Variables:
- v0 -- integer in [0,2]
- v1, ..., v4 -- binary

Constraints:
- v1 <= v0 <= 2 * v1
- v2 + v3 + v4 = 1
- 2(v2 + v3 + v4) - v0 = 1

It emits feasible when presolve mode is `off` and emits infeasible when presolve mode is `on`, where parallel mode is "on" with 8 threads.

jajhall:
Sorry about this. I've reproduced the behaviour and we will do our best to fix it.

h2oche:
Big thanks for your hard work. I'm not sure if it helps, but following MIP instance, similar to the reported MIP instance, worked well in both presolve modes.

Variables:
- v0 -- integer in [0,2]
- v1, ..., v3 -- binary

Constraints:
- v1 <= v0 <= 2 * v1
- v2 + v3  = 1
- 2(v2 + v3) - v0 = 1

jajhall:
@lgottwald , this behaviour is reproduced using 
[h2oche.mps.txt](https://github.com/ERGO-Code/HiGHS/files/10375653/h2oche.mps.txt)

fwesselm:
> HiGHS version: v1.4.1 OS: Ubuntu 20.04
> 
> Variables:
> 
>     * v0 -- integer in [0,2]
> 
>     * v1, ..., v4 -- binary
> 
> 
> Constraints:
> 
>     * v1 <= v0 <= 2 * v1
> 
>     * v2 + v3 + v4 = 1
> 
>     * 2(v2 + v3 + v4) - v0 = 1
> 
> 
> It emits feasible when presolve mode is `off` and emits infeasible when presolve mode is `on`, where parallel mode is "on" with 8 threads.

Hello!

I think the presolve code correctly identifies that variable v0 can be fixed. However, it assumes that the value that the variable is fixed to is one of its bounds. In this case, the variable needs to be fixed to 1, which is in between its bounds. The attached changes first adjust the bounds and then fix the variable. This resolved the issue on my side, but I am not sure if there is a better implementation.

Regards,
Franz

[presolveBug.txt](https://github.com/ERGO-Code/HiGHS/files/10743793/presolveBug.txt)

jajhall:
Well spotted: I'll look to integrate this into the MIP presolve

[...]

PR #1163, opened 2023-02-15
Fix 1098

jajhall:
Now handling case where variable is fixed between its bounds in `HPresolve::rowPresolve`. Thanks @fwesselm.


jajhall:
This will close #1098

PR #1182, opened 2023-02-24
Fix dualtol in HEkkDual::computeExactDualObjectiveValue()

ambros-gleixner:
In SCIP we sometimes use tighter dual feasibility tolerances; with 1e-10, I discovered this inconsistent use of dual tolerance in computeExactDualObjectiveValue() and the subsequent correctDualInfeasibilities().

Without this fix, the assert
```
      // no shifts should have occurred
      assert(!ekk_instance_.info_.costs_shifted);
```
is triggered on instance `milo-v12-6-r2-40-1` from MIPLIB 2017.

### DIFF
--- a/src/simplex/HEkkDual.cpp
+++ b/src/simplex/HEkkDual.cpp
@@ -2876,9 +2876,9 @@ double HEkkDual::computeExactDualObjectiveValue(HVector& dual_col,
     // flipping the sign of its dual and for a basic variable we may need to add
     // to the dual objective using one of the bounds when its dual is not zero.
     double active_value;
-    if (exact_dual > ekk_instance_.options_->small_matrix_value)
+    if (exact_dual > ekk_instance_.options_->dual_feasibility_tolerance)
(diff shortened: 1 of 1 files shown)


jajhall:
Strange: I'll look at this. If I remember correctly, `computeExactDualObjectiveValue()` is only used to determine the dual objective value with unperturbed costs to determine whether to terminate via the dual bound, so I'm puzzled how the assert is then avoided.

ambros-gleixner:
The reason is that using the correct (=tighter) tolerance makes `computeExactDualObjectiveValue()` return minus infinity, so `correctDualInfineasibilities()` is not called.

jajhall:
OK I guessed something significant had triggered your interest in this. I need to do some sanity performance checks on latest - since I don't think I've done anything to compromise it - and then I'll merge this.

ambros-gleixner:
Thanks!

jajhall:
Stale

PR #1184, opened 2023-02-25
Allow small matrix values introduced by presolve, also in debug mode

ambros-gleixner:
During presolve it can happen that matrix coefficients smaller than the threshold `small_matrix_value` in absolute value are introduced, even if in the original problem all coefficients have absolute value larger or equal. This leads to a warning in `assessLp()` and as a result the assert in `Highs::run()` fails in debug mode.

This PR relaxes this assert to accept this case, as would happen in optimized mode.

An example where the assert is triggered quickly is ` bin/scip -c "read /nfs/optimi/kombadon/IP/miplib/momentum1.mps.gz o"`, but it happens also for numerically nicer instances. I checked, and in this case the small values were introduced by the aggregator.

```
opt111:~/opti/devel/scip/optbuild/Debug-highs% gdb --args bin/scip -c "read /nfs/optimi/kombadon/IP/miplib/momentum1.mps.gz o"
GNU gdb (Ubuntu 8.1.1-0ubuntu1) 8.1.1
Copyright (C) 2018 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from bin/scip...done.
(gdb) r
Starting program: /scratch/opt/bzfgleix/devel/scip/optbuild/Debug-highs/bin/scip -c read\ /nfs/optimi/kombadon/IP/miplib/momentum1.mps.gz\ o
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
SCIP version 8.0.4 [precision: 8 byte] [memory: block] [mode: debug] [LP solver: HiGHS 1.5.1] [GitHash: 72e9bb731e]
Copyright (c) 2002-2023 Zuse Institute Berlin (ZIB)

External libraries:
  Readline 7.0         GNU library for command line editing (gnu.org/s/readline)
[...]

PR #1195, opened 2023-03-03
Allow small matrix values after presolve

ambros-gleixner:
As suggested in #1184, here is an attempt at a non-intrusive assessLp(). I suppose this is not exactly how you would do it, but maybe a start.


jajhall:
@ambros-gleixner Please leave me to write things like this

ambros-gleixner:
Okay, just wanted to help.

Issue #1191, opened 2023-03-02
Problem handling integer variables with floating point bounds


rma350:
For the problem:

minimize x + y
s.t. x + y <= 1
x, y >= 0.5 integer

As encoded by the following LP file

Minimize
 obj: x1 + x2
Subject To
 c1: x1 + x2 <= 1
Bounds
 0.5 <= x1
 0.5 <= x2
General
  x1
  x2
End

The problem is infeasible, but highs gives  back the bad solution with optimal status (x, y) = (0.5, 0.5). The problem appears to be solved in presolve, and then a warning is printed saying that the optimal solution from presolve is not feasible. 

You can reproduce this error by running the LP file here:

https://lovasoa.github.io/highs-js/

(I was using the C++ API). While I can manually round my lower bounds up and my upper bounds down, since I am using C++ and not giving you an LP file, this forces me to track the true bound if the variable is changed back to continuous for a subsequent solve. For context, I am working on the ortools mathopt integration.

jajhall:
Indeed, HiGHS makes a bit of a mess of this simple problem! The issue is, indeed, in presolve as it's handled fine with `--presolve=off`.

PR #1197, opened 2023-03-07
Now declares infeasiblilty in MIP presolve if forcing row fixes integer variable to fractional value

jajhall:
Should be uncontroversial, as it should only be relevant for infeasible MIPs.


jajhall:
This will close #1191

Issue #1187, opened 2023-03-01
In release, disallow small matrix values introduced by presolve 


jajhall:
Motivated by  #1184

ambros-gleixner:
See #1184 for an .lp file to reproduce on the command line:
[debug.lp.txt](https://github.com/ERGO-Code/HiGHS/files/10864518/debug.lp.txt)

Issue #1174, opened 2023-02-20
No presolve => assessLp(reduced_lp, options_) == HighsStatus::kOk' failed


Downsite:
If the following .mps file is run with the `--presolve "false"` flag, the following error occurs:

```
Running HiGHS 1.2.2 [date: 2023-02-20, git hash: 437c98b08]
Copyright (c) 2022 ERGO-Code under MIT licence terms
Number of BV entries in BOUNDS section is 35
MIP  out has 546 rows; 370 cols; 1302 nonzeros; 35 integer variables

Presolve is switched off

Solving MIP model with:
   546 rows
   370 cols (35 binary, 0 integer, 0 implied int., 335 continuous)
   1302 nonzeros

        Nodes      |    B&B Tree     |            Objective Bounds              |  Dynamic Constraints |       Work      
     Proc. InQueue |  Leaves   Expl. | BestBound       BestSol              Gap |   Cuts   InLp Confl. | LpIters     Time

         0       0         0   0.00%   -2000           inf                  inf        0      0      0         0     0.0s
highs: /.../dep/HiGHS/src/lp_data/Highs.cpp:1014: HighsStatus Highs::run(): Assertion `assessLp(reduced_lp, options_) == HighsStatus::kOk' failed.

Aborted (core dumped)
```

The same behavior is present on master
```
Running HiGHS 1.5.0 [date: 2023-02-20, git hash: ce351010e]
Copyright (c) 2022 HiGHS under MIT licence terms
```

[out.mps.txt](https://github.com/ERGO-Code/HiGHS/files/10785591/out.mps.txt)
[out.lp.txt](https://github.com/ERGO-Code/HiGHS/files/10785598/out.lp.txt)

Downsite:
Issues similar to #1080 and #1164 are the reason why I needed to set presolve to false.

SanPen:
Hi, 

I've found this issue as well.

My experience is the following:

- I'm setting the model using the C++ interface.
- I replaced `assessLp(reduced_lp, options_) == HighsStatus::kOk` by:
```
if(assessLp(reduced_lp, options_) != HighsStatus::kOk){
    writeModel("assessLp_exception.mps");
    throw std::runtime_error("assessLp failed :(");
}
```
- As you can see, I save the model for further inspection [assessLp_exception.mps](https://github.com/ERGO-Code/HiGHS/files/10861929/assessLp_exception.mps.txt)

- When I run this model on the console it works surprisingly.

[...]

PR #1199, opened 2023-03-09
Fix 1187

jajhall:
Although presolve can yield small values in the matrix, they were only stripped out (by assessLp) in debug. This suggests that they are no real danger to the simplex solver. The only danger is pivoting on them, but that implies that values of roughly that size have been chosen in the ratio test. Even with the filter, values of 1e-9 could be in the matrix, and these would be bad pivots. Hence, since the small values may play a meaningful role in postsolve, then it's better to keep them.

The fix only executes the assert if `kAllowDeveloperAssert` is true, otherwise it calls `assessSmallValues` to report on matrix values (when `log_dev_level` is positive)


jajhall:
This will fix #1187 and #1174

PR #1200, opened 2023-03-10
Improve formatting of docstrings in the C API

odow:
This is a very minor thing and I haven't finished yet, but I though I'd open this to start with.

I was working on https://github.com/jump-dev/HiGHS.jl/pull/152 and got annoyed at the inconsistency in capitalization and punctuation in the docstrings that were coming through.

Ideally, I think we should aim for the Gurobi API documentation: https://www.gurobi.com/documentation/10.0/refman/c_setdblattr.html.

### DIFF
--- a/src/interfaces/highs_c_api.h
+++ b/src/interfaces/highs_c_api.h
@@ -229,21 +229,21 @@ const char* Highs_version(void);
  *
  * @returns The HiGHS major version number.
  */
-HighsInt Highs_versionMajor(void);
+HighsInt Highs_versionMajor();
 
 /**
  * Return the HiGHS minor version number.
  *
  * @returns The HiGHS minor version number.
  */
-HighsInt Highs_versionMinor(void);
+HighsInt Highs_versionMinor();
 
 /**
  * Return the HiGHS patch version number.
  *
  * @returns The HiGHS patch version number.
  */
-HighsInt Highs_versionPatch(void);
+HighsInt Highs_versionPatch();
 
 /**
  * Return the HiGHS githash.
@@ -648,8 +648,8 @@ HighsInt Highs_getNumOptions(const void* highs);
  * Get the name of an option identified by index
  *
  * @param highs     A pointer to the Highs instance.
- * @param index     the index of the option
- * @param name      the name of the option
+ * @param index     The index of the option.
+ * @param name      The name of the option.
  *
  * @returns A `kHighsStatus` constant indicating whether the call succeeded.
  */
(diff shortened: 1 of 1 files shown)


src/interfaces/highs_c_api.h
```diff
@@ -211,80 +213,80 @@ void* Highs_create(void);
...
 
 /**
- * Return the HiGHS major version number
+ * Return the HiGHS major version number.
  *
- * @returns the HiGHS major version number
+ * @returns The HiGHS major version number.
  */
-HighsInt Highs_versionMajor();
+HighsInt Highs_versionMajor(void);
```
odow: Should I make this a separate PR? I got a warning parsing this in Julia.

jajhall: No need for a separate PR. I've no idea what this `void` does, anyway

jajhall:
Thanks for all this @odow . I guess that you need proper sentences in the docstrings for subsequent appearance in documentation.

Are the non-comment changes limited to inserting `void` into some method parameters? They seem to be, as I skimmed through everything. If so - and the CI tests pass anyway - I'm happy to approve the PR for whenever you's happy to merge.

jajhall:
As for aiming for the Gurobi API documentation: https://www.gurobi.com/documentation/10.0/refman/c_setdblattr.html, there's no chance of us giving this level of web-based documentation in the short or medium term. Or do you just mean that the C API docstrings should yield this if processed appropriately?

We've made great strides on [documentation](https://ergo-code.github.io/HiGHS/) in the past few weeks, but C and C++ users will have to read header files to get all the details for the foreseeable future

odow:
> Are the non-comment changes limited to inserting void into some method parameters? 

Yes. I'll move to a separate PR so it's easier to track in future if anyone goes looking.

> Or do you just mean that the C API docstrings should yield this if processed appropriately?

Yeah, I meant the format of the docstrings, not that we need a website with documentation.

The docstrings in HiGHS.jl show up as:

https://github.com/jump-dev/HiGHS.jl/blob/1754060e4b7d756749686a9b05583136e418efdd/src/gen/libhighs.jl#L208-L221

image

It'd be nicer if `the HiGHS compilation date` was a properly punctuated sentence :smile:

jajhall:
OK I see that you want a proper sentence following `@returns`, rather than the completion of a sentence beginning with "Returns" :-)

odow:
I've split the `void` issue: https://github.com/ERGO-Code/HiGHS/pull/1201


The other reason to improve these is that we could add the docstrings into the documentation. Here's something I mocked up locally:

image

(With the old docstrings available in HiGHS.jl)

jajhall:
Have you finished?

odow:
Yeah finished for now. Although once https://github.com/ERGO-Code/HiGHS/pull/1203 is merged and we can more easily review the formatting I might have another go.

So either merge, or wait for #1203, up to you.

PR #1201, opened 2023-03-13
Fix signature of Highs_versionXXX in highs_c_api.h

odow:
Stackoverflow tells me that this is the correct way to say that a function has zero arguments in C: https://stackoverflow.com/questions/693788/is-it-better-to-use-c-void-arguments-void-foovoid-or-not-void-foo

I got a warning when I tried to parse the header in Julia. x-ref: https://github.com/ERGO-Code/HiGHS/pull/1200


jajhall (approved):
OK

PR #1203, opened 2023-03-13
Add documentation to latest

odow:
This is a copy of the documentation currently in the highs-docs branch.

The make.jl script is slightly modified to enable auto-deployment via CI.

I don't have permission to do so, but we'll need to change the GitHub pages setting from building from `/docs` to the `gh-pages` branch.

The main benefit of having the documentation in `latest` is that it's available to people working from source, and that it makes it easier to modify in a PR. For example, if you add a new option, then you should also update the documentation to reflect that in the same PR. No need to open separate PRs to `latest` and `highs-docs`.

If you agree to merge this, then in a separate PR I can parse the C docstrings and add them into the docs.


jajhall:
Thanks. I'll confer with @galabovaa

docs/make.jl
```diff
@@ -0,0 +1,71 @@
...
+    ],
+)
+
+# ==============================================================================
+#  Deploy everything in `build`
+# ==============================================================================
+
+Documenter.deploydocs(;
+    repo = "github.com/ERGO-Code/HiGHS.git",
+    push_preview = true,
```
odow, line 70:
```suggestion
    push_preview = true,
    devbranch = "latest",
```

odow:
Here's the build script to incorporate the C API automatically from `highs_c_api.h`: https://github.com/ERGO-Code/HiGHS/pull/1204

odow:
cc @mortenpi Documenter.jl is breaking into the world of C libraries.

odow:
@galabovaa the docs are now sitting at https://github.com/ERGO-Code/HiGHS/tree/gh-pages

So the next step is to change the settings over. Here are the settings we have for JuMP

image

galabovaa:
Hi Oscar, our settings are now as yours, thank you for your help with the Julia documentation!

PR #1214, opened 2023-03-17
remove duplicate old copyright message

svigerske:
Removes duplicate copyright messages that I believe were kept on mistake.


jajhall (approved):
Thanks

Issue #1229, opened 2023-03-22
Too many copies while processing solution


mingodad:
I've noticed that while processing the solution through the `C` api (and probably elsewhere), for example:
```
HighsInt Highs_getSolution(const void* highs, double* col_value,
                           double* col_dual, double* row_value,
                           double* row_dual) {
  HighsSolution solution = ((Highs*)highs)->getSolution(); //!!!<<<< here an unnecessary copy of HighsSolution
...
}
```
It seems that this would be the desired way:
```
HighsInt Highs_getSolution(const void* highs, double* col_value,
                           double* col_dual, double* row_value,
                           double* row_dual) {
  const HighsSolution &solution = ((Highs*)highs)->getSolution(); //!!!<< no copy just get a const reference.
...
}
```

mingodad:
Here is my diff (notice it also include ` && !defined(__clang__)` to avoid warnings when compiling with `clang`):
```
diff --git a/src/interfaces/highs_c_api.cpp b/src/interfaces/highs_c_api.cpp
index 6ab7279d6..90d50b9a2 100644
--- a/src/interfaces/highs_c_api.cpp
+++ b/src/interfaces/highs_c_api.cpp
@@ -32,9 +32,8 @@ HighsInt Highs_lpCall(const HighsInt num_col, const HighsInt num_row,
   status = highs.run();
 
   if (status == HighsStatus::kOk) {
-    HighsSolution solution;
     HighsBasis basis;
-    solution = highs.getSolution();
+    const HighsSolution &solution = highs.getSolution();
     basis = highs.getBasis();
     *model_status = (HighsInt)highs.getModelStatus();
     const HighsInfo& info = highs.getInfo();
@@ -87,8 +86,7 @@ HighsInt Highs_mipCall(const HighsInt num_col, const HighsInt num_row,
   status = highs.run();
 
   if (status == HighsStatus::kOk) {
-    HighsSolution solution;
-    solution = highs.getSolution();
+    const HighsSolution &solution = highs.getSolution();
     *model_status = (HighsInt)highs.getModelStatus();
     const HighsInfo& info = highs.getInfo();
     const bool copy_col_value =
@@ -132,9 +130,8 @@ HighsInt Highs_qpCall(
   status = highs.run();
 
   if (status == HighsStatus::kOk) {
-    HighsSolution solution;
     HighsBasis basis;
-    solution = highs.getSolution();
+    const HighsSolution &solution = highs.getSolution();
[...]

PR #1230, opened 2023-03-22
Avoided unnecessary solution and basis copies in C API

jajhall:
Avoided unnecessary solution and basis copies in C API and `clang` compiler warnings in `util/HighsDataStack.h` and `util/HighsHash.h`


jajhall:
This will close #1229

PR #1243, opened 2023-04-05
update

Thiago-NovaesB:
add methods

### DIFF
--- a/src/interfaces/highs_csharp_api.cs
+++ b/src/interfaces/highs_csharp_api.cs
@@ -7,934 +7,982 @@ using System.Text;
 
 public enum HighsStatus
 {
-   kError = -1,
-   kOk,
-   kWarning
+    kError = -1,
+    kOk,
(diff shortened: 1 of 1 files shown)


jajhall (commented):
Since there are LOTS of changes due to adding or removing blank spaces, it's hard to see whether there are any changes that would affect execution. I assume that the blank space modifications matter. If they do, and there's nothing controversial that affects execution, the PR is OK by me

Thiago-NovaesB:
I don't know exactly why, but the original interface had a spacing of 3 whitespaces, but a tab has 4whitespaces, doesn't it? So there are many changes, but other than that, the following methods have been added in the interface:

Highs_clearSolver
Highs_passColName
Highs_passRowName
Highs_writeOptions
Highs_writeOptionsDeviations

jajhall (approved):
Thanks for the clarification on what's been added.

PR #1245, opened 2023-04-07
Highspy

galabovaa:
> Replace nose with pytest
> Move python interface to root folder
> Workflow to test pip install on macOS
> Bazel test target moved to root dir


jajhall (approved):
Well done! I guess that the essence of what nose did has been taken on by pytest

galabovaa:
> Well done! I guess that the essence of what nose did has been taken on by pytest

Thanks! Yes, exactly :)

PR #1262, opened 2023-04-18
Get column (row) by name

jajhall:
Responding to #1138

### DIFF
--- a/src/interfaces/highs_c_api.h
+++ b/src/interfaces/highs_c_api.h
@@ -1639,7 +1639,8 @@ HighsInt Highs_getColName(const void* highs, const HighsInt col, char* name);
 /**
  * Get the index of a column from its name.
  *
- * If multiple columns have the same name, one is chosen arbitrarily.
+ * If multiple columns have the same name, or if no column exists with `name`, this 
+ * function returns `kHighsStatusError`.
  *
  * @param name A pointer of the name of the column to query.
  * @param col  A pointer in which to store the index of the column
--- a/src/interfaces/highs_c_api.h
+++ b/src/interfaces/highs_c_api.h
@@ -1613,7 +1613,9 @@ HighsInt Highs_getRowsByMask(const void* highs, const HighsInt* mask,
 HighsInt Highs_getRowName(const void* highs, const HighsInt row, char* name);
 
 /**
- * Get the row for a name
+ * Get the index of a row from its name.
+ *
+ * If multiple rows have the same name, one is chosen arbitrarily.
  *
  * @param name A pointer of the name of the row to query.
  * @param row  A pointer in which to store the index of the row
--- a/src/interfaces/highs_c_api.h
+++ b/src/interfaces/highs_c_api.h
@@ -1615,7 +1615,8 @@ HighsInt Highs_getRowName(const void* highs, const HighsInt row, char* name);
 /**
  * Get the index of a row from its name.
  *
- * If multiple rows have the same name, one is chosen arbitrarily.
+ * If multiple rows have the same name, or if no row exists with `name`, this 
+ * function returns `kHighsStatusError`.
  *
  * @param name A pointer of the name of the row to query.
  * @param row  A pointer in which to store the index of the row
--- a/src/interfaces/highs_c_api.h
+++ b/src/interfaces/highs_c_api.h
@@ -1636,7 +1636,9 @@ HighsInt Highs_getRowByName(const void* highs, const char* name, HighsInt* row);
 HighsInt Highs_getColName(const void* highs, const HighsInt col, char* name);
 
 /**
- * Get the column for a name
+ * Get the index of a column from its name.
+ *
+ * If multiple columns have the same name, one is chosen arbitrarily.
  *
  * @param name A pointer of the name of the column to query.
  * @param col  A pointer in which to store the index of the column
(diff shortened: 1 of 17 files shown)


src/lp_data/Highs.cpp
```diff
@@ -2431,6 +2433,20 @@ HighsStatus Highs::getColName(const HighsInt col, std::string& name) const {
   return HighsStatus::kOk;
 }
 
+HighsStatus Highs::getColByName(const std::string& name, HighsInt& col) {
+  HighsLp& lp = model_.lp_;
+  if (!lp.col_names_.size()) return HighsStatus::kError;
+  if (!lp.name2col_.size()) {
+    for (HighsInt iCol = 0; iCol < lp.num_col_; iCol++)
+      lp.name2col_.emplace(lp.col_names_[iCol], iCol);
```
odow: So I guess this has the same behavior as Gurobi: https://www.gurobi.com/documentation/10.0/refman/c_getvarbyname.html

> Retrieves a variable from its name. If multiple variables have the same name, this routine chooses one arbitrarily.

src/interfaces/highs_c_api.h
```diff
@@ -1612,6 +1612,16 @@ HighsInt Highs_getRowsByMask(const void* highs, const HighsInt* mask,
  */
 HighsInt Highs_getRowName(const void* highs, const HighsInt row, char* name);
 
+/**
+ * Get the row for a name
```
odow:
```suggestion
 * Get the index of a row from its name.
 *
 * If multiple rows have the same name, one is chosen arbitrarily.
```

src/interfaces/highs_c_api.h
```diff
@@ -1623,6 +1633,16 @@ HighsInt Highs_getRowName(const void* highs, const HighsInt row, char* name);
  */
 HighsInt Highs_getColName(const void* highs, const HighsInt col, char* name);
 
+/**
+ * Get the column for a name
```
odow:
```suggestion
 * Get the index of a column from its name.
 *
 * If multiple columns have the same name, one is chosen arbitrarily.
```

src/lp_data/Highs.cpp
jajhall: Yes, if folk duplicate names then there's not much else to do. I've not tested what happens if there are duplicate names. There will be an "error" return from the second call to `lp.name2col_.emplace(lp.col_names_[iCol], iCol);` with the same name, but I can't see much value in doing anything. Logging a comment and returning a warning would be easy enough though. Knowing that there were duplicate names could also prompt a logging message and warning when someone writes a model or solution file.

odow: The alternative is what JuMP does:

https://github.com/jump-dev/JuMP.jl/blob/9428757d3399c95a44cb2b1f6042a3a6b00a3246/src/variables.jl#L447-L452

```
Returns the reference of the variable with name attribute `name` or `Nothing` if
no variable has this name attribute. Throws an error if several variables have
`name` as their name attribute.
```

jajhall: I see that this is more helpful than what Gurobi does, so will add it.

Other than having vectors of length `num_col` and `num_row` indicating whether a name is duplicate, is there a neat way to know whether duplicates exist?

odow: When creating `lp.name2col_`, if you detect a key is already added, then change the value to some sentinel value like `-1`. 

Then, on query, if the value is `-1`, throw an error.

odow: Because it's okay for multiple rows/columns to have duplicate names (e.g., during printing). The problem is only if someone tries to query them.

Not sure if you check for duplicate names when writing out LP/MPS files, but you need to take that into account too.

jajhall: > When creating lp.name2col_, if you detect a key is already added, then change the value to some sentinel value like -1.

That's neat, thanks

We don't currently check for duplicate names when writing out LP/MPS files - just empty column/row name vectors - but I realised that this would allow us to do so. Would you throw an error, or just a warning?

odow: On files, it's a bit involved, but we deduplicate names by appending `_i` to duplicate names, where `i` increments every time we see the same name:
https://github.com/jump-dev/MathOptInterface.jl/blob/539f1aa47c1805023502644b643e84aeb3d18435/src/FileFormats/utils.jl#L11-L178

For now you could just leave it, and rely on the consumer of the file to detect duplicate names. It's the user's fault.

src/interfaces/highs_c_api.h
```diff
@@ -1612,6 +1612,18 @@ HighsInt Highs_getRowsByMask(const void* highs, const HighsInt* mask,
  */
 HighsInt Highs_getRowName(const void* highs, const HighsInt row, char* name);
 
+/**
+ * Get the index of a row from its name.
+ *
+ * If multiple rows have the same name, one is chosen arbitrarily.
```
odow:
```suggestion
 * If multiple rows have the same name, or if no row exists with `name`, this 
 * function returns `kHighsStatusError`.
```

src/interfaces/highs_c_api.h
```diff
@@ -1623,6 +1635,18 @@ HighsInt Highs_getRowName(const void* highs, const HighsInt row, char* name);
  */
 HighsInt Highs_getColName(const void* highs, const HighsInt col, char* name);
 
+/**
+ * Get the index of a column from its name.
+ *
+ * If multiple columns have the same name, one is chosen arbitrarily.
```
odow:
```suggestion
 * If multiple columns have the same name, or if no column exists with `name`, this 
 * function returns `kHighsStatusError`.
```

PR #1263, opened 2023-04-19
Updated the python interface to support a simplified model builder.

mathgeekcoder:
Updates include changes to documentation, example files, test scripts, and highs python interface.

There's a lot of boilerplate code that overrides the default operators ```[+, -, *, <=, ==, =>]```.  One new feature that I've recently added is support for linear expressions with lower and upper bounds, i.e., 0 <= 2*x + 3*y <= 10.  I've tried to add plenty of test cases to make sure that it works as expected, but I might have missed something.

Happy for feedback and opinions regarding the name of functions etc.

# New API Documentation 
 
## Highs 
 
Highs class is an interface to the HiGHS solver. It provides methods to add variables and constraints, solve linear programming problems, and retrieve the solution. 

Previously existing functions remain unchanged.

### Methods 
 
- ```minimize(obj)```: Minimizes the given linear expression obj. 
- ```maximize(obj)```: Maximizes the given linear expression obj. 
- ```addVar(lb=0, ub=inf, obj=0, type=kContinuous, name=None)```: Adds a new variable to the solver, with given parameters.
- ```addIntegral(lb=0, ub=inf, obj=0, name=None)```: Adds a new integral variable to the solver.  
- ```addBinary(obj=0, name=None)```: Adds a new binary variable to the solver. 
- ```addConstr(cons, name=None)```: Adds a new constraint to the solver. The constraint is given as a linear expression. 
- ```update()```: Batch update the variables in the solver, i.e., provides performance improvement when adding many variables.
- ```getVars()```: Returns a list of all variables in the solver. 
- ```getConstrs()```: Returns a list of all constraints in the solver. 
- ```removeVar(var)```: Removes the given variable from the solver. 
- ```removeConstr(cons)```: Removes the given constraint from the solver. 
- ```chgCoeff(cons, var, val)```: Changes the coefficient of the given variable and constraint to the given value. 
- ```val(var)```: Returns the value of the given variable in the solution. 
- ```vals(vars)```: Returns a list of values of the given variables in the solution. 

### Properties 

- ```inf```: Returns the infinity constant kHighsInf. 
- ```numVars```: Returns the number of variables in the solver. 
- ```numConstrs```: Returns the number of constraints in the solver.

### DIFF
--- a/docs/src/interfaces/python/example-py.md
+++ b/docs/src/interfaces/python/example-py.md
@@ -37,6 +37,20 @@ subject to              x1 <= 7
             0 <= x0 <= 4; 1 <= x1
 ```
 
+Using the simplified interface, the model can be built as follows:
+
+```python
+x0 = h.addVar(lb = 0, ub = 4)
+x1 = h.addVar(lb = 1, ub = 7)
+
+h.addConstr(5 <=   x0 + 2*x1 <= 15)
+h.addConstr(6 <= 3*x0 + 2*x1)
+
+h.minimize(x0 + x1)
+```
+
+Alternatively, the model can be built using the more general interface, which allows the user to specify the model in a more flexible way.
+
 Firstly, one variable at a time, via a sequence of calls to `addVar` and `addRow`:s
 ```python
 inf = highspy.kHighsInf
--- a/examples/minimal.py
+++ b/examples/minimal.py
@@ -1,16 +1,11 @@
 import highspy
-import numpy as np
-inf = highspy.kHighsInf
+
 h = highspy.Highs()
-h.addVars(2, np.array([-inf, -inf]), np.array([inf, inf]))
-h.changeColsCost(2, np.array([0, 1]), np.array([0, 1], dtype=np.double))
-num_cons = 2
-lower = np.array([2, 0], dtype=np.double)
-upper = np.array([inf, inf], dtype=np.double)
-num_new_nz = 4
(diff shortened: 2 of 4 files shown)


jajhall (commented):
Excellent. I see that this is also a very much more Python style interface

If I understand correctly, this extends `highspy` with methods for which there is no equivalent in the C++ API. Which is fine in principle, but will have to be highlighted in the HiGHS documentation, as the documentation of `highspy` is also intended to document the C++ API. 

Also, I note that we now have an overloading of addVar in `highspy`. Presumably this is OK

I'll merge this into a local branch so that I can have a play with it before merging into latest

jajhall (approved):
OK to merge into highspy-modeller

fhk:
So looks like this hasn't made it to pypi @jajhall .

I see there is a 1.7 dev release, is there any plans to update?

Mainly as the docs have drifted.

https://or.stackexchange.com/questions/11125/starting-with-highs

vs. 

https://ergo-code.github.io/HiGHS/stable/interfaces/python/example-py/

galabovaa:
Yes, we are planning to release v1.7.2 soon with some method name modifications and an update of the documentation.

The pre-release version should be good to use until then. Not so much will change between 1.7.1.dev1 and 1.7.2 apart from the method names for addVar introduced here. These are planned to change to addVariable, so the addVar from the highs class in the python bindings can be accessed too. We are currently working on the updates.

PR #1264, opened 2023-04-19
Add writeBasis interface method to highspy

fuglede:
As per [this discussion](https://github.com/ERGO-Code/HiGHS/discussions/1198#discussioncomment-5648001), this makes it possible to access `Highs::writeBasis` through `highspy`.


jajhall (approved):
Thanks, particularly for the testing

PR #1272, opened 2023-04-25
Add option STDCALL for cmake under windows and extend build documentation

peno64:


CMakeLists.txt
```diff
@@ -92,6 +92,16 @@ endif()
 
 include(CTest)
 
+if(MSVC)
+	# This option is only available when building with MSVC. By default, highs
+	# is build using the cdecl calling convention, which is useful if you're
+	# writing C. However, the CLR and Win32 API both expect stdcall.
```
tobiass-sdl, line 98: I guess you are talking about P/Invoke. I always use cdecl and it works fine with explicite calling convention [DllImport(CallingConvention = CallingConvention.Cdecl)].
Therefore this description imho way to simplified.
Are you sure that "if(MSVC)" is sufficient? What's about mingw, clang and the intel compiler?

peno64:
Vb6 for example only accepts stdcall callibg convention not cdecl. I have no idea about mingw or gcc. Most probably they also would need different switches.

Peter


________________________________

tobiass-sdl:
Please note that I'm neither a maintainer of highs nor am Iarguing against stdcall. 
The documentation/comments may suit your case but in general they are more misleading than helpful. E.g. with mingw the library is named libhighs.dll. Whether using cmake's visual studio generator is a good advice is up to debate (if you open a cmake project in visual studio it will use ninja instead), I doubt that cmake will default to 64 bit  on a 32 bit windows, ... 

Another approach to this would be create a proper highs nuget package with the low level api of the bundled C# interface. Than a user from the .net ecosystem do not need to care about cdecl and stdcall. I do not use VB though.

galabovaa (approved):

PR #1318, opened 2023-06-09
Handle no spaces in quadratic expressions in LP files

arcondello:
This is done by disallowing `[` or `]` in variables names. This is a bit of an additional restriction on what was previously allowed, but it's consistent with [CPLEX's LP spec](https://www.ibm.com/docs/en/icos/12.8.0.0?topic=representation-variable-names-in-lp-file-format) so hopefully not *too* much of a restriction .

The specific failing case was expressions of the form `[ x0 * x1]`, but I added tests for a few more variations for completeness.


jajhall (commented):
Looks uncontroversial

feldmeier:
Looks good to me

PR #1319, opened 2023-06-12
MAINT: Cleanup to remove pyomo dependency

### LINKED ISSUE #1022: highspy unnecessarily requires pyomo
`highspy` [lists `pyomo` as a dependency](https://github.com/ERGO-Code/HiGHS/blob/master/src/interfaces/highspy/pyproject.toml#L6), just to use [a single small function from `pyomo` in it's setup.py](https://github.com/ERGO-Code/HiGHS/blob/master/src/interfaces/highspy/setup.py#L5).

I think it's best that requirements are kept as light as possible.

Towards this, I can open a PR to remove the `pyomo` dependency by simply copying [`find_library`](https://github.com/Pyomo/pyomo/blob/main/pyomo/common/fileutils.py#L293) in setup.py.

HaoZeke:
Closes #1022.

The `find_library` implementation here is much simpler than that of `pyomo`, however, for the use case here within `setup.py` it is sufficient.

I couldn't find any use of the `capture_output` class, so that was removed without any further changes.


jajhall:
Thanks for this. However, I thought that @galabovaa had removed the Pyomo dependency in v1.5.3

.github/workflows/test-python-api-ubuntu.yml
HaoZeke, line 1: This file seems to be testing the same version as that included in `test-python-platforms` and I think it can be safely removed.

HaoZeke:
> Thanks for this. However, I thought that @galabovaa had removed the Pyomo dependency in v1.5.3

Looking over the commit history it seems that it was removed from `pyproject.toml` but it is still imported in `setup.py` and the CI was pre-installing it (along with `pybind11`).

HaoZeke:
Closing in favor of fully reworked bindings in #1342

PR #1326, opened 2023-06-14
Issue 1325

### LINKED ISSUE #1325: Write out to MPS broken for Integer variables
In the current version on pypi v1.5.3 the MPS format given out by Highs seems to be broken. As far as I can see, integer variables are mistakenly written out as binary variables. 

The following is a reproducible code 

```python

import highspy

with open("test.mps", "w") as fn:
    fn.write("""
NAME
ROWS
 N  OBJ
 G  c0
 G  c1
 G  c2
COLUMNS
    MARKER    'MARKER'                 'INTORG'
    x0        OBJ       1
    x0        c0        1
    x1        OBJ       1
    x1        c1        1
    x2        OBJ       1
    x2        c2        1
    MARKER    'MARKER'                 'INTEND'
    x3        OBJ       2
    x3        c0        1
    x4        OBJ       2
    x4        c1        1
    x5        OBJ       2
    x5        c2        1
RHS
    RHS1      c0        10.99
    RHS1      c1        10.99
    RHS1      c2        10.99
BOUNDS
 LI BND1      x0        0
 LI BND1      x1        0
 LI BND1      x2        0
ENDATA
"""
)


h = highspy.Highs()
h.readModel("test.mps")
h.run()
print("Optimum at first run:", h.getObjectiveValue())


# now write out to MPS and load again 
h.writeModel("test2.mps")
h2 = highspy.Highs()
h2.readModel("test2.mps")
h2.run()
print("Optimum at second run (after writing and loading again):", h2.getObjectiveValue())
```

Returns
```
Optimum at first run: 33.0
Optimum at second run (after writing and loading again): 62.94
```

Comparing the new MPS with the original, we see that the `Bounds` section is missing, which leads to a wrong interpretation of the integer variables type.

jajhall:
This fixes a serious bug in the MPS write that was introduced by @jajhall 

This will close #1325


jajhall:
Another issue that I spotted is that of writing semi-variables with infinite upper bounds to MPS. The SC/SI bound has to be specified to define the variable as semi, and the infinite upper bound has to be stated. The use of kHighsInf causes the corresponding unit test to fail with mingw, so some large no-infinite value should be used.

galabovaa (approved):

Issue #1331, opened 2023-06-18
run does not reset timers


odow:
Each call to `run` does not reset the timers, so repeated calls (e.g., in a row or column generation scheme) eventually hit a time limit. 

Notice how the `HiGHS run time      :` keeps increasing each iteration, and eventually the time limit is hit:

```julia
julia> using HiGHS

julia> highs = Highs_create()
Ptr{Nothing} @0x00007fe078c24000

julia> Highs_readModel(highs, "/Users/Oscar/Downloads/bug.mps")  # Doesn't matter. Arbitrary model
Running HiGHS 1.5.3 [date: 1970-01-01, git hash: 45a127b78]
Copyright (c) 2023 HiGHS under MIT licence terms
0

julia> Highs_setIntOptionValue(highs, "time_limit", 1)
0

julia> for _ in 1:7
           println()
           Highs_run(highs)
       end

Iteration, Runtime, ObjVal, NullspaceDim
0, 0.019859, -0.684376, 0
1000, 0.126506, -196828.896682, 0
1152, 0.142994, -198524.875970, 0
Model   status      : Optimal
Simplex   iterations: 44
QP ASM    iterations: 1152
Objective value     : -1.9852487597e+05
HiGHS run time      :          0.14

Iteration, Runtime, ObjVal, NullspaceDim
0, 0.163659, -0.684376, 0
1000, 0.270918, -196828.896682, 0
1152, 0.286709, -198524.875970, 0
Model   status      : Optimal
Simplex   iterations: 44
QP ASM    iterations: 1152
Objective value     : -1.9852487597e+05
HiGHS run time      :          0.29

Iteration, Runtime, ObjVal, NullspaceDim
0, 0.305871, -0.684376, 0
1000, 0.420568, -196828.896682, 0
1152, 0.437243, -198524.875970, 0
Model   status      : Optimal
Simplex   iterations: 44
QP ASM    iterations: 1152
Objective value     : -1.9852487597e+05
HiGHS run time      :          0.44

Iteration, Runtime, ObjVal, NullspaceDim
0, 0.458031, -0.684376, 0
1000, 0.568664, -196828.896682, 0
1152, 0.584456, -198524.875970, 0
Model   status      : Optimal
Simplex   iterations: 44
QP ASM    iterations: 1152
Objective value     : -1.9852487597e+05
HiGHS run time      :          0.59

Iteration, Runtime, ObjVal, NullspaceDim
0, 0.605321, -0.684376, 0
[...]

PR #1332, opened 2023-06-18
Add Highs_zeroAllClocks to reset clocks between calls to Highs_run

odow:
Closes https://github.com/ERGO-Code/HiGHS/issues/1331

@jajhall let me know if you had something else in mind.

### DIFF
--- a/src/Highs.h
+++ b/src/Highs.h
@@ -419,6 +419,11 @@ class Highs {
   const HighsSolution& getSolution() const { return solution_; }
 
   /**
+   * @brief Zero all clocks in the internal HighsTimer instance
+   */
+  void zeroAllClocks() { timer_.zeroAllClocks(); };
+
+  /**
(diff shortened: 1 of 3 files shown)


jajhall (approved):
This is fine for your purposes. What's been discussed - in the context of the SCIP interface to HiGHS - is that the HiGHS clock is reset to zero every time run() is called. However, this implies housekeeping to be added whenever run() is called within a call to run(), as well as breaking the behaviour that some will assume. Hence the need for it to only appear in v2.0

odow:
Ah. I can see how the recursive use of `run` complicates things. We actually have a few meta-solvers in JuMP that have the opposite problem: they iteratively call solvers in a row or column generate scheme and want a time limit on the total solve, so they have to keep calling `set_time_limit(model, time_limit - current_time)` before each solve. In that context, documenting people to call `zeroAllClocks` isn't that bad. It's a fairly reasonable choice once the backstory is explained.

PR #1336, opened 2023-06-28
ENH: Add meson build system

HaoZeke:
This is almost at parity with the `cmake` build system, and should be simpler to maintain. It is also easier to make `shared` and `static` libraries from the commandline / at the subproject level. ~The only missing feature is that `csharp` is not a supported language so the interface cannot be built with `meson`.~ **Added** `csharp` interface build.

Usage:

```bash
meson setup build
meson compile -C build
meson install -C build
meson test -C build
```

The options are in `meson_options.txt` and also for static libraries:

```bash
meson setup --default-library static build
meson install -C build
```

This generates `libhighs.so`.


.github/workflows/build-meson.yml
```diff
@@ -0,0 +1,47 @@
...
+      - name: Build Static
+        shell: bash -l {0}
+        run: |
+          meson setup bbdir_static -Duse_zlib=True --default-library static --prefix $CONDA_PREFIX
+          meson install -C bbdir_static
+      - name: Build Fortran
+        shell: bash -l {0}
+        run: |
+          meson setup bbdir_fortran -Duse_zlib=True -Dwith_fortran=True
+          meson install -C bbdir_fortran
```
HaoZeke: These don't all need to be tested here in CI like this. Probably just a single one.

eli-schwartz:
> The only missing feature is that `csharp` is not a supported language so the interface cannot be built with `meson`.

Actually meson does support this :) the language name is `cs` in meson though. Test cases for it which demonstrate using it, can be found at: https://github.com/mesonbuild/meson/tree/master/test%20cases/csharp

meson.build
```diff
@@ -0,0 +1,63 @@
...
+_deps += [declare_dependency(link_args: '-lstdc++')]
+
+# Required
+threads_dep = dependency('threads',
+                        required: true)
+_deps += threads_dep
+
+# Optional
+zlib_dep = dependency('zlib',
+                      required: get_option('use_zlib'))
```
eli-schwartz, line 78: This option is declared as a boolean, so highs *always* tries to depend on zlib and merely permits falling back to disabling it if lookup fails. It would be better off as an option of `type: 'feature'` (defaulting to auto).

See e.g. https://wiki.gentoo.org/wiki/Project:Quality_Assurance/Automagic_dependencies

meson.build
```diff
@@ -0,0 +1,63 @@
...
+if zlib_dep.found()
+  _deps += zlib_dep
+  _incdirs += include_directories(['extern/zstr'])
+else
+    message('Couldn\'t find zlib')
```
eli-schwartz: (meson already prints a message of the form "Dependency zlib found: NO". Do we really need another message?)

src/meson.build
```diff
@@ -0,0 +1,256 @@
...
+import datetime
+
+today = datetime.date.today()
+today_str = today.strftime('%Y-%m-%d')
+print(today_str)
+'''
+today_cmd = run_command('python3', '-c',
+                        python_getdate,
+                       check: true)
+today_str = today_cmd.stdout().strip()
```
eli-schwartz: It probably makes sense to check https://reproducible-builds.org/docs/source-date-epoch/ here.

meson.build
HaoZeke: Good point, removing

src/meson.build
HaoZeke: Neat, modifying to check the variable

meson.build
```diff
@@ -26,6 +26,13 @@ if host_system == 'linux'
   _args += '-Wno-unused-const-variable'
 endif
 
+if cppc.get_id() == 'msvc'
+  add_project_arguments(cppc.get_supported_arguments('/W4'), language : 'cpp')
```
eli-schwartz: Note that meson can set /W4 for you if you use warning_level=3, which for some other compilers sets -Wpedantic so you may or may not want that.

We assume this option always works for the "msvc" compiler, any particular reason you're doing a supported-arguments check here?

meson.build
```diff
@@ -26,6 +26,13 @@ if host_system == 'linux'
...
+  add_project_arguments('/wd4018', '/wd4061', '/wd4100', '/wd4101', '/wd4127', '/wd4189', '/wd4244', '/wd4245', '/wd4267', '/wd4324', '/wd4365', '/wd4389', '/wd4456', '/wd4457', '/wd4458', '/wd4459', '/wd4514', '/wd4701', '/wd4820', language: 'cpp')
+  add_project_arguments('/MP', language: 'cpp')
```
eli-schwartz: The use of `/MP` seems wrong here. Isn't this supposed to instruct cl.exe to spawn multiple copies and compile each source file with one process, when given multiple source files on the command line?

Meson's ninja backend already creates one build edge per source file, it doesn't need a command-line option to cl.exe to unlock parallel compilation since ninja can already do this quite well.

Meson's Visual Studio backend uses whatever parallelism msbuild makes available, which is less... robust. It would benefit from /MP except that it doesn't actually work -- meson specifies output filenames for compiled object files, which surprisingly does mean that /MP gets ignored (!) so it does nothing. Meson does automatically set up a slightly newer approach to parallelized builds, though, see https://github.com/mesonbuild/meson/commit/14096260c434a6fcc80d06eeba006bd1965b213c. It apparently requires  Visual Studio 2019 16.3, see e.g. https://devblogs.microsoft.com/cppblog/improved-parallelism-in-msbuild/

tl;dr setting /MP does nothing but make command lines a bit bigger. I would suggest dropping this argument. For ninja builds, as well as for sufficiently *new* versions of VS, build parallelism works regardless, and you cannot really fix old versions of VS.

meson.build
HaoZeke: No, this was basically a straight port of the `CMakeLists.txt` ([here](https://github.com/ERGO-Code/HiGHS/blob/8d71e1ed3744b633da08a235c29f7d194b546ec6/CMakeLists.txt#L266-L275)), but since we do use `warning_level=3` here I'll drop it.

HaoZeke:
@galabovaa this is ready for review now, the documentation julia run is failing on `latest` as well. I will open the follow up `meson-python` PR targeting `latest` as well.

galabovaa (approved):

galabovaa:
Thank you, looks good to me! Merging in latest, so you can go ahead with the next RP

Issue #1328, opened 2023-06-15
File missing in  wheel for OS X on x86_64?


rschwarz:
When I look at the [uploaded wheels](https://pypi.org/project/highspy/1.5.3/#files) for `highspy` 1.5.3, I see that most files have a size around 2MB.

But the files for OS X for `x86_64`, eg [highspy-1.5.3-pp39-pypy39_pp73-macosx_10_9_x86_64.whl](https://files.pythonhosted.org/packages/93/ce/f9fc084399f51470c586fe4c8e3994038104e4c8a9a178776d3cbef3db2b/highspy-1.5.3-pp39-pypy39_pp73-macosx_10_9_x86_64.whl) are a lot smaller, at around 300KB.

It looks like it's missing the shared library for HiGHS itself. Compare this listing of the wheel content:
```
M Filemode      Length  Date         Time      File
- ----------  --------  -----------  --------  --------------------------------------------
  -rw-rw-r--       984  16-May-2023  17:09:08  highspy-1.5.3.dist-info/RECORD
  -rw-r--r--      1062  16-May-2023  14:09:04  highspy-1.5.3.dist-info/LICENSE
  -rw-r--r--        85  16-May-2023  14:09:04  highspy-1.5.3.dist-info/AUTHORS
  -rw-r--r--       116  16-May-2023  14:09:04  highspy-1.5.3.dist-info/WHEEL
  -rw-r--r--         8  16-May-2023  14:09:04  highspy-1.5.3.dist-info/top_level.txt
  -rw-r--r--       601  16-May-2023  14:09:04  highspy-1.5.3.dist-info/METADATA
  -rw-r--r--     41088   6-May-2023  16:21:04  highspy/highs_bindings.cpp
  -rw-r--r--       588  21-Apr-2023  13:40:46  highspy/__init__.py
  -rw-r--r--       672  21-Apr-2023  13:40:46  highspy/highs.py
  -rwxr-xr-x   1085544  16-May-2023  14:06:00  highspy/highs_bindings.pypy39-pp73-darwin.so
  -rw-r--r--     17826  16-May-2023  13:51:56  highspy/tests/test_highspy.py
  -rw-r--r--         0  21-Apr-2023  13:40:46  highspy/tests/__init__.py
- ----------  --------  -----------  --------  --------------------------------------------
               1148574                         12 files
```

with the listing of the corresponding wheel for `ARM64`, which has one additional file (`libhighs.1.5.3.dylib`):

```
[...]

Issue #1119, opened 2023-01-31
ImportError for `highspy`


goroderickgo:
Hello, I'm on an M1 Mac (MacOS Ventura, Python 3.10.9) and downloaded HiGHS via PyPI ([`highspy`](https://pypi.org/project/highspy/)) and am getting this import error. 

I didn't seen any similar issues asked previously, so I figured I'd ask here. Appreciate any help!

```
Python 3.10.9 (main, Jan 11 2023, 09:18:20) [Clang 14.0.6 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import highspy
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/e3-work/opt/anaconda3/envs/nmt-pins/lib/python3.10/site-packages/highspy/__init__.py", line 1, in <module>
    from .highs import (
  File "/Users/e3-work/opt/anaconda3/envs/nmt-pins/lib/python3.10/site-packages/highspy/highs.py", line 1, in <module>
    from .highs_bindings import (
ImportError: dlopen(/Users/e3-work/opt/anaconda3/envs/nmt-pins/lib/python3.10/site-packages/highspy/highs_bindings.cpython-310-darwin.so, 0x0002): symbol not found in flat namespace '__ZN5Highs10clearModelEv'

```

mckib2:
MacOS will use Clang to compile C++ sources, so we can use `llvm-cxxfilt` to find out exactly what it's complaining about:

    $ llvm-cxxfilt --strip-underscore '__ZN5Highs10clearModelEv'
    Highs::clearModel()

Looks like the library on PyPI either didn't compile Highs.cpp (unlikely) or something is not linked correctly (more likely) -- might be due to how conda is handling this wheel?  Try installing with pip to see if that resolves it.

If it doesn't, as a hacky workaround, try adding the path to libhighs: `sys.path.append("/Users/e3-work/opt/anaconda3/envs/nmt-pins/lib/python3.10/site-packages/highspy/.dylibs/")`  (also make sure there is indeed a library named `libhighs.1.5.0.dylib` somewhere in `/Users/e3-work/opt/anaconda3/envs/nmt-pins/lib/python3.10/site-packages/highspy`)

jajhall:
We've heard of people installing HiGHS via `(ana)conda¬, but we only support installation via `pip`.

chkwon:
I installed via `pip` and obtained the same error:


```bash
[...]

PR #1322, opened 2023-06-12
MAINT: Fixup pyb11 build

HaoZeke:
Closes #1320 and (apart from moving to a different repo, #1321). This should be much easier to maintain and use. Supersedes https://github.com/ERGO-Code/HiGHS/pull/1319

PR #1342, opened 2023-07-02
ENH: Meson pybind11

HaoZeke:
Needs #1336. Superseeds #1322. The wheels should be built correctly for `manylinux`, `musllinux` and `macos` (both `arm64` and `x86_64`) [as [seen here](https://github.com/HaoZeke/HiGHS/actions/runs/5434563842)] so it should fix the existing broken wheel issues, that is:

- Closes https://github.com/ERGO-Code/HiGHS/issues/1328
- Closes https://github.com/ERGO-Code/HiGHS/issues/1119

The windows build isn't perfect yet (linking issues, perhaps if @eli-schwartz could take a peek at the `meson` sections, and the [failed action log](https://github.com/HaoZeke/HiGHS/actions/runs/5434309985/jobs/9882644804)), am working on it.

**EDIT**: `cibuildwheel` is able to [generate wheels for windows as well](https://github.com/HaoZeke/HiGHS/actions/runs/5435337546), so this should be good to go


eli-schwartz:
```
  [163/165] Linking target src/highs.dll
  [164/165] Generating symbol file src/highs.dll.p/highs.dll.symbols
  [165/165] Linking target highspy/highspy.cp39-win_amd64.pyd
  FAILED: highspy/highspy.cp39-win_amd64.pyd
  "link"  /MACHINE:x64 /OUT:highspy/highspy.cp39-win_amd64.pyd highspy/highspy.cp39-win_amd64.pyd.p/highs_bindings.cpp.obj "/nologo" "/release" "/nologo" "/OPT:REF" "/DLL" "/IMPLIB:highspy\highspy.cp39-win_amd64.lib" "src\highs.lib" "C:\hostedtoolcache\windows\Python\3.9.13\x64\libs\python39.lib" "kernel32.lib" "user32.lib" "gdi32.lib" "winspool.lib" "shell32.lib" "ole32.lib" "oleaut32.lib" "uuid.lib" "comdlg32.lib" "advapi32.lib"
  LINK : fatal error LNK1181: cannot open input file 'src\highs.lib'
```

In my experience, missing implibs usually occur when there are missing -D defines that are supposed to set up dllexport/dllimport correctly, and as a result the *.dll contains no symbols. The link step for the DLL itself helpfully succeeds, but doesn't log a line stating that it created an implib, because it doesn't create one.

Apparently *.def files can export symbols without that being a compile-time option, thus serving as an alternative to dllexport, but provides no help to dllimport, so that's fun...

highspy/meson.build
```diff
@@ -2,7 +2,8 @@ py_mod = import('python')
 py = py_mod.find_installation(pure: false)
 pyb11_dep = [
   py.dependency(),
-  dependency('pybind11')
+  dependency('pybind11',
+            required: true)
```
eli-schwartz: Note that `required: true` is the default, you can omit it although some people would say "explicit is better than implicit". Just pointing out that this commit didn't *change* that, only made it... explicit.

meson.build
```diff
@@ -106,5 +106,6 @@ endif
 
 # --------------------- Bindings
 if get_option('with_pybind11')
+  subproject('pybind11')
```
eli-schwartz: ... but I'm not sure I see any point to this. You don't need to force the subproject to be used, it will already be automatically used unless pybind11 is found on the system with a suitable version matching the optional `version: '>='X.Y` kwargs.

If you just want to generally force the use of vendored dependencies / subprojects, then meson has a global option for this, `-Dwrap_mode=forcefallback`. It is preferrable to rely on that option, which you can set in the project() function's default_options, because it allows third-party builders to override it if they really prefer to use their own packaged versions of pybind11, e.g. linux distributions will greatly appreciate that flexibility. And yes, they can override the default_options by simply manually passing -Dwrap_mode=default (or nofallback) on the command-line, no patching of the build system needed.

So leaning on the global wrap_mode option allows everyone to be happy.

HaoZeke: Ah, makes sense, `wrap_mode` sounds like the way to go. Updated.

jajhall:
Thanks for your continued work on the meson build 🙂

HaoZeke:
> ```
>   [163/165] Linking target src/highs.dll
>   [164/165] Generating symbol file src/highs.dll.p/highs.dll.symbols
>   [165/165] Linking target highspy/highspy.cp39-win_amd64.pyd
>   FAILED: highspy/highspy.cp39-win_amd64.pyd
>   "link"  /MACHINE:x64 /OUT:highspy/highspy.cp39-win_amd64.pyd highspy/highspy.cp39-win_amd64.pyd.p/highs_bindings.cpp.obj "/nologo" "/release" "/nologo" "/OPT:REF" "/DLL" "/IMPLIB:highspy\highspy.cp39-win_amd64.lib" "src\highs.lib" "C:\hostedtoolcache\windows\Python\3.9.13\x64\libs\python39.lib" "kernel32.lib" "user32.lib" "gdi32.lib" "winspool.lib" "shell32.lib" "ole32.lib" "oleaut32.lib" "uuid.lib" "comdlg32.lib" "advapi32.lib"
>   LINK : fatal error LNK1181: cannot open input file 'src\highs.lib'
> ```
> 
> In my experience, missing implibs usually occur when there are missing -D defines that are supposed to set up dllexport/dllimport correctly, and as a result the *.dll contains no symbols. The link step for the DLL itself helpfully succeeds, but doesn't log a line stating that it created an implib, because it doesn't create one.
> 
> Apparently *.def files can export symbols without that being a compile-time option, thus serving as an alternative to dllexport, but provides no help to dllimport, so that's fun...

It seems to be an issue with the way libraries are looked up / handled on CI specifically. Pre-installing `highs` and then explicitly pointing `delvewheel` to it allows `cibuildwheel` to build and repair wheels correctly, so I think this should probably be good to go.

jajhall:
Not complaining, but two points

- Do you still need all the earlier PRs?
- Please PR to `latest` rather than `master`, as the latter only corresponds to releases and patches

Thanks!

HaoZeke:
> * Do you still need all the earlier PRs?

Conceptually there are two PRs, one which adds the `meson` build, and one which reworks the `pybind11` interface and fixes the wheels.


> * Please PR to `latest` rather than `master`, as the latter only corresponds to releases and patches
> 
> Thanks!


I'll close this and open two :)

Issue #1320, opened 2023-06-12
ENH: Simplify `pybind11` bindings (redundancy)


HaoZeke:
## Remove redundancy in bindings

The `highspy.highs_bindings` module defines the same set of attributes as the `highspy` and `highspy.highs` modules

`highspy.highs` has the same set of attributes as `highspy` so it doesn't make much sense to have them around together.

Essentially:

```python
In [2]: dir(highspy.highs_bindings)
Out[2]: 
['BasisValidity',
 'HessianFormat',
 'HighsBasis',
 'HighsBasisStatus',
 'HighsHessian',
 'HighsInfo',
 'HighsInfoType',
 'HighsLogType',
 'HighsLp',
 'HighsModel',
 'HighsModelStatus',
 'HighsObjectiveSolution',
 'HighsOptionType',
 'HighsOptions',
 'HighsPresolveStatus',
 'HighsRanging',
 'HighsRangingRecord',
 'HighsSolution',
 'HighsSparseMatrix',
 'HighsStatus',
 'HighsVarType',
 'MatrixFormat',
 'ObjSense',
 'SolutionStatus',
 '_Highs',
 '__doc__',
 '__file__',
 '__loader__',
 '__name__',
 '__package__',
 '__spec__',
 'kHighsIInf',
 'kHighsInf']

In [3]: dir(highspy)
Out[3]: 
['BasisValidity',
 'HessianFormat',
 'Highs',
 'HighsBasis',
 'HighsBasisStatus',
 'HighsHessian',
 'HighsInfo',
 'HighsInfoType',
 'HighsLogType',
 'HighsLp',
 'HighsModel',
 'HighsModelStatus',
 'HighsObjectiveSolution',
 'HighsOptionType',
 'HighsOptions',
 'HighsPresolveStatus',
 'HighsRanging',
 'HighsRangingRecord',
 'HighsSolution',
 'HighsSparseMatrix',
 'HighsStatus',
 'HighsVarType',
 'MatrixFormat',
 'ObjSense',
 'SolutionStatus',
 '__builtins__',
 '__cached__',
 '__doc__',
 '__file__',
 '__loader__',
 '__name__',
 '__package__',
 '__path__',
 '__spec__',
 'highs',
 'highs_bindings',
 'kHighsIInf',
 'kHighsInf']

In [4]: dir(highspy.highs)
Out[4]: 
['BasisValidity',
 'HessianFormat',
 'Highs',
 'HighsBasis',
[...]

PR #1343, opened 2023-07-02
ENH: Meson pybind11 build

### LINKED ISSUE #1321: BIND: Move python bindings and switch build systems
## Move away from `setup.py`

- A lot of issues (including the `pyomo` dep!) come from being unable to bundle `highs` the library with the bindings
- With `meson` as a build system this would be trivial (and it implements more robust library finding methods than the `find_library` stuff in `setup.py` or `pyomo` 

Personally (more divisively) I would prefer to have `meson` as a build system and then have the bindings in a different repository altogether, this approach has worked well for me (as a toy example there is https://github.com/TheochemUI/potlib and the bindings in https://github.com/TheochemUI/pypotlib)

Another reason to have a separation between binding code and the main library is that the issues pertaining to builds (especially for bindings) and general memory issues need not be relevant to the core library (and algorithm developers).

HaoZeke:
Needs #1336. Closes #1321, #1320, and #1022. The wheels should be built correctly (i.e. built with `cibuildwheel` and repaired by `auditwheel`, `delocate` and `delvewheel`) for `manylinux`, `musllinux`, windows (`arm_64`), and `macos` (both `arm64` and `x86_64`) [as [seen here](https://github.com/HaoZeke/HiGHS/actions/runs/5437445844)] so it should fix the existing broken wheel issues, that is it:

- Closes https://github.com/ERGO-Code/HiGHS/issues/1328
- Closes https://github.com/ERGO-Code/HiGHS/issues/1119

@galabovaa this will be hard to review in its current state, but after #1336 is merged and I rebase onto latest, the change-set will contain just the Python changes and it will be easier to review. @eli-schwartz had some great comments on the [original PR](https://github.com/ERGO-Code/HiGHS/pull/1342).

The windows issues were known problems discussed upstream as well:
- https://github.com/mesonbuild/meson-python/issues/265
- https://github.com/mesonbuild/meson-python/issues/259#issuecomment-1478539715

But for users consuming wheels from PyPI or building locally it isn't an issue.


.gitignore
```diff
@@ -1,4 +1,31 @@
...
+# Ninja
+build.ninja
+.ninja_deps
+.ninja_logs
+
+# Misc
+compile_commands.json
+
+# End of https://www.toptal.com/developers/gitignore/api/meson
+# #################
```
eli-schwartz: This almost entirely looks buggy, so I don't know what toptal.com's rationale here is.

`/subprojects/*` is reasonable enough -- meson subprojects will be downloaded there so you only want to check in `.wrap` files. Appropriately, it is rooted in `/`.

The rest are all files that are created inside of the build directory. Excluding any one of them -- or even a bunch of them -- is missing the point that it fails to gitignore every possible `*.o` file, or your `*.so` / `*.dll` outputs, or many many other things generated as part of a build.

... it's also redundant due to https://mesonbuild.com/FAQ.html#how-do-i-ignore-the-build-directory-in-my-vcs since this is one of the Quality of Life advantages of using meson; you do not need to gitignore the build directory at all.

That maybe should be fixed somewhere in this generator tool.

pyproject.toml
```diff
@@ -1,49 +1,51 @@
...
+
 [build-system]
-# Minimum requirements for the build system to execute.
-requires = [
-  "setuptools>=45",
-  "pybind11>=2.4",
-  "pyomo>=6.0",
-  "wheel>=0.2",
-] 
+requires = ["pybind11", "meson-python"]
```
eli-schwartz: Due to the use of forcefallback, the pybind11 requirement here is installed but not used -- meson doesn't bother checking for the pybind11-config script installed by the pybind11 wheel into the `pip` / `python -m build` isolated build venv. It goes straight to initializing pybind11 as a meson subproject and retrieves the pybind11 `declare_dependency()` object.

highspy/meson.build
```diff
@@ -0,0 +1,20 @@
+py_mod = import('python')
+py = py_mod.find_installation(pure: false)
+pyb11_dep = [
+  py.dependency(),
+  dependency('pybind11')
+]
+_deps += pyb11_dep
```
eli-schwartz: Due to ordering, this file is processed last in the build -- so `_deps` is not referred to again, after this, I think.

But if it were, then this seems to be a global array and would inject the py installation dependency as well as pybind11 into other targets, right?

I would probably have done something like:
```meson
pyb11_dep = [
  _deps,
  py.dependency(),
  dependency('pybind11')
]
```

eli-schwartz (changes requested):

highspy/meson.build
HaoZeke: Yup, I reworked the logic a bit to make it cleaner (hopefully).

pyproject.toml
HaoZeke: Ah neat, I didn't realize that. Updated :)

HaoZeke:
@galabovaa does this look alright?

eli-schwartz:
Looks fine to me from a meson perspective.

galabovaa (approved):

galabovaa:
Looks good to me! Merging in latest. Thank you for the time and effort you spent on this. 

I will now try to build highspy for all platforms with meson locally.

alspadoni:
This is great guys! When will the next release of highspy be uploaded on pypi with all these fixes?

jajhall:
I don't know, but it might not be long now that @galabovaa is back from holiday

AdeelK93:
Hi I'd love to see this on pypi! I'm building from source and working off an open PR on python-mip. Getting working wheels on pypi would make it a lot easier to use this excellent solver!

Issue #1354, opened 2023-07-17
LP reader Infinite Loop on malformed LP file


ecyrb:
A zero byte can cause the LP reader to enter an infinite loop. I acknowledge that this is user error, but it would be friendlier to the user to exit instead of entering an infinite loop.

Test case attached. Changing this assert in reader.cpp:

```cpp
    case '\0':  // empty line
      assert(this->linebufferpos == this->linebuffer.size());
      return false;
```

to an `lpassert` seems to fix the issue (i.e. cause an immediate exit).

[test.lp.txt](https://github.com/ERGO-Code/HiGHS/files/12073550/test.lp.txt)

As always, thank you for HiGHS.

feldmeier:
I'll take a look at this tomorrow, should be easy enough to fix. Thanks for reporting this issue.

jajhall:
I'm fixing it now. It looks as if the `assert(this->linebufferpos == this->linebuffer.size());` should have been `lpassert(this->linebufferpos == this->linebuffer.size());` all along

PR #1355, opened 2023-07-17
Fixed #1345 as suggested; formatted

### LINKED ISSUE #1345: Sporadic segfault when solving the same MIP
Hi!

I'm experiencing sporadic memory-related failures in a relatively simple MIP.
An example MPS file https://gist.github.com/grinya007/7219af99a70e9d5b7d3b8a3b677d4e9b
I built highs 1.5.3 and when I run it against this file it would fail with `corrupted size vs. prev_size` or `malloc(): invalid next size (unsorted)` in ~10% of runs. 90% of runs solve the MIP successfully. I'm not sure what more details are required but would be glad to answer any questions.

Thank you!

jajhall:
Change to `reader.cpp` as suggested so that `case '\0':  // empty line` is `lpassert`, not plain `assert`

### DIFF
--- a/check/TestAlienBasis.cpp
+++ b/check/TestAlienBasis.cpp
@@ -1,9 +1,10 @@
-#include "catch.hpp"
 #include <sstream>
-#include "util/HFactor.h"
-#include "util/HighsRandom.h"
+
 #include "HCheckConfig.h"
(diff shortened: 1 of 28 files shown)


jajhall:
This will close #1354

PR #1367, opened 2023-07-31
MAINT: More memory safety

HaoZeke:
This should make it easier to catch errors.

### DIFF
--- a/src/interfaces/highs_c_api.cpp
+++ b/src/interfaces/highs_c_api.cpp
@@ -1199,24 +1199,14 @@ void Highs_resetGlobalScheduler(HighsInt blocking) {
 }
 
 
-/**
- * Allocate memory of a given size. If allocation fails, print an error message
- * and terminate the program.
- *
- * This function tries to allocate memory of the given size using malloc. If the
- * allocation fails (i.e., malloc returns NULL), it prints an error message to
- * stderr indicating which variable it was trying to allocate memory for and
- * then calls exit to terminate the program.
- *
- * @param size      The number of bytes to allocate.
- * @param var_name  The name of the variable for which memory is being allocated.
- *                  This is used in the error message if the allocation fails.
- *                  This should be NULL terminated.
- *
- * @returns A pointer to the allocated memory. If the allocation fails, this
- *          function does not return.
- */
-void* Highs_mallocOrExit(size_t size, const char* var_name);
+void* Highs_mallocOrExit(unsigned long long size, const char* var_name) {
+    void* ptr = malloc(size);
+    if (ptr == NULL) {
+        fprintf(stderr, "Out of memory when allocating %s\n", var_name);
+        exit(EXIT_FAILURE);
+    }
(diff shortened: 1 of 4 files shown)


src/interfaces/highs_c_api.cpp
```diff
@@ -1198,6 +1198,16 @@ void Highs_resetGlobalScheduler(HighsInt blocking) {
   Highs::resetGlobalScheduler(blocking);
 }
 
+
+void* Highs_mallocOrExit(unsigned long long size, const char* var_name) {
+    void* ptr = malloc(size);
+    if (ptr == NULL) {
+        fprintf(stderr, "Out of memory when allocating %s\n", var_name);
+        exit(EXIT_FAILURE);
```
tobiass-sdl, line 1206: I'm no developer of HiGHS, but please do not add calls to exit() that might be reached when HiGHS is used as library - there is no way to handle the out of memory error in that case.

Example: HiGHS is used as library within a web server. With this code, the web server will simply stop if one request triggers an out of memory error...

HaoZeke: Wouldn't it crash if `malloc` failed anyway? This just makes it easier to debug instead of failing at a later stage due to a nullptr. Also although it's part of the C API, users can still make calls to `malloc`if they wish to handle memory differently.

jajhall: These `malloc` calls will only fail if there's insufficient memory available. Having HiGHS fail cleanly in such circumstances is on the ToDo list, but it is very unusual so we've not prioritised it.

As you see @HaoZeke (and also with reference to #1368), having HiGHS handle errors via exceptions and exit will meet with valid objections. 

There are also uses of HiGHS where output to `stderror` is not permitted. In this case, HiGHS will just exist without explanation if `Highs_mallocOrExit` fails.

src/util/HFactorDebug.cpp
```diff
@@ -96,20 +96,16 @@ void debugReportRankDeficiency(
...
+    const std::vector<HighsInt>& iwork, const HighsInt rank_deficiency,
+    const std::vector<HighsInt>& col_with_no_pivot,
+    const std::vector<HighsInt>& row_with_no_pivot) {
   if (highs_debug_level == kHighsDebugLevelNone) return;
   if (rank_deficiency > 10) return;
-  double* ASM;
-  ASM = (double*)malloc(sizeof(double) * rank_deficiency * rank_deficiency);
-  for (HighsInt i = 0; i < rank_deficiency; i++) {
-    for (HighsInt j = 0; j < rank_deficiency; j++) {
-      ASM[i + j * rank_deficiency] = 0;
```
jajhall, line 110: Fair point, I'll remove this `malloc`. I wrote it (in debug code) before I knew that a pointer could be created as `vector.data()`

jajhall:
Use of `exit()` and exception calling cannot be permitted in HiGHS

Issue #1365, opened 2023-07-30
Reject quadratic constraints in FilereaderLp


jajhall:
If a `.lp` file has quadratic constraints, HiGHS should return an error rather than ignoring them. @svigerske has provided a fix, but I want to have a look at the unit test before merging to `latest`

PR #1369, opened 2023-07-31
Fix 1365

jajhall:
Test file is now `qcqp.lp`, not `quadratic.lp`, and `Highs::readModel("qcqp.lp")` returns `kError`.


jajhall:
This will close #1365

Issue #1378, opened 2023-08-08
Missing required packages for highspy 


torressa:
When installing `highspy` from pip no required packages are installed with it (e.g. `numpy` is required as per [highspy/highs_bindings.cpp#L5](https://github.com/ERGO-Code/HiGHS/blob/9e519472fb3e630552178d8196b54e9b58dab27d/highspy/highs_bindings.cpp#L5)).
This was removed in https://github.com/ERGO-Code/HiGHS/commit/95b0a9cefa0bdee30ee5bd16818cfa7642fa9c5d.

jajhall:
OK I guess that @galabovaa did this for a good reason

jajhall:
Fixed by #1384

PR #1384, opened 2023-08-19
MAINT: Add numpy dep [highspy]

HaoZeke:
Closes https://github.com/ERGO-Code/HiGHS/issues/1378.

`numpy` is not a build dependency (because `pybind11` [doesn't need the package](https://pybind11.readthedocs.io/en/stable/advanced/pycpp/numpy.html), just the header), it is a runtime dependency, and it makes sense to add it.

Note that generally anyone using `highspy` probably has it at runtime anyway (and at a much higher version), but this makes it clearer.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,3 @@
-[tool.pdm]
-
 [project]
 name = "highspy"
 version = "1.5.3"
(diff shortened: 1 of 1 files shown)


HaoZeke:
cc @galabovaa

HaoZeke:
Documentation build failure is unrelated :)

PR #1387, opened 2023-08-23
add ndebug definition

lkapelevich:
I think this is enough to ensure NDEBUG is set... 

Separately, the indentations in CMakeLists are inconsistent. If it's OK with you I can autoindent everything.


galabovaa (approved):
Thank you! 

Yes, if you could auto indent that would be great. More cleaning up coming soon!

PR #1393, opened 2023-09-03
MAINT: Expose simplex constants [`highspy`]

HaoZeke:
These are needed by `scipy`'s interface `_linprog_highs`. As seen here:
- https://github.com/scipy/scipy/blob/main/scipy/optimize/_highs/cython/src/SimplexConst.pxd


HaoZeke:
Sorry, wrong branch.

PR #1394, opened 2023-09-03
ENH: Add access to simplex constants [`highspy`]

HaoZeke:
These are needed by `scipy`'s interface `_linprog_highs`. As seen here:
- https://github.com/scipy/scipy/blob/main/scipy/optimize/_highs/cython/src/SimplexConst.pxd


jajhall:
They haven't been asked for previously, and many of these are for development and/or redundant. Are you sure that they won't simply confuse people?

HaoZeke:
> They haven't been asked for previously, and many of these are for development and/or redundant. Are you sure that they won't simply confuse people?

The ones currently used in the `cython` interface (linked) are:
- `SimplexEdgeWeightStrategy`
- `SimplexStrategy`
- `SimplexAlgorithm`
- `SimplexCrashStrategy`
- `DualEdgeWeightMode`
- `PriceMode`

They were also manually exposed (not as enums) in the earlier candidate `scipy` [wrapper](https://github.com/scipy/scipy/pull/18642). I could minimize the exposed ones to just those. They are currently in a submodule, `highspy.simplex_constants` so I'm reasonably certain they won't be used by mistake / would confuse people.

highspy/highs_bindings.cpp
```diff
@@ -929,4 +929,198 @@ PYBIND11_MODULE(highspy, m) {
...
+      .value("kRebuildReasonPrimalInfeasibleInPrimalSimplex",
+             RebuildReason::kRebuildReasonPrimalInfeasibleInPrimalSimplex)
+      .value("kRebuildReasonChooseColumnFail",
+             RebuildReason::kRebuildReasonChooseColumnFail)
+      .value("kRebuildReasonForceRefactor",
+             RebuildReason::kRebuildReasonForceRefactor)
+      .value("kRebuildReasonExcessivePrimalValue",
+             RebuildReason::kRebuildReasonExcessivePrimalValue)
+      .value("kRebuildReasonCount", RebuildReason::kRebuildReasonCount)
+      .export_values();
```
HaoZeke: I suppose these could be dropped from the Python API.

HaoZeke:
Essentially, everything in the `cython` folder [in Scipy](https://github.com/scipy/scipy/tree/main/scipy/optimize/_highs/cython/src) needs to be exposed either via `highspy` (my preferred approach) or recreated from the library again within `scipy` itself (not a good idea).

highspy/highs_bindings.cpp
jajhall: Yes, these are utterly irrelevant to users

jajhall:
> The ones currently used in the `cython` [interface in Scipy](https://github.com/scipy/scipy/blob/main/scipy/optimize/_highs/cython/src/SimplexConst.pxd) (linked) are:
> 
> * `SimplexEdgeWeightStrategy`

OK

> * `SimplexStrategy`

OK

> * `SimplexAlgorithm`

As observed: these are utterly irrelevant to users and must not appear in highspy

> * `SimplexCrashStrategy`

This is redundant: there is no such strategy, so the associated constants must not appear in highspy

> * `DualEdgeWeightMode`

This string does not appear within HiGHS

> * `PriceMode`

This string does not appear within HiGHS

> 
> These are definitely required for Scipy, and were manually exposed (not as enums) in the earlier candidate `highspy-scipy` [wrapper](https://github.com/scipy/scipy/pull/18642). They are currently in a submodule, `highspy.simplex_constants` so I'm reasonably certain they won't be used by mistake / would confuse people. Perhaps some of them can be removed from the Python API, but the rest are options which can be set from within `scipy` and should be provided.

No, the answer is that these options should not be set from within `scipy` and the possibility to set them must be removed

> Since these are being exposed on the Python side as `enum` instances, it is fine to have the repeated values IMO, since users would use the strongly typed instance, not the integral representation. If (and when) the duplicates are removed from `highs` they should be removed from the wrapper, otherwise it would be more confusing for users to be expected to know which options are aliases.

Where are the "duplicate"  integral representations?

jajhall:
I'll got through the suggestions that you make, but most will be removed. Would you then work back and see that the options corresponding to what I've removed are not set in `scipy`?

HaoZeke:
> I'll go through the suggestions that you make, but most will be removed because they are redundant or relate to development-only options. Please could you then work back and see that the options corresponding to what I've removed are not set in `scipy`?

Yup, I'll make sure they're not present in the `highspy-scipy` wrapper, thanks a ton :)

jajhall:
> Most things are already exposed in `highspy`, [as I finish porting](https://github.com/HaoZeke/scipy/tree/toHighspy/scipy/optimize/_highs) the `linprog` interface to `highspy` there might be a few more PRs like this to round out the API needed for SciPy.

Why should the `linprog` interface be ported to `highspy`? Surely the  `linprog` interface is part of `scipy`

HaoZeke:
> > These are definitely required for Scipy, and were manually exposed (not as enums) in the earlier candidate `highspy-scipy` [wrapper](https://github.com/scipy/scipy/pull/18642). They are currently in a submodule, `highspy.simplex_constants` so I'm reasonably certain they won't be used by mistake / would confuse people. Perhaps some of them can be removed from the Python API, but the rest are options which can be set from within `scipy` and should be provided.
> 
> No, the answer is that these options should not be set from within `scipy` and the possibility to set them must be removed

Makes sense, corrected.

> > Since these are being exposed on the Python side as `enum` instances, it is fine to have the repeated values IMO, since users would use the strongly typed instance, not the integral representation. If (and when) the duplicates are removed from `highs` they should be removed from the wrapper, otherwise it would be more confusing for users to be expected to know which options are aliases.
> 
> Where are the "duplicate" integral representations?

For example:

```cpp
enum SimplexEdgeWeightStrategy {
  kSimplexEdgeWeightStrategyMin = -1,
  kSimplexEdgeWeightStrategyChoose = kSimplexEdgeWeightStrategyMin,
  kSimplexEdgeWeightStrategyDantzig,
  kSimplexEdgeWeightStrategyDevex,
  kSimplexEdgeWeightStrategySteepestEdge,
  kSimplexEdgeWeightStrategyMax = kSimplexEdgeWeightStrategySteepestEdge
};
```

When converted to integer form, kSimplexEdgeWeightStrategyMax and kSimplexEdgeWeightStrategySteepestEdge [will both be 2](https://onlinegdb.com/IQ9T_upQc).

> > Most things are already exposed in highspy, [as I finish porting](https://github.com/HaoZeke/scipy/tree/toHighspy/scipy/optimize/_highs) the linprog interface to highspy there might be a few more PRs like this to round out the API needed for SciPy.

> Why should the linprog interface be ported to highspy? Surely the linprog interface is part of scipy

Sorry I was unclear, I meant the API required by `_linprog_highs` (which is the wrapper used by SciPy's `linprog` when working with `highs`. That is, if `_linprog_highs` requires (as an example) a function or value not currently exposed in `highspy` it would be good to have it exposed in `highspy`.

It is of course not desirable nor planned to have a `linprog` like interface to be part of `highspy` directly.

jajhall:
> Sorry I was unclear, I meant the API required by `_linprog_highs` (which is the wrapper used by SciPy's `linprog` when working with `highs`. That is, if `_linprog_highs` requires (as an example) a function or value not currently exposed in `highspy` it would be good to have it exposed in `highspy`.
> 
> It is of course not desirable nor planned to have a `linprog` like interface to be part of `highspy` directly.

Ah, now  I parse your original comment correctly...

jajhall:
> For example:
> 
> ```c++
> enum SimplexEdgeWeightStrategy {
>   kSimplexEdgeWeightStrategyMin = -1,
>   kSimplexEdgeWeightStrategyChoose = kSimplexEdgeWeightStrategyMin,
>   kSimplexEdgeWeightStrategyDantzig,
>   kSimplexEdgeWeightStrategyDevex,
>   kSimplexEdgeWeightStrategySteepestEdge,
>   kSimplexEdgeWeightStrategyMax = kSimplexEdgeWeightStrategySteepestEdge
> };
> ```
> 
> When converted to integer form, kSimplexEdgeWeightStrategyMax and kSimplexEdgeWeightStrategySteepestEdge [will both be 2](https://onlinegdb.com/IQ9T_upQc). So if they were exposed to the Python API only as integers, as was being done in `scipy` (instead of `enum` objects) there could be quite a bit of confusion.

This duplication is quite deliberate, and for internal code clarity and robustness. I want to check that the value passed for an option is within the valid range, so check on the `min` and `max` values. So long as the enum is updated correctly, the use of  `min` and `max` values remains correct. 

I suggest that the `min` and `max` values are not exposed via `scipy`

PR #1420, opened 2023-09-19
Add getMipDualBound() to the C API

adrien-mogenet:
The main motivation was to allow wrappers such as HiGHS.jl, the Julia wrapper, to access the Dual Bound value. This is useful when the solver hits a time limit, and contains no solution available from the "objective value" or "objective bound" field, but we still want to retrieve the latest dual bound computed in the last iteration.


jajhall:
This isn't necessary: use 

`Highs_getDoubleInfoValue(highs, "mip_dual_bound", mip_dual_bound)`

PR #1435, opened 2023-09-28
Fixed error in Highs::callRunPostsolve

### LINKED ISSUE #1434: Error in TestPresolve.cpp: presolve-solve-postsolve-mip
There is an error in HiGHS, popping up in the second test of TestPresolve.cpp: presolve-solve-postsolve-mip 

The issue is at Highs::callRunPostsolve on line 3499: iCol has value 15 but solution.col_value only has 14 entries. So there is a mismatch of dimensions for the columns.

jajhall:
This will close #1434


jajhall:
All checks pass! <3

PR #1405, opened 2023-09-10
MAINT: `pybind11` and the modeling API

HaoZeke:
Adds back the modeling API for the documentation examples.

PR #1437, opened 2023-09-30
MAINT: Use HighsInt consistently

HaoZeke:
I decided to split https://github.com/ERGO-Code/HiGHS/pull/1405 , since the bugs addressed here are independent of the modeling API work there. This essentially handles:
- Using `HighsInt` in the Pybind11 bindings
- Fixes a rather odd bug with `HIGHSINT64` and `printf` ([b97083](https://github.com/ERGO-Code/HiGHS/commit/b97083c6acc4c177b95389825e812d31679c3f52))

The CI failure for the Python bindings has to do with the modeling API and is being worked on at #1405.

### DIFF
--- a/src/util/HighsTimer.h
+++ b/src/util/HighsTimer.h
@@ -19,10 +19,10 @@
 #include <chrono>
 #include <cstdio>
 #include <cstdlib>
-#include <string>
-#include <vector>
 #include <iomanip>
 #include <iostream>
+#include <string>
+#include <vector>
 
 #include "util/HighsInt.h"
 
@@ -277,10 +277,11 @@ class HighsTimer {
             double percent_ideal = 100.0 * time / ideal_sum_time;
             printf("; %5.1f%%", percent_ideal);
           }
-          std::cout << "; " << std::fixed << std::setprecision(1) << std::setw(5)
-                    << percent_sum_clock_times[i] << "%):";
+          std::cout << "; " << std::fixed << std::setprecision(1)
+                    << std::setw(5) << percent_sum_clock_times[i] << "%):";
           std::cout << std::setw(9) << clock_num_call[iClock];
-          std::cout << " " << std::scientific << std::setprecision(4) << time_per_call << "\n";
+          std::cout << " " << std::scientific << std::setprecision(4)
+                    << time_per_call << "\n";
         }
       }
       sum_time += time;
--- a/src/util/HighsTimer.h
+++ b/src/util/HighsTimer.h
@@ -19,8 +19,6 @@
 #include <chrono>
 #include <cstdio>
 #include <cstdlib>
-#include <iomanip>
-#include <iostream>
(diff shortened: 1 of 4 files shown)


src/util/HighsTimer.h
jajhall, line 1: I see that this is the only instance of `HIGHSINT_FORMAT` in this file but, rather than having one instance of logging done with `std::cout` (and two more include files), I'd rather that `clock_num_call[iClock]` were cast to int.

jajhall (changes requested):
Thanks for introducing `HighsInt` in all these places. I was aware of using `int` carelessly, and your changes are fine with me

I have to trust you on the `meson.build` changes, as I understand nothing of meson, but @galabovaa should approve them

There's a minor request relating to `HighsTimer.h`. For my own education, what was wrong with using `HIGHSINT_FORMAT`? [Note that `HIGHSINT_FORMAT` was inserted wholesale when we introduced `HighsInt` years ago, and values that can't be large are gradually being cast to `int` in the `printf` command so that the formatting is more transparent. Although `std::cout` is C++, and there's more scope for formatting, I find it horribly clunky to use, so wish to stick to `printf` as much as possible.

HaoZeke:
> Thanks for introducing `HighsInt` in all these places. I was aware of using `int` carelessly, and your changes are fine with me

Awesome, thanks :)

> I have to trust you on the `meson.build` changes, as I understand nothing of meson, but @galabovaa should approve them

👍🏾


> There's a minor request relating to `HighsTimer.h`. For my own education, what was wrong with using `HIGHSINT_FORMAT`? [Note that `HIGHSINT_FORMAT` was inserted wholesale when we introduced `HighsInt` years ago, and values that can't be large are gradually being cast to `int` in the `printf` command so that the formatting is more transparent. Although `std::cout` is C++, and there's more scope for formatting, I find it horribly clunky to use, so wish to stick to `printf` as much as possible.

I was running into a compiler error:
```bash
src/util/HighsTimer.h:278:34: error: expected ')' before 'PRId64'
  278 |           printf("; %5.1f%%):%9" PRId64 " %11.4e\n",
      |                  ^~~~~~~~~~~~~~~
# Or when simplified
User
../src/util/HighsTimer.h:279:23: error: expected ')' before 'PRId64'
  279 |           printf("%9" PRId64, clock_num_call[iClock]);
```

It was pretty difficult to reliably reproduce though, and as noted `printf` can be nicer to use, so I've taken your suggestion and `static_cast` `clock_num_call` instead.

In general, at some point it would be nice to look into C++20's `<format>` or the library which inspired it, [fmt](https://fmt.dev) for `printf` style printing with nicer (automatic) type deduction, but I didn't want to introduce a whole library to fix such a hard to find bug anyway (though its also faster apparently).

jajhall (approved):

jajhall (approved):

Issue #1277, opened 2023-04-28
Overhead using highspy


jajhall:
I'm just calling your attention to a potential pitfall for users of highspy, which is that accessing `solution.col_value[icol]` is a quite slow operation, I guess because of some overhead in pybind11. Doing over 100k independent accesses to retrieve a solution takes a few minutes, but can be reduced to milliseconds by first converting to a Python list and accessing that instead.

I made the relevant change in PuLP recently since whoever wrote the interface had made that mistake which was obviously a massive and avoidable overhead.
https://github.com/coin-or/pulp/pull/641

Perhaps it should remain the responsibility of consumers of highspy to be aware of and manage, but you could consider implementing something like a `solution.col_values()` function that returns a python list of values. Wouldn't even need to be cached as that introduces a risk of invalidation. I'd be happy to implement it and open a PR, but I wasn't sure what your policy would be on unsolicited feature PRs(?). In general I hope to be a contributor to HiGHS in the near future 🙂.

matt035343:
I would love to use highspy in favor of the open source solver we use now, but `solution.col_value[icol]` is way too slow for me. I found that the access time of `solution.col_value[icol]` scales with the number of variables in the program. I have in the order of `1^10` variables in my program which I can build and solve reasonably fast in highspy, but accessing `solution.col_value[icol]` takes around 0.16s (per variable!). It would take many days for me to retrieve all variable values. I hope this issue will get some attention!

@jajhall Thanks for the tip regarding the Python list. I will try that out if it can temporarily solve my performance issues.

jajhall:
This is very odd. Addressing this is on our ToDo list, and we've mentioned it in discussions, but that's as far as it's got. I
Your comment raises the priority!

aphi:
Yes that comment was from an email I sent and referring to the change we made in PuLP to avoid this overhead.

matt035343:
@aphi It did work, thanks. I now use it as a temporary workaround. For anyone experiencing this issue, you can use following snippet:
```
values = list(solution.col_value)
values[icol]
```

However, I do still believe it should be fixed permanently somewhere upstream

jajhall:
Anyone have an idea how to fix it upstream? 

I'd have thought it was a known pybind11 issue

jajhall:
I've submitted a query to pybind11

jajhall:
Only solution at the moment is a warning in the documentation

PR #1438, opened 2023-09-30
BUG: Fix `highspy` `__getitem__` speed

HaoZeke:
Closes https://github.com/ERGO-Code/HiGHS/issues/1277. For this test case (adapted from [here](https://gist.github.com/aphi/4b2f7cfc7ccc52465bead3c24ee9c3f4)):

```python
import highspy
import timeit

h = highspy.Highs()

h.readModel("80bau3b.mps")
h.setOptionValue("time_limit", 10)

# Solve
start_time = timeit.default_timer()
h.run()
elapsed = timeit.default_timer() - start_time

print(f"Time taken to solve: {elapsed:.4f}")

solution = h.getSolution()
num_vars = h.getNumCol()

# Extract values using __getitem__ N times
start_time = timeit.default_timer()
values = [solution.col_value[icol] for icol in range(num_vars)]
elapsed = timeit.default_timer() - start_time
print(f"__getitem__ method for {num_vars} vars: {elapsed:.4f}")


# Extract values by copying first to list
start_time = timeit.default_timer()
col_values = list(solution.col_value)
values = [col_values[icol] for icol in range(num_vars)]
elapsed = timeit.default_timer() - start_time
print(f"Copy-first method for {num_vars} vars: {elapsed:.4f}")
```

Where the `80bau3b.mps` is part of this repo, we have:

```bash
# Original
Time taken to solve: 0.1016
__getitem__ method for 9799 vars: 0.7423
Copy-first method for 9799 vars: 0.0003
```

As a first approximation, we can use the buff protocol and `numpy`:

```bash

# Numpy readonly (33b6634823b41252f28fdd9daa5b553c6b9b43cd)
Time taken to solve: 0.0974
__getitem__ method for 9799 vars: 0.0956
Copy-first method for 9799 vars: 0.0010
```

However, this isn't equivalent to `readwrite` and isn't very nice conceptually, so we can try to use an opaque type:

```bash
# Opaque type (standard) for std::vector<double> (2aa08d9c8471a86aafa544d7976de1e520c7357f)
Time taken to solve: 0.1007
__getitem__ method for 9799 vars: 0.0064
Copy-first method for 9799 vars: 0.0015
```

This can be optimized a bit more:
```bash
# Custom opaque type and class
Time taken to solve: 0.0919
__getitem__ method for 9799 vars: 0.0046
Copy-first method for 9799 vars: 0.0015
```

Note that this method has the caveats listed in [the documentation](https://pybind11.readthedocs.io/en/stable/advanced/cast/stl.html#making-opaque-types). I think the standard opaque type is probably about as far as can be optimized safely, though there seem to be some micro-optimizations [discussed on Gitter](https://matrix.to/#/!vYqYcTzqekPEfQIxEf:gitter.im/$DaLVlGr0PT27yS8vp0LGt5ap7QfqYqCM5YjKW-0seLc?via=gitter.im&via=matrix.org&via=chaos.jetzt) (which I admit are beyond me).

### DIFF
--- a/highspy/highs_bindings.cpp
+++ b/highspy/highs_bindings.cpp
@@ -543,26 +543,7 @@ std::tuple<HighsStatus, int> highs_getRowByName(Highs* h,
 }
 
 PYBIND11_MODULE(highspy, m) {
-  py::class_<std::vector<double>>(m, "VectorDouble")
-      .def(py::init<>())
-      .def("clear", &std::vector<double>::clear)
-      .def("pop_back", &std::vector<double>::pop_back)
-      .def("__len__", [](const std::vector<double>& v) { return v.size(); })
-      .def(
-          "__iter__",
-          [](std::vector<double>& v) {
-            return py::make_iterator(v.begin(), v.end());
-          },
-          py::keep_alive<0, 1>())
-      .def("__getitem__",
-           [](const std::vector<double>& v, size_t i) {
-             if (i >= v.size()) throw py::index_error();
-             return v.at(i);
-           })
-      .def("__setitem__", [](std::vector<double>& v, size_t i, double value) {
-        if (i >= v.size()) throw py::index_error();
-        v.at(i) = value;
-      });
+  py::bind_vector<std::vector<double>>(m, "VectorDouble");
   // enum classes
   py::enum_<ObjSense>(m, "ObjSense")
       .value("kMinimize", ObjSense::kMinimize)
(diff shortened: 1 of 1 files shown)


HaoZeke:
One of the biggest issues with this is that now users need to be careful to cast:

```python
# Before:
lp.col_cost_ = c

# After:
vector_double_c = highspy._highs.VectorDouble(c.tolist())
lp.col_cost_ = vector_double_c
```

This can be sidestepped with the NumPy approach (or just documented). Alternatively, since for the modeling API we have the Python wrapper over the C++ internals anyway, we can just extract it into a list before operating on it (as is done in the test case).

HaoZeke:
I don't like any of these here actually. They're a mess for downstream library integration (e.g. SciPy). The solution should be to rework the `highs` python module instead to extract the list and store it.

PR #1444, opened 2023-10-05
Restrict meson-python to 0.13.2 temporarily

metab0t:
meson-python 0.14.0 has a bug that handles rpath of python binding incorrectly. Use 0.13.2 temporarily as a fix.
The upstream fix has landed in https://github.com/mesonbuild/meson-python/pull/510

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,7 @@ license = {text = "MIT"}
 "Bug Tracker" = "https://github.com/ERGO-Code/HiGHS/issues"
 
 [build-system]
-requires = ["meson-python<0.14.0"]
+requires = ["meson-python<0.14.0", "meson>=1.2.0"]
 build-backend = "mesonpy"
 
 [tool.meson-python.args]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,7 @@ license = {text = "MIT"}
 "Bug Tracker" = "https://github.com/ERGO-Code/HiGHS/issues"
 
 [build-system]
-requires = ["meson-python"]
+requires = ["meson-python<0.14.0"]
 build-backend = "mesonpy"
 
 [tool.meson-python.args]
--- a/meson.build
+++ b/meson.build
@@ -1,5 +1,6 @@
 project('highs', 'cpp',
   version : '1.6.0',
+  meson_version: '>= 1.2.0',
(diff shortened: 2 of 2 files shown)


HaoZeke (changes requested):
As per https://github.com/scipy/scipy/pull/19255#issuecomment-1742585119 we will need `meson-python` to be `>=1.2.0` so might as well bump it upwards instead.

pyproject.toml
```diff
@@ -19,7 +19,7 @@ license = {text = "MIT"}
 "Bug Tracker" = "https://github.com/ERGO-Code/HiGHS/issues"
 
 [build-system]
-requires = ["meson-python"]
+requires = ["meson-python<0.14.0"]
```
HaoZeke, line 22:
```suggestion
requires = ["meson-python>=1.2.0"]
```

HaoZeke: Also make the corresponding change to `meson.build`:

```meson
project('highs', 'cpp',
....
  meson_version: '>= 1.2.0',
....)
```

metab0t:
> As per [scipy/scipy#19255 (comment)](https://github.com/scipy/scipy/pull/19255#issuecomment-1742585119) we will also need `meson` to be `>=1.2.0` so might as well bump it in this PR too perhaps.

Applied

pyproject.toml
```diff
@@ -19,7 +19,7 @@ license = {text = "MIT"}
 "Bug Tracker" = "https://github.com/ERGO-Code/HiGHS/issues"
 
 [build-system]
-requires = ["meson-python"]
+requires = ["meson-python<0.14.0", "meson>=1.2.0"]
```
HaoZeke, line 22: Minor nit, the `meson.build` file should also define `meson_version`:
```meson
project('highs', 'cpp',
  version : '1.6.0',
  meson_version: '>= 1.2.0',
...
```

HaoZeke (commented):
Almost there :)

pyproject.toml
metab0t: Done

HaoZeke (approved):
LGTM :)

metab0t:
Wish the 1.6.0 wheel can be released after this PR.

galabovaa (approved):

Issue #1372, opened 2023-07-31
HiGHS crash while adding Clique 


Dexterux:
Hi everyone,

I'm testing HiGHS and I encountered a crash while solving a problem on Windows with HiGHS v1.5.3, v1.5.1 and the code from branch master.

The output of the solver is the following:
```
Running HiGHS 1.5.3 [date: 2023-06-28, git hash: a29e36f11-dirty]
Copyright (c) 2023 HiGHS under MIT licence terms
Rows:        6713 lower bounds exceeding       -1e+20 are treated as -Infinity
Rows:         217 upper bounds exceeding        1e+20 are treated as +Infinity

Presolve is switched off
Objective function is integral with scale 1

Solving MIP model with:
   18772 rows
   47675 cols (47418 binary, 257 integer, 0 implied int., 0 continuous)
   2348243 nonzeros

        Nodes      |    B&B Tree     |            Objective Bounds              |  Dynamic Constraints |       Work      
     Proc. InQueue |  Leaves   Expl. | BestBound       BestSol              Gap |   Cuts   InLp Confl. | LpIters     Time

         0       0         0   0.00%   0               inf                  inf        0      0      0         0     1.0s
         0       0         0   0.00%   896.8010189     inf                  inf        0      0      2      1241     9.6s
         0       0         0   0.00%   996.3344742     inf                  inf     1037     56      6      1429    17.1s
         0       0         0   0.00%   1014.008907     inf                  inf     1762     97      8      1651    24.6s
         0       0         0   0.00%   1028.754132     inf                  inf     2304    128     10      1833    29.8s
 C       0       0         0   0.00%   1076.135025     16640             93.53%     2545    170     24      2041    62.8s
         0       0         0   0.00%   1130.986398     16640             93.20%     2919    292     24      2792    69.0s
         0       0         0   0.00%   1218.879657     16640             92.68%     3390    344     24      4407    74.9s
         0       0         0   0.00%   1232.826308     16640             92.59%     3984    395     24      5106    80.9s
         0       0         0   0.00%   1237.830329     16640             92.56%     4490    358     24      5626    86.1s
         0       0         0   0.00%   1247.74377      16640             92.50%     4542    326     24      6522    92.1s
 L       0       0         0   0.00%   1248.171066     1507              17.18%     3753    344     24      6714   171.4s
 L       0       0         0   0.00%   1259.645304     1505              16.30%     3664    302     24      7883   190.8s

[...]

PR #1452, opened 2023-10-10
Minor changes to HighsHashTree.h motivated by #1372

fwesselm:
1. Properly initialize an array.
2. Handle unexpected result from "entries_to_size_class".

### DIFF
--- a/src/util/HighsHashTree.h
+++ b/src/util/HighsHashTree.h
@@ -655,6 +655,9 @@ class HighsHashTree {
               for (int i = 0; i <= newNumChild; ++i)
                 mergeIntoLeaf(newLeafSize4, hashPos, branch->child[i]);
             }
+            default:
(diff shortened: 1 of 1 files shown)


fwesselm:
These changes do not fix #1372 - these are just two improvements found while debugging.

jajhall (approved):
Omitting `* sizeof(uint8_t)` does seem like an error. Shame it doesn't fix #1372 

Assessing the difference between
> `uint8_t sizes[InnerLeaf<4>::capacity() + 1] = {};` 

and
> ` uint8_t sizes[InnerLeaf<4>::capacity() + 1]; memset(sizes, 0, branchSize * sizeof(uint8_t));`

is beyond my C++, but I trust you.

I'll merge it to a local branch and check MIPs for which it is executed

PR #1454, opened 2023-10-10
Check 1452

jajhall:
So, @fwesselm, an assert after 

> // there are many collisions, determine the exact sizes first

is triggered by `check/TestHighsHash.cpp`, but not in any of the MIP instance tests

### DIFF
--- a/src/util/HighsHashTree.h
+++ b/src/util/HighsHashTree.h
@@ -843,6 +843,8 @@ class HighsHashTree {
                 hash, hashPos + 1, entry);
           } else {
             // there are many collisions, determine the exact sizes first
+	    assert(999==888);
+	    printf("removeChildFromBranchNode has hit @fwesselm's mod\n");
             uint8_t sizes[InnerLeaf<4>::capacity() + 1] = {};
             sizes[occupation.num_set_until(hashChunk) - 1] += 1;
             for (int i = 0; i < leaf->size; ++i) {


fwesselm:
Interesting, I cannot reproduce this on Windows. Could you tell me which assert is triggered? Thank you.

jajhall:
The assert that I added is not yet pushed to GitHub - and it probably won't. I've run the easiest 100 miplib problems (in release, with a print statement) to see how often it's used in reality. 

It's not that I don't trust what you've done: I'm just curious to see why an apparent bug hasn't caused more trouble, and what that may tell us about the error in the user's issue

src/util/HighsHashTree.h
```diff
@@ -655,6 +655,9 @@ class HighsHashTree {
               for (int i = 0; i <= newNumChild; ++i)
                 mergeIntoLeaf(newLeafSize4, hashPos, branch->child[i]);
             }
+            default:
+              throw std::logic_error(
+                  "Unexpected result from entries_to_size_class");
```
HaoZeke, line 660: Note the discussion on exceptions here: https://github.com/ERGO-Code/HiGHS/discussions/1368

The basic takeaway is that since they require catching and handling before being passed to the C API they might not be of much use without a more principled effort.

HaoZeke: At the very least it should be caught and rethrown to ensure correct propagation.

fwesselm:
> The assert that I added is not yet pushed to GitHub - and it probably won't. I've run the easiest 100 miplib problems (in release, with a print statement) to see how often it's used in reality.
> 
> It's not that I don't trust what you've done: I'm just curious to see why an apparent bug hasn't caused more trouble, and what that may tell us about the error in the user's issue

FYI, the new line (correctly zero out array) is hit on problem `fast0507` on Windows for example.

fwesselm:
I created #1470 to fix #1372. Therefore, this PR could be closed, I think.

PR #1468, opened 2023-10-22
Fix compiler warnings (part 2)

fwesselm:
- Second round of fixes for compiler warnings (includes warnings issued by compilers in github actions).
- Again, testing showed that these changes do not affect HiGHS' behavior (as expected).


jajhall:
Thanks!

PR #1473, opened 2023-10-28
Issue 1472 - Create branch-and-price.py with a complete example for the python interface

guifcoelho:
- Solves a bin packing problem
- Adds the compact model
- Adds a complete branch-and-price algorithm using the python interface


jajhall:
I've not tried to run this, but it looks very impressive!

PR #1478, opened 2023-10-30
Reduce amount of repetitive code in Highs::run()

fwesselm:
Added lambda (for solving LPs) in order to reduce amount of similar code


jajhall (approved):
Thanks. Not being one to write lambdas, I used to have an intermediate method that did the same job as you've done with your lambda. 

I'm sure that there are enough tests to rule out the very small probability that you've changed behaviour

PR #1480, opened 2023-11-02
Add optional flag to Highs::postsolve to avoid reoptimization

fwesselm:
Add optional flag to Highs::postsolve and related methods to avoid reoptimization in postsolve if it is not necessary.


jajhall:
> Thank you. I did not know about the style guide, sorry.

No problem - we never mentioned it!

Issue #1484, opened 2023-11-03
Improve logging when dual ratio test fails


NPC:
Hello,

I've encountered this issue that I can't find an explanation for, nor can I identify the cause by looking at the log, can you please help?

Here's the LP MPS file, running on Windows 11 from command line (same result called from the DLL via C# interface): [LP-problem-2023-11-03.zip](https://github.com/ERGO-Code/HiGHS/files/13254491/LP-problem-2023-11-03.zip)

Run via `highs.exe --parallel on LP-problem.mps`, output:  
```
Running HiGHS 1.6.0: Copyright (c) 2023 HiGHS under MIT licence terms
LP   LP-problem has 18247 rows; 57099 cols; 107808 nonzeros
Presolving model
9111 rows, 30394 cols, 53144 nonzeros
8939 rows, 29003 cols, 50965 nonzeros
Presolve : Reductions: rows 8939(-9308); columns 29003(-28096); elements 50965(-56843)
Solving the presolved LP
Using EKK parallel dual simplex solver - PAMI with concurrency of 8
  Iteration        Objective     Infeasibilities num(sum)
          0    -6.3260410194e+02 Ph1: 1(1); Du: 1(632.604) 0s
Model   status      : Not Set
HiGHS run time      :          0.14
```

I tried parallel off, presolve off, versions 1.5.3 and 1.4.0, no difference.

As I understand, “Not Set” means that the algorithm ended unexpectedly, before it could report on any errors or infeasibilities, is this correct? Could the values `Ph1: 1(1); Du: 1(632.604)` in “Infeasibilities” column provide some valuable insight?

Simpler problems continue to solve without issues, as well as complex but different problems, these's something off about this one.

I'll be grateful for any help.

jajhall:
How odd. I've reproduced this behaviour with the latest (dev) version of HiGHS

jajhall:
Your problem is badly scaled. You have very large costs 

> 5055 values satisfy 10^( 11) <= v < 10^( 12)

upper bounds on variables

> 12 values satisfy 10^( 10) <= v < 10^( 11)

and RHS values for equality constraints

> 4 values satisfy 10^( 10) <= v < 10^( 11)

The HiGHS dual simplex solver fails due to the excessive costs. I see that you're running HiGHS from the command line, so the logging output that you quote is all that you're told. If calling HiGHS from an application, you'd get an error code returned from Highs::run()

[...]

PR #1485, opened 2023-11-03
Fix 1484

jajhall:
Scaling dev logging now driven by `log_dev_level>0` rather than `highs_analysis_level>0`

Added logging when dual simplex fails due to excessive primal values or ratio test failure


jajhall:
This will close #1484

PR #1488, opened 2023-11-07
Updated branch and price example

mathgeekcoder:
Updated branch and price python example. Added greedy heuristic for tighter initialization, additional comments, and fixed some bugs.

Bugs include:
* Branching code stopped at first integer solution, rather than optimal solution
* Exception raised when optimal solution was found at root node
* Possible wrong "optimal" solution found due to incorrect stopping criteria in column generation

### DIFF
--- a/examples/branch-and-price.py
+++ b/examples/branch-and-price.py
@@ -1,52 +1,76 @@
+# An example branch-and-price algorithm for the bin packing problem
+#
+from collections import defaultdict
+from operator import itemgetter
 import highspy, random, time, math
 
+# Relative gap for column generation
 CG_GAP_REL = 1e-4
 
+# Random seeds for reproducibility
 random.seed(100)
+SEED = 100
 
-#Total number of items
-N = 60
+# Instance parameters
+NumberItems = 40
+ItemWeights = [round(random.uniform(1, 10), 1) for _ in range(NumberItems)]
+BinCapacity = 15
 
-#Item weight
-weights = [round(random.uniform(1, 10),4) for _ in range(N)]
 
-# Bin capacity
-capacity = 15
+#
+# Solve instance with greedy first fit decreasing (FFD) heuristic
+#
+def solveGreedyModel():
+    bins = defaultdict(float)
+    solution = defaultdict(list)
 
-# Maximum number of bins
-B = N # Easy upper bound on the number of bins
+    for item, w in sorted(enumerate(ItemWeights), reverse=True, key=itemgetter(1)):
+        index = next((i for i, W in bins.items() if W + w <= BinCapacity), len(bins))
+        bins[index] += w
+        solution[index].append(item)
+    
+    return list(solution.values())
 
+#
+# The compact bin-packing model (for comparison with branch-and-price)
+#
+# min sum_{j} y_{j}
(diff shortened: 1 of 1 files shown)


guifcoelho:
Hello everyone! Thanks for letting me know about these modifications. I ran through the code and I'm very confortable with everything you did.

About the bugs you caught, thanks for listing them. I was about to push a new PR to fix when the optimal solution was found at root, but I didn't know about the other ones.

On the branch-and-price part there is this note:

> Note: Branch selection has been chosen for simple implementation and it's ability to avoid symmetry. 
> This comes at the cost of an unbalanced search tree (other approaches may solve the problem with fewer branches), 
> and it is less likely to be generalizable to other problems.

Would you mind directing me anywhere to improve on that?

Again, let me know if I can help with anything.

jajhall:
As I understand, to change this behaviour would require a significant modification and complication of the code. This would detract from the code being a simple illustration of using HiGHS. You never pretended that it was a fully optimized B&C implementation 

What I feel is sufficient is to flag up the issue in a comment just to indicate we're aware of it.

guifcoelho:
Ow yes. I don't mean to change it anymore. In my opinion, you are good to go. I meant it as to improve my knowledge on the subject. I'm happy with the example as it is.

jajhall:
Having merged into `master` I was about to create an image with a tiny example for a LinkedIn post, and set 

> NumberItems = 10
> BinCapacity = 8

This creates an infeasible problem with

> ItemWeights = [2.3, 5.1, 7.9, 7.3, 7.6, 4.9, 8.2, 5.8, 1.7, 5.1]

The compact formulation MIP is infeasible, but Greedy and ColGen both claim 8-bin solutions

I tried looking at the greedy code

>     for item, w in sorted(enumerate(ItemWeights), reverse=True, key=itemgetter(1)):
        index = next((i for i, W in bins.items() if W + w <= BinCapacity), len(bins))
        bins[index] += w
        solution[index].append(item)
    
To me, `W` should be the size of the current item(s) in the bin under consideration, so that `W + w <= BinCapacity` makes sense but it would appear not to be.

jajhall:
So, out of curiosity, what is `W + w <= BinCapacity` doing? Surely the greedy algorithm has to assess whether there is enough space to pack an item when there is already an item in the bin.

mathgeekcoder:
Absolutely, the W+w <= BinCapacity checks to see if the item can fit on any **existing** bin, i.e., packed with other items. 

If we can't find a bin to pack it on, it is placed in a new bin by itself.  But the code doesn't check to see if it can actually fit in a bin by itself.

jajhall:
Got you! Thanks

mathgeekcoder:
Specifically, the trick is in the "next" function call. If it can't find any bin to pack the item in, the 2nd parameter suggests to create a new bin (by using a larger bin index).

Since we're using defaultdict, everything is initialized when first accessed, so that's all correct.

jajhall:
Thanks for the Python _and_ B&C tutorial!

mathgeekcoder:
FYI:  I've added a new PR with a feasibility check. https://github.com/ERGO-Code/HiGHS/pull/1492

PR #1492, opened 2023-11-07
Added instance feasibility check for branch-and-price python example

mathgeekcoder:


jajhall:
Thanks. Since I'd already put this example into master, I've incorporated your change into a branch off master that I'll now push

PR #1493, opened 2023-11-07
Incorporated immediate infeasibility test into branch-and-price.py

jajhall:


jajhall:
@galabovaa This is just an update to `examples\branch-and-price.py` that I want to have in `master`

Issue #1487, opened 2023-11-04
Add cost and bound scaling as an option 


jajhall:
It's handy - in a case like #1484 - to be able to experiment with and suggest cost and uniform constraint scaling to users who have badly scaled models.

PR #1498, opened 2023-11-10
Fix 1487

jajhall:
HiGHS can now be run with user scaling of costs and/or all bounds, and also warns of excessively large or small maximum nonzero finite costs and bounds.

Hence, users with excessive bounds and costs causing HiGHS to fail can be asked to run HiGHS with appropriate scaling


jajhall:
This will fix #1487

PR #1499, opened 2023-11-11
Fix #1334

### LINKED ISSUE #1334: The optimal solution ultimately found has bugs
Hello, I solved the problem pk1.mps in MIPLIB2017 with Highs and found the optimal value is 15， but this problem's optimal value should be 11, as shown in the picture， I think there may be some issues with the logic of highs in proving the optimal solution, FYI.
gurobi_pk1

highs_pk1

Thanks

fwesselm:
HiGHS infers variable lower / upper bound (VLB / VUB) constraints in the clique table code.  For example, a variable upper bound looks like x <= a * y + c where x is a cont. and y is a binary variable.

Fix #1334 
- If a = -kHighsInf and c = kHighsInf, the actual bound on x given by the VUB constraint above evaluates to NaN.
- I added assertions to detect this in the future.
- Qualification on publicly available instances is ongoing. However, I think this bug fix may be non-controversial, and thus I decided to create this PR now.

### DIFF
--- a/src/mip/HighsCliqueTable.cpp
+++ b/src/mip/HighsCliqueTable.cpp
@@ -1164,6 +1164,9 @@ void HighsCliqueTable::extractCliquesFromCut(const HighsMipSolver& mipsolver,
               coef = globaldom.col_upper_[col] - implcolub;
               constant = implcolub;
             } else {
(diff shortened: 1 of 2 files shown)


fwesselm:
I should have mentioned that HiGHS now finds the reported optimal solution for pk1.

PR #1508, opened 2023-11-15
Reduce amount of similar code in HighsPathSeparator.cpp

fwesselm:
- While debugging I realized that HighsPathSeparator.cpp contains repetitive code. 
- I added two lambdas to simplify the code.
- Extensive testing showed that the code changes do not affect HiGHS behavior / performance (as expected).

### DIFF
--- a/src/mip/HighsPathSeparator.cpp
+++ b/src/mip/HighsPathSeparator.cpp
@@ -227,62 +227,70 @@ void HighsPathSeparator::separateLpSolution(HighsLpRelaxation& lpRelaxation,
         return w <= maxWeight && w >= minWeight;
       };
 
-      auto skipCol = [&](const HighsInt& col, const auto& colArcs,
-                         const auto& arcRows, const auto& otherColArcs,
-                         const auto& otherArcRows) {
-        if (currPathLen == 1 && !tryNegatedScale) {
-          if (colArcs[col].second - colArcs[col].first <= currPathLen) {
-            for (HighsInt k = colArcs[col].first; k < colArcs[col].second;
-                 ++k) {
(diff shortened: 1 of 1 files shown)


jajhall (approved):
I know that duplicated code is to be avoided, but I don't see the point in eliminating it from code written a long time ago that is correct. I know that you're careful, and testing for changes in performance, but there's still the small chance of introducing errors. Ditto replacing conditional execution of the code in a loop with `continue`using the opposite logic. 

What are you trying to achieve?

fwesselm:
> I know that duplicated code is to be avoided, but I don't see the point in eliminating it from code written a long time ago that is correct. I know that you're careful, and testing for changes in performance, but there's still the small chance of introducing errors. Ditto replacing conditional execution of the code in a loop with `continue`using the opposite logic.
> 
> What are you trying to achieve?

Three reasons for eliminating duplicate / similar code that come to my mind are increased maintainability, reduced error-proneness, and potentially increased code coverage. The changes in this PR were qualified on 850+ instances.

jajhall:
There are excellent reasons for avoiding duplicate code when first writing it but, once it's running I have to question the value of "fiddling" with it. Fixing bugs is far a more valuable use of time and benchmarking resources.

On reflection, perhaps there's a very small chance that if you eliminate duplicated code and behaviour changes, then it indicates that there was an error due to having a difference between two code sections that should have been identical. Still seems like a long shot to me.

lkapelevich:
> Three reasons for eliminating duplicate / similar code that come to my mind are increased maintainability, reduced error-proneness, and potentially increased code coverage. The changes in this PR were qualified on 850+ instances.

Simplifying older code will also help future readers.

PR #1542, opened 2023-12-07
Updates to readme | Python installation and restructured headers/nesting

thesethtruth:
Hi guys,

First of all, awesome work on the solver. I use it on a regular basis for energy system modelling, and feel grateful to be able to use your work. 

Me and some other colleagues found it difficult to make up from the readme that installing the Python interface will also install the HiGHS-library. It's insinuated in the docs, and works in practice. 

Please let me know if this restructuring of the readme is useful to you. Personally I think it is more accessible like this to the inexperienced programmers (for sure those who can only use Python) and want to make use of HiGHS but are uncertain if compiling HiGHS themselves lays within their capabilities. 

Thanks again,

Seth

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@
   - [Google Colab Example](#google-colab-example)
   - [Reference](#reference)
 
-About HiGHS
+## About HiGHS
 -----------
 
 HiGHS is a high performance serial and parallel solver for large scale sparse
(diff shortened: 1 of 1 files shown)


jajhall:
Thanks for this. I don't like the bold lines that you've introduced, so I've removed them. It's also incorrect that `highspy` can be tested with `minimal.py`, since (as it says in the file itself) 

> This example cannot be run with the version (1.5.3) of highspy available from PyPI. It requires a local installation of highspy for (at least) HiGHS version 1.6.0

However, there may be merit in your more substantial changes, and I'll get a view on them from @galabovaa

thesethtruth:
Hi @jajhall, thanks for the response. I can't see any changes made to the 'bold lines', if you provide a little more feedback I can perhaps rewrite it to be more to your liking. 

I didn't check the original documentation, and the part about `minimal.py` is taken from there. 

Let me know if I can assists @galabovaa, I'd be happy to help.

jajhall:
The 'bold lines' have now been removed (again!) and I see that `minimal.py` was there originally. We've focused on the [main documentation](https://ergo-code.github.io/HiGHS/stable/) recently

thesethtruth:
Thanks, I can see it on your branch now. Might have been that you didn't push the changes before. I can see two commits now, the first removing the reference to the `minimal.py` and the second removing the 'chapter' lines. 

Again, I'd be happy to help if there's any other points that need rewriting.

Issue #1557, opened 2024-01-04
Duals not flipped when maximizing and not running crossover


jajhall:

jajhall:
Closed by #1559

PR #1559, opened 2024-01-04
Fix 1557

jajhall:
Now performs dual flip in `ipxSolutionToHighsSolution` when maximizing


jajhall:
This will close #1557

PR #1564, opened 2024-01-08
Problems reduced to empty by presolve

fwesselm:
I think it would it be reasonable to have `presolved_model_` reference `data_.reduced_lp_` if presolve reduced the problem to empty. For example, this would allow for easily accessing the optimal objective value (`offset_`) via `getPresolvedLp()` in this case.


jajhall:
Fair enough

PR #1567, opened 2024-01-09
Fix potential crash in postsolve

fwesselm:
Postsolve may access vectors `basis.col_status` and `basis.row_status` even though the basis is not valid (`!basis.valid`).

The proposed fix is to overload function `computeStatus` for the two cases.


fwesselm:
> I'm working on the "reviseBasis" anomaly at the moment, so please revert your change or we'll get a merge conflict

Done.

jajhall:
Well spotted.

PR #1572, opened 2024-01-11
Allow postsolve to run when presolve has not performed any reductions

fwesselm:
Currently, HiGHS would return an error in this case (instead of just letting postsolve run and do nothing).


jajhall:
Thanks, I've already made this modification as part of the other work that I'm doing - which is to add presolve, postsolve and getPresolvedLp to the C API, and improve the testing, which is how I spotted the same issues as you.

fwesselm:
> Thanks, I've already made this modification as part of the other work that I'm doing - which is to add presolve, postsolve and getPresolvedLp to the C API, and improve the testing, which is how I spotted the same issues as you.

Ok, sounds great.

Issue #1576, opened 2024-01-12
Pass tests with sanitizer


Coloquinte:
Using the meson build, it is easy to run address, undefined and thread sanitizers.

```
meson setup build_address -Dwith_tests=True -Db_sanitize=address; meson test -C build_address
meson setup build_undefined -Dwith_tests=True -Db_sanitize=undefined; meson test -C build_undefined
meson setup build_thread -Dwith_tests=True -Db_sanitize=thread; meson test -C build_thread
```

The address sanitizer fails due to leaks in the C API tests. The undef and thread sanitizers seem to uncover more serious errors.

(I'll get around to investigating it more at some point, but probably not this week)

jajhall:
Thanks, that's useful to know. I can well imagine that there are memory leaks in the testing of the C API, as we're C++ developers!

PR #1579, opened 2024-01-13
Fix minor sanitizer issues and add optional Meson workflow to run the sanitizers

Coloquinte:
Related to #1576 

Some very minor sanitizer issues remain, as can be seen in the logs that are uploaded [by the Github action here](https://github.com/Coloquinte/Highs-solver/actions/runs/7513161918)
* A small memory leak (could not reproduce locally)
* A typing issue when casting an option
* Null pointer arithmetic when manipulating an empty slice
* A data race between `HighsSplitDeque::WorkerBunk::waitForNewTask` and `HighsSplitDeque::injectTaskAndNotify(HighsTask*)` (could not reproduce on github)

### DIFF
--- a/.github/workflows/sanitizers.yml
+++ b/.github/workflows/sanitizers.yml
@@ -1,0 +1,43 @@
+name: Test with sanitizers
+on: [workflow_dispatch]
+
+jobs:
+  sanitizer:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest]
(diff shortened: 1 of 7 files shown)


galabovaa:
Thank you, these look great so far!

jajhall:
Thanks: I can see why the source code modifications sanitize what was originally written

Coloquinte:
Thanks! There is one minor leak left, but I'm not sure that's really a bug

jajhall:
Thanks for all these! Have a good weekend

Coloquinte:
Thank you! You too :)

PR #1582, opened 2024-01-15
Fix #1581 

### LINKED ISSUE #1581: Out-of-bound access in HighsLpUtils.cpp
I think there is an out-of-bound access in HighsLpUtils.cpp [here](https://github.com/ERGO-Code/HiGHS/blob/6c9fa1b23a3fc178588443ae1597bcd465deece3/src/lp_data/HighsLpUtils.cpp#L424) (access after the bounds of the container), which is picked by the address sanitizer in some conditions.
This can be shown by running the tests with [this commit](https://github.com/Coloquinte/Highs-solver/commit/05bcdd85b455f012616f6c0f1aa5f7055ca04d66), which checks the bounds explicitly.

Related to #1576, but a bit more involved to fix than the other issues.

fwesselm:
Fix #1581 - vector `integrality` may be empty.

### DIFF
--- a/src/lp_data/HighsLpUtils.cpp
+++ b/src/lp_data/HighsLpUtils.cpp
@@ -56,9 +56,9 @@ HighsStatus assessLp(HighsLp& lp, const HighsOptions& options) {
                                         return_status, "assessCosts");
     if (return_status == HighsStatus::kError) return return_status;
     // Assess the LP column bounds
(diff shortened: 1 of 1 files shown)


Coloquinte:
Very quick, and simpler than I thought :)

jajhall (approved):
Great!

Issue #1584, opened 2024-01-17
Added variables without calling Highs_changeColIntegrality are sometimes marked as integer or semi-cont


peno64:
I don't have a repro case atm, but could it be checked in the code that when a column is added via API call Highs_addCol, giving it a name via Highs_passColName, possibly setting its bounds via Highs_changeColBounds and then not calling Highs_changeColIntegrality to keep it a continiuous variable (but a previous variable was set to integer via Highs_changeColIntegrality) that 'sometimes' that variable is not seen as a continiuous variable but as integer or semi-cont?
I have this not after a first optimize but sometimes after a third or more build model/optimize model/destroy model but it is not consistent. It is as if there is some unitialized memory there.
It looks like if I always call Highs_changeColIntegrality for each added variable, so also for the continiuous variables where it should not be needed, that the problem is not there.

I know this is a difficult one but I guess it is easy to check in the code when a call is done to Highs_addCol if that added variable is at this moment also really initialised as being a continiuous variable and not something random.

jajhall:
Well spotted. What actually happens is that when a column is added, the `integrality_` data member of the internal model isn't even resized, so the `integrality_` values you're seeing for new columns are unassigned. When `changeColIntegrality` is called for new columns to make them `kInteger`, the `integrality_` values for earlier columns aren't set if the problem was previously an LP! 

Quite a mess. I'll fix it in our `master` branch. Look for the PR related to this issue

jajhall:
Closed by #1586

PR #1586, opened 2024-01-17
Applied fix-1584 mods to update master

jajhall:


jajhall:
This was a horribly bad error in master. For comments, see #1584 ; for modifications, see #1585

jajhall:
Closes #1584

PR #1589, opened 2024-01-23
Fix assertion failure in getVariableKktFailures

fwesselm:
When I set the flag `force_debug` [here](https://github.com/ERGO-Code/HiGHS/blob/6410710acf6ef49f2294a864c5127d2e54237bf7/src/lp_data/Highs.cpp#L1447) to `true` and solve the LP relaxation of problem `vpm1` from MIPLIB3.0, I get an assertion failure in `getVariableKktFailures` on win64:

```
    Highs highs;

    // Read problem vpm1 from MIPLIB3.0 and make it an LP
    highs.clearSolver();
    std::string filename = "vpm1.mps";
    highs.readModel(filename);
    std::vector<HighsInt> vars(highs.getNumCol(), 1);
    std::vector<HighsVarType> integral(highs.getNumCol(), HighsVarType::kContinuous);
    highs.changeColsIntegrality(vars.data(), integral.data());

    // Run
    highs.run();
```

Postsolve assumes that a doubleton row was an equality constraint. However, row bounds may have been modified (in the row presolve method).

### DIFF
--- a/src/presolve/HPresolve.cpp
+++ b/src/presolve/HPresolve.cpp
@@ -3063,13 +3063,14 @@ HPresolve::Result HPresolve::rowPresolve(HighsPostsolveStack& postsolve_stack,
     }
   }
 
-  if (rowsize[row] == 2 && model->row_lower_[row] == model->row_upper_[row] &&
-      analysis_.allow_rule_[kPresolveRuleDoubletonEquation])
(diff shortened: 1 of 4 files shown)


jajhall (approved):
Thanks. Since these have been tested extensively, they are good to go

PR #1596, opened 2024-01-28
Cleanup of continuous integration

Coloquinte:
A proposal to get rid of spurious failures in CI:
* Option to disable threading for Windows CI (so the Windows build does not break due to threading bad luck)
* Increase the timeout of the meson tests to match the CMake tests (so they do not timeout randomly)
* Found some tests that were only run in a very specific workflow (by mistake I think). Moved them with the other tests (but maybe they should just be removed)
* Minor cleanup of messages and warnings


jajhall:
Thanks: nice to see the checks all passing!

Are these changes OK with you @galabovaa?

galabovaa (approved):

galabovaa:
> Thanks: nice to see the checks all passing!
> 
> Are these changes OK with you @galabovaa?

Looks good to me!

Issue #1294, opened 2023-05-17
On 32-bit platforms libatomic should be linked to


barracuda156:
```
[ 77%] Linking CXX shared library ../lib/libhighs.dylib
cd /opt/local/var/macports/build/_opt_PPCRosettaPorts_math_HiGHS/HiGHS/work/build/src && /opt/local/bin/cmake -E cmake_link_script CMakeFiles/libhighs.dir/link.txt --verbose=ON
[...]

PR #1603, opened 2024-02-03
BLD: `meson` updates for SciPy

HaoZeke:
Also includes a fix for #1294.

Addresses the [latest review](https://github.com/scipy/scipy/pull/19255#issuecomment-1925424141) for SciPy integration.

### DIFF
--- a/check/meson.build
+++ b/check/meson.build
@@ -70,8 +70,10 @@ foreach test : test_array
       )
 endforeach
 
-test('test_capi',
-    executable('capi_unit_tests', 'TestCAPI.c',
-  include_directories: _incdirs,
-  link_with : _linkto ,
-))
+if get_option('with_c')
+  test('test_capi',
+      executable('capi_unit_tests', 'TestCAPI.c',
+    include_directories: _incdirs,
+    link_with : _linkto ,
+  ))
+endif
--- a/cmake/CheckAtomic.cmake
+++ b/cmake/CheckAtomic.cmake
@@ -1,0 +1,115 @@
+# atomic builtins are required for threading support.
+INCLUDE(CheckCXXSourceCompiles)
+INCLUDE(CheckLibraryExists)
+# Sometimes linking against libatomic is required for atomic ops, if
+# the platform doesn't support lock-free atomics.
+function(check_working_cxx_atomics varname)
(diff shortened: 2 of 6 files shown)


HaoZeke:
@galabovaa and @jajhall this should be ready to go.

galabovaa (approved):
looks good to me!

barracuda156:
@HaoZeke Why does it refer only to LLVM? GCC needs that check and `-latomic` flag.

HaoZeke:
> @HaoZeke Why does it refer only to LLVM? GCC needs that check and `-latomic` flag.

I believe that is covered by this section:

```cmake
if(MSVC)
  set(HAVE_CXX_ATOMICS_WITHOUT_LIB True)
elseif(LLVM_COMPILER_IS_GCC_COMPATIBLE OR CMAKE_CXX_COMPILER_ID MATCHES "XL")
  # First check if atomics work without the library.
  check_working_cxx_atomics(HAVE_CXX_ATOMICS_WITHOUT_LIB)
  # If not, check if the library exists, and atomics work with it.
  if(NOT HAVE_CXX_ATOMICS_WITHOUT_LIB)
    check_library_exists(atomic __atomic_fetch_add_4 "" HAVE_LIBATOMIC)
    if(HAVE_LIBATOMIC)
      list(APPEND CMAKE_REQUIRED_LIBRARIES "atomic")
      check_working_cxx_atomics(HAVE_CXX_ATOMICS_WITH_LIB)
      if (NOT HAVE_CXX_ATOMICS_WITH_LIB)
        message(FATAL_ERROR "Host compiler must support std::atomic!")
      endif()
    else()
      message(FATAL_ERROR "Host compiler appears to require libatomic, but cannot find it.")
    endif()
  endif()
endif()
```

Would be best if you could try it though, I don't have a 32-bit machine to check against.

barracuda156:
@HaoZeke If, counter-intuitively, `LLVM_COMPILER_IS_GCC_COMPATIBLE` evaluates as true for GCC, it probably will work.

HaoZeke:
> @HaoZeke If, counter-intuitively, `LLVM_COMPILER_IS_GCC_COMPATIBLE` evaluates as true for GCC, it probably will work.

Ah, my bad, that's unlikely, I opened #1604 with a modified fix.

PR #1619, opened 2024-02-16
Dual and basis postsolve for empty solution

fwesselm:
- Dual and basis postsolve do not work when presolve reduced problem to empty.
- I have added a test case.


jajhall (approved):
Users can't be expected to set the `.valid` flag, but I see that this is necessary in the case where postsolve is being used for a problem that was reduced to empty. I'll retain `basis_supplied`, but (also) set it true if the `.valid` flag is set. 

Ditto for the dual solution

fwesselm:
> Users can't be expected to set the `.valid` flag, but I see that this is necessary in the case where postsolve is being used for a problem that was reduced to empty. I'll retain `basis_supplied`, but (also) set it true if the `.valid` flag is set.
> 
> Ditto for the dual solution

Thank you

PR #1627, opened 2024-02-19
Windows version included in DLL + Bazel update

### LINKED ISSUE #1610: No version info on binaries in Windows
Under Windows, each exe and dll should have the version imbeded in the binary. This is done via a resource file used when compiling/linking the binaries. Then when the properties of the binary is retrieved, it should show this version. But even more. You can see a copyright and a company name also but the version is the most important and this when the application is distributed with an installer. Windows installers look at this version number to determine if a binary must be overwritten. If the version is older or equation, it will not be overwritten. When the version is newer is will be overwritten/updated.
I am not sure what happens if the binary has no version. Either it will always be overwritten or not.

The problem is now that HiGHS does not set the version information on its binaries. This is not on the precompiled binaries on JuliaBinaryWrappers and also not if you compile the binaries yourself (https://github.com/ERGO-Code/HiGHS/blob/master/docs/src/installation.md or https://github.com/ERGO-Code/HiGHS/issues/694)

Can this be added in future versions?

Coloquinte:
Fixes #1610 for DLL (not .exe)


galabovaa (approved):
Thank you for these! The bazel config file was a dirty hack for a WIP interface to or-tools, which we have yet to complete. This will be helpful as we return to the task.

Windows DLL version fix is much appreciated!

PR #1637, opened 2024-02-26
unscaleSolution

fwesselm:
I would like to be able to call `unscaleSolution`. The proposed solution is to move it from `src/simplex/HSimplex.cpp` to `src/lp_data/HighsLpUtils.cpp`.


jajhall:
OK, and hopefully no-one else is using it and expects it in 'simplex'

PR #1651, opened 2024-03-06
Merge latest for release

galabovaa:
Release 1.7.0


jajhall:
Version 1.7.0 of HiGHS contains several new developments. 

HiGHS now offers the exciting new first order LP solver ([**cuPDLP-c**](https://github.com/COPT-Public/cuPDLP-C)). Ours is currently only a serial CPU implementation, so performance isn't spectacular, but we aim to offer the GPU implementation in due course.  To run PDLP, choose the option `solver=pdlp`. Note that there is currently no time-out available for PDLP, and the termination criteria for cuPDLP-c may mean that the HiGHS optimality conditions are not satisfied. See [README.md](https://github.com/ERGO-Code/HiGHS/tree/master/src/pdlp/cupdlp/README.md)  . 

The HiGHS interior point solver now allows analytic **centre calculations** to be performed. This is done by setting the HiGHS option `run_centring` to be true. Other new options allow the centring calculations to be controlled.

When `Highs::run()` is called, ranges of the matrix, costs, bounds and RHS are reported, together with advice on setting the new HiGHS options `user_cost_scale` and `used_bound_scale` if there are extreme values. 

After solving an LP, the method `Highs::getIllConditioning` tries (in an approach similar to that used by Gurobi) to identify constraints or basic variables that contribute most to ill-conditioning of the basis matrix. 

Logging from the interior point solver is now consistent with logging from the simplex and MIP solvers.

The Python interface has been updated, with several overlooked methods having been added

Considerable modifications and fixes have been added to the build system, in particular for Python. `highspy` is no longer built with `meson`, rather a combination of `setuptools` and `cmake`.

Bugs in the QP and MIP solvers have been fixed

Thanks are due to due to @galabovaa , @fwesselm @feldmeier and @Coloquinte

jajhall (approved):
Hurray for the new release: a lot of work has been done since the last!

PR #1656, opened 2024-03-08
fix some compiler warnings

svigerske:
`#ifndef CUPDLP_CPU & USE_KERNELS` generates a compiler warning that cannot be suppressed. And it is indeed difficult to see for the compiler what could be meant by `CUPDLP_CPU & USE_KERNELS` not being defined.

Also make some functions that are used locally only static to get rid of missing prototype warnings in C++ code.

### DIFF
--- a/src/pdlp/cupdlp/cupdlp_step.c
+++ b/src/pdlp/cupdlp/cupdlp_step.c
@@ -17,7 +17,7 @@ void PDHG_primalGradientStep(CUPDLPwork *work, cupdlp_float dPrimalStepSize) {
   CUPDLPiterates *iterates = work->iterates;
   CUPDLPproblem *problem = work->problem;
 
-#ifndef CUPDLP_CPU & USE_KERNELS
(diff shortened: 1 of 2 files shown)


jajhall:
Thanks @svigerske. I'd meant to fix the warning relating to the compiler directives. 

The original directive (in [cupdlp/cupdlp_step.c](https://github.com/COPT-Public/cuPDLP-C/blob/main/cupdlp/cupdlp_step.c)) is 

> #if !(CUPDLP_CPU) & USE_KERNELS

and this caused issues - as you may have read in [src/pdlp/cupdlp/README.md](https://github.com/ERGO-Code/HiGHS/blob/master/src/pdlp/cupdlp/README.md)

Your fix is what I couldn't find!

As for 

> make some functions that are used locally only static to get rid of missing prototype warnings in C++ code.

that's way over my head programming-wise! I'm tempted to say that the function definitions have been there for a couple of years, but they were introduced by @fwesselm, who is helping us to debug the MIP presolve.

PR #1657, opened 2024-03-08
fix that optionChangeAction() was not called when reading options file

svigerske:
When reading an options file, the setOptionValue() functions are bypassed, so this call was missing, which resulted in the new user-scalings not being applied.

Btw, is it intentional, that if one sets `user_bound_scale`, then the solution returned is not scaled back, but given for the scaled problem? I suppose the same happens with `user_cost_scale` and dual solution.

### DIFF
--- a/src/lp_data/Highs.cpp
+++ b/src/lp_data/Highs.cpp
@@ -124,7 +124,7 @@ HighsStatus Highs::readOptions(const std::string& filename) {
     default:
       break;
   }
-  return HighsStatus::kOk;
+  return optionChangeAction();
 }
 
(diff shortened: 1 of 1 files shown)


jajhall:
Fair cop, thanks!

> Btw, is it intentional, that if one sets `user_bound_scale`, then the solution returned is not scaled back, but given for the scaled problem? I suppose the same happens with `user_cost_scale` and dual solution.

Yes, it was intentional, otherwise the solution returned won't satisfy the HiGHS feasibility tolerances. The idea is to return "the solution you would have got if you'd scaled your problem sensibly"

However, I can see that this needs to be documented.

PR #1661, opened 2024-03-10
scikit-build-core for python

galabovaa:
Python modelling language included


galabovaa:
closing until I fix the linux wheel test to avoid unnecessary action triggers

PR #1664, opened 2024-03-11
Feature: Add a simple nix setup to build the highs binary

silky:
I notice HiGHS [is not on nixpkgs](https://search.nixos.org/packages?channel=23.11&from=0&size=50&sort=relevance&type=packages&query=HiGHS), so until such time I thought I'd add a `flake.nix` to build the binary.

It also means it's much easier for nix users to just try it out.

Note that I don't do anything with the Python side here; it could be an interesting addition later to _also_ build the python package; but I'll leave that for another piece of work.

Hope this is of use!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -277,3 +277,7 @@ bazel*
 build_webdemo
 
 CMakeSettings.json
+
+# Nix
+.direnv/
+result
--- a/README.md
+++ b/README.md
@@ -83,6 +83,23 @@ _These binaries are provided by the Julia community and are not officially suppo
 
 See https://ergo-code.github.io/HiGHS/stable/installation/#Precompiled-Binaries.
 
+### Build with Nix
+
+There is a nix flake that provides the `highs` binary:
+
+```shell
+nix run .
+```
+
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -276,4 +276,8 @@ bazel*
 # webdemo
 build_webdemo
 
-CMakeSettings.json
```
silky, line 279: I have no idea what this means :/ (i.e. why it's added and removed and what the warning sign is about; maybe it's just newlines.)

README.md
```diff
@@ -98,14 +115,14 @@ The python package `highspy` is a thin wrapper around HiGHS and is available on
 $ pip install highspy
 ```
 
-Alternatively, `highspy` can be built from source.  Download the HiGHS source code and run
```
silky, line 101: these are just my text editor removing trailing spaces.

.gitignore
silky: ah yeah; it's just the EOF marker.

galabovaa:
Thank you for the nix addition, @silky!

Issue #1271, opened 2023-04-25
CMakeLists.txt RELWITHDEBINFO handling joins RELEASE and DEBUG flags causing msvc conflict /RTC and /O


Thell:
https://github.com/ERGO-Code/HiGHS/blob/d4809eae74a89e4bb36dd45dca2c7e08c14d5ef1/CMakeLists.txt#L187

![image](https://user-images.githubusercontent.com/2185020/234175977-dd48a4af-7276-43f4-a55f-25be906bf753.png)

I had originally opened this to the `highs-sys` project (https://github.com/rust-or/highs-sys/issues/11) but have now found where it comes from (linked above). 

`CMAKE_CXX_FLAGS_DEBUG:STRING=/Zi /Ob0 /Od /RTC1`
`CMAKE_CXX_FLAGS_RELEASE:STRING=/O2 /Ob2 /DNDEBUG`

This is what it should be according to CMakeCache.txt -> `CMAKE_CXX_FLAGS_RELWITHDEBINFO:STRING=/Zi /O2 /Ob1 /DNDEBUG` when configured with Ninja using

```
[...]

PR #1665, opened 2024-03-11
C flags

galabovaa:
Issue https://github.com/ERGO-Code/HiGHS/issues/1271


galabovaa:
A patch for IPO and removing old broken code for release with debug info

Issue #1251, opened 2023-04-12
nameclash on Windows for .pdb files


lkapelevich:
When building with MSVC the .pdb file for the .dll (which is needed for the debugger) is overwritten by the .pbd file for the executable. This happens because both are named highs.pdb. When I put an "if(NOT MSVC)" clause around [this line](https://github.com/ERGO-Code/HiGHS/blob/master/src/CMakeLists.txt#L327) we generate a libhighs.pdb, so the issue is resolved. 

Perhaps a better fix would be to check whether the system is not Windows for the same clause, or rename the executable instead.

lkapelevich:
Hi @galabovaa. That works, no issues. 

I still think it's strange that the shared library and executable get a .pdb with the same name. FWIW, I just found this flag https://cmake.org/cmake/help/latest/prop_tgt/PDB_NAME.html, and `PDB_NAME libhighs` in the same clause also works for me.

jajhall:
Has this issue been resolved?

lkapelevich:
I think adding `PDB_NAME` would resolve this fully. 
An analogous issue/fix in another repo: https://github.com/eclipse/mosquitto/issues/2714

galabovaa:
Thanks again Lea, I have just finished the big cmake refactoring we needed for python and looked at this again. I've added the fix to FAST_BUILD OFF for legacy code and will merge it in latest. Note, that we may remove the FAST_BUILD OFF option all together at some point, so users are strongly encouraged to use the default FAST_BUILD=ON and let us know if there are any issues with that!

lkapelevich:
That's great.
Is there a reason not to add the flag for FAST_BUILD ON, so that it can stay when FAST_BUILD OFF disappears?

galabovaa:
Yes, it is best not to add it at all! Removing it won't happen for another few months, but still, since FAST_BUILD=ON is default now, you don't need it. It was needed just before we made it default, as we were adding features and bug fixes to the "more modern cmake" and we didn't change much in FAST_BUILD OFF to avoid breaking what people were using at the time as we upgraded.

PR #1668, opened 2024-03-12
Pdb

galabovaa:
speed of testing 
closes https://github.com/ERGO-Code/HiGHS/issues/1251


jajhall (approved):
Nothing that I understand here, so go ahead!

Issue #1673, opened 2024-03-15
Various issues with highspy modelling language


jajhall:
- Need to be able to get variable value from the variable name
- Need to be able to get the variable name from its algebraic form
- Need to be able to get the variable dual in both ways
- Need to be able to get the constraint dual from its name
- Need to document that names cannot have spaces when defining MPS files
- 
[chip.py.txt](https://github.com/ERGO-Code/HiGHS/files/14618614/chip.py.txt)

jajhall:
Closed by #1675

PR #1675, opened 2024-03-17
Added methods to highspy modelling and two example models

jajhall:
Fleshed out the modelling methods in `highs.py` with

- setMinimize()
- setMaximize()
- setInteger(var)
- silent()
- solve()
- varName(var)
- varNames(vars)
- allVarNames()
- varValue(var)
- varValues(vars)
- allVarValues()
- varDual(var)
- varDuals(vars)
- allVarDuals()
- constrValue(constr_name)
- constrValues(constr_names)
- allConstrValues()
- constrDual(constr_name)
- constrDuals(constr_names)
- allConstrDuals()

### DIFF
--- a/examples/chip.py
+++ b/examples/chip.py
@@ -1,0 +1,53 @@
+# model and solve the LP
+#
+# maximize 10 Tables +  25 SetsOfChairs
+# s. t.       Tables +   2 SetsOfChairs <=  80: Assembly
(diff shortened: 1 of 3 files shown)


jajhall:
Will close #1673

jajhall:
I guess this is OK @galabovaa: all I've done is edit `highs.py`, which is executed like Python at run-time, so I don't imagine that it affects the Python build

PR #1698, opened 2024-03-28
Moved pdlp_iteration_count to last data member of HighsCallbackDataOut

### LINKED ISSUE #1697: Add accessor functions for the callback data structs from the C API
It doesn't seem possible to use callback data in a way that allows for swapping out HiGHS DLLs for newer versions safely.

In 1.7.0, it looks like a new `pdlp_iteration_count` member was added to the `HighsCallbackDataOut` struct ([see here](https://github.com/ERGO-Code/HiGHS/commit/2df50399c5123eaf8b350a423e1cd2d96ef5d3e8#diff-def8df6d86ea8e7443f442dc63040b38b2167c37235db1af09c26110058199cf)). This moves all subsequent fields, so we can't access the data in 1.7.0 using our application built for HiGHS 1.6.0 (and now to make it compatible with both, we'd have to check the HiGHS version and cast to the different structs).

Generally, I think the way to deal with this would be to add accessor functions to the C API for getting the values of the struct members. That way even if the layout of the struct changes, the accessor functions will still work.

jajhall:
This will close #1697

### DIFF
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -1,13 +1,21 @@
+### Build changes
+
 ## HiGHS on nixpkgs
 
 HiGHS now has a `flake.nix` to build the binary, allowing `nix` users to try it out
 
-#### Python build update
+## Python build update
 
 Highspy with setuptools from v1.7.0 only worked on Python 3.12
 For v1.7.0 we have dropped setuptools and switched to scikit-build-core
 
-#### Windows versions
+## Windows versions
 
 Fixed version info of shared library
 Added version info to executable
+
+### Code changes
+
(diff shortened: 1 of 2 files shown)


odow (approved):
Is it possible to get this released as a 1.7.1? Otherwise we should flag somewhere that using the CallbackStruct from C is still experimental and will break in the next release.

Although probably anyone attempting to use it will run into #1697 so maybe that's enough warning.

jajhall:
I guess we could get an interim v1.7.1 release without me fixing all the issues exposed in 1.7.0 - I don't have time ATM

It would mean that 'highspy' and other recent work by @galabovaa would be available.

PR #1700, opened 2024-03-29
master: Add tests to bazel

Mizux:


Mizux:
@jajhall should I remove it in favor of:
* #1701

jajhall:
Thanks for this. Note that we only push to master to create new releases or make important bug fixes, so please submit PRs to latest

PR #1726, opened 2024-04-15
Initialize substitution opportunities

fwesselm:
Presolve only updates `substitutionOpportunities` after a new implied bound has been found.

If no implied bounds are identified, `substitutionOpportunities` remain empty even though the model passed to presolve may contain equality constraints with free variables.

The proposed change is to initialize `substitutionOpportunities` before any reductions are performed.


fwesselm:
Testing on MIPLIB2017, etc. is ongoing.

PR #1730, opened 2024-04-22
Update bounds on dual variables when converting inequality into equality constraints in presolve

### LINKED ISSUE #1725: Potential issues with presolve: incorrect 'optimal' solution and incorrect infeasibiliy detection
Greetings,
I have two problems where HiGHS returns incorrect results. In the first problem the solution returned is suboptimal and in the second it is incorrectly reported as infeasible. Turning presolve off avoids the potential issue(s).
[TwoProblems.zip](https://github.com/ERGO-Code/HiGHS/files/14963087/TwoProblems.zip)

fwesselm:
Fix #1725 
- Row presolve may convert inequality into equality constraints. Currently, the code does not update the bounds on the dual variables accordingly. This causes presolve to do an invalid reduction (fixing) subsequently.
- I am currently qualifying the fix to make sure that there are no undesired side effects.
- @jajhall , do you think adding these problems as test cases would be worthwhile?


jajhall:
Well spotted. Whilst these problems have exposed a bug, once fixed they may be easy to solve, so not worth having as test cases. I can't imagine anyone reverting your fix.

jajhall:
I've merged into #fix-1725 since I put a minor debugging change into that branch, and so that I can see what other issues your fix may close

PR #1732, opened 2024-04-23
Fix 1725

jajhall:
To merge into latest once @fwesselm has tested it


fwesselm:
Testing revealed no issues.

PR #1737, opened 2024-04-26
Updates to the Meson build to fix portability issues and warnings 

rgommers:
Each commit has a commit message with more context. This reduces the build warnings to a small amount, only the ones that may need inspection or are perhaps better fixed rather than silenced (things like `-Wunused-variable` are safe to silence though).

The changes to compiler flag handling are necessary for portability. The symbol visibility one avoids re-exporting ~2,000 private symbols in the shared libraries that link against a static `libhighs.a`. It also reduced binary size, by ~300 kb on macOS in a release build.

### DIFF
--- a/check/meson.build
+++ b/check/meson.build
@@ -63,7 +63,7 @@ foreach test : test_array
                      test.get(1),
                      highs_conf_file],
           dependencies : _deps,
-          link_with : _linkto ,
+          link_with : highslib,
           cpp_args : _args,
           include_directories: _incdirs,
                  ),
@@ -75,6 +75,6 @@ if get_option('with_c')
   test('test_capi',
       executable('capi_unit_tests', 'TestCAPI.c',
     include_directories: _incdirs,
-    link_with : _linkto ,
+    link_with : highslib,
   ))
 endif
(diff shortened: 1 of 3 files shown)


rgommers:
Argh, missed that PRs are not supposed to go to the default branch. Closing - will rebase and resubmit.

rgommers:
And then that you already re-targeted, thanks @jajhall. Let me retest locally and fix the merge conflict.

jajhall:
Rather a lot of CI failures: are they expected @galabovaa ?

rgommers:
The reason CI jobs on macOS are failing is because the `macos-latest` runner image changed from macOS 12 (x86-64) to macOS 14 (arm64) in the last couple of days. This is probably best addressed in a separate PR. Changing from `macos-latest` to `macos-13` is perhaps the easiest short-term fix?

galabovaa:
Looks good to me, particularly since the changes only affect meson build code. The failures are due to different iteration counts on macOS arm64 and are not concerning. I can look at those if they still appear once this is merged in latest.

galabovaa (approved):

PR #1743, opened 2024-05-03
Deprecate compilation date

galabovaa:


jajhall:
Are you going to remove compilation date from the build system?

PR #1746, opened 2024-05-03
Addition to new tests, iterations on macOS arm64 in CMake

galabovaa:


jajhall (approved):
I'm still curious to know why HiGHS runs differently on this architecture

However, lets get the tests fixed first

galabovaa:
Recently, I added some tests on the CI for macOS arm: previously on GitHub we were only testing macOS Intel. 

I have been observing the differences on my Apple M1/M2 for a long time, last night I traced it back to 2022. At the time I didn't think it was significant and left it for later, the differences are on around 5/10 instances. This count is not related to anything that was added recently: I've been curious about it too. I guess I could try to compare the data structures at the start of solve

PR #1749, opened 2024-05-05
ENH: Meson highs 1.7.x

HaoZeke:
Should supersede #1405. Draft until:
- [X] It works locally (`linux`).
- [x] CI passes
- [x] This is checked against SciPy
- [ ] Bugfixes from #1405 are pulled in (as needed)

What's inside (why merge):
- Meson build for `highspy` (needed for SciPy)
- Minor fixes to reduce warnings, so the following are no longer needed:
  + `-Wno-return-type`
  + `-Wno-switch`
  + `-Wno-comment`
  + `-Wno-unused-const-variable`
  + `-Wno-unused-function`


HaoZeke:
For the windows python issues we might have to wait on https://github.com/mesonbuild/meson/issues/13167

HaoZeke:
I think, after the conversation with @galabovaa and @jajhall I'll just refocus this to work with SciPy. The Windows build "works" with careful handling of `delvewheel` but that's about as far as is needed since the team will keep `scikit-build` as the primary build system.

HaoZeke:
Closing in favor of another PR (in a few min)

PR #1752, opened 2024-05-07
Documentation: callback data structure comment 

galabovaa:


jajhall (commented):
Don't we also need documentation for `HighsCallbackDataIn`, too? 

And `HighsCCallbackType`?

PR #1753, opened 2024-05-07
Fix 1679 and 1681

jajhall:
Renamed

- `addVar` to `addVariable` to prevent shadowing of `highspy` binding to `Highs::addVar`
- Change the name of `removeVar` to `deleteVariable` for consistency with `Highs::deleteCol`
- Change the name of `getVars` to `getVariables`

Added a call of `self.update()` to `addVariable`, rather than deferring it to `addConstr`, to allow column data to be changed, and rows to be added, using `highspy` bindings to methods in the `Highs` class. These will assume that all variables that have been defined are in the incumbent model.

Unfortunately - see experiments with `examples/distillation.py` - this gets the column names wrong. It would appear as if the `names` stack isn't flushed when `self.update()` is called. Any thoughts @galabovaa ?


galabovaa:
OK it does seem that names are not cleared, this seems to be by design because the idea I think was to call update() once all the modelling is done, just before run. 

I tried to clear them naively at the end of update() in branch fixx, but that breaks one test: 

> TestHighsPy::test_deleteVariable FAILED  
https://github.com/ERGO-Code/HiGHS/actions/runs/9051630828/job/24868361542

Here is the failing test code: 
`
  def test_deleteVariable(self):  
        h = highspy.Highs()  
        x = [h.addVariable(), h.addVariable()]

        h.update()
        self.assertEqual(h.numVars, 2)

        h.deleteVariable(x[0])
        self.assertEqual(h.numVars, 1)
`

Natuarry, once I have cleared the list, the modelling language can not find the variable to delete. 

I'll investigate further

jajhall:
'update()' is called when each constraint is added - otherwise addRow() returns kHighsError. 

It seems as if just the names fail - possibly something Luke didn't test.

galabovaa:
I added a list of names to the batch and now call that with names: see the latest commit. 

I thought it would be easier than trying to keep track of batch indices in vars_. Now the batch names are cleared at the end of update() and the column names should look allright

src/highspy/highs.py
```diff
@@ -106,7 +106,9 @@ def update(self):
         current_batch_size = len(self._batch.obj)
         print("\nDEBUG update(self): On entry - current_batch_size = ", current_batch_size)
         if current_batch_size > 0:
-            names = [self._vars[i].name for i in range(current_batch_size)]
+            # names = [self._vars[i].name for i in range(current_batch_size)]
+            names = [self._batch.name[i] for i in range(current_batch_size)]
```
jajhall, line 108: Yes, this fixes it. It's actually nothing to do with adding update() to addVariables, since this would be wrong once update() has been called for the first non-empty batch.

jajhall:
> something is not right again because there is a test_var_name failing, will look further

We now can't change name thus

> 
    # change name before adding to the model
        x.name = 'y'
        self.assertEqual(x.name, 'y')

        # add to the model
        h.update()

TBH, I can't see a value in allowing this to be possible, so we can delete the test

Issue #1766, opened 2024-05-16
Protect `HPresolve::setInput` against memory allocation errors


jajhall:
Maximum memory usage in HiGHS, at least when solving LPs, may occur in `HPresolve::setInput`. It would be good if memory allocation errors in this method were caught, allowing HiGHS to return cleanly.

jajhall:
Closed by #1767

PR #1767, opened 2024-05-17
Fix 1766

jajhall:
Now catching memory allocation failures in `HPresolve` for all arrays except those in `fromCSR` and `fromCSC`, with these methods and `HPresolve::okSetInput` now returning `bool`


jajhall:
This closes  #1766

Issue #1689, opened 2024-03-21
QP Solver logging


jajhall:
The QP solver logging is too "chatty", and ragged in terms of field width. 

The frequency should be akin to the simplex and MIP solver logging

The original logging frequency should be available when debugging

jajhall:
Closed by #1774

PR #1774, opened 2024-05-24
QP solver logging less chatty; modifing Hessian when adding or deleting columns

jajhall:
QP solver only reports ten times per order of magnitude of iteration count


jajhall:
This will close #1689

jajhall:
This will close #1720

PR #1779, opened 2024-05-28
QP hot start

jajhall:
Takes the current HiGHS solution and basis and, if possible uses them to define the starting point of the QP solver.


jajhall:
- Hot start is well tested using `primal3.mps`, a non-trivial QP problem now added to `check/instances`. 
- Code style for QP solver identifiers relevant to the interface has been improved considerably. 
- The relation between the active set "basis" and the HighsBasis has been documented

PR #1782, opened 2024-05-30
Fix the O(N^2) bug of passColName and passRowName

metab0t:
Reland https://github.com/ERGO-Code/HiGHS/pull/1780

I also simplify some logic here. The result of `emplace` has contained the duplicate key-value pair if it detects duplication, so we can mark the value as duplicate directly instead of using the slow search-erase-insert combo.

### DIFF
--- a/src/lp_data/HighsLp.cpp
+++ b/src/lp_data/HighsLp.cpp
@@ -594,14 +594,13 @@ void HighsNameHash::form(const std::vector<std::string>& name) {
   size_t num_name = name.size();
   this->clear();
   for (size_t index = 0; index < num_name; index++) {
-    const bool duplicate = !this->name2index.emplace(name[index], index).second;
+    auto emplace_result = this->name2index.emplace(name[index], index);
+    const bool duplicate = !emplace_result.second;
     if (duplicate) {
       // Find the original and mark it as duplicate
-      auto search = this->name2index.find(name[index]);
-      assert(search != this->name2index.end());
+      auto& search = emplace_result.first;
       assert(int(search->second) < int(this->name2index.size()));
-      this->name2index.erase(search);
-      this->name2index.insert({name[index], kHashIsDuplicate});
+      search->second = kHashIsDuplicate;
     }
   }
 }
@@ -616,6 +615,19 @@ bool HighsNameHash::hasDuplicate(const std::vector<std::string>& name) {
   }
   this->clear();
   return has_duplicate;
+}
+
+void HighsNameHash::update(int index, const std::string& old_name,
+                           const std::string& new_name) {
+  this->name2index.erase(old_name);
+  auto emplace_result = this->name2index.emplace(new_name, index);
+  const bool duplicate = !emplace_result.second;
+  if (duplicate) {
+    // Find the original and mark it as duplicate
+    auto& search = emplace_result.first;
+    assert(int(search->second) < int(this->name2index.size()));
+    search->second = kHashIsDuplicate;
+  }
 }
 
 void HighsNameHash::clear() { this->name2index.clear(); }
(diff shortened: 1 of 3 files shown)


metab0t:
@jajhall CI tests pass. It is ready to be reviewed.

src/lp_data/HighsLp.cpp
```diff
@@ -594,14 +594,13 @@ void HighsNameHash::form(const std::vector<std::string>& name) {
...
+    const bool duplicate = !emplace_result.second;
     if (duplicate) {
       // Find the original and mark it as duplicate
-      auto search = this->name2index.find(name[index]);
-      assert(search != this->name2index.end());
+      auto& search = emplace_result.first;
       assert(int(search->second) < int(this->name2index.size()));
-      this->name2index.erase(search);
-      this->name2index.insert({name[index], kHashIsDuplicate});
+      search->second = kHashIsDuplicate;
```
jajhall, line 603: By  deleting

> this->name2index.erase(search);
> this->name2index.insert({name[index], kHashIsDuplicate});

and only setting 

> search->second = kHashIsDuplicate;

(which has no effect since search is immediately destroyed) the property of marking duplicate names is lost. hence errors due to the same name being given to different columns are not trapped.

src/lp_data/HighsLp.cpp
```diff
@@ -618,4 +617,17 @@ bool HighsNameHash::hasDuplicate(const std::vector<std::string>& name) {
...
+void HighsNameHash::update(int index, const std::string& old_name,
+                           const std::string& new_name) {
+  this->name2index.erase(old_name);
+  auto emplace_result = this->name2index.emplace(new_name, index);
+  const bool duplicate = !emplace_result.second;
+  if (duplicate) {
+    // Find the original and mark it as duplicate
+    auto& search = emplace_result.first;
+    assert(int(search->second) < int(this->name2index.size()));
+    search->second = kHashIsDuplicate;
```
jajhall, line 629: Ditto

jajhall (commented):
This may pass the CI tests, but that's because there's no unit test to check that duplicate names are marked as such

src/lp_data/HighsLp.cpp
metab0t: `emplace` will only insert when there is no duplication. When there is duplication, `search` points to the duplicated key and value, and setting the value will influence the original map.

This is an example: https://godbolt.org/z/YsYYrKrPb

metab0t:
> This may pass the CI tests, but that's because there's no unit test to check that duplicate names are marked as such

https://github.com/ERGO-Code/HiGHS/blob/13363c9f1252b015cf6527132eb9cf8f4b5bf020/check/TestNames.cpp#L68-L88

In the beginning, col0_name is unique, so it has corresponding column.

But after col0_name is assigned to `num_col_ / 2` column, it will be duplicate and cannot get column by the name.

src/lp_data/HighsLp.cpp
jajhall: Thanks, it did cross my mind that there was more to setting 

> search->second = kHashIsDuplicate;

than I gave you credit for, sorry.

As I said, I'm not fluent in the use of `std::unordered_map` :-)

Just spotted the `&` in

> auto& search = emplace_result.first;

Now I see why setting 

> search->second = kHashIsDuplicate;

achieves the desired result

metab0t: Never mind. I also learned about the meaning of return values of `emplace` today. It is a clever choice to use `emplace` to detect duplication.

metab0t:
There is still one edge case: if a name is marked as duplicate and the corresponding column is assigned with a new name, then the old name should not be deleted directly (there might be other columns with this old name). However, this is an edge case and checking it would require $O(N)$ iteration through the `col_names_`.

(In practice, I believe that few people will set the name of a variable twice)

The final and complete solution is to use https://en.cppreference.com/w/cpp/container/unordered_multimap to map name to column/row if we want to allow multiple columns/rows to have the same name.

jajhall:
Indeed, there's a limit to making this idiot-proof

PR #1783, opened 2024-05-30
OpenBSD - Update strict_fstream.hpp

ron-at-swgy:
The OpenBSD definition of "strerror_r", similar to FreeBSD, returns an `int` rather than a `char *`. This commit includes `__OpenBSD__` in the list of defines to check for this function call.


ron-at-swgy:
Here is the relevant man page, in case it helps to solidify understanding - https://man.openbsd.org/strerror_r

jajhall:
Thanks. @galabovaa will look at this in due course.

galabovaa:
Looks good to me! Merging in latest.

PR #1792, opened 2024-06-07
Add irreducible infeasibility set (IIS) calculation

jajhall:
Adding irreducible infeasibility set (IIS) calculation is great fun, and a good test of the simplex solvers


jajhall:
@galabovaa Definitely not for the new release! The CI failures that I've created look odd

kbrix2000:
@jajhall Very helpful feature, thank you very much!

Can I try IIS calculation already right now? I tried setting the option iis_strategy but did not see any effect. Is it possible to access IIS calculation from the command line?

jajhall:
> Can I try IIS calculation already right now? I tried setting the option iis_strategy but did not see any effect. Is it possible to access IIS calculation from the command line?

Sorry, no, I added `iis_strategy` in development, but this PR is just allows the `feasibility-relaxation` calculation - for which the first step of IIS calculation is a special case.

kbrix2000:
Thanks for your quick reply. Ok, I see.

Just let me know if you need another beta tester.

PR #1798, opened 2024-06-11
fix wrong declaration of highsCompilationDate(), some compiler warnings, and a comment header

svigerske:
- Move declaration of `highsCompilationDate()` back to original location, as it should not be part of class `Highs` (the implementation is for `highsCompilationDate()` and not `HiGHS::highsCompilationDate()`).
- Make new local functions static to fix missing declaration warnings.
- Fix header
- Avoid comparison of HighsInt and size_t.
- Fix order of initializers in constructor to align to order of member declaration.


galabovaa (approved):

galabovaa:
Thank you for these updates, @svigerske!

PR #1799, opened 2024-06-11
Correctly Report Invalid Row Names

jens-diewald:
Otherwise, HiGHS may silently not write an MPS file when it runs into invalid row names and still return kOK.

I ran into this with a model that has long names with spaces in it.
Presumably a silly typo or copy&paste error.

This is the minimal fix, but some log message indicating what went wrong may also be nice.
Maybe it would be even better to just include this in the `construct_names` case in `normaliseNames`. That way, the MPS would still be written with constructed names, a warning message would be printed and kWarning would be returned.
What do you think?


jajhall:
Thanks for spotting this. I'll look at this soon and consider your suggestion

PR #1809, opened 2024-06-21
Uninitialized member variables

fwesselm:
When compiling on Windows with the `/W4` flag, Visual Studio complains about uninitialized member variables:

Warning	C26495	Variable is uninitialized. Always initialize a member variable (type.6).

I fixed some of these by updating / adding constructors. 

The fact that some member variables are uninitialized has not created any issues. As expected, a testing run with 800+ problems showed that HiGHS' behavior is not affected by the code changes in this PR.


jajhall:
Thanks. Presumably all the values are initialised elsewhere, otherwise valgrind would fail

PR #1817, opened 2024-06-27
Amended conditional policy setting in CMake file (fixes #1816)

### LINKED ISSUE #1816: CMake configuration fails with versions >3.23.0 and <3.24.0
A condiotional policy setting in the CMake file makes builds fail with the above mentioned versions of CMake.
Policy CMP0138 was introduced in version 3.24.0 (https://cmake.org/cmake/help/latest/policy/CMP0138.html) but the policy is specified in: 
```
    if(CMAKE_VERSION VERSION_GREATER  "3.23.0") 
      cmake_policy(SET CMP0138 NEW)
    endif()
```
making the build fail with CMake 3.23.3 (which is the default in manylinux 2010).
I will submit a pull request shortly with the simply fix.

Great work, by the way!

mapgccv:
Fix for #1816

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -189,7 +189,7 @@ if (BUILD_CXX)
     set(ipo_supported NO)
     message(STATUS "IPO / LTO not currently supported building HiGHS on MinGW")
   else()
-    if(CMAKE_VERSION VERSION_GREATER "3.23.0") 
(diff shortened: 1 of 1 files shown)


galabovaa:
Looks good to me!

galabovaa (approved):

PR #1829, opened 2024-07-01
Assertion failure on neos4

fwesselm:
When solving problem `neos4` on `win64` using `latest` I get the following assertion failure:

`Assertion failed: std::fabs(double(activitymax_[mip->a_matrix_.index_[i]] - tmpmaxact)) <= mipsolver->mipdata_->feastol, file L:\HiGHS\myHiGHS\src\mip\HighsDomain.cpp, line 1685`

Updating bounds on row activities using `HighsCDouble` fixes this issue.


jajhall (approved):
Thanks @fwesselm

Issue #1811, opened 2024-06-23
`Highs::writeModel("")` uses `HighsIO`, but `Highs::writeSolution("")` uses `fprintf` for stdout


jajhall:
Hence logging callback cannot be used to redirect `Highs::writeSolution("")`

jajhall:
Closed by #1840

PR #1840, opened 2024-07-10
Fix 1811

jajhall:
Replaced `fprintf(file` in `HighsModelUtils.cpp` by `stringstream`s and calls to `highsFprintfString(file, log_options`


jajhall:
This will close #1811

PR #1848, opened 2024-07-18
Update substitutions if a column is transformed

### LINKED ISSUE #1578: Model status goes from infeasible to optimal when shutting down presolve
## Description

The attached model is infeasible when run with a presolve and optimal when run without a presolve. This only happens when I lower mip_feasibility_tolerance. Is this a bug or would it be expected somehow? Thanks

(note: attached as txt to be uploaded)
[model_min_custo_1.txt](https://github.com/ERGO-Code/HiGHS/files/13924824/model_min_custo_1.txt)

```python
import highspy

for presolve in ['on', 'off']:
    h = highspy.Highs()
    h.setOptionValue('log_to_console', False)
    filename = 'model_min_custo_1.mps'
    h.readModel(filename)
    h.setOptionValue('mip_feasibility_tolerance', 1e-7)
    h.setOptionValue('presolve', presolve)
    h.run()
    print(f'Model with presolve {presolve} has status {h.getModelStatus()}')
```

gives 

```console
Model with presolve on has status HighsModelStatus.kInfeasible
Model with presolve off has status HighsModelStatus.kOptimal
```

## System Information

- highspy version: 1.5.3 (with pip install)
- python version: 3.9.16
- operating system: win 11

fwesselm:
If a column is transformed in presolve (complemented and/or shifted), variable bound constraints are updated accordingly. However, substitutions that may be affected are currently not updated. I think this is a bug, and I fixed this.

This should partly fix #1578, in particular the incorrect answer on `Civ4.100.turns` (mentioned [here](https://github.com/ERGO-Code/HiGHS/issues/1578#issuecomment-2223324400)). 

I did not do large-scale testing so far. @jajhall, I am away for the next 2 weeks and I could do the testing when I return.

### DIFF
--- a/src/mip/HighsImplications.h
+++ b/src/mip/HighsImplications.h
@@ -109,6 +109,7 @@ class HighsImplications {
               double vlbconstant);
 
   void columnTransformed(HighsInt col, double scale, double constant) {
+    // Update variable bounds affected by transformation
     if (scale < 0) std::swap(vubs[col], vlbs[col]);
 
     auto transformVbd = [&](HighsInt, VarBound& vbd) {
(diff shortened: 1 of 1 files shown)


jajhall:
Well done @fwesselm ! Have a good break

PR #1851, opened 2024-07-22
Fix a broken link.

Mathemalsky:
The link to documentation subdirectory was broken. Probably because target branch `docsy` was deleted. I updated the Link to point to master branch.


jajhall:
Th

PR #1861, opened 2024-07-29
Fix error reading fixed format .mps files

### LINKED ISSUE #1860: Error reading Fixed format MPS files
Highs currently fails to read the MIPLIB 2017 collection instance "rvb-sub".
The issue appears to be that the free format incorrectly classifies the file as not fixable, even though the fixed format mps reader can read it just fine. In particular, the MPS file starts with: 

```
NAME          rvb-sub
ROWS
 E  s0        $t5                  0

```
but the last line is incorrectly classified as not being interpretable in fixed format by the free format reader.

I will submit a PR that fixes this soon.

rolfvdhulst:
Fixes #1860

Fixes a bug where the free format mps reader incorrectly classifies a fixed format mps file as not readible.


jajhall:
See comment in #1860

PR #1867, opened 2024-08-06
Add Python type stub file

### LINKED ISSUE #1866: Python type stubs for native extensions
The native extensions in the `highspy` package are currently untyped, which makes it difficult for many editors/language servers to discover available methods and data fields. In particular, in VS Code, I get `Any` as the return type for almost every method call into the native extensions in `_core`.

I can work around this for the most part by following the documentation, but not being able to discover the interface of many objects from my IDE makes for a frustrating development experience.

The canonical way to solve this with Python is via [type stub files](https://peps.python.org/pep-0484/#stub-files), in this case a `_core.pyi` file. They can be generated relatively easily using [mypy's stubgen](https://mypy.readthedocs.io/en/stable/stubgen.html) and then updated as needed. Looking at the bindings, this should not be too much work to set up and maintain. I could have a go at this if there's any interest?

Edit: `stubgen --parse-only -o . -m _core` seems to work. I need to check the output a bit and then I'll open a PR.

N-Wouda:
This PR closes #1866. It adds a Python type stub file that covers the content of the `_core` native extensions.

I generated these stubs using mypy's stubgen, as follows:
```shell
stubgen --parse-only -o . -m _core
```
I then checked the stub file for any errors, and although it can always be improved, I think this'll suffice as an initial default. It's certainly enough to provide Python users with meaningful type information and autocompletion when working with `highspy`.


N-Wouda:
One thing I'm not entirely sure of is how to keep the type stubs in sync with the actual extension going forward. In my own projects I write these files largely myself and thus also keep them up to date, but I can imagine you might not want to spend a lot of (human) time with this. In that case doing stub generation during Python package builds could be worth investigating?

galabovaa (approved):
Thank you! We will keep in mind to update this when we update highspy, and, as you say, it may be worth automating. Looks good to me!

Issue #1856, opened 2024-07-26
Getting incumbent solution in callback (highspy)


sertalpbilal:
I am trying to read the incumbent solution in a user callback whenever an improvement is made in a MIP model. Here is what I have:

```python
import highspy
import numpy as np

hscb = highspy.cb
h = highspy.Highs()
h.readModel('test.mps')

def myfunc(callback_type,
    message,
    data_out,
    data_in,
    user_callback_data):
    if callback_type == hscb.HighsCallbackType.kCallbackMipImprovingSolution:
        print(data_out.mip_solution)

h.setCallback(myfunc, None)
h.startCallback(hscb.HighsCallbackType.kCallbackMipImprovingSolution)
```

It seems like the "mip_solution" is just a numpy array with dimension of 3.
I was trying to investigate why I can't get the solution and have seen this comment here
https://github.com/ERGO-Code/HiGHS/blob/5ce7a27531a7f4166ee5a8343169a1014febb41a/src/highs_bindings.cpp#L1179

So, I was wondering if there is either a way to get the current solution by fixing the issue in this file, or at least a way to export the current solution to a file.

Thanks
-Sertalp

jajhall:
This is related to #1857 raised by @thell

Hello!

Consider the following options being passed to HiGHS:

"mip_improving_solution_save=on",
"mip_improving_solution_file=highs_improved_solution_{max_cost}_c.sol",

I can see that they are being accepted because a highs_improved_solution file was created and saved the first of the improved solutions.

That was ~6.5hrs ago. Since then over a dozen improvements have taken place yet none of them got saved. Much to my dismay because I just hit ctrl-c to copy something but my focus was on my terminal 🤦‍♂️ The real sad part was that I had realized this was taking place and the ctrl-c I used was to copy the options pasted above for this ticket. Anyhow, after getting down to a gap of 1.72% and 75.01% explored in those 6.5hrs I really would like to have seen the last incumbent.

PR #1882, opened 2024-08-17
python: Provide incumbent solution in callback

few:
An accessor for the incumbent solution was already present in the callback data in highspy. But as pointed out in #1856 it was not properly implemented. This PR fixes that.


few:
I updated the commit to fix the formatting issue and made a small edit to examples/call_highs_from_python.py

galabovaa:
This should be addressed in [PR #1942](https://github.com/ERGO-Code/HiGHS/pull/1942), which was just merged in latest. There are significant updates to highspy, including callbacks. 

Please, if possible, pull the updates from the latest branch before we can review/approve this.

PR #1884, opened 2024-08-18
highspy: Expose new feasibility relaxation feature

few:
This adds python bindings for the things introduced in #1876.

* I adapted some of the C++ tests as examples in examples/call_highs_from_python.py.
* src/highspy/_core.pyi contains lots of unrelated changes. Not sure why. I used the following command to generate it: stubgen --parse-only -o . -m highspy._core


jajhall (commented):
Thanks for this. The `_core.pyi` file is new, and wasn't written by us, but by @N-Wouda in response to #1866, for which the PR is #1867 

Perhaps @N-Wouda would like to comment, as we're not in a position to asses changes to this file.

N-Wouda:
@jajhall @few certainly! From the updated interface file I gather that the Python version is likely 3.8, since the use of `list` rather than `typing.List`, etc. for generic typing is supported since Python 3.9 onwards. Since `highspy` supports 3.8 and up - something I missed when I wrote this file - these changes are probably a good thing to have.

The changes to `__getstate__` and `__setstate__` have to do with `pickle`/serialisation support, and are something you get for free on enums. No worries there. This is often not documented, because the only way one would interact is indirectly, via `pickle`. The other changes also look innocuous and might be due to a different version of stubgen/mypy being used (I used mypy 1.11.1, the latest, with Python 3.11).

On the whole I don't see anything of concern here!

PR #1897, opened 2024-08-27
Initialize some arrays

fwesselm:
Valgrind complains about uninitialized values, for example, used by `snprintf`. These changes initialize some arrays to fix this.


jajhall (approved):
Fine

Issue #1900, opened 2024-08-28
Enable QP handling in C# and tidy up the use of HighsModel


jajhall:
The C# interface cannot handle QPs, so some mechanism needs to be added. 

I can add a C# call to `Highs::passHessian` that allows a Hessian term to be added to the incumbent LP

In C++ the Hessian is held as an instance of the `HighsHessian` class, and a QP is represented as an instance of the `HighsModel` class - that consists of a `HighsLp` and a `HighsHessian` instance.

In C# there is a "HighsModel" object that corresponds to (a subset of) the data members of the `HighsLp` class, which is a bit of a mess. This should be sorted so that the C# there is a "HighsModel" object corresponds to the C++ `HighsModel`  class, and something else (in C#) relates to `HighsLp`  class.

But this would break the C# API...

@Thiago-NovaesB: are you in a position to advise/help us on this

Thiago-NovaesB:
I believe that exposing the internal structures of HiGHS via API is not a good idea.
I really like the API created by SCIP, I recently made an interface in C# for it and we can use some ideas from there.
SCIP exposes pointers to its structures via API. The API methods receive these pointers. So these structures do not need to be replicated in each API. Some examples:

The Create method just give us the pointer to the main scip struct:
`private static partial ScipRetCode SCIPcreate(ref IntPtr scip);`
`// SCIP_RETCODE SCIPcreate(SCIP** scip)`

The get solution receive just the main struct pointer and returns the pointer to the solution struct:
`private static partial IntPtr SCIPgetBestSol(IntPtr scip);`
`// SCIP_SOL* SCIPgetBestSol(SCIP* scip)`

The get the value of a variable in the solution, just pass the main struct pointer, solution pointer and variable pointer:
`private static partial double SCIPgetSolVal(IntPtr scip, IntPtr sol, IntPtr v);`
`// SCIP_Real SCIPgetSolVal(SCIP* scip, SCIP_SOL* sol, SCIP_VAR* var)`

Thiago-NovaesB:
Do you think that we can do the same in HiGHS? Share the pointers to avoid replicate structs?

jajhall:
If I understand correctly, 

- In principle this would make the HiGHS C# API similar to its C API, where the parameter lists consist only of scalars and pointers to vectors.
[...]

PR #1902, opened 2024-08-28
Fixing #1900

jajhall:
Added public enum HessianFormat, public class HighsHessian, private static extern int Highs_passHessian, private static extern int Highs_passModel and private static extern int Highs_getHessianNumNz to highs_csharp_api.cs


LatonCypher:
Thanks @jajhall  this will really help my work

PR #1908, opened 2024-09-04
Some minor changes

fwesselm:
- Use `std::array` instead of C arrays
- Character strings produced by `snprintf` and `vsnprintf` are null-terminated, so no need to check
- Fix typo in `src/parallel/HighsSplitDeque.h`
- Testing on 400+ MIP instances revealed no issues


jajhall (approved):
All good with me. I'll merge into a new branch so that I can fix the formatting failures

fwesselm:
Maybe restarting `test-clang-format` would help, it looks like it did not run.

PR #1911, opened 2024-09-06
Precompute expensive computations

heshpdx:
Analysis showed some easy opportunities to gain performance through replacing expensive FP divides with FP multiplies. We can loft the computation outside the hot loops and hold inverse values instead.  This way we don't incur the cost of the divide on every loop iteration. Most CPUs are much slower on FP divides compared to FP multiplies.

This technique already exists in the codebase, for example:
https://github.com/ERGO-Code/HiGHS/blob/5ce7a27/src/util/HFactor.cpp#L726

I measured +4.5% performance with this patch, when solving  `i_n13.mps` or `netlarge.mps` from https://plato.asu.edu/ftp/lptestset/network/.

PR #1914, opened 2024-09-06
Replace expensive computations

heshpdx:
Replace hot FDIVs with FMULs through precomputing inverse values outside the loops. FDIVs are always more expensive than FMULs, although the amount varies per CPU microarchitecture.

This is a new commit, since I don't know how to revert a reverted commit.

### DIFF
--- a/src/simplex/HEkk.cpp
+++ b/src/simplex/HEkk.cpp
@@ -2124,6 +2124,7 @@ void HEkk::updateDualSteepestEdgeWeights(
 
   const double col_aq_scale = simplex_nla_.variableScaleFactor(variable_in);
   const double col_ap_scale = simplex_nla_.basicColScaleFactor(row_out);
+  const double inv_col_ap_scale = 1.0 / col_ap_scale;
 
   const bool DSE_check = false;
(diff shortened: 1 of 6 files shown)


heshpdx:
I added some more data here: https://github.com/ERGO-Code/HiGHS/pull/1911#issuecomment-2338591502

What are the next steps? Should I open a new PR to master?

jajhall:
> I added some more data here: [#1911 (comment)](https://github.com/ERGO-Code/HiGHS/pull/1911#issuecomment-2338591502)

Noted, thanks. This looks good. 

> What are the next steps? Should I open a new PR to master?

No, I'll test the `consider-1911` branch on my desktop now that I'm back in the UK

Note that we only push to `master` from `latest`, and then only when we create a release, so PRs should be made to `latest`.

jajhall:
So, this gives a 2.5% performance improvement on the 32 Mittelmann LPs that HiGHS dual simplex can solve in 3600s.

I'm surprised, as I'd have expected compiler optimization to pre-compute the reciprocals.

Thanks!

Issue #1915, opened 2024-09-09
Add documentation page listing and naming available solvers?


pierre-haessig:
Hello,

I wonder whether it may be useful to add a documentation page listing the available solvers in HiGHS. In particular, this may clarify some aspects:
- provide fully developed names to the short names (`simplex`, `ipm` and `pdlp`) used in the `solver` setting option https://ergo-code.github.io/HiGHS/dev/options/definitions/#solver
- is the "Dual simplex" mentioned on the  Parallelism page (https://ergo-code.github.io/HiGHS/dev/parallel/#Dual-simplex) the same as `pdlp`? (seems yes from previous [discussion](https://github.com/ERGO-Code/HiGHS/discussions/1836#discussioncomment-9958225))
  - if yes, add the discussion comment that it is meant for GPU usage?
- finally (but this is a separate issue), a parallel interior point method is mentioned as future plans https://ergo-code.github.io/HiGHS/dev/parallel/#Future-plans. Is it indeed a work in progress?

jajhall:
> I wonder whether it may be useful to add a documentation page listing the available solvers in HiGHS. In particular, this may clarify some aspects:

There is already some information on [Solution-algorithms](https://ergo-code.github.io/HiGHS/dev/#Solution-algorithms)

> * provide fully developed names to the short names (`simplex`, `ipm` and `pdlp`) used in the `solver` setting option https://ergo-code.github.io/HiGHS/dev/options/definitions/#solver

This isn't necessary, as these are three wholly distinct solution techniques for LP. Variants within these techniques are 

> * is the "Dual simplex" mentioned on the  Parallelism page (https://ergo-code.github.io/HiGHS/dev/parallel/#Dual-simplex) the same as `pdlp`? (seems yes from previous [discussion](https://github.com/ERGO-Code/HiGHS/discussions/1836#discussioncomment-9958225))
>   
>   * if yes, add the discussion comment that it is meant for GPU usage?

No, "Dual simplex" is a variant of the simplex algorithm, but "PDLP" is a first order method. 

I've realised that [Solution-algorithms](https://ergo-code.github.io/HiGHS/dev/#Solution-algorithms) doesn't say anything about what PDLP is, so I'll add that, stressing that PDLP is currently only available for CPUs, and may be totally uncompetitive.

[...]

PR #1916, opened 2024-09-10
Fix 1915

jajhall:
Added the page `solvers.md`, with a little information on what's available and how to override the default behaviour


jajhall:
Fails due to [R](https://cran.r-project.org/) not loading

This will close #1915

PR #1918, opened 2024-09-11
Add function to compute fractional part

fwesselm:
- Add function `fractionality` to compute fractional part (to reduce amount of duplicate code)
- Fix some typos along the way
- Testing on 400+ MIPs revealed no regressions
- If the new function should live somewhere else (i.e. not in `src/util/HighsUtils.h`), let me know.


jajhall (approved):
Thanks. Looks fine to me

PR #1920, opened 2024-09-11
Fix 1860

jajhall:
WIP


jajhall:
MPS read now erases trailing comments - so `rvb-sub` is read, and much duplicated code is removed by using `getMpsLine` rather than `getline`

This closes #1860

PR #1922, opened 2024-09-12
Clang format update

galabovaa:


jajhall (approved):
Thanks. Hopefully there won't be merge conflicts with work being done in other branches. Looking through , I think that the chance is small

PR #1940, opened 2024-09-20
Tsan test bool

galabovaa:
Address hanging issue. Fix by @mathgeekcoder with minor modifications 

Add sanitizer tests: WIP 

Add workaround for cmake regex error


src/parallel/HighsTaskExecutor.h
```diff
@@ -30,9 +30,12 @@ class HighsTaskExecutor {
  public:
   using cache_aligned = highs::cache_aligned;
   struct ExecutorHandle {
-    cache_aligned::shared_ptr<HighsTaskExecutor> ptr{nullptr};
+    HighsTaskExecutor* ptr{nullptr};
+    bool isMain{false};
+    bool isStopped{false};
```
mathgeekcoder, line 35: I would move `isStopped` into `HighsTaskExecutor` and wrap it in an `std::atomic<bool>`.

src/parallel/HighsTaskExecutor.h
```diff
@@ -85,47 +90,75 @@ class HighsTaskExecutor {
...
-      if (currentTask != nullptr) continue;
-
-      currentTask = workerBunk->waitForNewTask(localDeque);
+  static void run_worker(int workerId, HighsTaskExecutor* ptr) {
+    auto& executorHandle = threadLocalExecutorHandle();
+    executorHandle.ptr = ptr;
+
+    // check if main thread has shutdown before thread has started
+    // if (ptr->mainWorkerId.load() != std::thread::id()) {
+    if (!(executorHandle.isMain)) {
```
mathgeekcoder, line 99: This should be a check to the shared `isStopped`, i.e., have we stopped before the thread has started.  The `isMain` check is not needed.  

I was using the `mainWorkerId == thread::id` to implicitly check if we had stopped - the explicit `isStopped` is much better!

src/parallel/HighsTaskExecutor.h
```diff
@@ -85,47 +90,75 @@ class HighsTaskExecutor {
...
+        localDeque->runStolenTask(currentTask);
+
+        currentTask = ptr->random_steal_loop(localDeque);
+        if (currentTask != nullptr) continue;
+
+        currentTask = ptr->workerBunk->waitForNewTask(localDeque);
+      }
     }
+
+    threadLocalExecutorHandle().dispose();
```
mathgeekcoder, line 114: Since we have the `executorHandle` above, it might be nicer to use it instead, i.e., `executorHandle.dispose();`

src/parallel/HighsTaskExecutor.h
```diff
@@ -85,47 +90,75 @@ class HighsTaskExecutor {
...
+      workerThreads.emplace_back(
+          std::move(std::thread(&HighsTaskExecutor::run_worker, i, this)));
+    }
+  }
+
+  void stopWorkerThreads(bool blocking = false) {
+    // auto id = mainWorkerId.exchange(std::thread::id());
+    // if (id == std::thread::id()) return;  // already been called
+    auto& executorHandle = threadLocalExecutorHandle();
+    if (executorHandle.isStopped) return;
```
mathgeekcoder, line 141: Need to "safely" only call stop once. i.e., 
```cpp
if (executorHandle.ptr == nullptr || executorHandle.ptr->isStopped.exchange(true)) 
   return;
```

This will set `isStopped` to `true` and return the previous result.  If it had already been set to `true` we will exit, otherwise we'll proceed to shut things down.  This should be guaranteed to avoid race conditions and only shut down once.

src/parallel/HighsTaskExecutor.h
```diff
@@ -85,47 +90,75 @@ class HighsTaskExecutor {
...
+
+    // now inject the null task as termination signal to every worker
+    for (auto& workerDeque : workerDeques) {
+      workerDeque->injectTaskAndNotify(nullptr);
+    }
+
+    // only block if called on main thread, otherwise deadlock may occur
+    // if (blocking && std::this_thread::get_id() == id) {
+    // threadLocalExecutorHandle();
+    if (blocking && executorHandle.isStopped) {
```
mathgeekcoder, line 151: Need to check `isMain` instead of `isStopped`.  If we are within a worker thread the call to `.join()` doesn't make sense, i.e., we are joining on ourselves?

According to docs it would throw `resource_deadlock_would_occur` if `workerThread.get_id() == std::this_thread::get_id()` (deadlock detected).

src/parallel/HighsTaskExecutor.h
```diff
@@ -85,47 +90,75 @@ class HighsTaskExecutor {
...
+      for (auto& workerThread : workerThreads) {
+        workerThread.join();
+      }
+    } else {
+      for (auto& workerThread : workerThreads) {
+        workerThread.detach();
+      }
+    }
+
+    if (executorHandle.isMain) executorHandle.isStopped = true;
```
mathgeekcoder, line 161: This line is no longer needed if the other `isStopped` changes are made.

src/parallel/HighsTaskExecutor.h
```diff
@@ -138,34 +171,21 @@ class HighsTaskExecutor {
...
     auto& executorHandle = threadLocalExecutorHandle();
-    if (!executorHandle.ptr) {
-      executorHandle.ptr =
-          cache_aligned::make_shared<HighsTaskExecutor>(numThreads);
-      executorHandle.ptr->mainWorkerHandle.store(&executorHandle,
-                                                 std::memory_order_release);
+    if (executorHandle.ptr == nullptr) {
+      executorHandle.ptr = new (cache_aligned::alloc(sizeof(HighsTaskExecutor)))
+          HighsTaskExecutor(numThreads);
+      executorHandle.isMain = true;
```
mathgeekcoder, line 177: This is absolutely fine, but I'd feel more comfortable if `isMain = true` was set before `ptr` was constructed.

src/parallel/HighsTaskExecutor.cpp
```diff
@@ -21,7 +21,19 @@ thread_local HighsTaskExecutor::ExecutorHandle
     HighsTaskExecutor::globalExecutorHandle{};
 #endif
 
-HighsTaskExecutor::ExecutorHandle::~ExecutorHandle() {
-  if (ptr && this == ptr->mainWorkerHandle.load(std::memory_order_relaxed))
-    HighsTaskExecutor::shutdown();
+void HighsTaskExecutor::ExecutorHandle::dispose() {
+  if (ptr != nullptr) {
+    // check to see if we are the main worker and if so, shut down the executor
+    // if (std::this_thread::get_id() == ptr->mainWorkerId.load()) {
```
mathgeekcoder, line 27: Feel free to remove my commented code (here and elsewhere)

src/parallel/HighsTaskExecutor.h
```diff
@@ -52,9 +55,11 @@ class HighsTaskExecutor {
   }
 #endif
 
-  std::vector<cache_aligned::unique_ptr<HighsSplitDeque>> workerDeques;
+  std::atomic<int> referenceCount;
+
```
mathgeekcoder, line 59: insert: `std::atomic<bool> isStopped{false};`  or rename it to `hasStopped`?

mathgeekcoder (changes requested):
Looking good!  I've made some suggestions for changes, they're important but fairly small.

galabovaa:
Closing, changes are now in https://github.com/ERGO-Code/HiGHS/pull/1948

Issue #1905, opened 2024-09-01
Missing callbacks


few:
There are currently three callbacks to interrupt highs:
`kCallbackSimplexInterrupt`
`kCallbackIpmInterrupt`
`kCallbackMipInterrupt`

1) There is no interrupt callback for `presolve`
2) A single "interrupt everything" callback would be nice. That way a user who is only interested in interrupting highs without caring for the solver state (i.e. on keyboard interrupt), wouldn't have to keep track of newly added interrupt callbacks.
3) A "finished" callback, that gets called after the solving process has stopped. Currently one does not know when highs will finish after a callback sets `user_interrupt=true`.

few:
See the EDIT on https://github.com/ERGO-Code/HiGHS/pull/1886#issuecomment-2323191530 for why 3) is a real problem.

jajhall:
OK These are soon added.

jajhall:
If I understand correctly, as well as a blanket "enable all callbacks", you also want a callback that is executed just before the return from `Highs::run()`?

few:
> If I understand correctly, as well as a blanket "enable all callbacks", 

While an "enable all callbacks" might be useful for some, I want an "enable all interrupt callbacks".

> you also want a callback that is executed just before the return from `Highs::run()`?

That sounds right.

jajhall:
>  I want an "enable all interrupt callbacks".

Of course, sorry 

I may be able to implement them this week

Issue #1904, opened 2024-09-01
highspy: Passing user_callback_data to setCallback raises exception


few:
The following code raises an exception (at least on the `latest` branch):

```import highspy

h = highspy.Highs()

def callback(callback_type, message, data_out, data_in, user_callback_data):
    pass

user_callback_data = {}
h.setCallback(callback, user_callback_data)
```

```h.setCallback(callback, callback_data)
TypeError: setCallback(): incompatible function arguments. The following argument types are supported:
    1. (self: highspy._core._Highs, arg0: Callable[[int, str, HighsCallbackDataOut, HighsCallbackDataIn, capsule], None], arg1: capsule) -> highspy._core.HighsStatus

Invoked with: <highspy.highs.Highs object at 0x7f9b12ffa160>, <function callback at 0x7f9b132b7d90>, {}
```
Passing None as user_callback_data works.

jajhall:
This is done so that users can pass data required in a callback. It works fine in C++ (and C, I think). I've got no idea how to make it work in Python

few:
I found a workaround:
```
import highspy

h = highspy.Highs()
h.readModel("30n20b8.mps")
h.setOptionValue("time_limit", 90)

class CallbackManager:
    mip_solution = None

    def callback(self, callback_type, message, data_out, data_in, user_callback_data):
        self.mip_solution = data_out.mip_solution

callback_manager = CallbackManager()
h.setCallback(callback_manager.callback, None)
h.startCallback(highspy.cb.HighsCallbackType.kCallbackMipImprovingSolution)
h.run()
h.stopCallback(highspy.cb.HighsCallbackType.kCallbackMipImprovingSolution)

print("best solution so far:", callback_manager.mip_solution)
```

This way one can store arbitrary data in the CallbackManager object. Maybe user_callback_data is not needed in highspy?

jajhall:
Closed by #1942

Issue #1903, opened 2024-08-30
highspy (direct calls via pybind11) incorrectly handles sliced numpy arrays


mathgeekcoder:
Adding issue for visibility, although I've already fixed the code in #1891.

Consider the following:

```python
h = highspy.Highs()

zero, ones = np.zeros(10), np.ones(10)
h.addCols(10, zero, zero, ones, 0, [], [], [])

x = np.arange(10, dtype=np.int32) # [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
tmp = x[1::2]                     # [1, 3, 5, 7, 9]

h.addRow(1, 1, len(tmp), tmp, ones)
```

We should expect that `[1, 3, 5, 7, 9]` is added as a row, but currently `[1, 2, 3, 4, 5]` is added instead.

The issue is in `highs_bindings.cpp` where we use `py::array_t<T>`.  I've changed this to `py::array_t<T, py::array::c_style | py::array::forcecast>` instead.  This will ensure that the underlying raw data is contiguous and of correct type.

jajhall:
Closed by #1942

Issue #1892, opened 2024-08-24
highspy: Highs.val should work for highs_linear_expression


few:
In my modeling, I'm carrying around several linear expressions where sometimes I'm only interested in the value of the expressions and not the individual variables. It would be nice if that could be supported.

```import highspy
h = highspy.Highs()
x = h.addVariable(ub=3, obj=1)
y = h.addVariable(ub=4, obj=1)
h.maximize()

assert abs(h.val(x)-3) < 1e-6
assert abs(h.val(y)-4) < 1e-6
assert abs(h.val(x+y)-7) < 1e-6 #AttributeError: 'highs_linear_expression' object has no attribute 'index'
```

mathgeekcoder:
This is a simple feature (if we limit to linear expressions).  I've already added support in #1891.  Will commit when I've finished with all the new tests and suitable highspy coverage.

jajhall:
Closed by #1942

PR #1942, opened 2024-09-21
Significant enhancements and stability for highspy

mathgeekcoder:
Major highspy update:
* changed `highs_linear_expression` to be immutable by default
* improved callback support
* improved test coverage (98%)
* performance and usability enhancements
* Support `__iadd__`, `__imul__`, etc.
* Updated chain comparison support in immutable setting
* `h.val()` can take `highs_linear_expression`
* `expr == [lb,ub]` -> `lb <= expr <= ub` syntax
* `qsum`
* added pretty print `__repr__` and `__str__`
* added KeyboardInterrupt support
* added user interrupt
* fixed slicing issues with numpy and highs
* added `resetGlobalScheduler`
* released GIL for `Presolve`
* fixed issues with deadlock on Windows
* fixed MIP solution callback issue
* support `getExpr` that creates a `highs_linear_expression` from existing row
* fixed all issues with static analysis (pyright)
* improved typing and overloads
* added new `HighspyArray` numpy subclass for improved usability

Replaces #1891
Should address multiple issues: #1865, #1882, #1888, #1892, #1903, #1904, and perhaps #1905


src/highspy/_core/simplex_constants.pyi
jajhall, line 1: Some constants in this file are needed by Python users, but many others are for development options that I'd rather not encourage any users to play with, and may confuse them. 

Shall I try to edit them out?

mathgeekcoder: This file was actually auto-generated by [pybind11-stubgen](https://github.com/sizmailov/pybind11-stubgen).  

I believe it should be fine to remove any option from the .pyi file.  It'll still be accessible by the user if they really really want.

jajhall: OK. We'll have them for now, and remove some of them later. I've got other pressing issues and don't want to hold up this PR

galabovaa (approved):

jajhall:
Closes the original #992, #1044 and #1670

few:
There are breaking changes in highspy:
highs_linear_expression had properties LHS, RHS and vars which I used and are now gone. I can easily change my code, just mentioning it in case backwards compatibility is important for you.

few:
You mentioned #1905. I don't see new callbacks in this merge request, and highspy still doesn't handle CTRL+C on its own. Am I missing something?

few:
Regarding #1882: data_out.mip_solution in the kCallbackMipImprovingSolution callbacks is not an array, but prints as this:
<highspy._core.readonly_ptr_wrapper_double object at 0x7f921d7b9af0>

galabovaa:
@few 

> There are breaking changes in highspy: highs_linear_expression had properties LHS, RHS and vars which I used and are now gone. I can easily change my code, just mentioning it in case backwards compatibility is important for you.

Please update your code with the new highs_linear_expression syntax, the highspy modelling language is not very mature yet, and this is indeed one breaking change, but we decided it would be better to have it in this format.

Thank you for your comments, we will look into 1905 and 1882 agian.

jajhall:
Thanks @galabovaa. 

Sorry @few if I was a bit enthusiastic about closing issues, but I'm not in a position to assess Python modifications, and took the lead from the comments in the PR

mathgeekcoder:
@few:

> highs_linear_expression had properties LHS, RHS and vars which I used and are now gone. I can easily change my code, just mentioning it in case backwards compatibility is important for you.

Apologies about the breaking change. I never imagined people would be using the internal LHS/RHS/vars etc.  As @galabovaa mentioned, please update your code.  

BTW: it's likely that you don't need to reference these internals anymore anyhow.

> You mentioned https://github.com/ERGO-Code/HiGHS/issues/1905. I don't see new callbacks in this merge request, and highspy still doesn't handle CTRL+C on its own. Am I missing something?

For Ctrl-C support you'll need to set `h.HandleKeyboardInterrupt = True`.  It is false by default, as it needs to spawn a new thread.

For #1905:  
1. Presolve callbacks are not yet implemented the C++ HiGHS code
2. The python callbacks will automatically enable any callback on subscription.  If you `disableCallbacks`/`enableCallbacks` it will disable/enable all callbacks that have been subscribed - and will remember past subscriptions - unless they are explicitly cleared.
3. The "finished" callback shouldn't be required, as I'm using thread sync objects to avoid any issues.

> Regarding https://github.com/ERGO-Code/HiGHS/pull/1882: data_out.mip_solution in the kCallbackMipImprovingSolution callbacks is not an array, but prints as this:
<highspy._core.readonly_ptr_wrapper_double object at 0x7f921d7b9af0>

Correct. I didn't want to change the internal HiGHS code, but I wanted to give access to the array.  This array is guaranteed to be the size of `h.numVariables`.  You can get a copy of the array via:
```python
copy = data_out.mip_solution.to_array(h.numVariables)
```

But that's not required.  You can also access the values with your variables, e.g.,
```python
x = h.addVariables(...)

def my_callback(e):
   copy = e.val(x)
```

PR #1948, opened 2024-09-24
Bug fix and some tests

galabovaa:
- Fixes deadlock issue on windows
    -  Updates to HighsTaskExecutor
- Adds sanitizers in CMake
    - Linux only so far
    - Workflow WIP 
- Fixes cmake regex error for githash


jajhall (approved):
There's no way that I can make any comment on this!

I have to trust you and Leona!

PR #1961, opened 2024-10-05
Fix some more compiler warnings

fwesselm:
These changes fix some more compiler warnings reported by Visual Studio. 

Testing on 850+ MIPs showed that the changes do not have any effect on HiGHS' behavior.

### DIFF
--- a/src/lp_data/HighsDeprecated.cpp
+++ b/src/lp_data/HighsDeprecated.cpp
@@ -177,7 +177,7 @@ HighsStatus Highs::writeSolution(const std::string& filename,
 }
 #endif
 
-const HighsModelStatus& Highs::getModelStatus(const bool scaled_model) const {
+const HighsModelStatus& Highs::getModelStatus(const bool) const {
   deprecationMessage("getModelStatus(const bool scaled_model)",
                      "getModelStatus()");
   return model_status_;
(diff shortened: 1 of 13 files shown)


jajhall (commented):
Most of this is fine and I'm happy with the checks

However, @galabovaa must approve the changes to `HighsTaskExecutor.h/cpp` as a longstanding bug in them was fixed recently with great difficulty, and testing on Linux/MacOS isn't sufficient. If there's any doubt, then we won't make the changes, as it's vastly more important to have fixed this bug, than eliminate compiler warnings.

fwesselm:
I realize that I made a minor change in `HighsTaskExecutor::ExecutorHandle::dispose()` along the way. The method does nothing when `ptr == nullptr`. So, I thought having `if (ptr == nullptr) return;` would be easier than having an if-block `if (ptr != nullptr) {...}` that basically includes the entire method code. I can revert this, if it's controversial.

galabovaa:
OK, the HighsTaskExecutor is more of a style change. There is a style argument to not have extra "return" statements, unless it really cleans up the code. In this case, your change does not affect performance so we are happy to accept to modify it as you have suggested.

galabovaa (approved):

PR #1963, opened 2024-10-08
Presolve: Add utility to convert variable into an implied integer

fwesselm:
- Add `HPresolve::convertImpliedInteger` to convert a variable into an implied integer variable.
- Testing on 850+ MIPs showed that changes do not affect HiGHS' behavior.


jajhall (approved):
Fine!

Issue #1966, opened 2024-10-10
No infeasibilities for infeasible LP 


jajhall:
In the OR-Tools unit tests, is an infeasible LP for which the HiGHS IPM solver finds no infeasibilities due to excessive primal residuals.

> c = 2 -1

> col_lower = 0 0
col_upper Inf inf

> row_lower -inf, 2
row_upper 1, inf

> a_value
1, -1
1, -1

jajhall:
Closed by #1978?

jajhall:
Closed by #2370

PR #1978, opened 2024-10-15
Fix 1966a

jajhall:
When IPM or PDLP thinks that the LP is solved to optimality, `correctResiduals` identifies whether there are primal or dual residual errors greater than `primal_residual_tolerance` or `dual_residual_tolerance`, and corrects them by adjusting `row_value` or `col_dual` in the solution, possibly introducing primal or dual infeasibilities.


jajhall:
Closes #1966

odow:
Nice nice nice. I'll make sure we document this for BT.

PR #1982, opened 2024-10-16
Correct residuals

galabovaa:


jajhall (approved):
This is fine. I didn't know what model status was wrong in the OR-TOOLS test, and when I first tested `correctResiduals` I got 6 unit tests failure in non-optimal cases. So I restricted the call to `solver_object.model_status_ == HighsModelStatus::kOptimal`.

However, they were due to errors in lines 90-92. Once they were fixed, I see that `correctResiduals` can always be called.  :-)

Hence your fix is fine. I've just updated the comment before the call to `correctResiduals`

PR #1989, opened 2024-10-22
Phase 2 primal simplex being forced inadvisedly

jajhall:
As exposed by #1865, after the objective bound is reached for the scaled LP, primal simplex phase 2 for the unscaled LP was being forced when there will be large primal infeasibilities. Whilst not strictly incorrect, it can be a very bad idea,

It should be that primal simplex phase 2 for the unscaled LP is forced after the objective `_target_` is reached, since the primal infeasibilities should be small. 

After the objective bound is reached for the scaled LP, _dual_ simplex phase 2 for the unscaled LP should be forced, since the dual infeasibilities should be small.

### DIFF
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -2,35 +2,6 @@
 
 ## Code changes
 
-Added `int64_t mip_total_lp_iterations` to `HighsCallbackDataOut` and modified accessor function
-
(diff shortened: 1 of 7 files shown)


jajhall:
Closes #1865

PR #1994, opened 2024-10-23
Update cxxopts

galabovaa:
Update the outdated version of the cpp options library for command line options for the highs executable 
Move the runtime options header out of the library and into the executable directory

old version of cxxopts 2.2.0
new version of cxxopts: 3.2.0


jajhall (approved):
Can't see anything wrong with this

Issue #1990, opened 2024-10-23
Segfault with small time limit


odow:
Originally reported as https://github.com/jump-dev/HiGHS.jl/issues/240

MPS file: [time_crash.mps.txt](https://github.com/user-attachments/files/17484098/time_crash.mps.txt)

Very reproducible in HiGHS v1.7.2 and v1.8.0, on both Windows and Mac, and every time.
```julia
julia> HiGHS_jll.highs() do exe
       run(`$exe time_crash.mps --time_limit 0.1`)
       end
Running HiGHS 1.8.0 (git hash: fcfb53414): Copyright (c) 2024 HiGHS under MIT licence terms
QP   time_crash has 8440 rows; 5871 cols; 24016 matrix nonzeros; 5871 Hessian nonzeros
Coefficient ranges:
  Matrix [2e-01, 1e+03]
  Cost   [8e+02, 7e+03]
  Bound  [4e-03, 1e+03]
  RHS    [2e-03, 1e+00]
  Iteration        Objective     NullspaceDim
ERROR: failed process: Process(`/Users/oscar/.julia/artifacts/6c052094bcce703903faeccc27a66f3453fd061e/bin/highs time_crash.mps --time_limit 0.1`, ProcessSignaled(11)) [0]
```

The Julia log makes it look like is related to `HighsTimerClock` during the `buildSimple` call in factorization:
```julia
[1610] signal (11.1): Segmentation fault: 11
in expression starting at REPL[6]:1
_ZN7HFactor11buildSimpleEv at /Users/oscar/.julia/artifacts/6c052094bcce703903faeccc27a66f3453fd061e/lib/libhighs.1.8.dylib (unknown line)
_ZN7HFactor5buildEP15HighsTimerClock at /Users/oscar/.julia/artifacts/6c052094bcce703903faeccc27a66f3453fd061e/lib/libhighs.1.8.dylib (unknown line)
_ZN5Basis5buildEv at /Users/oscar/.julia/artifacts/6c052094bcce703903faeccc27a66f3453fd061e/lib/libhighs.1.8.dylib (unknown line)
_ZN5BasisC2ER7RuntimeNSt3__16vectorIiNS2_9allocatorIiEEEENS3_I11BasisStatusNS4_IS7_EEEES6_ at /Users/oscar/.julia/artifacts/6c052094bcce703903faeccc27a66f3453fd061e/lib/libhighs.1.8.dylib (unknown line)
[...]

PR #1995, opened 2024-10-23
Fix 1990

jajhall:
Now handles interrupts and timeouts, ensuring that the time limit is positive


jajhall:
This will close #1990

PR #1997, opened 2024-10-24
fix 2 compiler warnings from 1.8.0

svigerske:
Fixed order of initializers in `HighsOptionsStruct` constructor.

Avoid `std::move` on temporary  `std::thread`, which gave a clang warning.
The idea of `emplace_back` is that one can construct the object that is placed into the `workerThreads` container directly in that container, so one can directly pass the arguments for the constructor of the object to be constructed. (@fwesselm ok?)


jajhall (commented):
Fixing the order of initializers in the `HighsOptionsStruct` constructor is obviously fine.

These modifications were also spotted by @HaoZeke, but I'm wary of making the other in case it compromises the behaviour of the task scheduler, since @lgottwald and @galabovaa have only just fixed a serious issue in it.

I need to see a clear explanation as to why the task scheduler is unchanged before this PR will be approved.

Also, we don't accept PRs to `master`.

svigerske:
If it were
```cpp
workerThreads.emplace_back(
          std::thread(&HighsTaskExecutor::run_worker, i, this));
```
then it would create a `std::thread` object A, where the constructor is called with arguments `&HighsTaskExecutor::run_worker, i, this`. Then it would extend the space in the `workerThreads` container to hold another `std::thread` object. To initialize this object, it would call the `std::thread` copy constructor, passing it object A. Then A would be destroyed.

With the `std::move`, it may put A into the container directly and not destroy A, at least that seems to be the intension of the code.

With the new code, it will extend the space in the `workerThreads` container to hold another `std::thread` object. To initialize this object, it will call its constructor with arguments `&HighsTaskExecutor::run_worker, i, this`. The result should be the same, but the code is more straight forward and a compiler warning on the use of `std::move` is avoided.

Maybe @lgottwald can just review this part.

galabovaa (approved):
What you say makes sense and if @lgottwald is happy I'm happy as well!

svigerske:
```
In file included from ipx_control.cpp:1:
In file included from src/parallel/HighsParallel.h:16:
In file included from src/parallel/HighsMutex.h:18:
src/parallel/HighsTaskExecutor.h:132:11: warning: moving a temporary object prevents copy elision [-Wpessimizing-move]
          std::move(std::thread(&HighsTaskExecutor::run_worker, i, this)));
          ^
src/parallel/HighsTaskExecutor.h:132:11: note: remove std::move call here
          std::move(std::thread(&HighsTaskExecutor::run_worker, i, this)));
          ^~~~~~~~~~                                                    ~
1 warning generated.
```

lgottwald:
Ah ok, that means the compiler suggests to not use move semantics because it the copy would be optimized away which is even better than a move and using a move prevents that.Adding std::move around the expression just makes the move explicit and thereby suppresses the warning.Am 24.10.2024 um 09:10 schrieb Stefan Vigerske ***@***.***>:﻿
In file included from ipx_control.cpp:1:
In file included from src/parallel/HighsParallel.h:16:
In file included from src/parallel/HighsMutex.h:18:
src/parallel/HighsTaskExecutor.h:132:11: warning: moving a temporary object prevents copy elision [-Wpessimizing-move]
          std::move(std::thread(&HighsTaskExecutor::run_worker, i, this)));
          ^
src/parallel/HighsTaskExecutor.h:132:11: note: remove std::move call here
          std::move(std::thread(&HighsTaskExecutor::run_worker, i, this)));
          ^~~~~~~~~~                                                    ~
1 warning generated.

jajhall:
Happy to merge this now

PR #2005, opened 2024-10-28
ENH: Update `highspy` bindings and cleanup `meson.build` for SciPy

HaoZeke:
One part of #1987. Basically after #1997 there are still some more warnings needed for SciPy to be happy https://github.com/scipy/scipy/pull/21565; but that'll be in a follow up as per @galabovaa 's suggestion https://github.com/ERGO-Code/HiGHS/pull/1987#issuecomment-2427021065 and.

The changes here are (as seen in each commit message, summarized here):
- Fixes to remove unused variables within `highs_bindings.cpp`
  + Also cleaned up the `export_values` situation with a code comment 
- Adding enums used by SciPy but not currently present in `highspy`:
  + `HighsDebugLevel`
  + `SimplexCrashStrategy`
- Removed `HighsOptionsManager`, which I added explicitly for the SciPy bindings
  + It isn't used by `highspy` internally, so should live in `scipy` anyway

The only other changes (not really enumerated) are those to the `meson.build`; and are essentially just to keep the built libraries up to date and compatible with SciPy.

**Next steps:**
- I'll open a series of PRs, one per warning to be removed (since they can seem intimidating otherwise)
  + If any of them seem dodgy we can close them and I'll just suppress them in the SciPy `meson.build`
- I'll then rebase the SciPy HiGHS fork and have that merged in
  + Unfortunately as noted by @rgommers https://github.com/scipy/HiGHS/pull/68#issuecomment-2425199994 the fork and upstream cannot get much closer
    - The branch history in the fork includes a lot of other commits which aren't relevant so can't be part of a PR to upstream (see https://github.com/scipy/HiGHS/pull/67)


jajhall (commented):
Happy for the Meson build changes to be made, as this is no longer supported by HiGHS

In `src/highs_bindings.cpp`, I'd like @mathgeekcoder to review the removal of `data` from `highs_setCallback`. 

The code to make `enum SimplexCrashStrategy` available to `highspy` should be removed, as there is no longer any code in HiGHS for a simplex crash strategy. I'll delete it from `lp_data/SimplexConst.h` - which, I believe, doesn't need to be included, since the values of `enum SimplexEdgeWeightStrategy` were obtained via `#include "Highs.h"`

mathgeekcoder:
> In `src/highs_bindings.cpp`, I'd like @mathgeekcoder to review the removal of `data` from `highs_setCallback`.

Thanks Julian.  The change to `setCallback` looks good to me!

PR #2009, opened 2024-10-29
Pdlp ctest

galabovaa:
WIP need to switch presolve off 

add instance tests for pdlp to ctest 

slightly different objectives on platforms / architectures


galabovaa:
replaced by https://github.com/ERGO-Code/HiGHS/pull/2018

PR #2010, opened 2024-10-29
Update GitHub macOS runner 

galabovaa:
macos-12 is being deprecated, update to 13 for Intel and 14 for M series


jajhall (approved):
Happy Mac users with this!

PR #2018, opened 2024-10-31
Add cuPDLP to ctest on all OS

galabovaa:
add instance tests for pdlp to ctest

slightly different objectives on platforms / architectures

presolve off


galabovaa:
Failure is documentation and unrelated, link check for the IPM paper is failing 

_Implementation of an interior point method with basis
preconditioning_, Mathematical Programming Computation, 12, 603-635, 2020. [DOI:
10.1007/s12532-020-00181-8] 

https://link.springer.com/article/10.1007/s12532-020-00181-8

Issue #2016, opened 2024-10-30
HiGHS misreads MIPLIB solution files


jajhall:
HiGHS is designed to read only its own solution file format. However, when given a MIPLIB solution file, it doesn't fail! It reads in most of the values. 

HiGHS should identify solution files beginning =obj= as being in the MIPLIB format and read them as such. This  can provide the user solution file format of #1971

jajhall:
Closed by #2020

PR #2020, opened 2024-11-01
Supportcase22

jajhall:
Rather than misreading a MIPLIB solution file (without reporting failure!), HiGHS now identifies solution files beginning `=obj=` as being in the MIPLIB format and reads them as such.


jajhall:
This will close #2016

Issue #2021, opened 2024-11-01
HiGHS doesn't check time limit often enough in presolve


jajhall:
Find out why!

jajhall:
Closed by #2022

PR #2022, opened 2024-11-01
Added time limit check to HPresolve::strengthenInequalities

jajhall:
This may be an edge case of presolve taking much too long to time out (because it's exposed when a problem has dense rows that aren't too dense) but HPresolve::strengthenInequalities now has a time limit check


jajhall:
This will close #2021

PR #2023, opened 2024-11-03
BLD: Silence all warnings in SciPy

HaoZeke:
I thought about silencing these only at the SciPy end, but on whole I think it makes sense to be explicit here, since not having a clean build (w.r.t warnings) is a project level issue.

This is the last (mandatory) PR for the SciPy update. Warnings fixes can still be ported from #1987 later without blocking the SciPy update.

Please consider merging as soon as the SciPy CI is green : https://github.com/scipy/scipy/pull/21565

Then the only thing left is over at SciPy (I'll just update the fork) and we'll finally have an up-to-date HiGHS!

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -20,17 +20,54 @@ host_system = host_machine.system()
 is_windows = host_system == 'windows'
 is_mingw = is_windows and cc.get_id() == 'gcc'
 
+# Granular flags
+# C++
+_cpp_Wno_invalid_offsetof = cppc.get_supported_arguments('-Wno-invalid-offsetof')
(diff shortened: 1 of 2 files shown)


HaoZeke:
The UCRT64 failure shouldn't be related to this PR (since none of the sources have changed, and AFAIK the CI run doesn't even use `meson`)

jajhall (approved):
Fine by me

jajhall:
> The UCRT64 failure shouldn't be related to this PR (since none of the sources have changed, and AFAIK the CI run doesn't even use `meson`)

CI test failures like this are due to GitHub, and pass on re-running (as I'm now trying)

PR #2040, opened 2024-11-15
Fix incorrect bound checks

fwesselm:
Fix incorrect signs when comparing with `kHighsInf`.

### DIFF
--- a/src/presolve/HPresolve.cpp
+++ b/src/presolve/HPresolve.cpp
@@ -5568,17 +5568,21 @@ HPresolve::Result HPresolve::detectParallelRowsAndCols(
       // compensating column is integral
       bool checkColImplBounds = true;
       bool checkDuplicateColImplBounds = true;
+      auto isLowerStrictlyImplied = [&](HighsInt col) {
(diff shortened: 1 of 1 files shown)


jajhall (approved):
Interesting. I guess the first case doesn't matter, as rounding -inf gives -inf.

If the second was used in error, I guess it might have shown up spectacularly, but I'd have to follow through the subsequent actions properly

fwesselm:
> Interesting. I guess the first case doesn't matter, as rounding -inf gives -inf.
> 
> If the second was used in error, I guess it might have shown up spectacularly, but I'd have to follow through the subsequent actions properly

I can also run the entire test suite for this change, if you like. Just to see if any problem is affected...

fwesselm:
> > Interesting. I guess the first case doesn't matter, as rounding -inf gives -inf.
> > If the second was used in error, I guess it might have shown up spectacularly, but I'd have to follow through the subsequent actions properly
> 
> I can also run the entire test suite for this change, if you like. Just to see if any problem is affected...

I started testing and will report back tomorrow.

jajhall:
Failure is due to odd behaviour for new CI test

PR #2043, opened 2024-11-18
Treating `|x|<1` as a special case in `floor(HighsCDouble x)` and `ceil(HighsCDouble x)`

### LINKED ISSUE #2041: Error with `ceil` operating on a `HighsCdouble`
For the following, the assert fails.
> #include "util/HighsCDouble.h"
#include <cassert>
int main() {
    HighsCDouble a = 1e-23;
    double ceil_a;
    double double_a;
    ceil_a = double(ceil(a));
    double_a = double(a);
    assert(ceil_a >= double_a);
   // Original assert reported by user
    assert(ceil(a) >= a);
}

I'll investigate and debug, but maybe you're interested to see this @lgottwald?

jajhall:
This will close #2041

### DIFF
--- a/check/CMakeLists.txt
+++ b/check/CMakeLists.txt
@@ -257,9 +257,10 @@ if ((NOT FAST_BUILD OR ALL_TESTS) AND NOT (BUILD_EXTRA_UNIT_ONLY))
         "standgub\; 1.25769949\;"
         )
   elseif(WIN32)
-    # on windows e226 model status is unknown, rel gap e00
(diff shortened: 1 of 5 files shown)


fwesselm:
I was wondering whether it would be worth implementing `ceil` using `floor` as follows:

`friend HighsCDouble ceil(const HighsCDouble& x) { return -floor(-x); }`

jajhall:
Ceil and floor are only a few lines of code, so it's more transparent to have them separate

PR #2061, opened 2024-11-25
Fix compiler warnings from HiGHS 1.8.1

svigerske:
- `possInfRelDiff` had no previous declaration, but it seems it can be made local (static) to the compilation unit
- fix order of constructor arguments to match member definition order


jajhall (approved):
Thanks @svigerske

PR #2063, opened 2024-11-27
MAINT: declare that `highspy._core` supports free-threaded CPython

rgommers:
This allows `highspy` to be imported in a free-threaded CPython interpreter (see [PEP 703](https://peps.python.org/pep-0703/), https://py-free-threading.github.io/) without raising a warning and re-enabling the GIL. We've tested thread-safety of this extension module reasonably well in SciPy now, and everything seems fine on all tested platforms (Linux, macOS, Windows and x86-64/arm64).

The one-line change matches the Pybind11 example at https://py-free-threading.github.io/porting/#declaring-free-threaded-support.

### DIFF
--- a/src/highs_bindings.cpp
+++ b/src/highs_bindings.cpp
@@ -642,7 +642,7 @@ HighsStatus highs_setCallback(
         data.ptr());
 }
 
-PYBIND11_MODULE(_core, m) {
+PYBIND11_MODULE(_core, m, py::mod_gil_not_used()) {
   // To keep a smaller diff, for reviewers, the declarations are not moved, but
   // keep in mind:
   // C++ enum classes :: don't need .export_values()


jajhall (commented):
This appears to be very welcome, but what's your view @galabovaa and @mathgeekcoder ?

mathgeekcoder:
If I'm understanding correctly, we're basically just turning off the GIL for everyone.  I don't quite know the implications, but it sounds good to me - as long as we assume the user isn't wanting to try anything pathological.  We make the same assumption for C++ users etc. anyhow.

I'm guessing if we make this change, we no longer need to disable the GIL in ```run``` and ```presolve``` via ```py::call_guard<py::gil_scoped_release>()```?

rgommers:
> If I'm understanding correctly, we're basically just turning off the GIL for everyone.

Almost. I'd phrase it more like "we avoid re-enabling the GIL for everyone globally when using a `cp313t` free-threaded interpreter". For any other Python interpreter, nothing changes.

> I'm guessing if we make this change, we no longer need to disable the GIL in `run` and `presolve` via `py::call_guard<py::gil_scoped_release>()`?

No, that would affect users of all Python versions, so I would not touch that.

jajhall:
Thanks @mathgeekcoder and @rgommers , I'm happy with this as it is

jajhall:
> Great, thanks for the quick review!

You're welcome. It should be in the next release of HiGHS, which will come out in a couple of weeks

PR #2074, opened 2024-12-06
fix #1947

### LINKED ISSUE #1947: Acknowledge feasible simplex solutions
When the simple solvers are interrupted, indicate whether the primal or dual solution is feasible

quantresearch1:
@jajhall Hello, I was not sure if you want you wanted to have a way to expose the solution status or if you just wanted to log them ? Let me know if this is not what you want, I am happy to change it.


src/highs_bindings.cpp
```diff
@@ -537,8 +537,8 @@ highs_getColsEntries(Highs* h, HighsInt num_set_entries,
                          py::cast(value));
 }
 
-std::tuple<HighsStatus, HighsVarType>
-highs_getColIntegrality(Highs* h, HighsInt col) {
+std::tuple<HighsStatus, HighsVarType> highs_getColIntegrality(Highs* h,
```
quantresearch1: Looks like I messed up the formatting... I believe VSCode might have done that automatically, let me check how to revert. Is there a developer guidelines about the formatting tools used in this project ? Apologies if obvious, I am a python developer and this is my first foray in C++ land

jajhall:
Primal/dual solution status is available from HighsInfo, and can't generally be pulled from the simplex primal/dual solution status, as simplex isn't necessarily used to solve problems - even for LP

We use version 14 of

> clang-format -style=file -i foo.cpp/h/cc/hh

PR #2082, opened 2024-12-13
Simplify for loops

fwesselm:
I accidentally found two for loops in `src/simplex/HEkkDualRow.cpp` that could be rewritten as ranged-based for loops, which would simplify the code, I think.

### DIFF
--- a/src/presolve/HPresolve.cpp
+++ b/src/presolve/HPresolve.cpp
@@ -2050,8 +2050,7 @@ void HPresolve::storeRow(HighsInt row) {
   rowpositions.clear();
 
   auto rowVec = getSortedRowVector(row);
-  auto rowVecEnd = rowVec.end();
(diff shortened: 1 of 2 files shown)


jajhall (commented):
I've never used iterators, so can't comment on the correctness of the change. What's the gain in making the change?

jajhall (commented):
Since both changes occur in lines that are heavily used, any errors would manifest themselves immediately. 

I still wonder what's gained by changing the code in this way

fwesselm:
> Since both changes occur in lines that are heavily used, any errors would manifest themselves immediately.
> 
> I still wonder what's gained by changing the code in this way

Ranged-based for loops were introduced with C++11 (and are used in the presolve code, for example).

The code here iterates over the entire `std::set<HighsInt> freeList`. I think this change makes the code syntactically simpler and easier to read and understand. Of course, there won't be performance gains.

PR #2092, opened 2024-12-21
fix compiler warnings in 1.9.0

svigerske:
- clarify precedence in if condition
- declare local function static


jajhall (commented):
Thanks, there are a couple of other comparisons in the pipeline that I'll declare static

PR #2094, opened 2024-12-25
submit changes necessary for cran

FlorianSchwendinger:
The repo for **R** packages (CRAN) has very strict requirements (enforced by custom checks) which have to be fully filled otherwise the package is removed from CRAN.

The following small changes are required from CRAN currently I apply them at each update. It would be nice if you could apply the changes upstream since it would simplify the maintenance of the **R** package and the changes are quite simple.

1. In the files
- `extern/filereaderlp/reader.cpp` and
- `extern/filereaderlp/reader.hpp`
 CRLF line endings are replaced with CR to address the CRAN warning
 "Found the following sources/headers with CR or CRLF line endings".
  
2. Changed
`void PDHG_PrintUserParamHelper()` to `void PDHG_PrintUserParamHelper(void)` and
`void PDHG_PrintHugeCUPDHG()` to `void PDHG_PrintHugeCUPDHG(void)` 
to address the CRAN warning
"function declaration isn't a prototype"

3. Added a newline at several files to address the CRAN warning
"Found the following sources/headers not terminated with a newline:"


jajhall:
Thanks, I'll look at these in due course, but I'm on holiday without access to a Linux machine

jajhall (commented):
I now see properly what the changes are. The CRLF to CR changes are uncontroversial. Do they come from editing on Windows creating the "^M" that's seen when viewing files on a Linux editor?

I'll leave the PDLP modifications to @galabovaa, but I guess they are fine

FlorianSchwendinger:
Thank you!

> I now see properly what the changes are. The CRLF to CR changes are uncontroversial. Do they come from editing on Windows creating the "^M" that's seen when viewing files on a Linux editor?

Yes exactly.

> I'll leave the PDLP modifications to @galabovaa, but I guess they are fine

You can reproduce the warnings by adding the flag `-Wstrict-prototypes` to the C compiler.
This [link](https://softwareengineering.stackexchange.com/questions/286490/what-is-the-difference-between-function-and-functionvoid) gives detailed information on this topic and references to the corresponding sections in the C and C++ Standard.

galabovaa (approved):
Looks good to me!

Issue #2083, opened 2024-12-13
Expose getRays and basis solve methods to highspy 


jajhall:

jajhall:
Closed by #2102

PR #2102, opened 2025-01-05
Fix 2083

jajhall:
Added basis solve methods to highspy

Added methods to get primal/dual ray and dual unboundedness direction to highspy


jajhall:
Closes #2083

PR #2103, opened 2025-01-06
Bazel update

galabovaa:
remove no bzlmod from bazelrc 
update of highs in the bazel central registry: I am now specified as the maintainer of HiGHS


jajhall (approved):
I can only say "yes"

PR #2107, opened 2025-01-07
Fix 2095

### LINKED ISSUE #2095: Segfault in presolve
With [issue-2095.mps](https://github.com/user-attachments/files/18278073/issue-2095.mps.txt), HiGHS segfaults in presolve.

See https://github.com/scipy/scipy/issues/22188

fwesselm:
Fix #2095:
- Pointers stored in `HighsMatrixSlice` may get invalidated if the coefficient matrix is reallocated (e.g. when non-zeros are added in `HPresolve::addToMatrix`).
- The proposed fix is to directly iterate over the non-zero elements (positions in the matrix) instead of using `HighsMatrixSlice` when the matrix is modified while iterating.


fwesselm:
> Looks correct, and more robust than my fix. I guess you'll regression test it
> 
> Are there other places in presolve where such an issue could be exposed?

Regression testing is ongoing.

I looked for calls to `HPresolve::addToMatrix` while iterating over a `HighsMatrixSlice` in presolve, and I think the two occurrences in this PR are the only problematic ones. I hope I did not overlook something.

fwesselm:
Testing shows that, while HiGHS behavior is affected, the overall results look reasonable.

PR #2112, opened 2025-01-08
Code coverage and python workflows 

galabovaa:
Updated code coverage and a workflow to generate stats
Uploading to CodeCov will be added later
Some python workflow updates


jajhall (commented):
Wow! So impressive!

jajhall (approved):

Issue #2109, opened 2025-01-08
Segfault in sequence of solves


odow:
I fiiiinnally isolated a reproducible example from https://github.com/jump-dev/HiGHS.jl/issues/202

Here's the file [log2.txt](https://github.com/user-attachments/files/18341455/log2.txt) (rename as `log2.mps`)

```c
#include "interfaces/highs_c_api.h"
#include <stdlib.h>

int main(int argc, char *argv[]) {
    void* highs = Highs_create();
    int ret;
    ret = Highs_readModel(highs, "log2.mps");
    ret = Highs_run(highs);
    double DBL_MAX = Highs_getInfinity(highs);
    int index1[] = {160, 210, 211};
    double value1[] = {0.027716058468715117, 0.09140473855744354, -0.14285714401611166};
    ret = Highs_addRow(highs, 0.0, DBL_MAX, 3, index1, value1);
    int index2[] = {168, 210, 212};
    double value2[] = {0.08645287577455307, -0.07110776325022987, -0.14285714401611166};
    ret = Highs_addRow(highs, 0.0, DBL_MAX, 3, index2, value2);
    int index3[] = {176, 210, 213};
    double value3[] = {0.18014292280865904, -0.17598648905497463, -0.14285714401611166};
    ret = Highs_addRow(highs, 0.0, DBL_MAX, 3, index3, value3);
[...]

PR #2113, opened 2025-01-08
Fix 2109

jajhall:
Now running postsolve after model status of LP reduced by presolve is unknown. This fixes the bug exposed by #2109 but means that postsolve is working with a non-optimal solution. Since there's no guarantee that an optimal solution remains so through postsolve, I guess it doesn't matter, and can lead to HiGHS finding a definitive model status for LPs whose reduced LP did not have such a status.


jajhall:
Closes #2109

PR #2116, opened 2025-01-09
BUG: Localize pybind11 bindings to prevent import crashes

HaoZeke:
Fixes https://github.com/scipy/scipy/issues/22257

and is basically an upstream variant of https://github.com/scipy/HiGHS/pull/71

Now this will work:

```python
python dev.py python
(InteractiveConsole)
>>> import scipy.optimize
>>> import highspy
>>> exit()
```

Note that this is not just related to `scipy` but basically any package which needs to import `highspy` along with any other package with bindings built from the HiGHS sources.

Relevant documentation: https://pybind11.readthedocs.io/en/stable/advanced/classes.html#module-local-class-bindings


HaoZeke:
Failure seems unrelated.

PR #2130, opened 2025-01-14
Fix 2129

### LINKED ISSUE #2129: Uninitialised variable causing UB
Hi there,

thanks for your library! We are using this in a project which is compiled with `-fsanitize=undefined` and noticed complaints. I took a look at your tests and they seem to reproduce the issue. I am using Clang 19.1.1 on Ubuntu.
1. Build with `LDFLAGS=-fsanitize=undefined CXXFLAG=-fsanitize=undefined CFLAGS=-fsanitize=undefined cmake -S . -B build -DALL_TESTS=ON && cmake --build build`
2. Run unit tests via `build/bin/unit_tests` (but other tests are affected as well)

Output:
```
/tmp/HiGHS/src/mip/HighsCutGeneration.cpp:1427:29: runtime error: load of value 52, which is not a valid value for type 'bool'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /tmp/HiGHS/src/mip/HighsCutGeneration.cpp:1427:29
/tmp/HiGHS/src/mip/HighsCutGeneration.cpp:1428:35: runtime error: load of value 249, which is not a valid value for type 'bool'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /tmp/HiGHS/src/mip/HighsCutGeneration.cpp:1428:35
===============================================================================
All tests passed (1258621 assertions in 235 test cases)
```
This complaint is about this piece of code: https://github.com/ERGO-Code/HiGHS/blob/bdf95f2e0786914c1bfe15b20e1ba534e8f35d6b/src/mip/HighsCutGeneration.cpp#L1427-L1428
And as far as I can tell, the variables `integralSupport` and `integralCoefficients` are just uninitialised. At least the constructor does not initialise them.

(The above was done with latest master, commit bdf95f2e0786914c1bfe15b20e1ba534e8f35d6b)

-------

I saw #1576 and and #1579, but I am not sure what the state of the workflows are. According to https://github.com/ERGO-Code/HiGHS/actions/workflows/sanitizers-meson.yml, the meson workflow is never run. And https://github.com/ERGO-Code/HiGHS/actions/workflows/sanitizers-cmake.yml says that the workflow definition is incorrect. The message `No event triggers defined in on` sounds like https://github.com/ERGO-Code/HiGHS/blob/bdf95f2e0786914c1bfe15b20e1ba534e8f35d6b/.github/workflows/sanitizers-cmake.yml#L2 is incorrect.

fwesselm:
Fix #2129 
- Initialize `integralSupport` and `integralCoefficients` in `HighsCutGeneration::separateLiftedMixedBinaryCover`, `HighsCutGeneration::separateLiftedMixedIntegerCover` and `HighsCutGeneration::cmirCutGenerationHeuristic`.
- These change will not affect HiGHS' behavior since these variables were only uninitialized when the respective cut generation method failed (and hence `integralSupport` and `integralCoefficients` were not used).

### DIFF
--- a/src/mip/HighsCutGeneration.cpp
+++ b/src/mip/HighsCutGeneration.cpp
@@ -215,6 +215,10 @@ void HighsCutGeneration::separateLiftedKnapsackCover() {
 }
 
 bool HighsCutGeneration::separateLiftedMixedBinaryCover() {
+  // initialize indicators
+  integralSupport = false;
+  integralCoefficients = false;
+
   HighsInt coversize = cover.size();
   std::vector<double> S;
   S.resize(coversize);
@@ -275,6 +279,10 @@ bool HighsCutGeneration::separateLiftedMixedBinaryCover() {
 }
 
 bool HighsCutGeneration::separateLiftedMixedIntegerCover() {
+  // initialize indicators
+  integralSupport = false;
+  integralCoefficients = false;
+
   HighsInt coversize = cover.size();
 
   HighsInt l = -1;
@@ -500,6 +508,10 @@ bool HighsCutGeneration::cmirCutGenerationHeuristic(double minEfficacy,
   using std::floor;
   using std::max;
   using std::sqrt;
+
+  // initialize indicators
+  integralSupport = false;
+  integralCoefficients = false;
 
   double continuouscontribution = 0.0;
   double continuoussqrnorm = 0.0;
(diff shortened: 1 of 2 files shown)


jajhall (commented):
Fine, thanks

fwesselm:
I ran MIPLIB2010, and this confirmed that these changes do not affect HiGHS behavior.

PR #2144, opened 2025-01-21
Update HConfig.h.bazel.in

noajshu:
Use latest release number (1.9.0) and commit hash


galabovaa:
Thank you! Just FYI, From the next release onwards, we will be uploading an archive to reference from bazel as well.

galabovaa (approved):

PR #2153, opened 2025-01-27
fix format of log message

svigerske:
`highsLogUser()` is given `int(this->iis_.col_index_.size()),  int(this->iis_.row_index_.size()), int(num_lp_solved)`, but the format string of the message seems to miss the corresponding `%d`'s.


jajhall (commented):
Thanks

jajhall (approved):
The CI failures are not related to the change. Hopefully @galabovaa can identify what's causing them

PR #2157, opened 2025-01-29
Use && for logical and

fwesselm:
I think that using `&&` for logical `and` in `src/mip/HighsMipSolver.cpp` would make the code clearer.


jajhall (approved):
Absolutely: my error!

PR #2161, opened 2025-01-30
Now only recording written_log_header_ = true if output_flag is true …

jajhall:
Now only recording `written_log_header_ = true` if `output_flag` is true so log header is printed if `output_flag` is first false and later set true

### DIFF
--- a/src/Highs.h
+++ b/src/Highs.h
@@ -1471,7 +1471,7 @@ class Highs {
   bool called_return_from_run = true;
   HighsInt debug_run_call_num_ = 0;
 
-  bool written_log_header = false;
+  bool written_log_header_ = false;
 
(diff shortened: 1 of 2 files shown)


maurerle:
Hi @jajhall
I even see the `Running HiGHS 1.11.0 (git hash: 364c83a): Copyright (c) 2025 HiGHS under MIT licence terms` line when calling Highs (from linopy) using:

`m.solve(output_flag=False, log_to_console=False)`

This was the case with 1.9 as well (which does not include this commit).
But this does not feel right, and is quite cumbersome when using highspy programatically.

jajhall:
I assume that it will always be possible to obtain the logging from HiGHS if desired, otherwise it makes support much more difficult.

PR #2162, opened 2025-01-30
Moved linux/aarch64 build Python wheel build to native Github runners 

merschformann:
# Description

This PR moves the _linux/aarch64_ build of the Python wheels to the recently released [Github _linux/aarch64_ runners](https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/). This greatly simplifies and speeds up the Python wheel build, also, it addresses a current [QEMU bug](https://github.com/pypa/cibuildwheel/issues/2257#issuecomment-2619161281) causing build errors.

## Changes

* Moves linux/aarch64 Python wheels build to native Github runners.
* Harmonizes Python wheel for Linux to `ubuntu-24.04*`.
* Harmonizes Python wheel tests to `cp311` (same as macOS.

## Notes/Questions

* Addresses the [issue at hand](https://github.com/ERGO-Code/HiGHS/actions/runs/13030619124/job/36348899659).
* Is this all that is needed right now? Happy to extend the fix, if necessary.
* This uses `ubuntu-24.04-arm`, since `ubuntu-20.04` is not supported. I feel like the runners/builders should be aligned in their Ubuntu version. I didn't want to incur more change than necessary though. Should I align them?


merschformann:
I attempted to harmonize the Python wheel builds in the workflows a bit more using `ubuntu-24.04*` and `cp311`. I hope this makes sense. Let me know, if something is missing / needs to be changed. :relaxed:

galabovaa:
Thank you, @merschformann! This is great. I think I tested it once manually, and the wheels built on Ubuntu 24 still worked on older versions of Ubuntu, at least the Ubuntu 22 I tested on. Let's merge this as is, and if necessary, later on, we will add the previous versions.

galabovaa (approved):

PR #2170, opened 2025-02-08
Added `kImplicitInteger` to definition of `HighsVarType` for `highspy`

jajhall:


odow:
Question: when should we use `kImplicitInteger` instead of `kInteger`?

jajhall:
Users shouldn't use `kImplicitInteger` when defining models. It's an artefact of presolve. It's a continuous variable that, by implication, must take an integer value. For example, if it is the only continuous variable in an equation with integer coefficients and RHS. The fact that it can only take integer values is exploited in the MIP solver. For example, if a domain placed it in an interval interior to consecutive integer, then the corresponding node would be infeasible.

PR #2176, opened 2025-02-11
Zero out vectors with uplocks and downlocks

fwesselm:
Vectors `uplocks` and `downlocks` are not initialized with zeros before iterating over the matrix elements and computing the locks. Therefore, the number of locks may be incorrect after a restart, for example.


fwesselm:
Testing is ongoing. I will report back once I have the results.

jajhall:
Sorry @fwesselm, I thought this was #2177

fwesselm:
> > Sorry @fwesselm, I thought this was #2177
> 
> @jajhall, since the locks were incorrect previously the fix should affect behavior. I will let you know when I have the test results.

The results look reasonable, average speedup is 2 percent.

PR #2181, opened 2025-02-15
Fix cppcheck performance passedByValue warnings

MaximSmolskiy:


jajhall:
I understand your motivation for making these changes. Passing strings by value is of no great importance, but I see that there are a few vectors that you've picked up on.

I've merged it into a branch rather than latest so I can study your changes and fix the formatting

PR #2188, opened 2025-02-24
Update model_status when checkOptimality fails

odow:
I'm not sure if this is the correct fix, but see https://github.com/jump-dev/JuMP.jl/issues/3947

Without this PR, I get `Model status        : Optimal`:
```raw
Running HiGHS 1.9.0 (git hash: 05db5c231): Copyright (c) 2025 HiGHS under MIT licence terms
Coefficient ranges:
  Cost   [0e+00, 0e+00]
  Bound  [1e+00, 1e+00]
  Iteration        Objective     NullspaceDim
ERROR:   QP solver claims optimality, but with num/max/sum primal(2/1/2) infeasibilities
Model status        : Optimal
HiGHS run time      :          0.00
```
which is very confusing for users.

### DIFF
--- a/src/lp_data/Highs.cpp
+++ b/src/lp_data/Highs.cpp
@@ -3783,8 +3783,13 @@ HighsStatus Highs::callSolveLp(HighsLp& lp, const string message) {
   return_status = solveLp(solver_object, message);
   // Extract the model status
   model_status_ = solver_object.model_status_;
-  if (model_status_ == HighsModelStatus::kOptimal)
-    checkOptimality("LP", return_status);
+  if (model_status_ == HighsModelStatus::kOptimal) {
+    HighsStatus check_status = checkOptimality("LP", return_status);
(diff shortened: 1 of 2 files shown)


src/lp_data/Highs.cpp
```diff
@@ -4024,8 +4034,13 @@ HighsStatus Highs::callSolveMip() {
                                       ? -1
                                       : HighsInt(mip_total_lp_iterations);
   info_.valid = true;
-  if (model_status_ == HighsModelStatus::kOptimal)
-    checkOptimality("MIP", return_status);
+  if (model_status_ == HighsModelStatus::kOptimal) {
+    HighsStatus check_status = checkOptimality("QP", return_status);
```
odow:
```suggestion
    HighsStatus check_status = checkOptimality("MIP", return_status);
```

PR #2199, opened 2025-02-28
MPS read mods

jajhall:
Limited excess printing of warnings in the MPS file by doing so with decreasing frequency, adding reporting of total number of each type of warning issued.


jajhall:
Allowed MPS section names to be used as names for columns, RHS, ranges, bounds etc by only accepting them as section names if nothing follows on the line. 

Doesn't work for  NAME, OBJSENSE, QCMATRIX and QSECTION, since they can be followed by text.

No sane modeller would use "RHS" as a column name, but I've found one that does!

Issue #2185, opened 2025-02-20
HiGHS crashes with assertion failure


zxt5:
HiGHS (build from `latest` branch) crashes when solving [seed.mps.txt](https://github.com/user-attachments/files/18879838/seed.mps.txt) with options_file [highs_options.txt](https://github.com/user-attachments/files/18879843/highs_options.txt)

```
root@b9f8d0525121:/tmp# cat highs_options.txt
presolve=off
run_crossover=off
threads=0
mip_detect_symmetry=true
mip_allow_restart=true
mip_max_start_nodes=291
mip_max_leaves=2147480008
mip_max_improving_sols=2147481624
mip_lp_age_limit=2
mip_pool_age_limit=47
mip_heuristic_effort=0.8312576260210047
iis_strategy=1
mip_rel_gap=0
mip_abs_gap=0
root@b9f8d0525121:/tmp# highs --presolve off --model_file seed.mps --options_file highs_options.txt
Running HiGHS 1.9.0 (git hash: 1fd953d7a): Copyright (c) 2025 HiGHS under MIT licence terms
Set option presolve to "off"
Set option run_crossover to "off"
Set option log_file to "HiGHS.log"
Set option mip_max_start_nodes to 291
Set option mip_max_leaves to 2147480008
Set option mip_max_improving_sols to 2147481624
Set option mip_lp_age_limit to 2
Set option mip_pool_age_limit to 47
Set option mip_heuristic_effort to 0.831258
Set option mip_rel_gap to 0
Set option mip_abs_gap to 0
Set option iis_strategy to 1
MIP  seed has 172 rows; 17 cols; 2922 nonzeros; 13 integer variables
Coefficient ranges:
  Matrix [2e-02, 2e+01]
  Cost   [6e+00, 1e+02]
  Bound  [2e+02, 2e+02]
  RHS    [9e-02, 1e+02]

Presolve is switched off

Solving MIP model with:
   172 rows
   17 cols (0 binary, 13 integer, 0 implied int., 4 continuous)
   2922 nonzeros

Src: B => Branching; C => Central rounding; F => Feasibility pump; H => Heuristic; L => Sub-MIP;
     P => Empty MIP; R => Randomized rounding; S => Solve LP; T => Evaluate node; U => Unbounded;
     z => Trivial zero; l => Trivial lower; u => Trivial upper; p => Trivial point

        Nodes      |    B&B Tree     |            Objective Bounds              |  Dynamic Constraints |       Work
[...]

PR #2203, opened 2025-03-03
Fix 2185

fwesselm:
Fix issue #2185 
- Do not call `RENS` or `RINS` heuristics when the domain they operate on is infeasible (i.e. `mipdata_->domain.infeasible()`) since this creates inconsistencies in the bounds on constraint activities.
- Unfortunately, the reproduction example takes relatively long to solve, so I did not use it as an automated test.
- Testing is ongoing.


jajhall:
This will close #2185

fwesselm:
A different node count is reported for `neos-1112787`. Otherwise, the fix does not seem to affect HiGHS' behavior at all.

PR #2211, opened 2025-03-07
Replace command line parsing library for Cpp11

galabovaa:
cxxopts upgrade introduced one line from C++17 causing warnings. 
CLI11 is only C++11


jajhall (approved):
Fine with me

Issue #2220, opened 2025-03-09
Extend the MPS reader to accept OBJSENSE\n MAXIMIZE, not just MAX (and ditto for MINIMIZE)


jajhall:

jajhall:
Closed by #2221

PR #2221, opened 2025-03-10
Now testing just the first three chatacters of upper_word for "MAX"/"MIN" in MPS read

jajhall:
Allows the lines

> OBJSENSE
 MAXIMIZE

produced by (eg) Xpress to be read, not just

> OBJSENSE
 MAX


jajhall:
This will close #2220

Issue #2231, opened 2025-03-14
Report model dimensions on entry to `Highs::run()`


jajhall:
Currently the only model dimensions that are reported are those following presolve, or on entry to the MIP solver. This can hinder development and communication with users

jajhall:
Closed by #2232

PR #2232, opened 2025-03-14
Fix 2231

jajhall:
Added to `FEATURES.md` 

Removed `reportModelStatsOrError` from `RunHighs.cpp` and created `Highs::reportModelStats()`, which is now called before `Highs::run()` and `Highs::presolve()`


jajhall:
This will close #2231

PR #2241, opened 2025-03-19
Generate release source code artefact

galabovaa:
And updates to READMEs and FEATURES


jajhall (approved):
All in order, and nothing I can think of adding

"Added GPU support for cuPDLP-C" is so understated... :-)

galabovaa:
> "Added GPU support for cuPDLP-C" is so understated... :-)

It is a bit, yes!

PR #2245, opened 2025-03-20
Docs GPU

galabovaa:
Added a page with instructions about building with GPU support.


jajhall (commented):
I think that there should be some technical documentation relating to the tolerances and accuracy of PDLP. I've started adding some

jajhall:
Hard to give a short but precise answer to the tolerances issue, and thinking about it has prompted #2249 

Let's go with this for now

PR #2255, opened 2025-03-25
Minor fix to CLI11 to support UWP builds.

### LINKED ISSUE #2254: highs CLI no longer supports uwp builds (v1.10.0)
CLI11 does not currently support uwp (Universal Windows Platform).  Since v1.10.0 switched to using CLI11, it broke the uwp builds.

I'm just flagging this as an issue - doesn't necessarily need to be fixed - but would be good to check your thoughts.  Worst case, we can hack some support for uwp.

mathgeekcoder:
Fixes #2254.  This change will only impact UWP (Universal Windows Platform) builds (via #ifdefs).

CommandLineToArgvW does not seem to be available for UWP, so I replaced it with __argc and __argv directly.

### DIFF
--- a/app/CLI11.hpp
+++ b/app/CLI11.hpp
@@ -434,6 +434,19 @@ CLI11_INLINE std::vector<std::string> compute_win32_argv();
 namespace detail {
 
 #ifdef _WIN32
+    #ifdef WINAPI_FAMILY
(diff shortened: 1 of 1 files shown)


galabovaa (approved):

galabovaa:
Thank you @mathgeekcoder

PR #2264, opened 2025-03-29
expose multiobjective in highspy

### LINKED ISSUE #2263: highspy is missing HighsLinearObjective and addLinearObjective
Hello,

Is there a plan to add these to highspy in the future ? I was looking into this as part of an issue someone opened in the pyomo project to add multi objective optimization with HiGHs: https://github.com/Pyomo/pyomo/issues/3454

Many thanks

quantresearch1:
Fixes #2263
@jajhall After looking into it, the bindings already exist so it's as simple as exposing them. The unit test looks correct.
I’m still getting familiar with the codebase, so please feel free to suggest any improvements or modifications. I’m happy to adjust anything to better align with the project’s goals and coding standards.

Thank you for your feedback :)


jajhall (commented):
Looks fine to me, but I'm not the Python expert

jajhall:
This will close #2263

galabovaa:
Looks good to me, just needs a sanity check from @mathgeekcoder

mathgeekcoder:
It's a great start. I have some suggestions, but I want to double check something before I say them. I'll do this first thing tomorrow (Monday) morning.

quantresearch1:
> It's a great start. I have some suggestions, but I want to double check something before I say them. I'll do this first thing tomorrow (Monday) morning.

Good to hear. I have noticed I forgot to add a test for clearLinearObjectives, will add when I have a min

mathgeekcoder:
FYI: What I wanted to double check is whether this PR was _actually needed._  From my quick experiments, I can already access the multi-objective interface with highspy 1.10.0.  For example, the provided unit test passes for me without any of the other changes.

Note: Since multi-objective is fairly new to HiGHS, I would like to see a larger PR (I can work on this too) that adds additional support and syntactic sugar.  e.g., `highs.addObjective(3*x1 + x2)`, as well as `highs.getObjectiveValueN(2)` etc.

That said, it is good practice to provide the corresponding type information for `HighsLinearObjective`, as given by this PR.  **I suggest removing:** `highs.addLinearObjective` and `highs.clearLinearObjectives`, since they are only calling the base class and are not needed.  And then I'd be happy to approve this PR.

jajhall:
Sounds as if it's time to give up having highspy shadow the C++ API as much as possible - which was for simplicity and documentation purposes. I can see that Python folk need their own parameter lists and, let's be honest, far more people are using the Python interface directly than the C or C++ interface 😢

src/highspy/_core/__init__.pyi
```diff
@@ -788,6 +788,7 @@ class _Highs:
         indices: numpy.ndarray[typing.Any, numpy.dtype[numpy.int32]],
         values: numpy.ndarray[typing.Any, numpy.dtype[numpy.float64]],
     ) -> HighsStatus: ...
+    def addLinearObjective(self, linear_objective: HighsLinearObjective) -> HighsStatus: ...
```
quantresearch1, line 791: @mathgeekcoder As discussed, I have removed addLinearObjective and clearLinearObjectives from highs.py. However, I think it's worth having them here for the IDEs. In VSCode for example, without this, the code still works but the function is white and not clickable. I would assume it's also important for mypy

galabovaa:
Great, thank you!

Issue #2269, opened 2025-04-02
Ensure that if run-time options are set when `Highs::run()` is called, then they are actioned


senhalil:
Hello, I am having trouble exporting the model unless I call `writeModel` explicitly like [here](https://github.com/senhalil/cvxpy/blob/d6f63960b49e664c202174967ec4adf30f7987f6/cvxpy/reductions/solvers/conic_solvers/highs_conif.py#L213-L220) but since I call use cvxpy, directly calling `writeModel` is not an option. Nevertheless, this doesn't look like a cxvpy issue (because I verified that both options are passed to highspy via [this line](https://github.com/senhalil/cvxpy/blob/d6f63960b49e664c202174967ec4adf30f7987f6/cvxpy/reductions/solvers/conic_solvers/highs_conif.py#L211) and I managed to create the following minimal working example that demonstrates the issue solely with `highspy`

```python
import highspy

solver = highspy.Highs()

solver.readModel("check/instances/25fv47.mps") # A random instance from the repo

options = solver.getOptions()  # same result if `options = highspy.HighsOptions()`
options.write_model_file = "/tmp/highs_model.lp" # Absolute path (but same result with a relative one)
options.write_model_to_file = True

solver.passOptions(options)

solver.run() # I expected this to write the model to options.write_model_file path
```

In case this is useful I am on  `"version": "1.10.0"`. Please let me know if I can help with debugging.

jajhall:
No, run() solves the model writeModel("foo.mps") writes it

senhalil:
Thanks for the rapid response @jajhall . In that case what is the functionality of `write_model_file` and `write_model_to_file` options? I assumed they were telling the solver whether to call the `writeModel` or not. Would this be a feature that you would be interested in? Afaik, most solvers do it this way (which made me assume that this was a bug). I'd be willing to contribute such a feature based on the existing `write_model_file` and `write_model_to_file` options if you are interested.

jajhall:
These options apply when running HiGHS from the command line, and are used to convert between MPS and LP format. 

When there is a callable method to write a model to a file, I see no reason to change the actions of Highs::run()

senhalil:
Well that is a bummer :) I'd consider "behavior of major solvers" (i.e., established user expectations) and "users calling highs via other AMLs" (i.e., not possible to call the writer directly) as good reasons but I understand -- I'll follow it up with cvxpy.

[...]

PR #2274, opened 2025-04-04
Fix 2269

jajhall:
This responds to issue #2269, in which I realised that

1.  There are AMLs (such as CVX) that can (effectively) only call `Highs::run()`, so options that can only be applied when running from the command line or by method call, are unavailable. 
2. Gurobi will write the appropriate file in a call to `optimize()` if the `ResultFile` option is set

Now read/write basis/solution and write model are executed in the (external) call to `Highs::run()`, rather than in `app/RunHighs.cpp`, so long as the corresponding file name is not empty. The model file name must have the extension ".mps" or ".lp", too.

It's also observed that the following options have been redundant for some time. They were only relevant when writing models and the solution to the screen from the command line

> bool write_model_to_file;
  bool write_presolved_model_to_file;
  bool write_solution_to_file;

As part of the coding, I rearranged the declarations of the command line options so that all the files are listed together. The change to `docs/src/executable.md` has been made, and soon README.md will be updated. I don't know how to update the documentation for other builds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,7 +152,11 @@ options:
       --model_file file          File of model to solve.
       --options_file file        File containing HiGHS options.
       --read_solution_file file  File of solution to read.
-      --presolve text            Set presolve option to: 
+      --read_basis_file text     File of initial basis to read. 
+      --write_model_file text    File for writing out model.
+      --solution_file text       File for writing out solution.
+      --write_basis_file text    File for writing out final basis.
+      --presolve text            Set presolve option to:
                                    "choose" * default 
                                    "on"
                                    "off"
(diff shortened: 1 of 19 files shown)


jajhall:
@galabovaa , I'm struggling to get the meson build CI test to pass, due to trying to use the files written in `TEST_CASE("highs-files-mip"`.

galabovaa:
looks good to me! 

> I don't know how to update the documentation for other builds.

What do you mean? We don't need to update docs for older versions, do you mean python / dotnet? We don't have much docs there to update yet

The meson build is passing at your latest commit

jajhall:
I mean python / dotnet. 

The meson build is only passing because I've disabled tests. TBH I can't spend much more time trying to get to the bottom of what's going on. There are probably enough tests for now

kutlay:
@jajhall what would be the new alternative to `write_presolved_model_to_file` command which is removed?
I am trying to use HiGHS to presolve my mps file and then I pass it to another solver. Is there still a way to achieve this? When I use `--write_model_file` I observed that the saved mps file has the same size as the original mps file, despite presolve removing thousands of rows.

jajhall:
It should be sufficient to set the name of the presolved model file to a non-empty string with suffix ".mps" or ".lp".

PR #2275, opened 2025-04-04
CMake updates

galabovaa:
Better paths 
Use of CMake variables more consistent 
Some clean up following more recent approaches in CMake


galabovaa:
WIP, needs more tests

PR #2276, opened 2025-04-04
Fixed the meson build failures in TestLpSolvers.cpp

jajhall:
Using file names local to the test fixed the meson build failures in TestLpSolvers.cpp, suggesting strongly that there is a delay in deleting these files in one test, so that after being overwritten by files in the next test, they are being deleted before they need to be used in that test.


galabovaa:
Great!

galabovaa (approved):

PR #2277, opened 2025-04-04
Fix docs by removing links to HighsInt

odow:


jajhall (approved):
Thanks

odow:
I don't know why it didn't skip it, like it does for all the other enums and constants... but this seemed to work for me locally.

Issue #2270, opened 2025-04-03
Bug in python callbacks user_solution and mip_solution


hlinsen:
- ` to_array` accessor of `readonly_ptr_wrapper` returns wrong values after the first one.

```
and solution[0] = 14.0
and solution[1] = 118.0
to_array(2) = [14.0, 1.7991808152773074e-307]
```
see https://github.com/ERGO-Code/HiGHS/blob/latest/examples/call_highs_from_python.py#L258

- user_solution is not fully integrated at the python layer
https://github.com/ERGO-Code/HiGHS/blob/latest/src/highspy/_core/cb.pyi#L28

mathgeekcoder:
FYI: I can reproduce, and I think I might know the fix.  Testing now.

mathgeekcoder:
@jajhall - the issue with `readonly_ptr_wrapper` is an easy (one line) fix, however the whole premise behind `readonly_ptr_wrapper` is a bit of a hack that becomes more obvious when trying to support `user_solution` and `cut_pool`.

Basically, since the callbacks are C style, there's some potential issues with memory safety.  I have a working solution to clean it all up... but has a minor breaking changes to the C API and highspy (removing to_array since no longer needed).  

I'll push a PR to your fix2270 branch, but happy to discuss/modify if you want to avoid breaking changes.

jajhall:
> Basically, since the callbacks are C style, there's some potential issues with memory safety. I have a working solution to clean it all up... but has a minor breaking changes to the C API and highspy (removing to_array since no longer needed).

Thanks. I wrote the callbacks C-style because it made it simpler to add them to the C API. Clearly it wasn't the best idea....

hlinsen:
> > Thanks. I wrote the callbacks C-style because it made it simpler to add them to the C API. Clearly it wasn't the best idea....
> 
> I can appreciate the idea behind it. It's a design choice with its own pros/cons. Technically, it should be possible to use that C-style safely, but moving "ownership" of the memory will just make it easier for users to avoid issues.
> 
> [@hlinsen](https://github.com/hlinsen) I hope to have an updated PR before end of the week, but in the meantime, I suggest not to use the `to_array` function. It's a little clunky anyhow, and there's several better alternatives in highspy. With any luck I'll be able to deprecate that function.

Thanks @mathgeekcoder! For the second part of the issue, can I pull the changes now and set a user_solution through python?

mathgeekcoder:
[...]

PR #2278, opened 2025-04-04
Fixes + callback updates

mathgeekcoder:
Changed *internal* callback code to use C++ rather than C for increased memory safety and ownership.  e.g., avoid memory leaks, etc.

* Includes a small breaking change for the C API: The `HighsCallbackDataIn::user_solution` memory pointer is allocated and managed by HiGHS.  The user can modify the elements of the array - but they should not try to reallocate the memory.

* Added `setSolution` and `repairSolution` functions that can accept partial solutions and (naively) find feasible solutions

* Added usability changes to highspy callbacks + a `setObjective` function that should've been previously added.

* Minor breaking change to highspy, the `mip_solution` no longer needs/supports `to_array(n)` - as it is now a numpy array.

### QUESTIONS:
* Should we change `kHighsUndefined` so that it is different to `kHighsInf`?  Is it important to be able to tell the difference?
* Is there already a function that finds a feasible solution that I can call in `repairSolution`?  
* Are you happy with `repairSolution` naming, or would you prefer Gurobi's `useSolution`?


odow:
> Includes a small breaking change for the C API

No breaking changes please. People are using HiGHS in a variety of contexts, so it's really hard to know what the impact of this is.

mathgeekcoder:
> No breaking changes please. People are using HiGHS in a variety of contexts, so it's really hard to know what the impact of this is.

Absolutely understood, but the current memory access with callbacks is quite fragile.  I don't know how to improve this without any breaking changes.

The smallest breaking change would be to require a `user_solution_size` for `HighsCallbackDataIn`.  That's probably not too bad?  

Even with that, the C++ / highspy API would be less robust since the C-style structs ignore enum types and pass around raw pointers.  We'd want to make sure the code is suitably documented about **who** owns the pointers etc. so we don't have memory leaks or access violations.

Happy to take suggestions?

odow:
Perhaps I missed something in a different issue, but what's the problem with the current callback?

mathgeekcoder:
> Perhaps I missed something in a different issue, but what's the problem with the current callback?

It's related to #2270, updating support for callbacks with highspy.  When adding support for `user_solution` I noticed that the consuming code assumes length is `num_col` (see code below).  
```cpp
if (mipsolver.callback_->data_in.user_solution) {
    std::vector<double> user_solution(mipsolver.orig_model_->num_col_);

    for (HighsInt iCol = 0; iCol < mipsolver.orig_model_->num_col_; iCol++)
      user_solution[iCol] = mipsolver.callback_->data_in.user_solution[iCol];
```

This is a reasonable assumption, but there's no bound checks - and there's a lot of assumptions about memory ownership.  Unless the user really knows what they're doing, it'd be easy to make a mistake (especially if making changes to the model, e.g., adding columns).  

By converting the callback structure to C++, many of these potential issues are avoided.  In addition, this additional copy (above) is no longer needed for at least C++/highspy users since highs keeps ownership of the memory.

Alternatively - at a minimum, by having a `user_solution_size`, we make the user declare how much memory they've allocated to avoid potential access violations.  This is clearly not foolproof but might help avoid some issues. Honestly, it'd be better to have the C++ callback structure.

Similarly, with `HighsCallbackDataOut`, there's `mip_solution`, `cutpool_start`, `cutpool_index`, `cutpool_value`, `cutpool_lower`, and `cutpool_upper` are currently raw pointers.

odow:
> Unless the user really knows what they're doing, it'd be easy to make a mistake

I think a strong assumption is that callbacks are extras for experts. Callback authors are potentially novice, but typically very experienced because they understand what a callback is, when it gets called by the solver, and why it might be useful.

`highspy` could check bounds etc.

Before we make changes, let me try and wrap the callback with HiGHS.jl and I'll report back.

odow:
Oh. I see what you mean. The ownership of `user_solution` is particularly problematic for GC based languages.

We're also missing a `kHighsCallbackMipUserSolution` constant in the C API.

I'd settle for breaking changes to the way we provide `user_solution` in the next release. @jajhall can comment, but it looks like this was added for someone particular without thinking through the broader challenges?

Other questions: how can I provide a partial solution? Is there an equivalent to `GRB_UNDEFINED`?

Broader: I'd be in favor of a new callback interface that followed Gurobi. You cannot interact with the `in` and `out` pointers except via things like `Highs_getDoubleCallbackDataItem` and `Highs_setDoubleCallbackDataItem`.

mathgeekcoder:
Thanks, I appreciate the 2nd opinion!

I'm not quite happy with my approach in the PR.  I feel like HiGHS should consistently either own the memory, or the user does. 

Looking at Gurobi API, from the users perspective, they own the memory - and they can delete at end of scope since they get/set the solution with function calls. Likely these calls are just copying to/from the internally owned buffers in Gurobis cbdata struct.

With this in mind, and since HiGHS currently owns the *out solution* data, it probably should also own the *in solution* data.

If we do this, it's easy to support a `set solution()` function syntax (or the user can modify the buffer directly).  And I agree - it would be great to support partial solutions with UNKNOWN etc!

I'll have a crack at it this week, but I'd certainly appreciate @jajhall thoughts too!

jajhall:
Thanks for your work on this, @mathgeekcoder and @odow 

> it looks like this was added for someone particular without thinking through the broader challenges?

Yes, the user solution callback was created by me for NVIDIA. My C++/C/Python skills aren't up to thinking through broader challenges in this case. 

> it would be great to support partial solutions with UNDEFINED etc!

At the moment, partial solutions can only be defined by passing arrays of values and indices for the defined values. However, it looks simple to allow a full-length solution to contain values set to `kHighsUndefined`.

I'm OK with breaking changes to the user_solution callback, as it's only recently been introduced. I'm not 100% happy with having it anyway, as it generally makes the HiGHS MIP solver non-deterministic. 

I'm happy with you guys doing what you think is best with HiGHS callbacks, as I've already shown that I don't have the skills to code them up the best way, although please keep @galabovaa in the loop.

In passing, I'm curious that values of `data_in` set in (eg) `examples/call_highs_from_python.py` that are passed back through the parameter list of `user_interrupt_callback` are recognised by the C++ method (ultimately) calling the callback, since I didn't think Python could pass parameters by reference. Maybe I should just accept that it works, and not try to understand why!

odow:
> or the user can modify the buffer directly

Yes, if HiGHS provided a valid HighsCallbackDataIn that we could modify, that would be great.

> it looks simple to allow a full-length solution to contain values set to kHighsUndefined

:+1:

mathgeekcoder:
Just FYI, I'm still working on this (made good progress already), but had some other commitments earlier this week.

On my local clone, I've rolled back most of my breaking changes to the callback C API, the main one now is that `user_solution` is "owned" by HiGHS.  I've been playing with some other design choices (can elaborate more when I update the PR).  Would love to get feedback on it. 
 
> I'm happy with you guys doing what you think is best with HiGHS callbacks, as I've already shown that I don't have the skills to code them up the best way, although please keep @galabovaa in the loop.

Honestly, this stuff is *hard* to get "right" for everyone - especially since what you've written *will* work and can be very efficient.  It just makes me nervous to ensure that users "do the right thing".  

I'm also worried about how this will work when we have multiple threads etc.  I've some ideas, but haven't really thought too much ahead yet. @jajhall - I know you've started looking at the parallel branch and bound - do you know if this might impact our approach?  We might need to synchronize the threads or have multiple copies of the data.

 
> In passing, I'm curious that values of `data_in` set in (eg) `examples/call_highs_from_python.py` that are passed back through the parameter list of `user_interrupt_callback` are recognised by the C++ method (ultimately) calling the callback, since I didn't think Python could pass parameters by reference. Maybe I should just accept that it works, and not try to understand why!

I swear there's some arcane magic behind the scenes with pybind!  I probably don't understand half of it, but it does a great job.

jajhall:
A callback to feed user solutions to HiGHS is very low priority, so don't compromise other things to make it work. Feeding even the optimal solution to an exact MIP solver frequently doesn't help much - and tossing in user solutions makes the MIP solver non-deterministic. If only NVIDIA hadn't suggested it would be useful when we first spoke to them!

odow:
On `latest` I've added a Julia CI job: https://github.com/ERGO-Code/HiGHS/blob/latest/.github/workflows/julia-tests-ubuntu.yml. I'm okay with any changes that don't break that build. I don't see it showing up in the current jobs so I guess this branch needs rebasing to pick up the commit.

galabovaa:
Thank you for getting back to this @mathgeekcoder!

Just a note, the valgrind test is fixed on a branch not yet merged in latest, and I'm still tracking the occational bazel tsan failure

mathgeekcoder:
> Thank you for getting back to this @mathgeekcoder!
> 
> Just a note, the valgrind test is fixed on a branch not yet merged in latest, and I'm still tracking the occational bazel tsan failure

Thanks @galabovaa!  Sorry it's taken me so long to get back to it.

I appreciate the note!  I was going to investigate that today... you've saved me some time :)   I'll try to track down the issue in julia, but afterwards I'd love to get your opinion on the PR.

galabovaa (approved):
Looks good to me! Perhaps @jajhall should have a final look too before we merge this.

galabovaa:
A comment on the CI: 
I have more work to do on the bazel tsan test and the valgrind fix will be merged soon. 
I have just re-run the two other failing tests

darrylmelander:
I appreciate that effort was made to avoid breaking changes. However, at least one snuck in, the introduction of `void* cbdata` at the beginning of the `HighsCallbackDataOut` struct. Language bindings often have to essentially repeat the struct definition to specify the binary layout of composite data types like this, so a surprise addition at the start of a struct breaks things. I'd encourage new fields to be placed at the end of existing structs instead of at the start or the middle.

I'm not complaining - this PR is a good change - just raising awareness.

mathgeekcoder:
Thanks for your understanding @darrylmelander. This (breaking) change was somewhat intentional by me - not to cause people trouble, but to test the waters (so to speak). 

I'm aware that it's best practice for closed-source APIs to add new fields to the end of a struct, but I figured that open-source projects shouldn't be limited to keep same ordering. 

There can be some performance benefits when reordering structs, but for this case it doesn't really matter.

We might be changing how the callbacks are used in the near future - but we'll try not to break things too much.

Issue #2283, opened 2025-04-07
CMake Error when trying to linstall HiGHS as LP solver in SCIP (`highs::highs` not found)


adbuerger:
Hello everyone,

I am putting this issue here since I think the problem might be related to HiGHS.

I am trying to build SCIP with HiGHS as LP solver. I followed the instructions at https://github.com/ERGO-Code/HiGHS to build HiGHS (1.10.0) which worked fine. Then, I followed the instructions given at https://scipopt.org/doc/html/md_INSTALL.php to build SCIP (9.2.1) with HiGHS as LP solver using the command 

```
cmake .. -DREADLINE=false -DIPOPT=false -DZIMPL=false -DLPS=highs -DHIGHS_DIR=/path/to/HiGHS-1.10.0/build
```

which fails with the following error:

```
CMake Error at scip/src/CMakeLists.txt:1069 (target_link_libraries):
  Target "libscip" links to:

    highs::highs

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.



CMake Error at scip/src/CMakeLists.txt:1097 (target_link_libraries):
  Target "scip" links to:

    highs::highs

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.

```

My experience with CMake is limited, but when looking at the contents of `HiGHS-1.10.0/build/highs-targets.cmake` I can see multiple items being called `highs::highshighs`. I suspect those should read `highs::highs` instead?

I would be very happy if you can let me know whether this is actually a bug or if I am just missing the right configuration. Thank you very much for your help and for your work on this great project!

jajhall:
Sounds strange, but maybe @galabovaa can shed some light on this

galabovaa:
Thank you @adbuerger, yes, this is a bug and I will add a test and fix for it for our upcoming release. 

In the meantime, linking with HiGHS from CMake works if you install HiGHS in a custom installation directory.

```
cmake -S. -B build -DCMAKE_INSTALL_PREFIX=/path/to/highs_install 
cmake --build build --parallel
cmake --install build
```

And then 

```
-DHIGHS_DIR=path_to_highs_install/lib/cmake/highs
```

Please let us know if this works!

adbuerger:
[...]

PR #2253, opened 2025-03-24
Include cmake bux fix

galabovaa:
closes https://github.com/ERGO-Code/HiGHS/issues/2250

PR #2286, opened 2025-04-08
CMake updates 

galabovaa:
Apologies for a rather large PR. Build system updates and a rename of the source directory.

Added several tests for linking with CMake in https://github.com/ERGO-Code/highs-tests

- CMake targets in build dir now works: closes https://github.com/ERGO-Code/HiGHS/issues/2283
- More relocatable package using BUILD_INTERFACE.. in CMake https://github.com/ERGO-Code/HiGHS/pull/2253
- long awaited rename of source dir from src/ to highs/ for better style and cleaner integration 
- src/test is now highs/test_kkt to avoid the test/ in gitignore for CMake FetchContent


jajhall (approved):
Only 452 files changed, but nothing that affects any of the source code files, so fine by me.

PR #2289, opened 2025-04-11
Features doc mods

jajhall:
Added use of `Catch::getResultCapture().getCurrentTestName()` to make temporary files unique to unit test

Updated `FEATURES.md` and `docs/src/parallel.md`


jajhall:
NB @galabovaa I've reverted the change from `https://github.com/ERGO-Code/HiGHS/blob/master/src/Highs.h` to `https://github.com/ERGO-Code/HiGHS/blob/master/highs/Highs.h`  in `src/index.md` and `src/interfaces/cpp/library.md` so that the documenter CI test passes.

The change will have to be made again when we push `latest` to `master`

galabovaa (approved):

PR #2291, opened 2025-04-14
Reset global scheduler

galabovaa:
In binary and unit tests, valgrind warnings about HighsTaskExecutor cleared


jajhall (approved):
Great, thanks!

Issue #1423, opened 2023-09-21
Add feasibility jump...


jajhall:
... with reference to https://github.com/SINTEF/feasibilityjump

PR #2298, opened 2025-04-17
Fix 1423: Add Feasibility Jump

BenChampion:
Add the Feasibility Jump primal heuristic as per #1423

Testing across the MIP2017 benchmark set revealed a small (~1%) decrease in the shifted geometric mean running time (10s shift). [Note that these results exclude the two instances `momentum1` and `var-smallemery-m6j6`, due to some problems with running them via the testing infrastructure that we are investigating.]


galabovaa (approved):
Looks good to me!

jajhall (approved):
Quite a few changes unrelated to FJ that I feel I've made somewhere, but merging into the branch I'm working on yields minimal conflict, so I guess the changes I see _are_ the ones that I made!

So, looks good to me.

Where are the lines that @BenChampion mentioned I should look at?

highs/mip/feasibilityjump.hh
```diff
@@ -0,0 +1,775 @@
...
+
+struct FJStatus {
+  size_t totalEffort;
+  size_t effortSinceLastImprovement;
+  int numVars;
+  double solutionObjectiveValue;
+  double* solution;
+};
+
+double violationTolerance = 1.0e-5;
```
tosttost: either this can be const or we have a threading issue if HiGHS is used as a library and asked to solve several instances in parallel.

jajhall: Thanks. I guess it can be const, but if it's never changed, how it can it cause problems in a multi-threaded application?

tosttost: if it is not written to it is not an issue. I guess it's a matter of personal preference. I find it nice to let the compiler validate the assumption that it is not written to.
In this case the compiler disagrees due to HighsFeasibilityJump.cpp line 31 and 32.

jajhall: Thanks, I see the issue now. @galabovaa and @BenChampion , how can this most easily be fixed?

highs/lp_data/HighsInterface.cpp
```diff
@@ -3445,29 +3452,62 @@ bool Highs::infeasibleBoundsOk() {
...
-      // Semi-variables can have inconsistent bounds
+      // Semi-variables cannot have inconsistent bounds
       if (lp.integrality_[iCol] == HighsVarType::kSemiContinuous ||
           lp.integrality_[iCol] == HighsVarType::kSemiInteger)
         continue;
+      if (perform_inward_integer_rounding &&
+          lp.integrality_[iCol] == HighsVarType::kInteger) {
+        // Assess bounds after inward integer rounding
+        double integer_lower = std::ceil(lower);
+        double integer_upper = std::floor(upper);
```
fwesselm: Should a tolerance be used when bounds are rounded?

highs/mip/HighsFeasibilityJump.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+    assert(model->integrality_[col] == HighsVarType::kContinuous ||
+           model->integrality_[col] == HighsVarType::kInteger ||
+           model->integrality_[col] == HighsVarType::kImplicitInteger);
+    external_feasibilityjump::VarType fjVarType;
+    if (model->integrality_[col] == HighsVarType::kContinuous) {
+      fjVarType = external_feasibilityjump::VarType::Continuous;
+    } else {
+      fjVarType = external_feasibilityjump::VarType::Integer;
+      lower = std::ceil(lower);
```
fwesselm: Should a tolerance be used when bounds are rounded?

highs/mip/feasibilityjump.hh
BenChampion: The easiest potential fix I see at the moment is to make `violationTolerance` a instance variable of `Problem`, `equalityTolerance` a instance variable of both `Problem` and `JumpMove`, and `eq` a member function of both `Problem` and `JumpMove`. Then we can pass both `violationTolerance`  and `equalityTolerance`  through the `FeasibilityJumpSolver` constructor to its `Problem` and `JumpMove` members.

It doesn't seem particularly elegant, but I think it would do the job?

galabovaa: Sounds alright to me, we need some way of keeping the values attached to the solver instances and this sounds OK

BenChampion: Cool, I've updated the PR. (I added an additional tolerance argument to `eq` meaning I don't have to make it a member function anywhere.)

highs/lp_data/HighsInterface.cpp
BenChampion: I defer to @jajhall here (and on the related comment [here](https://github.com/ERGO-Code/HiGHS/pull/2298#discussion_r2053677960))

BenChampion: After chatting with @jajhall in person, I've updated the PR for both cases.

galabovaa:
The failing valgrind test is new and will be fixed elsewhere

PR #2299, opened 2025-04-20
Take into account discreteness for no cost zero variables in bounds section

MaximSmolskiy:
It seems that there was inconsistency in https://github.com/ERGO-Code/HiGHS/blob/master/src/io/HMPSIO.cpp#L879 with https://github.com/ERGO-Code/HiGHS/blob/master/src/io/HMPSIO.cpp#L724-L728

As I understand, for example, https://github.com/ERGO-Code/HiGHS/blob/master/src/io/HMPSIO.cpp#L925-L927 - for no cost zero integer variable with `lb = 0` and `ub = +inf` `LI BOUND` will be written (unless `write_no_cost_zero_columns = false`) and then maybe it is worth taking this into account in `num_no_cost_zero_columns_in_bounds_section` variable and corresponding log


jajhall:
We've not forgotten about this: just rather busy ATM

PR #2301, opened 2025-04-22
Fix data races in thread sanitizer 

galabovaa:
Add sanitizers for CMake, bazel, clang and gcc (linux) 
Add specifiers to thread sanitizer for warnings
Upload bazel-test logs to CI
Add highs binary to bazel build


jajhall (approved):
Wow! Looks great!

PR #2312, opened 2025-05-01
Docs: fix size of reduced row

mmghannam:
Shouldn't the size of a row of B^-1 A be the number of columns? please correct me if I'm wrong.


jajhall (commented):
The error is in saying that it's computing a row of B^-1A. It's solving B^Tx=B, so the size references are correct. Also, although it's only documentation, on principle we don't push to master, but to "latest", unless it's a critical bug fix.

I'll see that the docstring is updated

jajhall (approved):
I now see the error in my earlier comment - I was reading the PR on my phone, so missed that the docstrings came before the method header definition.

jajhall:
Sorry, there are merge conflicts, so I'll make the changes via #fix-2312

mmghannam:
Thank you!

PR #2316, opened 2025-05-05
Support empty objective

sedefe:
When a model has empty objective (e.g. generated in https://docs.gurobi.com/projects/examples/en/current/examples/python/sudoku.html ), it might not be as garbage as we suspect.


jajhall (approved):
This works fine - in that `num_objective_section = 0` doesn't seem to cause any downstream issues. Also tested with .lp file beginning

>Maximize
Subject to

as well as one beginning

> Subject to

PR #2319, opened 2025-05-05
Applied fix of 2316, but exposed isssue 2318

jajhall:


jajhall:
@galabovaa , the valgrind failures appear to be just due to calling `settings_clear(settings);` before `timers_clear(w->settings->nLogLevel, timers);`.

galabovaa (approved):
Great!

Issue #2326, opened 2025-05-09
Set cover like instance with strange presolver behavior & LpIters bug


Athlici:
While solving some set cover like instances with HiGHS we have encountered some undesirable behaviors in regards to presolving, all reproducible with the instance [dump.zip](https://github.com/user-attachments/files/20121682/dump.zip).

- For reference, with presolve disabled the instance is solved in ~6min [HiGHSPoff.log](https://github.com/user-attachments/files/20121738/HiGHSPoff.log).
- Presolve however takes very long (~4,5 hours), uses a lot of memory (~22GB) and the resulting instance then takes longer to solve than without presolve [HiGHSPon.log](https://github.com/user-attachments/files/20121770/HiGHSPon.log).
- With presolve enabled but presolve_reduction_limit=0 the instance is solved quickly (30s) but the LpIter counter takes on huge values (~10^9) [HiGHSRed0.log](https://github.com/user-attachments/files/20121829/HiGHSRed0.log).

Our problem structure is relative simple and is essentially given by
$$\min \sum_{i=1}^n c_i*x_i\quad\text{s.t.}$$
$$\sum_{i=1}^n \gamma_{i,j}*x_i = d_j\quad\text{for}\quad j=1,\ldots,m$$
$$\sum_{i=1}^n x_i\leq 10000$$
$$0\leq x_i\leq u_i\quad\text{for}\quad i=1,\ldots,n$$
with m small, n large and the upper bounds generally superfluous for the solution.
The coefficients $\gamma_{i,j}$ represent configurations and are non-negative integers, $d_i$ are integer amounts and $c_i$ are positive costs.
With higher variable counts the problems mentioned above become more pronounced.

Unfortunately presolve enabled with presolve_reduction_limit=0 leads to issues on other instances, see [issue 1578](https://github.com/ERGO-Code/HiGHS/issues/1578#issuecomment-2841493389).
For the moment we are just disabling presolve completely to work around these issues.
Ideally though presolve should probably not exhibit these detrimental effects, which is why we thought that you might be interested in this instance.

jajhall:
The culprit (@fwesselm) is probing. I've added another bit to the `presolve_rule_off` option so that probing can be prevented. This leads to presolve taking 30s, and the MIP is solved in 147s. I don't know why probing can have such a large memory requirement.

jens-diewald:
Hi @jajhall, thanks for looking into this.

[...]

PR #2330, opened 2025-05-10
Avoided false claim of optimality when presolve is terminated by `presolve_reduction_limit` when reduced model has no columns but zero row activity is infeasible.

jajhall:
This addresses the (separate) issue raised within #1578, and adds the "probing off" bit to `presolve_rule_off` option that allowed probing to be identified as the reason for the excessive time and memory requirement of #2326

### DIFF
--- a/highs/presolve/HPresolve.cpp
+++ b/highs/presolve/HPresolve.cpp
@@ -6898,8 +6898,8 @@ bool HPresolve::zeroRowActivityFeasible() const {
   // has no columns to assess whether the HighsModelStatus returned is
   // kOptimal or kInfeasible (as was required for 2326)
   for (HighsInt iRow = 0; iRow < model->num_row_; iRow++)
-    if (model->row_lower_[row] > primal_feastol ||
-        model->row_upper_[row] < -primal_feastol)
+    if (model->row_lower_[iRow] > primal_feastol ||
+        model->row_upper_[iRow] < -primal_feastol)
       return false;
   return true;
 }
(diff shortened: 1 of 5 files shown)


highs/presolve/HPresolve.cpp
```diff
@@ -6885,6 +6892,16 @@ HPresolve::Result HPresolve::sparsify(HighsPostsolveStack& postsolve_stack) {
   return Result::kOk;
 }
 
+bool HPresolve::zeroRowActivityFeasible() const {
+  // Check that zero row activity is feasible - called when reduced model
+  // has no columns to assess whether the HighsModelStatus returned is
+  // kOptimal or kInfeasible (as was required for 2326)
+  for (HighsInt iRow = 0; iRow < model->num_row_; iRow++)
+    if (model->row_lower_[iRow] > 0 || model->row_upper_[iRow] < 0)
```
fwesselm: Should the feasibility tolerance be used here? For example, `HPresolve::rowPresolve` has this check:

`if (model->row_upper_[row] < -primal_feastol || model->row_lower_[row] > primal_feastol)`

Otherwise, could you just use the existing code?

`HPRESOLVE_CHECKED_CALL(rowPresolve(postsolve_stack, iRow));`

jajhall: I wondered about extending the margin this way. Although I think that using `HPresolve::rowPresolve` obscures the nature of the check, I know you don't like duplicated code!

jajhall: On second thoughts, since my new method is const, and `postsolve_stack` is not const in `Hpresolve::rowPresolve` (and not needed anyway for the case I'm handling), I'll just modify my new method

fwesselm (approved):

PR #2333, opened 2025-05-12
Stronger CMIR cuts and correct efficacy calculation

### LINKED ISSUE #2334: Weak CMIR cuts
The CMIR cuts in Highs are weaker than necessary. For the j'th integer variable, HiGHS is using the formula:
floor(a_j) +  max(0, (fj-f0)) 
where a_j is the coefficient in the original row, fj is the fractionality of variable j and f0 is the fractionality of the right hand side.

Other solvers are using the following formula, which computes strictly better coefficients:

floor(a_j) +  max(0, (fj-f0) / (1 - f0) ) 

Furthermore, the efficacy calculation in HighsCutGeneration.cpp is not always correct, as it does not correctly scale the contribution of the continuous variables by  1 / (1 - f0)

rolfvdhulst:
This PR fixes some details in the calculation of the MIR cuts in HiGHS. Fixes #2334.

For the j'th integer variable, HiGHS is using the formula:
floor(a_j) +  max(0, (fj-f0)) 
where a_j is the coefficient in the original row, fj is the fractionality of variable j and f0 is the fractionality of the right hand side.

Other solvers are using the following formula, which computes strictly better coefficients:

floor(a_j) +  max(0, (fj-f0) / (1 - f0) ) 

This PR also contains some fixes for the efficacy calculation in the CMIR code, which was also incorrect because of a similar error where the scaling factor 1/(1-f0) was left out.
The changes here should be tested for performance before they are merged. Likely, the stronger cuts and the corrected efficacy calculation should give better performance.

### DIFF
--- a/src/mip/HighsCutGeneration.cpp
+++ b/src/mip/HighsCutGeneration.cpp
@@ -589,14 +589,15 @@ bool HighsCutGeneration::cmirCutGenerationHeuristic(double minEfficacy,
     double oneoveroneminusf0 = 1.0 / (1.0 - f0);
     if (oneoveroneminusf0 > maxCMirScale) continue;
 
-    double sqrnorm = scale * scale * continuoussqrnorm;
-    double viol = scale * continuouscontribution * oneoveroneminusf0 - downrhs;
+    double contscale = scale * oneoveroneminusf0;
+    double sqrnorm = contscale * contscale * continuoussqrnorm;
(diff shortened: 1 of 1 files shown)


Opt-Mucca:
Double nice spot on the norm calculation!

jajhall (commented):
I'll run the CI tests to see what happens. There's a merge conflict with `latest`, so I'll look at this locally

PR #2337, opened 2025-05-14
Add max scale factor for generated inequalities

### LINKED ISSUE #2322: HiGHS generates inconsistent answer with `mip_pscost_minreliable=1`
For [seed.mps.txt](https://github.com/user-attachments/files/20095669/seed.mps.txt), HiGHS generates the correct answer (`3610..`) by default,

but returns with an incorrect answer (`3620..`) with parameter `mip_pscost_minreliable=1`

```
highs --model_file seed.mps
Running HiGHS 1.10.0 (git hash: 75ddbe092): Copyright (c) 2025 HiGHS under MIT licence terms
MIP  seed has 128 rows; 10 cols; 1280 nonzeros; 5 integer variables (0 binary)
Coefficient ranges:
  Matrix [6e-02, 1e+02]
  Cost   [4e+00, 8e+01]
  Bound  [2e+02, 2e+02]
  RHS    [2e+01, 5e+03]
Presolving model
128 rows, 10 cols, 1280 nonzeros  0.012862s
128 rows, 10 cols, 1280 nonzeros  0.015607s

Solving MIP model with:
   128 rows
   10 cols (0 binary, 5 integer, 0 implied int., 5 continuous, 0 domain fixed)
   1280 nonzeros

Src: B => Branching; C => Central rounding; F => Feasibility pump; J => Feasibility jump; H => Heuristic;
     L => Sub-MIP; P => Empty MIP; R => Randomized rounding; Z => ZI Round; I => Shifting;
     S => Solve LP; T => Evaluate node; U => Unbounded; z => Trivial zero; l => Trivial lower; u => Trivial upper; p => Trivial point; X => User solution

         0       0         0   0.00%   -101700         inf                  inf        0      0      0         0     0.2s
         0       0         0   0.00%   3568.304521     inf                  inf        0      0      0        20     0.2s
 C       0       0         0   0.00%   3574.745409     3704.607624        3.51%       21      5      0        25     0.3s
 L       0       0         0   0.00%   3578.233136     3678.842839        2.73%       52     13      0        35     0.6s
 L       0       0         0   0.00%   3578.233136     3641.72362         1.74%       52      3      0        45     1.0s
 T      12       1         4  34.38%   3578.233136     3610.306598        0.89%       59      3      8       162     1.1s

Restarting search from the root node
Model after restart has 127 rows, 5 cols (0 bin., 0 int., 0 impl., 5 cont., 0 dom.fix.), and 635 nonzeros

        27       0         0   0.00%   3583.879779     3610.306598        0.73%        2      0      0       249     1.7s
        28       0         1 100.00%   3610.306598     3610.306598        0.00%        2      0      0       256     1.7s

Solving report
  Model             seed
  Status            Optimal
  Primal bound      3610.30659784
  Dual bound        3610.30659784
  P-D integral      0.0278207709693
  Solution status   feasible
                    3610.30659784 (objective)
                    0 (bound viol.)
                    0 (int. viol.)
                    0 (row viol.)
  Timing            1.69 (total)
                    0.00 (presolve)
                    0.00 (solve)
                    0.00 (postsolve)
  Max sub-MIP depth 1
  Nodes             28
  Repair LPs        0 (0 feasible; 0 iterations)
  LP iterations     256 (total)
                    84 (strong br.)
                    15 (separation)
                    61 (heuristics)
$>
$>
$>
$> highs --model_file seed.mps --options_file options.txt
Running HiGHS 1.10.0 (git hash: 75ddbe092): Copyright (c) 2025 HiGHS under MIT licence terms
Set option mip_pscost_minreliable to 1
MIP  seed has 128 rows; 10 cols; 1280 nonzeros; 5 integer variables (0 binary)
Coefficient ranges:
  Matrix [6e-02, 1e+02]
  Cost   [4e+00, 8e+01]
  Bound  [2e+02, 2e+02]
  RHS    [2e+01, 5e+03]
Presolving model
128 rows, 10 cols, 1280 nonzeros  0.002199s
128 rows, 10 cols, 1280 nonzeros  0.007048s

Solving MIP model with:
   128 rows
   10 cols (0 binary, 5 integer, 0 implied int., 5 continuous, 0 domain fixed)
   1280 nonzeros

Src: B => Branching; C => Central rounding; F => Feasibility pump; J => Feasibility jump; H => Heuristic;
     L => Sub-MIP; P => Empty MIP; R => Randomized rounding; Z => ZI Round; I => Shifting;
     S => Solve LP; T => Evaluate node; U => Unbounded; z => Trivial zero; l => Trivial lower; u => Trivial upper; p => Trivial point; X => User solution

         0       0         0   0.00%   -101700         inf                  inf        0      0      0         0     0.2s
         0       0         0   0.00%   3568.304521     inf                  inf        0      0      0        20     0.2s
 C       0       0         0   0.00%   3574.745409     3704.607624        3.51%       21      5      0        25     0.3s
 L       0       0         0   0.00%   3578.233136     3678.842839        2.73%       52     13      0        35     0.6s
 L       0       0         0   0.00%   3578.233136     3641.72362         1.74%       52      3      0        45     1.0s
 T      28       2        11  39.84%   3578.233136     3625.010513        1.29%       62      3     14       169     1.1s
 T      50       1        24  90.62%   3583.879779     3620.365759        1.01%       96     11     33       232     1.3s

Restarting search from the root node
Model after restart has 127 rows, 5 cols (0 bin., 0 int., 0 impl., 5 cont., 0 dom.fix.), and 635 nonzeros

        51       0         0   0.00%   3605.782369     3620.365759        0.40%        9      0      0       235     1.4s
        52       0         1 100.00%   3620.365759     3620.365759        0.00%        9      0      0       242     1.4s

Solving report
  Model             seed
  Status            Optimal
  Primal bound      3620.36575872
  Dual bound        3620.36575872
  P-D integral      0.0239785568714
  Solution status   feasible
                    3620.36575872 (objective)
                    0 (bound viol.)
                    0 (int. viol.)
                    0 (row viol.)
  Timing            1.43 (total)
                    0.00 (presolve)
                    0.00 (solve)
                    0.00 (postsolve)
  Max sub-MIP depth 1
  Nodes             52
  Repair LPs        0 (0 feasible; 0 iterations)
  LP iterations     242 (total)
                    17 (strong br.)
                    31 (separation)
                    61 (heuristics)
```

Opt-Mucca:
This should fix #2322 

Explanation: HiGHS always scales a generated inequality so that the largest coefficient is approximately 1.0 before running a separator from `HighsCutGeneration`. This results in problems when all the coefficients are small, e.g. near `feastol = 1e-6`, and the inequality violates the optimal solution by some minor value, e.g., `1e-10`, because the violation grows substantially after scaling, and the resultant cut then removes the solution.

Solution: Add a max scaling factor (you can still scale from 1e+12 to 1.0, but not from 1e-6 to 1.0). Talked with Leona and some other developers on a sensible value (`2**10 = 1024`). 

Instances affected: Any instance where HiGHS generates an inequality with largest absolute coefficient less than `5e-4` and then calls a separator from `HighsCutGeneration`. I don't see these being frequent and it's not like we're stopping a cut from being generated (just scaling it by less), but I can imagine a tiny performance hit. 

TLDR: Puts up some additional guard rails for numerically instable cuts

### DIFF
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -22,5 +22,5 @@ Corrected docstrings for `Highs_getReducedRow`, motivated by [#2312](https://git
 
 LP file reader no longer fails when there is no objective section. Fix is [#2316](https://github.com/ERGO-Code/HiGHS/pull/2316)), but this exposes code quality issue [#2318](https://github.com/ERGO-Code/HiGHS/issues/2318))
 
-
+Added a max scale factor (+1024) when scaling up coefficients in `preprocessBaseInequality` and `postprocessCut`. Fix is [#2337](https://github.com/ERGO-Code/HiGHS/pull/2337).
 
--- a/highs/mip/HighsCutGeneration.cpp
+++ b/highs/mip/HighsCutGeneration.cpp
@@ -781,6 +781,8 @@ bool HighsCutGeneration::postprocessCut() {
     }
   }
 
+  if (rowlen == 0) return false;
+
   if (integralSupport) {
     // integral support -> determine scale to make all coefficients integral
     double intscale =
@@ -861,6 +863,8 @@ bool HighsCutGeneration::postprocessCut() {
     int expshift;
     std::frexp(maxAbsValue - epsilon, &expshift);
     expshift = -expshift;
+    // Don't scale the coefficients by more than +1024 (violations can increase)
+    expshift = std::min(10, expshift);
     rhs = std::ldexp((double)rhs, expshift);
 
     for (HighsInt i = 0; i != rowlen; ++i)
@@ -895,6 +899,8 @@ bool HighsCutGeneration::preprocessBaseInequality(bool& hasUnboundedInts,
   int expshift = 0;
   std::frexp(maxAbsVal, &expshift);
   expshift = -expshift;
+  // Don't scale the coefficients by more than +1024 (violations can increase)
+  expshift = std::min(10, expshift);
   initialScale = std::ldexp(1.0, expshift);
   rhs *= initialScale;
   for (HighsInt i = 0; i < rowlen; ++i) vals[i] = std::ldexp(vals[i], expshift);
--- a/highs/mip/HighsCutGeneration.cpp
+++ b/highs/mip/HighsCutGeneration.cpp
@@ -900,7 +900,7 @@ bool HighsCutGeneration::preprocessBaseInequality(bool& hasUnboundedInts,
   for (HighsInt i = 0; i < rowlen; ++i)
     maxAbsVal = std::max(std::abs(vals[i]), maxAbsVal);
 
-  scale(maxAbsVal);
+  initialScale = scale(maxAbsVal);
 
   isintegral.resize(rowlen);
   for (HighsInt i = 0; i != rowlen; ++i) {
--- a/highs/mip/HighsCutGeneration.cpp
+++ b/highs/mip/HighsCutGeneration.cpp
@@ -717,6 +717,25 @@ bool HighsCutGeneration::cmirCutGenerationHeuristic(double minEfficacy,
   return true;
 }
 
+double HighsCutGeneration::scale(double val) {
+  int expshift = 0;
+  std::frexp(val, &expshift);
+  expshift = -expshift;
+
+  // Don't scale the coefficients by more than +1024 (violations can increase)
+  expshift = std::min(10, expshift);
+
+  // Scale rhs
+  rhs = std::ldexp(static_cast<double>(rhs), expshift);
+
+  // Scale row
+  for (HighsInt i = 0; i != rowlen; ++i)
+    vals[i] = std::ldexp(vals[i], expshift);
+
+  // Return scaling factor
+  return std::ldexp(1.0, expshift);
+}
+
 bool HighsCutGeneration::postprocessCut() {
   // right hand sides slightly below zero are likely due to numerical errors and
   // can cause numerical troubles with scaling, so set them to zero
@@ -848,27 +867,12 @@ bool HighsCutGeneration::postprocessCut() {
       for (HighsInt i = 0; i != rowlen; ++i)
         minAbsValue = std::min(std::abs(vals[i]), minAbsValue);
 
-      int expshift;
-      std::frexp(minAbsValue - epsilon, &expshift);
-      expshift = -expshift;
-
-      rhs = std::ldexp((double)rhs, expshift);
-
-      for (HighsInt i = 0; i != rowlen; ++i)
-        vals[i] = std::ldexp(vals[i], expshift);
+      scale(minAbsValue - epsilon);
     }
   } else {
     // the support is not integral, scale cut to have the largest coefficient
     // around 1.0
-    int expshift;
-    std::frexp(maxAbsValue - epsilon, &expshift);
-    expshift = -expshift;
-    // Don't scale the coefficients by more than +1024 (violations can increase)
-    expshift = std::min(10, expshift);
-    rhs = std::ldexp((double)rhs, expshift);
-
-    for (HighsInt i = 0; i != rowlen; ++i)
-      vals[i] = std::ldexp(vals[i], expshift);
+    scale(maxAbsValue - epsilon);
   }
 
   return true;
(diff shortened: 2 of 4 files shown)


fwesselm (commented):
Thank you for fixing this! 

- There is another occurrence of std::frexp in line 850. Is an upper bound needed there as well?

- Would it be OK to have a method for scaling (to avoid having to repeat the fix)? For example, I tried this [here](https://github.com/fwesselm/HiGHS/tree/cutScaling):

```
double HighsCutGeneration::scale(double val) {
  int expshift = 0;
  std::frexp(val, &expshift);
  expshift = -expshift;

  // Don't scale the coefficients by more than +1024 (violations can increase)
  expshift = std::min(10, expshift);

  // Scale rhs
  rhs = std::ldexp((double)rhs, expshift);

  // Scale row
  for (HighsInt i = 0; i != rowlen; ++i)
    vals[i] = std::ldexp(vals[i], expshift);

  // Return scaling factor
  return std::ldexp(1.0, expshift);
}
```

jajhall (commented):
Looks good to me, but @Opt-Mucca, what about the query from @fwesselm and suggestion that duplicated code is avoided by having a method to do the scaling?

Once this query is resolved, I'm happy for the PR to be merged

Opt-Mucca:
@fwesselm No problem! Interested to see if this helps minimise the amount of issues created by the fuzzer.

- Yes it should. Nice spot! I was wrongly interpreting that as part of the integral scale block.
- Am a big fan of that change! Is clean and easy. I'll add it now and also call it at line 850

fwesselm (approved):
Thanks!

Opt-Mucca:
Added suggested changes. Two things:

- I suspect some instances will be impacted heavily by this. Coin flip if it's for the better or for the worse. Would recommend performance testing this separately to the CMIR / ImpliedBound changes.
- In `preprocessBaseInequality` the `rhs` scaling was handled slightly different than the other cases. That is we were doing `rhs *= scale` instead of `rhs = std::ldexp((double)rhs, expshift)` (where `rhs` is type `HighsCDouble`). We now do the later case everywhere. I guess the scaling is now safer, but we lose precision when casting. Not sure if this was an active design choice and just wanted to note the change.

fwesselm:
> Added suggested changes. Two things:
> 
> * I suspect some instances will be impacted by this (likely some non-standard instances). Coin flip if it's for the better or for the worse. Would recommend performance testing this separately to the CMIR / ImpliedBound changes.
> * In `preprocessBaseInequality` the `rhs` scaling was handled slightly different than the other cases. That is we were doing `rhs *= scale` instead of `rhs = std::ldexp((double)rhs, expshift)` (where `rhs` is type `HighsCDouble`). We now do the later case everywhere. I guess the scaling is now safer, but we lose precision when casting. Not sure if this was an active design choice and just wanted to note the change. @fwesselm @jajhall (just so this second point isn't missed)
> 
> Edit: Some test are also currently failing

Yes, I would also be fine with having something like

```
double scale = std::ldexp(1.0, expshift);
rhs *= scale;

...
return scale;
```

in the new utility. The scaling of the coefficients would still be done using `std::ldexp`, not sure why they are handled differently.

The test failures seem to be due to the integration with the cMIR fixes that were just merged into `latest`.

Opt-Mucca:
> 
> Yes, I would also be fine with having something like
> 
> ```
> double scale = std::ldexp(1.0, expshift);
> rhs *= scale;
> 
> ...
> return scale;
> ```
> 
> in the new utility. The scaling of the coefficients would still be done using `std::ldexp`, not sure why they are handled differently.
> 

I think this is where I need some more C++ background. So the `rhs *= scale` here would be strictly better from a numerics POV? If so then I'd change it.

> The test failures may be due to the integration with the cMIR fixes that were just merged into latest.

I accidentally removed some initialisation of the `initialScale`. Have added it back. (Still may be a problem with the cMIR fixes)

Opt-Mucca:
@fwesselm no problem on the back and forth, especially if it comes out with a solution that convenient. I'll add that change now.

galabovaa (approved):

galabovaa:
I think these are already included in https://github.com/ERGO-Code/HiGHS/pull/2346 @jajhall

PR #2338, opened 2025-05-14
Fix bug in separateImpliedBounds

Opt-Mucca:
There is a bug when calculating a coefficient in one of the cuts in `separateImpliedBounds`. It only seems to affect implications of the form `x=1 => y >=M` and is fixed by rearranging the terms in one of the coefficients. 

Take the implication `x=1 => y >=3`, where `x in [0,1] && y in [0,4]`. The current code calculates the inequality `-y -3x <= 0`, which is true but is substantially weaker than it should be. The correct inequality should be `-y + 3x <= 0`.

This should hopefully improve performance by some minor amount.

### DIFF
--- a/highs/mip/HighsImplications.cpp
+++ b/highs/mip/HighsImplications.cpp
@@ -635,7 +635,7 @@ void HighsImplications::separateImpliedBounds(
           vals[0] = -1.0;
           inds[0] = implics[i].column;
           vals[1] =
-              globaldomain.col_lower_[implics[i].column] - implics[i].boundval;
+              implics[i].boundval - globaldomain.col_lower_[implics[i].column];
           inds[1] = col;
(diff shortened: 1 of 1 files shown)


BenChampion:
Looks good to me, but will let one of the others sign off just to be sure. (There's a small typo in your comment above. It should say implications of the form `x=1 => y >= M`. The code looks fine though.)

BenChampion:
(And nice spot, of course!)

galabovaa:
Looks good to me! Happy to merge once @jajhall has confirmed as well.

galabovaa (approved):

jajhall (approved):
No problem. It's a tightening, so only reason for testing is to observe any performance gain and to update the benchmark performance of latest. Since there seem to be a succession of such corrections that tighten cuts, I don't think that testing each for incremental performance gain is necessary.

PR #2339, opened 2025-05-15
FindCUDAConf on windows

galabovaa:
This only affects -DCUPDLP_FIND_CUDA=ON. 
For all of our test setups CUPDLP_FIND_CUDA was not required at all on Windows.
I modified it anyway, just in case there are problems with the default CMake find CUDA on Windows, possibly Windows server since CUPDLP_FIND_CUDA was required for linux servers. 
Users should set -DCMAKE_CUDA_PATH="..." when configuring CMake if they use -DCUPDLP_FIND_CUDA=ON


jajhall (approved):
I can believe that WIN32 and CUDA aren't things we want to combine, and otherwise you're just adding places to look

So, looks fine to me

galabovaa:
WIN32 in cmake is just checking for Windows, including Win64 😅 

I didn't even think to disable it on 32bit platforms yet :D 

On Windows, ENV{CUDA_HOME} is not propagating as expected but if CMAKE_CUDA_PATH is set it works fine. 

Thank you! Merging this

PR #2344, opened 2025-05-17
fix compiler warnings from HiGHS 1.10.0

svigerske:
This fixes again compiler warnings
- on defining public functions without declaring them first, by either making them static or adding the declaration, and
- on unused variables (`i_clock` is assigned by not used if `NDEBUG` is defined)

### DIFF
--- a/highs/lp_data/HighsInterface.cpp
+++ b/highs/lp_data/HighsInterface.cpp
@@ -674,10 +674,10 @@ HighsStatus Highs::addRowsInterface(HighsInt ext_num_new_row,
   return return_status;
 }
 
-void deleteBasisEntries(std::vector<HighsBasisStatus>& status,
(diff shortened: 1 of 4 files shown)


galabovaa:
Hi Stefan, 

Thank you for these! Just for reference, which compiler / OS combination do you use to check?

galabovaa (approved):

svigerske:
We compile HiGHS with GCC 14 with `-Wall -Wreturn-type -Wmissing-declarations -std=c++11 -Wno-unused-variable -Wno-unused-but-set-variable -Wno-comment` on Linux.

That we enable `-Wmissing-declarations` raised new warnings with every HiGHS update, which I then fix. But some of these `-Wno-...` are only added because it seemed a lot of work to fix all the warnings in the code.

This `HighsTimer.h` gets compiled in another code as well, one where we don't use `-Wno-unused-variable`.

Issue #2306, opened 2025-04-24
Large dual infeasibilities/residuals after postsolve


jajhall:
When postsolve is applied to accurate solutions from cuPDLP-C or IPX without crossover, sometimes this results in very large dual infeasibilities/residuals after postsolve. Indeed, this was probably the cause of the very large residuals in the issue raised by @odow 

The driver file [TestPostsolve.cpp](https://github.com/user-attachments/files/19897076/TestPostsolve.cpp.txt) allows postsolve to be tested without the overhead of solving the presolved model.

odow:
Here's an example
```
Maximize
 obj: x + y + 2 z
Subject To
 c1: x + y + z <= 4
 c2: x - y + 2 z <= 3
 c3: z <= 1
Bounds
 0 <= x <= 3
 0 <= y <= 10
 0 <= z <= 5
End
```

With `run_crossover=off`, `solver=ipm`:
```
% julia --project=/tmp/hgh scratch/run.jl scratch/max1.lp 
Running HiGHS 1.15.1 (git hash: 04024d701f): Copyright (c) 2026 under MIT licence terms
Includes third-party software components, see THIRD_PARTY_NOTICES.md for full details
Using BLAS: Apple Accelerate 
LP has 3 rows; 3 cols; 7 nonzeros
Coefficient ranges:
  Matrix  [1e+00, 2e+00]
  Cost    [1e+00, 2e+00]
  Bound   [3e+00, 1e+01]
  RHS     [1e+00, 4e+00]
Presolving model
2 rows, 3 cols, 6 nonzeros 0s
0 rows, 0 cols, 0 nonzeros 0s
Presolve reductions: rows 0(-3); columns 0(-3); nonzeros 0(-7) - Reduced to empty
Performed postsolve
WARNING: Solution optimality conditions
num/max      0 /        0 (relative      0 /        0) primal infeasibilities     (tolerance = 1e-07)
num/max      2 /        1 (relative      2 /    0.333)   dual infeasibilities     (tolerance = 1e-07)
num/max      0 /        0 (relative      0 /        0) primal residual errors     (tolerance = 1e-07)
num/max      3 /        4 (relative      3 /     1.33)   dual residual errors     (tolerance = 1e-07)
                                         1 /    0.909  P-D objective error        (tolerance = 1e-07)
WARNING: LP solver claims optimality, but with
WARNING:    num/max      2 /    0.333 relative   dual infeasibilities (tolerance = 1e-07)
[...]

PR #2346, opened 2025-05-18
Introduced relative KKT error data to `HighsInfo`, and using it to assess optimality claims from PDLP and IPM without crossover

jajhall:
PDLP and IPM without crossover terminate with optimality using relative measures of KKT conditions. Hence it's unreasonable to expect the absolute KKT conditions used to determine optimality in simplex and IPM with crossover - when a basis is available. However, using measures of the bounds (costs) of the LP in relative measures of primal (dual) KKT conditions can flatter the solution obtained (particularly in the case of PDLP).

Hence, this PR introduces relative KKT error data - available to the user via `HighsInfo` - that are computed using the optimal solution, and ignore large costs and bounds that play no role in the optimal solution. The measures of the costs and bounds are computed for bounds that are active, and costs for which the corresponding dual is small.

The fix in https://github.com/ERGO-Code/HiGHS/pull/1978 to remove residual errors (that, previously, were not being checked for) after IPM without crossover by transferring them to primal and dual activities, was a bad idea. If nothing else, it led to a lot of erroneous model status values of `HighsModelStatus::kUnknown` when the problem had been solved satisfactorily using relative tolerances. This correction of residuals has now been removed. The relative residuals are assessed using the new measures. The large residuals that were reported, probably correspond to #2306


jajhall:
The documentation of `HighsOptions` and `HighsInfo` has been updated to reflect the new options and data introduced. 

`FEATURES.md` has been update with respect to recent merged PRs, and _this_ PR!

jajhall:
I've added @odow as a reviewer, mainly to flag up this significant change.

odow:
If the Julia tests pass, I'm okay with it. We don't really have a good solver-independent way for users to check the quality of the returned solutions, but @joaquimg's tool that he is building should help.

galabovaa (approved):

PR #2348, opened 2025-05-19
Added MATLAB to `docs/src/interfaces/other.md`

jajhall:


jajhall:
Documenter failure is strange. It builds fine locally

rdzman:
I'm am glad to see this as I am quite interested in a full MATLAB and Octave interface to HiGHS on multiple platforms and just discovered [HiGHSMEX](https://github.com/savyasachi/HiGHSMEX/) myself 3 days ago.

- Is there a possibility of including this as an officially supported interface?
- I plan to follow up with the author directly, but do you know if anyone has used this to build a MEX file for Octave? An initial glance would indicate it is more comprehensive than what was is found [here](https://github.com/ERGO-Code/HiGHS/discussions/884).

jajhall:
> * Is there a possibility of including this as an officially supported interface?

The HiGHS team doesn't have the experience or resources to do this. However, our relationship with MathWorks may mean that they'd be interested in fixing it if necessary

> * I plan to follow up with the author directly, but do you know if anyone has used this to build a MEX file for Octave? An initial glance would indicate it is more comprehensive than what was is found [here](https://github.com/ERGO-Code/HiGHS/discussions/884).

I don't know. Since Octave is open-source, I was keen for it to use HiGHS, so reached out three years ago. However, it needed more help from the Octave community that didn't come. Since HiGHS is deeply committed to open-source, if someone from Octave reached out, we'd try to meet them half way.

PR #2354, opened 2025-05-21
remove highsparallel from bazel tests

galabovaa:
The failing test will need more time and the issue is likely to be in the test, rather than HiGHS


galabovaa:
Locally for me 500 tests pass and also 2000. Merging this anyway

PR #2362, opened 2025-05-24
Add setSparseSolution() to .NET interface

### LINKED ISSUE #2361: setSparseSolution() missing from .NET interface
The `setSparseSolution()` function has not yet been added to the .NET interface.

darrylmelander:
Resolves #2361

### DIFF
--- a/highs/interfaces/highs_csharp_api.cs
+++ b/highs/interfaces/highs_csharp_api.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Runtime.InteropServices;
 using System.Text;
@@ -460,6 +461,9 @@ public class HighsLpSolver : IDisposable
 
     [DllImport(highslibname)]
     private static extern int Highs_setSolution(IntPtr highs, double[] col_value, double[] row_value, double[] col_dual, double[] row_dual);
+
+    [DllImport(highslibname)]
+    private static extern int Highs_setSparseSolution(IntPtr highs, int num_entries, int[] index, double[] value);
 
     [DllImport(highslibname)]
     private static extern int Highs_getColsByRange(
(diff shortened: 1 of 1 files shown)


jajhall:
@darrylmelander Do you understand the reason for the CI failures?

jajhall:
> > @darrylmelander Do you understand the reason for the CI failures?
> 
> It can't find IReadOnlyDictionary, a type that was added in 2012. What version of .NET is it compiling against?

Can you help @galabovaa ?

galabovaa:
@darrylmelander We are using .NET 2.1 by default. Is it possible for you to modify the IReadOnlyDictionary to just a Dictionary, which is available in older .NET versions too? 

I will look at the other failures, they are likely to be unrelated. 

Thank you!

darrylmelander:
That's a little confusing, as there was no .NET Framework 2.1. There is a .NET Standard 2.1 and .NET Core 2.1, but both have IReadOnlyDictionary support and wouldn't cause this error. I would recommend explicitly targeting .NET standard 2.0, as it can be consumed by all currently supported versions of .NET Framework and .NET Core, and is the version targeted by the HiGHS nuget package.

At any rate, I *can* switch this particular argument to IDictionary, although it's not the solution I would choose.

Also, as I was looking into this I noticed that the windows-2019 github actions runner used by some of your build actions is about to be deprecated and removed. See https://github.com/actions/runner-images/issues/12045. You'll need to update those actions in the next month, regardless of what you do with this PR.

galabovaa:
Right, sorry, I did see the .NET standard when I looked in the cmake dotnet file. 

Thank you for the clarifications and tips!

I will try to add 2.0 explicitly and update the runners

jajhall:
Thanks @darrylmelander , this has been valuable for us!

galabovaa:
I checked again in the .csproj file used to generate the nuget package and it seems that .netstandard2.0 is already specified. The failing workflows were due to the 2019 runner, I checked updating them on another branch and all tests pass.  

unnamed

Merging this in latest. Thank you again, @darrylmelander!

PR #2370, opened 2025-05-28
Fix 1966

jajhall:
So, what failed in OR-Tools before failed again after the changes to assessing optimality after IPX without crossover and PDLP. Unfortunately the test that failed in OR-Tools before wasn't added to the unit test `test-1966` in HiGHS when @jajhall fixed things first time round!

@galabovaa I've not tested that this fixes the OR-Tools failure, as I can't link the `fix-1966` branch to your OR-Tools build, but I'm pretty confident that it will be OK

Note that after `fix-1966`, Hurst's second goal against Germany still stands, so England still win the World Cup


galabovaa:
Great! I will test this with OR-Tools

galabovaa:
Even greater for the World Cup!! :star:

galabovaa:
The test passed :partying_face:

PR #2371, opened 2025-05-28
Increase MPS precision when writing to file

Opt-Mucca:
This increases the precision of coefficients and bounds written to an MPS file to what is currently written to LP files. This was requested by the Xpress team.


jajhall (approved):
Fine by me, I guess we were just keeping consistency with fixed format MPS, but for no great reason.

rschwarz:
I believe that previously HiGHS already supported both the fixed MPS format (with truncated coefficients) and a free format, with a boolean flag. But that flag was only available in C++, not in the `highspy` API (and the default was fixed).

See discussion here: https://github.com/ERGO-Code/HiGHS/discussions/2059#discussioncomment-11485514

I bring this up because changing the default on that boolean flag might also help with longer row and column names.

PR #2373, opened 2025-05-29
Fix 2326

Opt-Mucca:
This MR does three things:
- Changes the order of some clauses that some if statements check. Each operation there is cheap and in general it doesn't matter, but it can be called millions of times, so extreme cases exists where this can help performance.
- Adds a limit on the number of vubs / vlbs (`x <= Mz + b`, where `x` is some variable, `y` is some binary, and `M /b` are some real values). I made the upper limit fairly large so I only see this affecting extreme cases, but the performance should still definitely be checked.
- There was an error in calculating the amount of "effort" / "work" a submip did. After this fix heuristics will essentially be given a larger budget as previously the "work" was being overestimated. I doubt the previous limit was being hit frequently, but for instances where it was then this is a potentially heavy performance change.


jajhall (commented):
Make sense, but needs regression tests to be run, please @fwesselm

jajhall:
I thought I'd fixed the formatting

jajhall:
> The documentation fix is awaiting review in [Documenter.jl](https://github.com/JuliaDocs/Documenter.jl/pulls)

Meanwhile, in #2360 I've switched off link checking

Opt-Mucca:
@jajhall testing by @fwesselm shows some minor performance improvements. Is always nice for a bug fix. I'd be for merging it now.
I'm going to leave a comment if we ever need to revisit this PR too: When I wrote this I decided on a big-M value of `5000000 + 10 * ncols` for how many items can be stored in the variable bound tree data structure. That helps avoid some worst-case scenarios, where the solve time is dominated by inserting elements into the tree. I believe the value's large enough to not affect any reasonable solve, although it's potentially possible that some instance out there might benefit from these extreme values.

PR #2375, opened 2025-05-29
Fix code to match comment

Opt-Mucca:
This is a relatively minor code fix, but it affects which variables get fixed by RINS / RENS, so will have big changes on plenty of instances.

There's no guarantee that this improves performance, and if it doesn't then the complement of this should be done (remove the redundant logic and comment). Hopefully it does though as the original intention seems to sound decent.

### DIFF
--- a/highs/mip/HighsPrimalHeuristics.cpp
+++ b/highs/mip/HighsPrimalHeuristics.cpp
@@ -444,7 +444,7 @@ retry:
           fixval = std::ceil(fracval);
         else if (rootchange <= -0.4)
           fixval = std::floor(fracval);
-        if (mipsolver.model_->col_cost_[col] > 0.0)
+        else if (mipsolver.model_->col_cost_[col] > 0.0)
           fixval = std::ceil(fracval);
         else if (mipsolver.model_->col_cost_[col] < 0.0)
           fixval = std::floor(fracval);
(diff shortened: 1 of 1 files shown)


jajhall (commented):
Yes, small change, but we'll need a separate testing run for this, so that its effects aren't obscured by the effects of #2373 (and vice versa)

fwesselm:
@Opt-Mucca, I am going to qualify this now and wanted to ask whether it would be worthwhile to add the following function (and not have the duplicate code)? 

```
static double getFixValue(double rootchange, double fracval, double cost) {
  // reinforce direction of this solution away from root
  // solution if the change is at least 0.4
  // otherwise take the direction where the objective gets worse
  // if objective is zero round to nearest integer
  if (rootchange >= 0.4)
    return std::ceil(fracval);
  else if (rootchange <= -0.4)
    return std::floor(fracval);
  else if (cost > 0.0)
    return std::ceil(fracval);
  else if (cost < 0.0)
    return std::floor(fracval);
  else
    return std::floor(fracval + 0.5);
}
```

Opt-Mucca:
I'm not against putting it into a function. Want me to do this now or after the experiments?

fwesselm:
> I'm not against putting it into a function. Want me to do this now or after the experiments?

Later is fine, thank you. I'll add a note here when I have the results.

fwesselm:
> > I'm not against putting it into a function. Want me to do this now or after the experiments?
> 
> Later is fine, thank you. I'll add a note here when I have the results.

105 instances (out of 854) are affected by these changes in terms of numbers of branching nodes explored. The ratios between the shifted geometric means are around 0.99 for solution times and branching nodes. Therefore, I think it would make sense to merge this.

PR #2376, opened 2025-05-29
Clean up for release 

galabovaa:
GPU objective values recently changed due to tolerances update
- Cleaned up in CMake so `ctest` passes with GPU ON

Build directory for Windows MSVC now `Release/bin` replacing `bin/Release`
- fixes new bug I got with `TARGET_FILE` in CMake
- consistent with OR-Tools, a patch there should not be required any more 

Clean up of `HConfig.h` and addition to `HCheckConfig.h`
- one less patch in vcpkg 
- `HCheckConfig.h` should be added to Catch2 test files when files are being read from the `check/` 

Fixes a warning by adding default: to a switch

### DIFF
--- a/.github/workflows/test-csharp-win.yml
+++ b/.github/workflows/test-csharp-win.yml
@@ -53,4 +53,4 @@ jobs:
         working-directory: ${{runner.workspace}}/build
         run: |
           ls 
-          ./bin/Debug/csharpexample.exe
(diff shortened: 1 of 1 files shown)


jajhall (approved):
All good. I must have been looking at @BenChampion 's PR when I said this was already in [docs-tolerances](https://github.com/ERGO-Code/HiGHS/tree/docs-tolerances)

PR #2377, opened 2025-05-30
Clean up MIP heuristic options

BenChampion:
I noticed our option for using Feasibility Jump had a name inconsistent with those for other heuristics. I've renamed the Feasibility Jump option accordingly. (Theoretically this could affect anyone who used the old name in an options file. Since this hasn't been in a release yet, and practically speaking I don't think anyone is changing it from the default, I think it's fine. )

While I was fixing this, I noticed the other heuristic options were uninitialized. I also cleaned up the documentation strings and reordered things for consistency. (In particular, I removed the default option values within the documentation strings. The HTML documentation displays the default values, and if these were ever to change it would be easy to end up with a stale string.)


jajhall (approved):
FJ is in v1.10.0, so renaming `mip_allow_feasibility_jump` to `mip_heuristic_run_feasibility_jump` is a breaking change, but few will be using it, and documenting he change in `FEATURES.md` should be sufficient.

`docs/src/options/definitions.md` will also need to be changed - it's not (quite) auto-generated - but I don't want to create conflicts with changes in #2360.

Note that I'll complete #2360 over the weekend, as it must be merged before v1.11.0 can be released

BenChampion:
@jajhall 

> FJ is in v1.10.0 [...]

I checked, and in fact my statement above stands. FJ didn't quite make it into v1.10.0

I'll put up a PR with a `FEATURES.md` blurb. I can also make the necessary changes to `docs/src/options/definitions.md`

(I'll base it on your branch for #2360 so there won't be conflicts.)

jajhall:
> I checked, and in fact my statement above stands. FJ didn't quite make it into v1.10.0

Indeed, sorry. In this case we need to add FJ to `FEATURES.md`

> I can also make the necessary changes to docs/src/options/definitions.md
(I'll base it on your branch for https://github.com/ERGO-Code/HiGHS/pull/2360 so there won't be conflicts.)

OK: this is how I do this

- Uncomment the following line in `app/RunHighs.cpp`

>   // highs.writeOptions("Options.md");

- Compile and run with any of the test instances to create `Options.md`
- Run

> meld Options.md ../docs/src/options/definitions.md

- Merge the necessary changes from `Options.md` into `../docs/src/options/definitions.md`
- Re-comment the uncommented line in `app/RunHighs.cpp`

jajhall:
I've just pushed my most recent changes of `docs-tolerances`  to #2360 

I won't want to work on this branch until tomorrow (at least) as I'm marking LPMS

BenChampion:
> Indeed, sorry. In this case we need to add FJ to `FEATURES.md`
#2378 is the PR (I could include it here, but I lean towards keeping PRs focused)

> OK: this is how I do this
Nice, thanks. I'll do this in the other PR. (And back out the FEATURES.md changes here.)

BenChampion:
> @BenChampion I've just pushed my most recent changes of `docs-tolerances` to #2360
> 
> I won't want to work on this branch until tomorrow (at least) as I'm marking LPMS

@jajhall: Noted, I'll wait until you've finished, then I can rebase my other PR's branch on top of that before running the steps you describe.

Opt-Mucca (approved):

docs/src/options/definitions.md
```diff
@@ -185,11 +185,6 @@
 - Type: boolean
 - Default: "true"
 
-## solution\_file
```
BenChampion, line 188: In case you're wondering about this deletion: this option is documented below (line 230)

BenChampion:
@jajhall documentation was easy enough to rebuild; let me know when your changes are in `latest` and I can rebase, rebuild again and handle any conflicts. (Unless you want to do so and I can merge this PR)

PR #2380, opened 2025-06-02
Describe supported model file formats in the documentation

pierre-haessig:
Hello,
While reading the documentation about the HiGHS executable (https://ergo-code.github.io/HiGHS/dev/executable/), I noticed that the supported file format are not described there.

I found in the [README](https://github.com/ERGO-Code/HiGHS/blob/master/README.md#running-highs) that MPS and CPLEX LP file formats are supported. I've added this to the doc.

To make the document a bit more easy for newcomers, I've added links to the file format description. I've used the same as used by [JuMP](https://jump.dev/JuMP.jl/stable/manual/models/#Write-a-model-to-file):
- MPS: Wikipedia https://en.wikipedia.org/wiki/MPS_(format)
- LP: MOSEK API doc https://docs.mosek.com/latest/capi/lp-format.html

Finally, while reviewing the entire doc, I noticed the following sentence:

> When HiGHS is run from the command line, some fundamental option values may be specified directly. **Many more may be specified via a file**.

and I'm not sure I understand what it means. Does this refer to the `--options_file file` option? Is that's correct, I can add an extra commit for this, but then I'd like to describe the option file format (or put a link if it's elsewhere in the doc).

### DIFF
--- a/docs/src/executable.md
+++ b/docs/src/executable.md
@@ -1,9 +1,13 @@
 # [Executable](@id executable)
 
-For convenience, the executable is assumed to be `bin/highs`.
+HiGHS can run as a standalone program with a command-line interface. It solves an optimization problem provided by a model file. HiGHS supports the following model file formats:
+
+- [MPS file format](https://en.wikipedia.org/wiki/MPS_(format))
+- (CPLEX) [LP file format](https://docs.mosek.com/latest/capi/lp-format.html)
 
 ### Running the executable
 
+For convenience, the executable is assumed to be `bin/highs`.
 The model given by the MPS file `model.mps` is solved by the command:
 
 ```shell


pierre-haessig:
> Thanks, I'll add a comment about the options file

OK. I see that you have added the comment:

> The list of options section gives a full list of options, and the format in which they are specified.

However, I think it is still missing the description of how these options should be specified in a file. Is it one option per line or comma separated? Perhaps adding an example of a config file snippet in a code block would be the easiest explanation.

jajhall:
> > Thanks, I'll add a comment about the options file
> 
> OK. I see that you have added the comment:

Not yet - you're looking at the original comment

> > The list of options section gives a full list of options, and the format in which they are specified.
> 
> However, I think it is still missing the description of how these options should be specified in a file. Is it one option per line or comma separated? Perhaps adding an example of a config file snippet in a code block would be the easiest explanation.

I'll soon push the description, but thanks for suggesting that one option per line (rather than comma separated) should be clarified

pierre-haessig:
OK I see.

In the mean time, I just came across the [Options/Introduction page](https://github.com/ERGO-Code/HiGHS/blob/latest/docs/src/options/intro.md) which pretty much does the job. I guess that adding a link to it from the Executable page would solve things.

The only missing bit I see is what I imagined as a short snippet of option file. It could be placed in the https://ergo-code.github.io/HiGHS/dev/options/intro/#Options-file section.

PR #2384, opened 2025-06-03
Tolerance fixes

jajhall:
Added `HighsOption::qp_regularization_value` so that QP regularization can be switched off in `TestQpSolver.cpp`. 

Hence exact solutions of simple QPs are now obtained, validating code to compute dual objective value for QPs (and primal-dual objective error), and tightening up QP unit tests; fixed error in `HighsHessian::product`.


jajhall:
Thanks @filikat

Issue #2385, opened 2025-06-03
HiGHS doesn't accept spaces in options/solution filepath parameters (Windows)


NPC:
Hello,

A new issue since version 1.10.0. If the path to options/solution file (passed in the command line to highs.exe) contains spaces, then I get these errors:
```
--options_file: Multiple files not implemented.
--solution_file: Multiple files not implemented.
```

Sample command line (I am careful to wrap paths in double quotes, but it doesn't help):
```
highs.exe --parallel=on "Log\LP-problem.mps" --options_file="D:\work\trunk\Middle Layer\NET\highs-options.txt" --solution_file="D:\work\trunk\Middle Layer\NET\Log\lp-solution.txt"
```

I can't change paths in our project (especially in the customer environment), so may have to revert to the previous version.

(this has to do with [this change](https://github.com/ERGO-Code/HiGHS/commit/9fd975b045bb12a6df5f7be4d582a978a6c4ea85) in `checkSingle` in `HighsRuntimeOptions.h`, right?)

jajhall:
That's unfortunate. I'm surprised that it worked before @galabovaa 

How do you suggest we identify the case where someone tries to specify multiple files?

Shame you can't remove the spaces in the file path/name.

jajhall:
Closed by #2387

PR #2387, opened 2025-06-04
CLI11 fix spaces

galabovaa:
Closes https://github.com/ERGO-Code/HiGHS/issues/2385

CLI11 handles multiple files with 
```
catch (const CLI::ExtrasError& e) 
```
and 
```
catch (const CLI::ArgumentMismatch& e) 
``` 
in `RunHighs.cpp`,  so `checkSingle` is not required in `HighsRuntimeOptions.h`


jajhall (approved):
Beautiful!

I've checked it locally

I did wonder about having a test to verify that this facility isn't broken in future. I guess that we have some "command line" CI tests to which is could be added

PR #2389, opened 2025-06-05
Mention Rust bindings in the documentation

KnorpelSenf:
There are excellent Rust bindings available at https://github.com/rust-or/highs. That's a selling point for HiGHS so it makes sense to mention this.


jajhall:
Thanks, I'm making mods to documentation now, so have added Rust to the list on docs/src/index.md

Rust is also mentioned on

docs/src/interfaces/other.md

KnorpelSenf:
Thanks!

PR #2397, opened 2025-06-06
Fix 2388

### LINKED ISSUE #2388: Presolve: HiGHS v1.10.0 incorrectly declares infeasibility
During presolve, HiGHS v1.10.0 declares as infeasible a MILP that can be solved by:

- HiGHS v1.7.0:  [highs_v170__log_optimal.txt](https://github.com/user-attachments/files/20612196/highs_v170__log_optimal.txt)
- Gurobi: [gurobi_v1103_log_optimal.txt](https://github.com/user-attachments/files/20612200/gurobi_v1103_log_optimal.txt)
- CPLEX

This is the model file:  [Infeasible_presolve_highs.txt](https://github.com/user-attachments/files/20612177/Infeasible_presolve_highs.txt)

HiGHS v1.10.0 is able to solve the problem under either of the following conditions:
- Presolve is deactivated.
- Some unbounded variables in the problem are explicitly declared as _Free_. For instance in [Feasible_presolve_highs.txt](https://github.com/user-attachments/files/20612205/Feasible_presolve_highs.txt), the variable _x49_ is now declared as free, yet other unbounded variables, such as _x11_ and _x12_, are not.

fwesselm:
Fix #2388 
- Use existing utility methods to change column bounds in `HPresolve::detectParallelRowsAndCols`.
- The current code does not update the lower and upper bounds on constraint activities correctly for the reproduction example.
- Added a test point.


jajhall:
I've just spotted that with `master`, using presolve off and presolve_reduction_limit=0, HiGHS gets an objective value of 1, which seems to be independent of a bug in presolve.

With `fwesselm:fix-2388-2`, this doesn't happen, but I wonder how fixing a presolve bug can cause this

PR #2399, opened 2025-06-07
fix new compiler warnings

svigerske:
There was a warning on a maybe uninitialized variable, which seems a false positive.

The other is about passing `this` to a lambda which does not need access to `this`. It raised an unused lambda capture warning from clang.


jajhall (approved):
Thanks @svigerske

PR #2407, opened 2025-06-11
Adds StrongCG cuts

Opt-Mucca:
While running `cmirCutGenerationHeuristic`, HiGHS now also tries to generate a strongCG cut.

Summarised current logic:
- Given a row aggregation, HiGHS tries multiple scale factors (`delta`) and for each generates the potential CMIR cut. It keeps track of the `delta` that produced the most efficacious cut, and at the end computes the actual cut using higher precision.

Changed logic:
- Given the best `delta` for the CMIR generation, now also compute the associated strongCG cut (in cases where it can be computed). If it is more efficacious (above some threshold of the CMIR cut) then use that instead.

Remaining potential changes (would require multiple runs)
- In theory we could be adding both the best CMIR cut and the best strongCG cut. One doesn't necessarily dominate the other, so having both can be helpful. It would require a design change of how cuts are generated though, and I didn't want to do this before getting more familiar with the code.
- I am only trying to compute the strongCG cut for the best `delta`. There may be a better cut that we are missing that uses a `delta` that was sub-optimal for CMIR. (No gut feeling on how helpful this would be)

Running on smaller instances 10s-600s this seems to have a nice improvement. It should not be merged before being tested on something larger. 

Edit: There were still some numerical issues because I hadn't added appropriate epsilons when taking `ceil`. I've removed the point on playing with tolerances a bit after having done some additional experiments.


jajhall (approved):
I don't like the use of all lower case identifier names like `oneoveroneminusf0` or camel case identifiers like `minEfficacy` - except for constants that are `kConstantValue` - all as per the Google style guide.

Otherwise, subject to regression testing, this is fine

highs/mip/HighsCutGeneration.cpp
```diff
@@ -500,6 +500,8 @@ bool HighsCutGeneration::separateLiftedMixedIntegerCover() {
   return true;
 }
 
+static double fast_ceil(double x) { return (int64_t)x + (x > (int64_t)x); }
```
fwesselm, line 502: Sorry, this is a stupid question, but it it really necessary to have separate `fast_ceil` and `fast_floor` functions? Would `ceil` and `floor` be considerably slower?

highs/mip/HighsCutGeneration.cpp
```diff
@@ -687,13 +692,64 @@ bool HighsCutGeneration::cmirCutGenerationHeuristic(double minEfficacy,
...
 
+  // Calc strongcg cut as potential alternative to CMIR
+  if (bestefficacy <= 0 || rowlen == 0) {
+    strongcg = false;
+  }
+  if (strongcg) {
+    double delta = bestdelta;
+    double scale = 1.0 / delta;
+    double scalrhs = double(rhs) * scale;
+    double downrhs = fast_floor(scalrhs);
```
fwesselm, line 704: I am wondering why no tolerance is needed when the right-hand side is rounded (similar question for cMIR code). I would have expected:

`double downrhs = floor(scalrhs + feastol);`

highs/mip/HighsCutGeneration.cpp
```diff
@@ -687,13 +692,64 @@ bool HighsCutGeneration::cmirCutGenerationHeuristic(double minEfficacy,
...
+      for (HighsInt j : integerinds) {
+        double scalaj = vals[j] * scale;
+        double downaj = fast_floor(scalaj + kHighsTiny);
+        double fj = scalaj - downaj;
+        if (fj <= f0 + 1e-6) {
+          double aj = downaj;
+          updateViolationAndNorm(j, aj, viol, sqrnorm);
+        } else {
+          double pj =
+              fast_ceil(k * (fj - f0) * oneoveroneminusf0 - (10 * epsilon));
```
fwesselm: Would just having `feastol` here be too conservative? I suspect this could avoid cutoff issues...

fwesselm (commented):
Thank you for adding this, @Opt-Mucca. I have added some comments regarding the tolerances used in `HighsCutGeneration::cmirCutGenerationHeuristic`.

highs/mip/HighsCutGeneration.cpp
Opt-Mucca: I've got genuinely no clue here. I would say that even if it's considerably slower, the function takes up an incredibly insignificant amount of the solve time. I simply followed the existing standard when making this function.

highs/mip/HighsCutGeneration.cpp
Opt-Mucca: I guess the logic is that these cases have been filtered previously by constantly checking if
```
double f0 = scalrhs - downrhs;
    if (f0 < f0min || f0 > f0max) continue / flipComplementation;
```
Edit: My logic above is correct. you're confident that `rhs` is always at least `0.05` away from an integer

highs/mip/HighsCutGeneration.cpp
Opt-Mucca: I wouldn't be against increasing this value (although for these cuts I observed a clear performance improvement line the smaller the tolerances got).
I'm against tying it to `feastol` though. I don't think users intend on getting numerically less stable cuts when they make the tolerances tighter.

Edit: Can try. Other bits of the code are also using feastol for similar purposes. The thing I want to avoid most is overestimating the cut strength, and having the cut vanish when we go to higher precision. (Because then we'd deny the CMIR cut)

Opt-Mucca: `feastol` is now used here

Opt-Mucca:
Current status: The change is correct (subject to occasional noise on numerically troublesome instances). The average performance improvement is slightly negative however, with the change helping many instances (especially larger ones), but harming too many others (mostly smaller ones, but also some larger ones). This should not be merged yet.
At some point we need to test multiple combinations of "when should this be applied (e.g. ignore for quicker cuts)?", "What should the requirement of the score to cmir be (e.g. use relative scoring)?", and "what tolerance values are safe while maximising performance?".

jajhall:
Is this PR still active @Opt-Mucca ?

Opt-Mucca:
@jajhall This is similar to the flow-cover PR, but even more noisy (with less potential and less tuning possibilities). I'd be in favour of bench-marking this one last time and then closing / merging it or putting it behind a default-disabled-parameter (is really not that many lines, and will be even less if I revert the attempt at complementing everything).

jajhall:
> @jajhall This is similar to the flow-cover PR, but even more noisy (with less potential and less tuning possibilities). I'd be in favour of bench-marking this one last time and then closing / merging it or putting it behind a default-disabled-parameter (is really not that many lines, and will be even less if I revert the attempt at complementing everything).

So my comment is the same, advising latest to be merged into it to avoid conflicts and have CI tests pass

Opt-Mucca:
I'm closing this PR. This branch is still active, and following some advice from guests at SIAM, I've tried to strengthen the cuts further. It is a much smaller change than the flow cover cuts and has extremely low overhead. If I can get this performance improving on the new cluster, then I will re-open a PR.

Issue #2411, opened 2025-06-12
macos: Wrong usage of MACOSX_RPATH (when using -DFAST_BUILD=OFF)


Mizux:
In your CMake based build, `MACOSX_RPATH` is wrongly used here:
https://github.com/ERGO-Code/HiGHS/blob/364c83a51e44ba6c27def9c8fc1a49b1daf5ad5c/highs/CMakeLists.txt#L46

note: Spotted in https://github.com/google/or-tools/issues/4674#issuecomment-2963390936

## Protocol

```sh
cmake -S. -Bbuild -DFAST_BUILD=OFF
cmake --build build --config Release -j8
# Check path inside the shared library
otool -L build/lib/libhighs.dylib  
```

## Observed

```
build/lib/libhighs.dylib:
**/Users/corentinl/work/HiGHS/build/lib/libhighs.1.11.dylib (compatibility version 1.11.0, current version 1.11.0)**
/usr/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.12)
/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 1900.180.0)
/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1351.0.0)
```

## Expected

```
**@rpath/libhighs.1.11.dylib (compatibility version 1.11.0, current version 1.11.0)**
```

## Explanation

1. `MACOS_RPATH` is a boolean property which must be set to `TRUE` or `FALSE`
   If TRUE, the default, cmake will use @rpath as directory portion (aka prefix) of the  install_name (otool LC_ID_DYLIB)
   note: CMP0042 set it to TRUE by default
2. Here, They are storing a "path" (i.e. a `string`) into a variable which should be use to store `TRUE` or `FALSE` and CMake `if(string)` is always evaluated to `FALSE` (unless store a "TRUE" constant which is not the case here) leading to **NOT** using `@rpath` directory prefix...
3. Your projects already use `set(CMAKE_MACOSX_RPATH TRUE)` in root CMakeLists.txt, which already set  `MACOS_RPATH` to `TRUE`, so you should just stop playing with `MACOSX_RPATH` *Target* property imho...
4. To change the runpath (otool LC_RPATH) (e.g. to set it to @loader_path) you must use the INSTALL_RPATH property.

ref:
ref: https://cmake.org/cmake/help/latest/command/if.html#string
[...]

PR #2412, opened 2025-06-12
Fix MACOSX_RPATH usage (#2411)

Mizux:


Mizux:
> Thanks. 
> 
> The formatting changes are a distraction: am I right in thinking that the only meaningful change is in line 46 of highs/CMakeLists.txt?
> 
> Please submit PRs to `latest`, not `master`

Yes, that's why I've sent two commits.
Note: my vim setup automatically trims trailing whitespace upon saving....

Granted for latest branch ! Currently commuting -> AFK

jajhall:
@galabovaa is away this week, so her review won't be until next week at the earliest

galabovaa:
Great, thank you @Mizux! I had not changed the rpath settings of FAST_BUILD OFF. Over time, we should migrate the use cases to use FAST_BUILD on, which may require some more CMake edits. It was a variable I used for updating CMake without breaking any applications.

Issue #2423, opened 2025-06-19
Wrong assertion in HighsSparseMatrix::getCol


samiran-kawtikwar:
The assertion check at: https://github.com/ERGO-Code/HiGHS/blob/364c83a51e44ba6c27def9c8fc1a49b1daf5ad5c/highs/util/HighsSparseMatrix.cpp#L530
should be `assert(iCol >= 0 && iCol < this->num_col_)` (instead of num_row_)

Tagging @galabovaa as mentioned in Contributing instructions

jajhall:
Well spotted! Thanks

jajhall:
Closed by #2425

PR #2425, opened 2025-06-20
Fixed one-line error

jajhall:
Fixes a simple 1-line error


jajhall:
This will close #2423

odow:
Does this indicate that we're missing a lot of tests for this area of the code? I guess all the JuMP tests query row-wise, not col-wise.

PR #2426, opened 2025-06-23
Simplify some for loops (part 1)

fwesselm:
Some of the raw `for` loops can be simplified by rewriting them as ranged-based `for` loops. This does not affect behavior, but simplifies the code and makes it more readable.

### DIFF
--- a/highs/io/HMpsFF.cpp
+++ b/highs/io/HMpsFF.cpp
@@ -108,8 +108,8 @@ FreeFormatParserReturnCode HMpsFF::loadProblem(
 
   // Only set up lp.integrality_ if non-continuous
   bool is_mip = false;
-  for (const auto& integrality : col_integrality) {
-    if (integrality != HighsVarType::kContinuous) {
+  for (const auto& var_type : col_integrality) {
+    if (var_type != HighsVarType::kContinuous) {
       is_mip = true;
       break;
     }


jajhall (approved):
It's a matter of opinion whether these changes simplify the code, as it's a loop syntax I don't fully understand. Whilst it uses fewer characters, what's being executed is less transparent. I've also found it frustrating when I've been debugging code written this way. It's fair to say that I won't be adopting this style. 

However, if you're confident that you've not introduced any errors in making cosmetic changes, I won't veto the PR

fwesselm:
Range-based for loop syntax is described here, for example:

https://en.cppreference.com/w/cpp/language/range-for.html

jajhall:
> Range-based for loop syntax is described here, for example:
> 
> https://en.cppreference.com/w/cpp/language/range-for.html

If you're looping over the values in a vector of costs, how do you print the index corresponding to the value used in each pass through the loop - for debugging?

For example, what's the index of each name when you loop as follows?

    for (const auto& name : lp.row_names_) f << name << std::endl;

I guess it's a const reference to avoid memory allocation?

fwesselm:
> > Range-based for loop syntax is described here, for example:
> > https://en.cppreference.com/w/cpp/language/range-for.html
> 
> If you're looping over the values in a vector of costs, how do you print the index corresponding to the value used in each pass through the loop - for debugging?
> 
> For example, what's the index of each name when you loop as follows?
> 
> ```
> for (const auto& name : lp.row_names_) f << name << std::endl;
> ```
> 
> I guess it's a const reference to avoid memory allocation?

Correct, this is a const reference to avoid potential copying.

Regarding your question on debugging, the range-based loops are not designed for this. Ranged-based notation is just syntactically easier and avoids handling loop indices or iterators (which may be error-prone).

So, as far as I understand, there is no direct way to get the index for a given element. However, there are some potential workarounds if the underlying container is contiguous (by computing the distance of the memory address of the respective element from the beginning of the container). `C++23` added a nice way to get the index and corresponding element:

https://en.cppreference.com/w/cpp/ranges/enumerate_view.html

jajhall:
> Regarding your question on debugging, the range-based loops are not designed for this. 

This is sufficient reason not to use ranged-based loops

I see that this PR is "Part 1". Please don't prepare "Part 2" until we've decided whether to accept "Part 1"

galabovaa:
Looks good to me! 

For debugging, we could investigate by changing the loop back to the previous, less-compact way, which allows us to see the offending indices

fwesselm:
> > Regarding your question on debugging, the range-based loops are not designed for this.
> 
> This is sufficient reason not to use ranged-based loops
> 
> I see that this PR is "Part 1". Please don't prepare "Part 2" until we've decided whether to accept "Part 1"

Ok, I see. I mistakenly thought that simplifying these loops would be non-controversial (since most C++ guidelines advise to avoid raw loops, if possible). I have a [second set of changes](https://github.com/ERGO-Code/HiGHS/compare/latest...fwesselm:HiGHS:rawLoops2) that I will just shelve then.

galabovaa:
Looking at your second set of changes, most of them are non-controversial. Only for a few would we possibly need to debug manually, if something goes wrong. 

I am happy to have the second set of changes merged as well. As long as @jajhall reviews them as well

jajhall (commented):
OK, I'll move with the times and accept code that has to be modified in order to study it or debug it

Issue #2414, opened 2025-06-12
Build-wheel jobs are failing


Opt-Mucca:
cibuildwheel just pushed a new major version. The jobs are automatically downloading the new version (3.0.0) and our pyproject.toml is now longer valid.

The (proper) fix:
- They've deprecated one of the parameters we were using. I tried replacing it with what the documentation said was the new replacement (`enable = ["cpython-freethreading"]`), however that seems to have wider consequences. Specifically, it does not work off the bat with 32 bit.

The easier temporary fix if this needs to be handled quickly:

- Add a version requirement of `<3` when `pip` installing `cibuildwheel` in the GitHub actions for now. That should be a hacky 15 line change that would kick the problem down the road.

jajhall:
One for you @galabovaa, in due course

jajhall:
Closed by #2428

PR #2428, opened 2025-06-23
Cibw 3.0.0

galabovaa:
Update python build and CI: new version of cibuidwheel 

Closes https://github.com/ERGO-Code/HiGHS/issues/2414

### DIFF
--- a/.github/workflows/build-wheels-push.yml
+++ b/.github/workflows/build-wheels-push.yml
@@ -1,12 +1,12 @@
 name: build-wheels-push
 
 # on: []
-on: push
+# on: push
  
-# on:
-#  release:
-#    types:
-#      - published
+on:
+ release:
+   types:
+     - published
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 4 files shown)


Opt-Mucca:
I think I'm missing something.... The only diff between the logic of the 32-bit builds is the line (`environment: pypi`)? Do I have that right? (Far from an expert on GitHub actions)

As for the 32-bit fails I had when I was playing around last week: Is the `archs = ["auto64", "auto32"]` now skipping building on architectures that would have been failing (because of the environment or available machines)? Or the separate 32-bit logic and new include fix everything? Either way, very cool fix.

galabovaa:
> I think I'm missing something.... The only diff between the logic of the 32-bit builds is the line (environment: pypi)? Do I have that right? (Far from an expert on GitHub actions)

Well spotted, we actually need the environment for the push to pypi! Closing this, will re-open with a simplified version. I made them into separate jobs because I was trying to pass extra CIBW arch arguments for the 32bit build, but it may not be necessary now that we have archs = ["auto64", "auto32"] . 

One issue was that in the new version of cibw, 32bit is no longer built by default, hence we need to add it to archs. There are two ways to do this: from pyproject and from environment variables. So, the separate jobs were needed for the latter approach, but I got the pyproject one to work.

Issue #2042, opened 2024-11-16
Interrupt solving from .NET and show result


josef-hak:
Please, how can I interrupt ongoing solver run when calling solver from .NET? I found there are some callbacks but I haven't found how could I leverage them from .NET. I would very appreciate. Trying to integrate Highs to our universal free optimization tool [JD](https://github.com/coolfreetool/JD/pull/10). Thanks in advance for any help!

galabovaa:
Added this to feature requests, still on the to-do list!

https://github.com/ERGO-Code/HiGHS/issues/3225

PR #2433, opened 2025-06-24
Callbacks as CSharp Events

darrylmelander:
Give access to some callbacks in C#, as events on the C# HighsLpSolver class. Only a handful of callbacks are made available by this commit. The callbacks, and the new events they map to, are as follows:

    kCallbackLogging => LogMessageReceived
    kCallbackMipLogging => MipStatusReported
    kCallbackMipImprovingSolution => MipImprovingSolutionFound
    kCallbackMipInterrupt => MipInterruptCheck
    kCallbackSimplexInterrupt => SimplexInterruptCheck
    kCallbackIpmInterrupt => IpmInterruptCheck

Clients subscribe to and unsubscribe from these events as they would for any other event, using syntax such as `solver.LogMessageReceived += MyLoggingFunc`. Event data only includes what's relevant to the type of event. The `*InterruptCheck` events allow you to set the interrupt flag, the others don't.

The events in this PR address #2042.

### DIFF
--- a/highs/interfaces/highs_csharp_api.cs
+++ b/highs/interfaces/highs_csharp_api.cs
@@ -73,6 +73,31 @@ public enum HighsIntegrality
     kImplicitInteger = 4,
 }
 
+/// <summary>A category of log message</summary>
+public enum HighsLogType
+{
+    Info = 1,
+    Detailed,
+    Verbose,
+    Warning,
+    Error
+}
+
+/// <summary>A category of callback</summary>
+internal enum HighsCallbackType
+{
+    Logging = 0,
+    SimplexInterrupt = 1,
+    IpmInterrupt = 2,
+    MipSolution = 3,
+    MipImprovingSolution = 4,
+    MipLogging = 5,
+    MipInterrupt = 6,
+    MipGetCutPool = 7,
+    MipDefineLazyConstraints = 8,
+}
+
+
 public class HighsModel
 {
     public HighsObjectiveSense sense;
@@ -180,9 +205,25 @@ public class HighsLpSolver : IDisposable
 {
     private IntPtr highs;
 
+    /// <summary>Read-only access to Highs instance pointer</summary>
+    /// <remarks>Allows sub-classes to do meaningful things using the HiGHS C API.</remarks>
+    protected IntPtr HighsObject => this.highs;
+
     private bool _disposed;
 
     private const string highslibname = "highs";
+
+    /// <summary>Signature of functions that can be called by HiGHS when callback events occur</summary>
+    private delegate void CallbackDelegate(
+        HighsCallbackType cbType, IntPtr messagePtr, [In] ref HighsCallbackDataOut cbDataOut,
+        ref HighsCallbackDataIn cbDataIn, IntPtr cbUserData);
+
+    /// <summary>Pointer to function that is called when HiGHS callbacks occur</summary>
+    private CallbackDelegate _cbDelegate;
+
+    /// <summary>A C function pointer to the event-generating callback delegate</summary>
+    /// <remarks>This property's primary purpose is to improve ability to test callback-triggered events</remarks>
+    protected IntPtr CallbackFunctionPtr => Marshal.GetFunctionPointerForDelegate(_cbDelegate);
 
     [DllImport(highslibname)]
     private static extern int Highs_call(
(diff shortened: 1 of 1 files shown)


darrylmelander:
I'm submitting this as a draft PR to get feedback from the HiGHS team, and to give me more time to run it through its paces.

In particular I'd like to discuss testing of this particular feature. I'd love for it to be included in CI so we can detect if future callback changes break the integration. Locally, I tested by adding a function to the C interface that lets you trigger callbacks with supplied data to a specified callback function. This function is only useful for testing purposes and wouldn't belong in the released C interface. I also added a C# test library that calls the callback trigger function. Without testing for specific callbacks with specific data, we can't be confident data is getting across the C/C# boundary correctly. My testing approach is a lot more intrusive than just adding some code to existing test files, so I'd be interested in hearing how the team would like to approach regression testing for this feature.

galabovaa:
Thank you, @darrylmelander!

Yes, we would like to add such tests to our CI too. We would possibly try to follow or extend what we already have in terms of testing.

1. Examples

So far, we have a csharp executable, which calls the csharp API: https://github.com/ERGO-Code/HiGHS/blob/master/examples/call_highs_from_csharp.cs

Perhaps we could have another example, which uses some callbacks from C#, to illustrate how to do it. This is built by cmake.

2. Nuget 

Additionally, we also test the nuget package: https://github.com/ERGO-Code/HiGHS/blob/master/.github/workflows/test-nuget-win.yml

It is worth adding callback examples tests to the nuget as well. 

3. External repos 

We are developing some code in external repositories with additional tests to be run before we tag a new release. I could add more C# tests there too, containing specific callbacks with specific data.

Perhaps it is worth having a short call about this. We are busy this week organising our workshop, but we could arrange a call next week. Please send me an email at galabovaa@gmail.com with your availability if you have the time and would like to discuss some of the details.

cpiber:
@darrylmelander Are you still working on this? Otherwise I would be interested in expanding further and testing to get it merged

jajhall:
> @darrylmelander Are you still working on this? Otherwise I would be interested in expanding further and testing to get it merged

I think so. @darrylmelander emailed us yesterday saying "I will send you my code as soon as I get a chance, hopefully soon."

darrylmelander:
> @darrylmelander Are you still working on this? Otherwise I would be interested in expanding further and testing to get it merged

I have tested it locally, but to do so I had to modify the core HiGHS library to expose some additional functions. Otherwise I didn't have a way to trigger specific events with specific data my tests could check against. Those changes, plus the tests that use those changes, is what I will be sending to jajhall when I get some time in the next few days.

I would welcome additional testing and follow-on development.

cpiber:
Any updates on this? If not, I could write some tests. I had a look already, I haven't seen any existing tests for C#, so I'd appreciate a pointer where to get started (any preferences on libraries, should this be hooked into existing tests [though py is not, so I assume no], ...)

jajhall:
Unfortunately no-one in the HiGHS team is a C# developer, so we've relied on external contributors to do the groundwork, with us just adding easy things by pattern-matching. For libraries in general, please ask @galabovaa

jajhall:
Unfortunately no-one in the HiGHS team is a C# developer, so we've relied on external contributors to do the groundwork, with us just adding easy things by pattern-matching. For libraries in general, please ask @galabovaa

highs/interfaces/highs_csharp_api.cs
```diff
@@ -636,6 +685,8 @@ public static HighsStatus call(HighsModel model, ref HighsSolution sol, ref High
     public HighsLpSolver()
     {
         this.highs = HighsLpSolver.Highs_create();
+        _cbDelegate = this.callbackFunction;
+        Highs_setCallback(this.highs, Marshal.GetFunctionPointerForDelegate(_cbDelegate), IntPtr.Zero);
```
cpiber, line 689: > You must manually keep the delegate from being collected by the garbage collector from managed code. The garbage collector does not track references to unmanaged code.

https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.marshal.getfunctionpointerfordelegate?view=net-9.0

You should store the delegate pointer, not `_cbDelegate`.

(*Removed reference to UnmanagedCallersOnlyAttribute since it's not available in the target version*)

darrylmelander: The `_cbDelegate` member variable stores a reference to the delegate, which prevents the delegate from being garbage collected for the lifetime of the HighsModel. The whole reason `_cbDelegate` exists is to satisfy the guidance you quoted. By my understanding, whether you store the return value of `Marshal.GetFunctionPointerForDelegate` has zero impact on garbage collection.

cpiber: I see, I think you may be right, thanks

jajhall:
Is this PR still active @darrylmelander / @cpiber

galabovaa:
This needs to be updated and tested. 

Added to Feature requests here https://github.com/ERGO-Code/HiGHS/issues/3225

PR #2434, opened 2025-06-25
Make some more methods const

fwesselm:
Make some more methods `const`.


fwesselm:
There seems to be a behavior change on `build-intel` for `rgn`. I did not see this in my fork.

fwesselm:
Interesting. I need to have a closer look.

PR #2436, opened 2025-06-26
CMake: Add support for UNITY_BUILD

jschueller:


jschueller:
hi @galabovaa, could you trigger the CI ?

jajhall:
@jschueller Thanks for these. Note that we push to latest, as we just use master for new releases

jschueller:
ok I rebased to latest

PR #2441, opened 2025-06-28
Concurrent solver for highspy

mathgeekcoder:
Implements a simple concurrent solver for highspy, that launches several threads with different random seeds and synchronizes the current best incumbent across all threads via callbacks.

In the log output, you'll see improved solutions found by other threads with the 'X' tag (i.e., user solution).

**Initial prototype, need to add tests, documentation etc.  PR is currently for discussion/demonstration only.**

```python
model = Highs()
model.ConcurrentSolve = True
model.setOptionValue('threads', 5) # optional
...
model.solve()
```

### DIFF
--- a/highs/highspy/highs.py
+++ b/highs/highspy/highs.py
@@ -4,6 +4,7 @@ import numpy as np
 from numbers import Integral
 from itertools import product
 from threading import Thread, local, RLock, Lock
+from multiprocessing import cpu_count
 from typing import Optional, Any, overload, Callable, Sequence, Mapping, Iterable, SupportsIndex, cast, Union
 
 from ._core import (
@@ -29,17 +30,19 @@ class Highs(_Highs):
     HiGHS solver interface
     """
 
-    __handle_keyboard_interrupt: bool = False
-    __handle_user_interrupt: bool = False
-    __solver_should_stop: bool = False
-    __solver_stopped: RLock = RLock()
-    __solver_started: Lock = Lock()
-    __solver_status: Optional[HighsStatus] = None
-
     def __init__(self):
         super().__init__()
         self.callbacks = [HighsCallback(cb.HighsCallbackType(_), self) for _ in range(int(cb.HighsCallbackType.kCallbackMax) + 1)]
         self.enableCallbacks()
+        
+        self.__handle_keyboard_interrupt: bool = False
+        self.__handle_user_interrupt: bool = False
+        self.__use_concurrent_solve: bool = False
+
+        self.__solver_should_stop: bool = False
+        self.__solver_stopped: RLock = RLock()
+        self.__solver_started: Lock = Lock()
+        self.__solver_status: Optional[HighsStatus] = None
 
     # Silence logging
     def silent(self, turn_off_output: bool = True):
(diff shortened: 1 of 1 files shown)


Opt-Mucca:
That was extremely quick!

Just to check that I'm following the code: Currently the multiple solves are sharing solutions, but the "concurrent solve" is not yet terminating when one of the threads has proven optimality / infeasibility?  Or does the `joinSolve` call automatically enforce this?

mathgeekcoder:
> That was extremely quick!

Thanks! Needed to do something while waiting for my flight :) 

> Just to check that I'm following the code: Currently the multiple solves are sharing solutions, but the "concurrent solve" is not yet terminating when one of the threads has proven optimality / infeasibility?  Or does the `joinSolve` call automatically enforce this?


Correct, it's pretty basic so far. It should be easy to check if one of the other thread terminates (opt/inf), but I'm not sure how to sync that information back to the main thread/model for the user to interact with.  So at the moment it's just a fancy parallel primal heuristic.

mathgeekcoder:
> Very neat, thanks. Needs a warning that its behaviour is non-deterministic, but it will be interesting to experiment with it. For greater usability, it would be better to reproduce it within the C++ class.

Thanks. Good call with the warning, do you think it's sufficient to add that to the documentation - or should it print to screen (if logging enabled)?

Happy to add to C++ if desired (code design should be similar). I didn't want to interfere with plans for the parallel implementation.  I just wanted to have a quick prototype to stimulate discussion.

jajhall:
> Thanks. Good call with the warning, do you think it's sufficient to add that to the documentation - or should it print to screen (if logging enabled)?

It needs to be added to both documentation and the logging - with a clear statement that debugging any failures may not be possible. 

> Happy to add to C++ if desired (code design should be similar). I didn't want to interfere with plans for the parallel implementation. I just wanted to have a quick prototype to stimulate discussion.

As I understand it, the work-stealing nature of the HiGHS multi-threading should mean that there's no conflict between the two activities.

mathgeekcoder:
> It needs to be added to both documentation and the logging - with a clear statement that debugging any failures may not be possible.

Sounds good!
 
> As I understand it, the work-stealing nature of the HiGHS multi-threading should mean that there's no conflict between the two activities.

That's fair.  However, it might not be the most efficient use of threads. Perhaps we want to add hierarchical support to our work-stealing. i.e., this "sub-task" can only use X threads. 

For example, we could run 3 concurrent solves, with each solve using 4 threads.  In addition, it might be worthwhile restricting parallelism on the sub-MIPs (i.e., use fewer threads in each recursion).

Work stealing is great when you have idle threads and each task has equal priority, but I'm worried about generating so many tasks that we become less efficient due to the order of processing and context switching etc.

That said, it's probably something that can be added later - without impacting this design too much.

mvernacc:
@mathgeekcoder I would be very excited to use the concurrent solve feature once it's released 😀
Thank you for working on this!

Thell:
I played with this using highspy 1.11.0 and needed to change 

`num_threads = self.getOptionValue("threads")`

to

`num_threads = self.getOptionValue("threads")[1]`

to get the num threads as an int to kick off the concurrent runs.

Also, a print statement when log to console is on indicating `n` concurrent processes are being used is nice.

It would be interesting to figure out what the overhead of the callback/solution sharing is...

Using a single thread, with default random seed I tested 2 quickly solved incidents and 2 taking a little bit longer:

| Incident | 1 proc| 8 proc shared|num 'Better'|8 proc not shared|
|----|----|----|---|---|
| 165A| 17.4 | 8.9| 0 |8.1|
| 165B| 13.8 | 11.4| 5|11.2|
| 355A| 259.1| 57.2|1| 51.19 |
| 355B| 310.2 |113.7|2| 67.2 |

I did notice that when the `Better incumbent found` set the solution it didn't show in the output as having been set, so I'm not sure if it was (or if this is related to the issue 2450 above). The 8 proc not shared time is the first to finish time.

The testing of the non-shared also revealed that my incident 355B highlights the issue of the random seed yielding different optimal results. From the 8 processes...
```
303 (objective)
303 (objective)
303 (objective)
303 (objective)
317 (objective)
303 (objective)
314 (objective)
319 (objective)
```
😬 

edit: I forgot to mention that on the 355B shared solution test that the shared solutions happened quite early:
```

         0       0         0   0.00%   281             inf                  inf        0      0      0         0     4.6s
         0       0         0   0.00%   291.4725379     inf                  inf        0      0      3      1191     5.6s
         0       0         0   0.00%   294.7974065     inf                  inf     2338    150    289      2079    11.2s
Better incumbent found 5 309.0
 X       0       0         0   0.00%   299.1892586     309                3.17%     3784    256    354      5083    14.6s
Better incumbent found 1 303.0
         0       0         0   0.00%   299.308362      309                3.14%     2233    197    869      9689    22.8s
```
         
The 303 better incumbent is the optimal and I'd wager that that process is also the one that completed in the 67 seconds on the non-shared test which I obviously don't know which process obtained it but its in the non-shared test log too.

`H       0       0         0   0.00%   299.3376535     303                1.21%     2225    326    356      7329    21.9s`

jajhall:
Thanks, I'll look to avoid this behaviour when I implement the same approach in C++

Thell:
@jajhall If I could I'd like to request that the c++ implementation also include a tunable option along the lines of a 'solution gap' that will allow the user to control if a solution gets shared, thereby reducing the number of calls to set solution for the other threads as both the checks within the set solution and the resetting of the target thread are pure overhead. Also, another request is don't wait for the non-incumbent threads to join/cleanup before returning the first-to-finish.

I played with this locally using HiGHSpy and have posted some findings on the Discord but suffice it to say that depending on the problem a `solution_gap` (where I used `abs(a-b)/b < solution_gap`), `abs_solution_gap` (where I used `abs(a-b) < abs_solution_gap` to limit the updates and let the `cancelSolve` do its job without waiting for a join dramatically improved my testing results for times on solving to optimality.

I'm sure you already have a strong direction and model in mind but I am hoping its early enough in the process that these ideas might seem plausible if you haven't already included/excluded them.

Thanks. :)

mathgeekcoder:
Thanks for the feedback @Thell!  A couple of comments:

> @jajhall If I could I'd like to request that the c++ implementation also include a tunable option along the lines of a 'solution gap' that will allow the user to control if a solution gets shared, thereby reducing the number of calls to set solution for the other threads as both the checks within the set solution and the resetting of the target thread are pure overhead. 

I believe the C++ implementation will be much more efficient at passing solutions between threads.  So, this tunable parameter is likely not necessary.

> Also, another request is don't wait for the non-incumbent threads to join/cleanup before returning the first-to-finish.

I'm not sure if the _initial_ C++ version will have first-to-finish.  Even if it does, there can be issues with returning without cleaning up all the threads. e.g., when calling solve again.  That said, it's quite likely that the C++ version could be faster (than my highspy prototype) at terminating the threads safely.

jajhall:
> Thanks for the feedback @Thell! A couple of comments:
> 
> > @jajhall If I could I'd like to request that the c++ implementation also include a tunable option along the lines of a 'solution gap' that will allow the user to control if a solution gets shared, thereby reducing the number of calls to set solution for the other threads as both the checks within the set solution and the resetting of the target thread are pure overhead.
> 
> I believe the C++ implementation will be much more efficient at passing solutions between threads. So, this tunable parameter is likely not necessary.

I agree, this parameter won't be necessary

> > Also, another request is don't wait for the non-incumbent threads to join/cleanup before returning the first-to-finish.
> 
> I'm not sure if the _initial_ C++ version will have first-to-finish. Even if it does, there can be issues with returning without cleaning up all the threads. e.g., when calling solve again. That said, it's quite likely that the C++ version could be faster (than my highspy prototype) at terminating the threads safely.

The C++ version will not have first-to-finish. However, each thread will check regularly for a "terminate" flag being issued by other threads, so there will be no meaningful overhead in having them all reach a synchronisation barrier.

Note that, in HiGHS, multithreading is achieved by having multiple threads running continuously and "stealing" work when it becomes available, rather than by spawning multiple threads when parallel work is to be done.

jajhall:
The performance gain from the MIP race didn't justify introducing a non-deterministic solver into HiGHS

PR #2444, opened 2025-06-30
Strong-branching refactor

Opt-Mucca:
This creates a lambda expression `strongBranch` that gets called when evaluating branching candidates. It is purely there to improve code readability and remove some duplicate code. Previously, there was a large amount of extended logic for `x <= b` and that was mostly repeated for `x >= b+1`. I've put them into a single function where you give a boolean flag on which one you want to evaluate. That should not change change the solution path whatsoever (completely unsure if there is a potential performance impact for semi-large lambda expressions in C++)

The single line change that may change the solution path of some instances: If there is a single branching candidate then I just now select it (I don't bother evaluating it).

### DIFF
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -13,3 +13,6 @@ The irreducible infeasibility system (IIS) facility now detects infeasibility du
 Prompted by [#2463](https://github.com/ERGO-Code/HiGHS/issues/2463), the HiGHS solution and basis files now match data to any column and row names in the model, only assuming that the data are aligned with column and row indices if there are no names in the model. This requires a new version (v2) of the HiGHS basis file. Basis files from v1 are still read, but deprecated. Now, when writing out a model, basis or solution, column and row names are added to the model - previously they were created temporarily and inconsistentyly on the fly. If the model has existing names, then distinctive names are created to replace any blank names, but names with spaces or duplicate names yield an error status return.
 
 Refactored strong branching to minimize duplicated code
+
+As per [#2487](https://github.com/ERGO-Code/HiGHS/issues/2487), trivial heuristics now run before feasibility jump (FJ), and FJ will use any existing incumbent. FJ will clip any finite variable values in the incumbent to lower and upper bounds, and falls back to the existing logic (lower bound if finite, else upper bound if finite, else 0) for any infinite values in the incumbent.
+
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ linear optimization problems of the form
 
 $$ \min \quad \dfrac{1}{2}x^TQx + c^Tx \qquad \textrm{s.t.}~ \quad L \leq Ax \leq U; \quad l \leq x \leq u $$
 
-where Q must be positive semi-definite and, if Q is zero, there may be a requirement that some of the variables take integer values. Thus HiGHS can solve linear programming (LP) problems, convex quadratic programming (QP) problems, and mixed integer programming (MIP) problems. It is mainly written in C++, but also has some C. It has been developed and tested on various Linux, MacOS and Windows installations. No third-party dependencies are required.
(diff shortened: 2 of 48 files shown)


jajhall (approved):
Fine by me. One more minor modification that can be accumulated before baseline testing to assess the effectiveness of expected performance enhancements sitting in PRs

highs/mip/HighsSearch.cpp
```diff
@@ -254,6 +254,7 @@ HighsInt HighsSearch::selectBranchingCandidate(int64_t maxSbIters,
   std::vector<double> downbound;
 
   HighsInt numfrac = lp->getFractionalIntegers().size();
+  if (numfrac == 1) return 0;
```
BenChampion: Why can we return early here? Worth a comment or not?

highs/mip/HighsSearch.cpp
```diff
@@ -530,18 +531,13 @@ HighsInt HighsSearch::selectBranchingCandidate(int64_t maxSbIters,
...
-    if (!downscorereliable[candidate] &&
-        (upscorereliable[candidate] ||
-         std::make_pair(downscore[candidate],
-                        pseudocost.getAvgInferencesDown(col)) >=
-             std::make_pair(upscore[candidate],
-                            pseudocost.getAvgInferencesUp(col)))) {
-      // evaluate down branch
-      // if (!mipsolver.submip)
-      //   printf("down eval col=%d fracval=%g\n", col, fracval);
+    auto strongBranch = [&](bool upbranch) {
```
BenChampion: Consider adding the return type of this lambda with `->` for clarity.

[C++ Reference on Lambda expressions](https://en.cppreference.com/w/cpp/language/lambda.html#:~:text=ret%2C%20where%20ret%20specifies%20the%20return%20type.)

BenChampion (commented):
Nice! Always in favor of removing unnecessary duplicated code. I'm not sure what the overhead of the lambda would be, but my intuition (possibly wrong!) is that it should be minimal. Worth testing, of course.

fwesselm:
I ran 424 instances from MIPLIB with these changes and observed different node counts on bell3a, bell5, blp-ar98, harp2, lectsched-4-obj, leo1, lseu, map18, mine-166-5, neos-1440460, neos-476283, noswot, opm2-z7s2, pg, pg5_34, pp08a, ran14x18-disj-8, ran16x16, reblock166, reblock67, rentacar, roll3000 and rout. I am not sure if all these are affected by the single line change you described.

highs/mip/HighsSearch.cpp
Opt-Mucca: The logic here would be: If there's only one branching candidate, then why waste time evaluating it when we're going to branch on it anyways? Thinking more on this though, I guess the overhead is negligible, and it could stop us taking the worse branching path because we skipped strong-branch-evaluations of both sides. I'll revert the change.

highs/mip/HighsSearch.cpp
Opt-Mucca: Added

Opt-Mucca:
@fwesselm I have just simply removed that line (Thought about it a bit more and there could be some fringe case where such a change slows down an instance by a fair bit)

fwesselm:
> @fwesselm I have just simply removed that line (Thought about it a bit more and there could be some fringe case where such a change slows down an instance by a fair bit)

Ok, I will re-run the instances where the behavior changed shortly.

fwesselm:
> > @fwesselm I have just simply removed that line (Thought about it a bit more and there could be some fringe case where such a change slows down an instance by a fair bit)
> 
> Ok, I will re-run the instances where the behavior changed shortly.

I re-ran the instances, and there are no behavior changes with the updated code.

fwesselm:
I tried to resolve the merge conflict and hope this is ready to be merged now.

Opt-Mucca:
Assuming the merge conflict was just for the README? I'm all for this being merged.

fwesselm:
> Assuming the merge conflict was just for the README? I'm all for this being merged.

Yes, FEATURES.md had a conflict.

PR #2456, opened 2025-07-06
IIS "light"

jajhall:
The irreducible infeasibility system (IIS) facility now detects infeasibility due to bounds on constraint activity values being incompatible with constraint bounds. 

A `kIisStrategyLight` mode for the `iis_strategy` option has been introduced so that only infeasibility due to incompatible variable/constraint bounds and constraint activity values is checked for. 

The LP corresponding to any known IIS is now formed and held as a data member of the HighsIis class. 

`Highs::checkIis` has been added for unit testing and beta testing. It takes the IIS LP, checks that it's infeasible and then relaxes each finite variable/constraint bound in turn, and solves the corresponding LP, checking that the model status is `HighsModelStatus::kOptimal` or `HighsModelStatus::kUnbounded`. Since the IIS LP is so small, the cost of the check is trivial compared with finding an IIS, so maybe the checking should be done a priori.


check/TestIis.cpp
```diff
@@ -19,6 +19,8 @@ void testFeasibilityRelaxation(
     const double rhs_penalty,
     const double require_feasibility_objective_function_value);
 
+void checkIisLp(HighsLp& lp, const HighsIis& iis);
```
odow:
```suggestion
void checkIISLp(HighsLp& lp, const HighsIis& iis);
```
or
```suggestion
void checkIIS(HighsLp& lp, const HighsIis& iis);
```

Currently it reads as "check I is LP"

check/TestIis.cpp
```diff
@@ -42,10 +44,21 @@ TEST_CASE("lp-incompatible-bounds", "[iis]") {
   Highs highs;
   highs.setOptionValue("output_flag", dev_run);
   highs.passModel(lp);
+  // Perform the light IIS check
+  highs.setOptionValue("iis_strategy", kIisStrategyLight);
+  HighsIis iis;
+  REQUIRE(highs.getIis(iis) == HighsStatus::kOk);
```
odow, line 47: I could put it below, but shouldn't we capitalize IIS everywhere?
```suggestion
  REQUIRE(highs.getIIS(iis) == HighsStatus::kOk);
```

check/TestIis.cpp
```diff
@@ -42,10 +44,21 @@ TEST_CASE("lp-incompatible-bounds", "[iis]") {
   Highs highs;
   highs.setOptionValue("output_flag", dev_run);
   highs.passModel(lp);
+  // Perform the light IIS check
+  highs.setOptionValue("iis_strategy", kIisStrategyLight);
+  HighsIis iis;
```
odow, line 46:
```suggestion
  HighsIIS iis;
```

odow:
How would you see this being used from the C API?

Can we tie the original rows and columns to the rows and columns in the IIS?

JuMP wants to be able to call some sort of `Highs_computeIIS(highs)` method, and then check an enum of `IN_CONFLICT`, `NOT_IN_CONFLICT`, `MAYBE_IN_CONFLICT` for each variable, bound, and constraint row.

odow:
We should have strong reasons to not follow the Gurobi API:

https://docs.gurobi.com/projects/optimizer/en/current/reference/c/solving.html#c.GRBcomputeIIS

check/TestIis.cpp
jajhall: I see how you're reading it, but with camel case it's "check IIS LP"

check/TestIis.cpp
jajhall: No, with a couple of exceptions - created many years ago! - in HiGHS initialisms appear in camel case with only the first letter capitalised. So, we have HighsLp, for example. One never knows when something would be less readable in camel case if initialisms were all in upper case.

jajhall:
> We should have strong reasons to not follow the Gurobi API:
> 
> https://docs.gurobi.com/projects/optimizer/en/current/reference/c/solving.html#c.GRBcomputeIIS

> Can we tie the original rows and columns to the rows and columns in the IIS?

Thanks, I'll add this. I was working from the Xpress IIS API

> JuMP wants to be able to call some sort of Highs_computeIIS(highs) method, and then check an enum of IN_CONFLICT, NOT_IN_CONFLICT, MAYBE_IN_CONFLICT for each variable, bound, and constraint row.

I'll add this. I assume that `MAYBE_IN_CONFLICT` is used when the IIS calculation has terminated prematurely.

The `FORCE` attribute is interesting. I won't add it yet, as the full IIS calculation is WIP, and I'm yet to identify precisely what needs to be completed/debugged/improved.

jajhall:
> an enum of `IN_CONFLICT`, `NOT_IN_CONFLICT`, `MAYBE_IN_CONFLICT` for each variable, bound, and constraint row.

Do you mean one enum for each variable and each constraint, or do you want to know whether a specific variable/constraint lower/upper bound has this status? 

Corresponding to each variable/constraint in the IIS, there's a bound status to indicate whether just the lower, just the upper, or both bounds are finite in the IIS LP.

@odow, there are a few more IIS issues that it would be good to discuss at the next HiGHS-JuMP meeting

highs/interfaces/highs_c_api.h
```diff
@@ -2350,6 +2377,39 @@ HighsInt Highs_feasibilityRelaxation(void* highs,
...
+ * @param const HighsInt* row_status n array of length [num_col], to be
+ *                                   filled with the IIS status of all original
+ *                                   constraints.
+ *
+ * @returns A `kHighsStatus` constant indicating whether the call succeeded.
+ */
+HighsInt Highs_getIis(void* highs, HighsInt* iis_num_col, HighsInt* iis_num_row,
+                      HighsInt* col_index, HighsInt* row_index,
+                      HighsInt* col_bound, HighsInt* row_bound,
+                      HighsInt* col_status, HighsInt* row_status);
```
odow, line 2412: What is the difference between `_bound` and `_status`? The `_bound` is free/lower/upper/boxed and the `_status` is in conflict/not in conflcit/maybe in conflict?

If we're returning only a subset of the rows and columns, the status would never be "not in conflict", so are both vectors needed?

Perhaps to answer my own question: if the IIS is not minimal, then there might be a "maybe in conflict" entry. If all are "in conflict" then the IIS is minimal.

jajhall:
- IIS LP is now tested in `Highs::getIis`, so `Highs::checkIis` has been removed
- `Highs::writeIisModel` can take a `foo.mps`, `foo.lp` or `""` argument
- `HighsIis`now has data members 
1. `std::vector<HighsInt> col_status_` and `std::vector<HighsInt> row_status_` giving `kIisStatusInConflict` or `kIisStatusNotInConflict` for each column and row of the original LP
2. `HighsModel model_` rather than `HighsLp lp_` 
3. `Highs::getIisLp()` to return a const reference to `Highs::iis_.model_.lp_`, any incumbent IIS LP

Methods and docstrings have been added to the C API

highs/interfaces/highs_c_api.h
jajhall: > What is the difference between `_bound` and `_status`? The `_bound` is free/lower/upper/boxed and the `_status` is in conflict/not in conflcit/maybe in conflict?

Yes
 
> If we're returning only a subset of the rows and columns, the status would never be "not in conflict", so are both vectors needed?

These vectors aren't needed at the moment, but if the IIS calculation is interrupted by timeout, say, then (as you guessed) the IIS is not minimal, so there might be a "maybe in conflict" entry. If all are "in conflict" then the IIS is minimal.

filipforro:
Hi, thanks for the work on this! Do you have any expectation when this will be available? We're interested in having the IIS available for use through the C API.

jajhall:
> Hi, thanks for the work on this! Do you have any expectation when this will be available? We're interested in having the IIS available for use through the C API.

I can't see a new release before mid-September, as we're doing a lot of urgent development work at the moment. 

However, the necessary methods are already in the C API - and tested with our other C API tests.

filipforro:
Thanks for the update! 

> However, the necessary methods are already in the C API - and tested with our other C API tests.

I was looking in the [interfaces](https://github.com/ERGO-Code/HiGHS/blob/master/highs/interfaces/highs_c_api.h) and have not seen the IIS methods exposed in the API, and saw them being added in this PR (https://github.com/ERGO-Code/HiGHS/pull/2456/commits/2d0bb198022099d4f75acbb142a1654a1608faf5).

jajhall:
> Thanks for the update!
> 
> > However, the necessary methods are already in the C API - and tested with our other C API tests.
> 
> I was looking in the [interfaces](https://github.com/ERGO-Code/HiGHS/blob/master/highs/interfaces/highs_c_api.h) and have not seen the IIS methods exposed in the API, and saw them being added in this PR ([2d0bb19](https://github.com/ERGO-Code/HiGHS/commit/2d0bb198022099d4f75acbb142a1654a1608faf5)).

This is because the ["interfaces"](https://github.com/ERGO-Code/HiGHS/blob/master/highs/interfaces/highs_c_api.h) you quote are in the master branch, and #2456 has not even been merged into [latest](https://github.com/ERGO-Code/HiGHS/tree/latest).

FEATURES.md
```diff
@@ -8,6 +8,8 @@ Now handling correctly the case where an infeasible MIP has a feasible relaxatio
...
 
+
 The irreducible infeasibility system (IIS) facility now detects infeasibility due to bounds on constraint activity values being incompatible with constraint bounds. A kIisStrategyLight mode for the iis_strategy option has been introduced so that only infeasibility due to incompatible variable/constraint bounds and constraint activity values is checked for. The LP corresponding to any known IIS is now formed and held as a data member of the HighsIis class. 
 
+Prompted by [#2463](https://github.com/ERGO-Code/HiGHS/issues/2463), the HiGHS solution and basis files now match data to any column and row names in the model, only assuming that the data are aligned with column and row indices if there are no names in the model. This requires a new version (v2) of the HiGHS basis file. Basis files from v1 are still read, but deprecated. Now, when writing out a model, basis or solution, column and row names are added to the model - previously they were created temporarily and inconsistentyly on the fly. If the model has existing names, then distinctive names are created to replace any blank names, but names with spaces or duplicate names yield an error status return.
```
BenChampion, line 13: Happened to notice a typo here: inconsistent**y**ly

**EDIT**: sorry, did not see you already merged this

PR #2457, opened 2025-07-07
Fix 2446

### LINKED ISSUE #2446: Incorrect identification of infeasibility in MIP presolve
For this [model](https://github.com/user-attachments/files/20987410/2446.mps.txt) HiGHS v1.11 claims infeasibility incorrectly, but without presolve the correct solution is obtained. The costs are randomized from another instance that exhibited the same behaviour, suggesting that the error derives from primal presolve

fwesselm:
Fix #2446 
- `HPresolve::rowPresolve` implements "simple probing on a single equation" (see section 3.6, Achterberg et al., Presolve Reductions in Mixed Integer Programming, INFORMS Journal on Computing 32(2):473-506).
- The current code uses bounds on constraint activities `sumLower` and `sumUpper` that were obtained using implied column bounds (vectors `implColLower` and `implColUpper`), but then uses the standard bounds (vectors `col_lower_` and `col_upper_`) to perform substitutions. I think this is inconsistent.
- I think there are two potential ways to resolve this: (a) use vectors `sumLowerOrig` and `sumUpperOrig` that were computed using only the standard bound vectors, or (b) keep using `sumLower` and `sumUpper`, but correct the bounds used in the substitutions. I implemented the latter in this PR.
- A test has been added.
- I think this should be qualified on a large set of MIPs and, unfortunately, I am not able to do this currently.
- Once I have the test results, I will report back.


highs/presolve/HPresolve.cpp
```diff
@@ -3213,48 +3213,52 @@ HPresolve::Result HPresolve::rowPresolve(HighsPostsolveStack& postsolve_stack,
...
-    // row is redundant
-    if (logging_on) analysis_.startPresolveRuleLog(kPresolveRuleRedundantRow);
-    postsolve_stack.redundantRow(row);
-    removeRow(row);
-    analysis_.logging_on_ = logging_on;
-    if (logging_on) analysis_.stopPresolveRuleLog(kPresolveRuleRedundantRow);
-    return checkLimits(postsolve_stack);
+  // handle special cases directly via a call to the specialized procedure
+  switch (rowsize[row]) {
+    default:
```
Opt-Mucca, line 3252: Is this now missing the 0 case that was handled previously?

highs/presolve/HPresolve.cpp
```diff
@@ -3323,12 +3327,18 @@ HPresolve::Result HPresolve::rowPresolve(HighsPostsolveStack& postsolve_stack,
   if (rowsizeInteger[row] != 0 || rowsizeImplInt[row] != 0) {
     if (rowLower == rowUpper) {
       // equation
+      double impliedRowLower = impliedRowBounds.getSumLower(row);
+      double impliedRowUpper = impliedRowBounds.getSumUpper(row);
+      // see section 3.6 "Simple probing on a single equation", Achterberg et
+      // al., Presolve Reductions in Mixed Integer Programming, INFORMS Journal
+      // on Computing 32(2):473-506.
```
Opt-Mucca, line 3334: That is a beautiful comment! (I also need to start adding more of these)

highs/presolve/HPresolve.cpp
```diff
@@ -3816,31 +3827,32 @@ HPresolve::Result HPresolve::rowPresolve(HighsPostsolveStack& postsolve_stack,
...
-      if (model->row_lower_[row] == -kHighsInf &&
-          impliedRowUpper != kHighsInf) {
+      if (model->row_lower_[row] == -kHighsInf) {
         // <= constraint: try to strengthen coefficients
-        HighsCDouble rhs = model->row_upper_[row];
-        strengthenCoefs(rhs, HighsInt{1},
-                        static_cast<HighsCDouble>(impliedRowUpper) -
-                            model->row_upper_[row]);
-        model->row_upper_[row] = static_cast<double>(rhs);
+        double impliedRowUpper = impliedRowBounds.getSumUpper(row);
```
Opt-Mucca, line 3832: Any strong reason why the single level if statement has to be changed to two? (Am fine with it but worried I'm missing something)

highs/presolve/HPresolve.cpp
fwesselm: > Is this now missing the 0 case that was handled previously?

Correct, I moved the check for infeasibility to the beginning of this method (i.e. before this switch statement). The check for redundancy is done after the switch statement. This avoids duplicate code in the "0" case.

highs/presolve/HPresolve.cpp
fwesselm: Thank you. I hope this will make it easier to understand the code.

highs/presolve/HPresolve.cpp
fwesselm: Previously, there were two local variables named `impliedRowUpper` and `impliedRowLower`, which were used throughout the row presolve. These were used to store the row activity bounds obtained with the standard bounds (`impliedRowBounds.getSumUpperOrig(row)` and `impliedRowBounds.getSumLowerOrig(row)`) and in some cases also the activity bounds obtained when taking into account implied bounds (`impliedRowBounds.getSumUpper(row)` and `impliedRowBounds.getSumLower(row)`). I hope my changes make it clearer which version of the activity bounds is used in which part of row presolve.

Opt-Mucca: Ahhhhhhhhhhhhhh. I need to get more used to the terminology used in presolve.

fwesselm:
The results look OK. The node counts on three problems are affected by this bug fix (blp-ic97, neos-1593097, p80x400b).

PR #2458, opened 2025-07-07
Replace some more "raw" by range-based loops

fwesselm:
@galabovaa, @jajhall, I know that the first PR to remove raw loops was controversial. Nevertheless, I hope that maybe a subset of the changes in this PR are acceptable. If not, feel free to just close this PR.

### DIFF
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -6,4 +6,6 @@ Forcing column reduction now checks the bound on the column dual rather than whe
 
 Now handling correctly the case where an infeasible MIP has a feasible relaxation, so no ray is computed fixing [#2415](https://github.com/ERGO-Code/HiGHS/issues/2415)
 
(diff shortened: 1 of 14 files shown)


jajhall:
I don't see the point of fiddling with code that works, just to satisfy some questionable C++ coding practice retrospectively. 

Didn't the last batch lead to a change of behaviour with some compiler on the MIP `rgn`?

To me there are far more important things to work on, but if you're confident that what's been done is correct - I don't understand the new loop syntax well enough to comment on correctness - then I'm not going to close the PR.

It needs checking by someone who is fluent in this range-based loops syntax - @galabovaa - although the time spent checking could be put to better use.

galabovaa:
I think these are all reasonable changes and I am happy to merge the PR.

galabovaa (approved):

Issue #1862, opened 2024-07-29
Encountered ERROR:   HighsSearch::selectBranchingCandidate


Thell:
First time I've seen this. I hadn't changed the logical model from prior runs, that didn't encounter this, with the same parameters. The final solution still looks to be valid.

I did add to the input data between the previous runs and this one.

```
      3020      82       651  66.21%   423991228.5359  402736871.5479     5.28%     6161    912   7613    542737   389.9s
ERROR:   HighsSearch::selectBranchingCandidate Error fracval = 0.444444 <= 1 = 1 + 1e-06 = localdom.col_lower_[col] + mipsolver.mipdata_->feastol: Residual -0.555557
ERROR:   HighsSearch::selectBranchingCandidate Error fracval = 0.666667 <= 1 = 1 + 1e-06 = localdom.col_lower_[col] + mipsolver.mipdata_->feastol: Residual -0.333334
ERROR:   HighsSearch::selectBranchingCandidate Error fracval = 0.444444 <= 1 = 1 + 1e-06 = localdom.col_lower_[col] + mipsolver.mipdata_->feastol: Residual -0.555557
ERROR:   HighsSearch::selectBranchingCandidate Error fracval = 0.75 <= 1 = 1 + 1e-06 = localdom.col_lower_[col] + mipsolver.mipdata_->feastol: Residual -0.250001
ERROR:   HighsSearch::selectBranchingCandidate Error fracval = 0.444444 <= 1 = 1 + 1e-06 = localdom.col_lower_[col] + mipsolver.mipdata_->feastol: Residual -0.555557
ERROR:   HighsSearch::selectBranchingCandidate Error fracval = 0.466667 <= 1 = 1 + 1e-06 = localdom.col_lower_[col] + mipsolver.mipdata_->feastol: Residual -0.533334
ERROR:   HighsSearch::selectBranchingCandidate Error fracval = 0.444444 <= 1 = 1 + 1e-06 = localdom.col_lower_[col] + mipsolver.mipdata_->feastol: Residual -0.555557
ERROR:   HighsSearch::selectBranchingCandidate Error fracval = 0.285714 <= 1 = 1 + 1e-06 = localdom.col_lower_[col] + mipsolver.mipdata_->feastol: Residual -0.714287
[...]

PR #2462, opened 2025-07-10
Add check for local domain during clique substitution

### LINKED ISSUE #2205: HiGHS crashes with assertion failure
HiGHS (build from `latest` branch) crashes when solving [seed.mps.txt](https://github.com/user-attachments/files/19096127/seed.mps.txt) with option file [highs_options.txt](https://github.com/user-attachments/files/19096142/highs_options.txt)

```
root@b9f8d0525121:/tmp# cat highs_options.txt
mip_rel_gap=0
mip_abs_gap=0
presolve=off
mip_lp_age_limit=19
mip_pool_age_limit=2
mip_heuristic_effort=0.321638279322984
root@b9f8d0525121:/tmp# highs --model_file seed.mps --options_file highs_options.txt
Running HiGHS 1.9.0 (git hash: acf51510d): Copyright (c) 2025 HiGHS under MIT licence terms
Set option presolve to "off"
Set option mip_lp_age_limit to 19
Set option mip_pool_age_limit to 2
Set option mip_heuristic_effort to 0.321638
Set option mip_rel_gap to 0
Set option mip_abs_gap to 0
MIP  seed has 370 rows; 19 cols; 7029 nonzeros; 11 integer variables
Coefficient ranges:
  Matrix [1e-02, 2e+01]
  Cost   [5e-01, 1e+02]
  Bound  [2e+02, 2e+02]
  RHS    [5e-02, 1e+02]

Presolve is switched off

Solving MIP model with:
   370 rows
   19 cols (0 binary, 11 integer, 0 implied int., 8 continuous)
   7029 nonzeros

Src: B => Branching; C => Central rounding; F => Feasibility pump; H => Heuristic; L => Sub-MIP;
     P => Empty MIP; R => Randomized rounding; S => Solve LP; T => Evaluate node; U => Unbounded;
     z => Trivial zero; l => Trivial lower; u => Trivial upper; p => Trivial point; X => User solution

         0       0         0   0.00%   -187950         inf                  inf        0      0      0         0     0.0s
         0       0         0   0.00%   2478.046103     inf                  inf        0      0      0        62     0.0s
 C       0       0         0   0.00%   2508.606812     3024.888376       17.07%       29      5      0        75     0.3s
 T      16       0         5   3.71%   2513.575625     3010.091182       16.50%       27      4     27      1920     4.4s
 T      57       2        24   8.01%   2513.575625     2963.942415       15.19%       59      4     60      2352     5.2s
 T     151      12        64  10.96%   2513.711157     2961.310011       15.11%      100      4    112      2746     6.3s
 T     286      29       123  60.80%   2513.739403     2949.737102       14.78%      126      4    201      3312     8.1s
 T     351      19       155  61.24%   2513.739403     2891.23908        13.06%      156      4    244      3552     8.4s
 T     449      24       201  61.62%   2513.86308      2861.092199       12.14%      155      4    308      3914     9.3s
 T     474      25       207  61.65%   2513.86308      2809.125929       10.51%      139      4    319      4016     9.6s
 T     521      23       232  62.26%   2513.86308      2795.344773       10.07%      164      4    359      4216     9.9s
 T     654      25       296  69.89%   2513.86308      2776.400062        9.46%      144      4    500      4972    11.9s
highs: /tmp/HiGHS/src/mip/HighsSearch.cpp:301: HighsInt HighsSearch::selectBranchingCandidate(int64_t, double&, double&): Assertion `lower_residual > -1e-12 && upper_residual > -1e-12' failed.
Aborted (core dumped)
```

Opt-Mucca:
This fixes #2205 and I suspect #1862 

Explanation:
In a sub-milp HiGHS is managing to tighten the global bounds of x4 from a dual proof. That is, given `x4 + x6 <= 1`, where `0 <= x6 <= 1` & `0 <= x4 <= 2` it can conclude that `x4 <= 1`. It also stores a clique (let's assume equality to make it easier) and manages to derive that `x4 = 1 - x6`. The problem is that this bound update (`x4 <= 1`) is only given to the global domain and NOT to the domain attached to the HighsSearch object!
This leads to a problem later when HiGHS during its search has branched upwards on `x4`, i.e., `1 <= x4 <= 2`. HiGHS is going to use the clique substitution (that is supposed to be globally valid) and replace some valid LP value of `x4` by `1 - x6`. Let's say `x6 = 0.05` and therefore `x4 = 1 - x6 = 0.95`  (violates the local domain `1 <= x4 <= 2`)!!

The fundamental problem is simply that the search is using some global information that has not yet been passed to it. I've added a band-aid if statement to catch these rare cases that should have no practical overhead and cannot imagine many instances at all get affected by this change.

I removed the more robust fix because that would essentially be adding more sync points of the global domain and the domain attached to the search object. This may not be worth it because of the overhead, and would have to be removed if we ever want to parallelise the search. (For parallel search: This clique is a great example of something we're going to have to disable to make parallelism valid or we would have to create an entirely new clique buffer data structure)

### DIFF
--- a/highs/mip/HighsLpRelaxation.cpp
+++ b/highs/mip/HighsLpRelaxation.cpp
@@ -1301,6 +1301,19 @@ HighsLpRelaxation::Status HighsLpRelaxation::resolveLp(HighsDomain* domain) {
                   break;
               }
 
+              // Spot cases where a global domain change during the search has
+              // led to a clique and the local domain has branched into a state
+              // where the clique subst. would be invalid for the local domain.
+              // TODO: Turn into assert when search becomes parallel
+              if (domain) {
+                double replace_val = subst->replace.val == 0 ? 1.0 - val : val;
+                double replace_lb = domain->col_lower_[subst->replace.col];
+                double replace_ub = domain->col_upper_[subst->replace.col];
+                double feastol = mipsolver.mipdata_->feastol;
+                if ((replace_val < replace_lb - feastol) ||
+                    (replace_val > replace_ub + feastol)) break;
+              }
+
               col = subst->replace.col;
               if (subst->replace.val == 0) val = 1.0 - val;
 


highs/mip/HighsLpRelaxation.cpp
```diff
@@ -1301,6 +1301,21 @@ HighsLpRelaxation::Status HighsLpRelaxation::resolveLp(HighsDomain* domain) {
...
+                double replace_ub = domain->col_upper_[subst->replace.col];
+                double feastol = mipsolver.mipdata_->feastol;
+                if ((replace_val < replace_lb - feastol) ||
+                    (replace_val > replace_ub + feastol)) {
+                  break;
+                }
+              }
+
               col = subst->replace.col;
               if (subst->replace.val == 0) val = 1.0 - val;
```
fwesselm, line 1321: You could move the new code block here. The local variables could then be used to simplify the code a little (but maybe that's a question of taste 😄 ):

 ```
              col = subst->replace.col;
              if (subst->replace.val == 0) val = 1.0 - val;

              // Spot cases where a global domain change during the search has
              // led to a clique and the local domain has branched into a state
              // where the clique subst. would be invalid for the local domain.
              // TODO: Turn into assert when search becomes parallel
              if (domain != nullptr &&
                  (val <
                       domain->col_lower_[col] - mipsolver.mipdata_->feastol ||
                   val > domain->col_upper_[col] + mipsolver.mipdata_->feastol))
                break;
```

But, your current code is fine. Just a suggestion...

fwesselm (approved):
Thank you for fixing this, @Opt-Mucca. Your fix looks correct. I tried to debug this some time ago and failed spectacularly. 😄 

Would it be worth adding one of the reproduction examples as tests? But, maybe they take too much time to solve.

jajhall (approved):
Well done: fine by me!

highs/mip/HighsLpRelaxation.cpp
Opt-Mucca: I think that change wouldn't work though, because then you'll have overwritten `col` and `val` with the invalid values already. (they're used outside of the `while` loop).

And thanks! It took me the better part of yesterday to track it down and properly understand the full range of fixes.

fwesselm: Oh, you are right! Thanks.

Issue #2455, opened 2025-07-06
HiGHS crashes on small MIP


hextrace1:
HiGHS crashes on [case1.lp.txt](https://github.com/user-attachments/files/21089663/case1.lp.txt) with assertion failed:

```
highs/mip/HighsDomain.cpp:3220: bool HighsDomain::ConflictSet::explainInfeasibility(): Assertion `ninfmin == 0' failed.
```

jajhall:
Thanks, this is for development. Please compile in "release"

Opt-Mucca:
This is also happening on #2439 when I debug without presolve....... (Not 100% sure these are related yet as I'd need to the correct HiGHS version)

hextrace1:
Thanks. This failure happens on both trunk and the latest release.

jajhall:
Closed by #2464

Issue #2439, opened 2025-06-27
Suboptimal objective returned for MILP


uhrm:
I'm using HiGHS to solve a set of MILP subproblems as part of a Lagrangian decomposition procedure. The subproblems are essentially multiple-choice knapsack problems with an additional epigraph variable to model a concave objective function (to be maximized).

All subproblems have a trivial lower bound of 0 (when all variables are zero), but for some instances HiGHS reports a negative objective value (claiming it is solved to optimality).

An example to reproduce the problem is attached ([highs_36.mps.gz](https://github.com/user-attachments/files/20946379/highs_36.mps.gz)). The log when solving this problem is
```
Running HiGHS 1.11.0 (git hash: n/a): Copyright (c) 2025 HiGHS under MIT licence terms
Set option log_file to "HiGHS.log"
Number of BV entries in BOUNDS section is 6504
MIP  highs_36 has 57 rows; 6505 cols; 19514 nonzeros; 6504 integer variables (6504 binary)
Coefficient ranges:
  Matrix [1e+00, 8e+01]
  Cost   [1e-01, 7e+00]
  Bound  [1e+00, 1e+00]
  RHS    [1e+00, 1e+00]
Presolving model
57 rows, 6505 cols, 19514 nonzeros  0s
57 rows, 933 cols, 2798 nonzeros  0s

Solving MIP model with:
   57 rows
   933 cols (932 binary, 0 integer, 1 implied int., 0 continuous, 0 domain fixed)
   2798 nonzeros

Src: B => Branching; C => Central rounding; F => Feasibility pump; J => Feasibility jump;
     H => Heuristic; L => Sub-MIP; P => Empty MIP; R => Randomized rounding; Z => ZI Round;
     I => Shifting; S => Solve LP; T => Evaluate node; U => Unbounded; X => User solution;
     z => Trivial zero; l => Trivial lower; u => Trivial upper; p => Trivial point

        Nodes      |    B&B Tree     |            Objective Bounds              |  Dynamic Constraints |       Work
Src  Proc. InQueue |  Leaves   Expl. | BestBound       BestSol              Gap |   Cuts   InLp Confl. | LpIters     Time

 J       0       0         0   0.00%   inf             -0                 Large        0      0      0         0     0.1s

99.1% inactive integer columns, restarting
Model after restart has 4 rows, 9 cols (8 bin., 0 int., 1 impl., 0 cont., 0 dom.fix.), and 20 nonzeros

[...]

PR #2464, opened 2025-07-11
Fix 2439 and 2455

Opt-Mucca:
This contains two changes:

1. It adds an additional check that the solution is actually improving on the current upper bound. In #2439 HiGHS was given a solution to try that was not feasible in the original space. After some repair LP the solution does become feasible, however the objective value becomes worse. Importantly, this objective change places it above the current best bound. As there is no objective value based check within the function though, it ends up overrides some global data (as seen in the output of the log). This is completely incorrect, and if it were to have happened in the middle of the solve process would have potentially terminated the solve with a fake optimal solution. All I've done is throw an additional sanity check inside of the function.
2. There is an assert that there are no infinite terms in the objective when propagating objective based infeasibility reasons. The logic is clear: If there is a `-inf` term in the objective how are you arguing that `-inf >= b` when b is finite?
The logic break when you are handling two different domains. Let's say the reason was found on the local domain, where a bound for `x1 <= inf` has potentially been changed to `x1 <= 1` in something like a fix-and-propagate heuristic. This means that after enough fixes and propagation rounds you can already see that the best possible remaining solution is worse than the current incumbent. The issue is that when you try and make a globally valid conflict out of this, you're going to use the global bounds `x1 <= inf`, for which the objective is unbounded, and this will hit the current assert. I think this is a simple error and should be replaced by a `return false`. This happens on both #2439 without presolve and on #2455

### DIFF
--- a/highs/mip/HighsDomain.cpp
+++ b/highs/mip/HighsDomain.cpp
@@ -3215,6 +3215,8 @@ bool HighsDomain::ConflictSet::explainInfeasibility() {
       HighsInt ninfmin;
       HighsCDouble minAct;
       globaldom.computeMinActivity(0, len, inds, vals, ninfmin, minAct);
+      // This case should only happen when a globally unbounded column is bounded
+      // in the local domain, e.g., by a branching choice in some heuristic.
       if (ninfmin > 0) return false;
 
       return explainInfeasibilityLeq(inds, vals, len, rhs, double(minAct));
--- a/highs/mip/HighsDomain.cpp
+++ b/highs/mip/HighsDomain.cpp
@@ -3215,8 +3215,9 @@ bool HighsDomain::ConflictSet::explainInfeasibility() {
       HighsInt ninfmin;
       HighsCDouble minAct;
       globaldom.computeMinActivity(0, len, inds, vals, ninfmin, minAct);
-      // This case should only happen when a globally unbounded column is bounded
-      // in the local domain, e.g., by a branching choice in some heuristic.
+      // This case should only happen when a globally unbounded column is
+      // bounded in the local domain, e.g., by a branching choice in some
+      // heuristic.
       if (ninfmin > 0) return false;
 
       return explainInfeasibilityLeq(inds, vals, len, rhs, double(minAct));
--- a/highs/mip/HighsDomain.cpp
+++ b/highs/mip/HighsDomain.cpp
@@ -3215,7 +3215,7 @@ bool HighsDomain::ConflictSet::explainInfeasibility() {
       HighsInt ninfmin;
       HighsCDouble minAct;
       globaldom.computeMinActivity(0, len, inds, vals, ninfmin, minAct);
-      assert(ninfmin == 0);
+      if (ninfmin > 0) return false;
 
       return explainInfeasibilityLeq(inds, vals, len, rhs, double(minAct));
     }
--- a/highs/mip/HighsMipSolverData.cpp
+++ b/highs/mip/HighsMipSolverData.cpp
@@ -1120,8 +1120,10 @@ try_again:
     }
   }
 
-  const double transformed_solobj = static_cast<double>(static_cast<HighsInt>(mipsolver.orig_model_->sense_) *
-    mipsolver_quad_objective_value - mipsolver.model_->offset_);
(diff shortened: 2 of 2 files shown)


highs/mip/HighsDomain.cpp
```diff
@@ -3215,7 +3215,7 @@ bool HighsDomain::ConflictSet::explainInfeasibility() {
       HighsInt ninfmin;
       HighsCDouble minAct;
       globaldom.computeMinActivity(0, len, inds, vals, ninfmin, minAct);
-      assert(ninfmin == 0);
+      if (ninfmin > 0) return false;
```
BenChampion, line 3221: Worth adding a brief comment explaining why this can happen, and why we fail to explain the infeasibility? (Along the lines of your PR comment)

BenChampion (commented):
Minor inline comment.

highs/mip/HighsMipSolverData.cpp
```diff
@@ -1120,6 +1120,15 @@ double HighsMipSolverData::transformNewIntegerFeasibleSolution(
     }
   }
 
+  double transformed_solobj;
```
jajhall: I always disliked this conditional expression, as it's not as simple as it might be 

Since `HighsInt(mipsolver.orig_model_->sense_)` is +1 (-1) if it's `ObjSense::kMinimize` (`ObjSense::kMaximize`), a much simpler expression is

> `const double transformed_solobj = static_cast<double>(HighsInt(mipsolver.orig_model_->sense_)*mipsolver_quad_objective_value - mipsolver.model_->offset_);`

jajhall (approved):
This ` assert(ninfmin == 0);` has popped up a few times in issues/questions. I always assumed that it was there as an assert so that @lgottwald could pick up cases to validate what to do if it occurs - but didn't get any, so the code was never finished. 

Now that @Opt-Mucca has identified that ignoring the assert can (always?) leads to an error, I agree with @BenChampion that an explanation would be useful.

highs/mip/HighsMipSolverData.cpp
Opt-Mucca: I think it was used because you can't just multiply the expression by `ObjSense`. You need to account for the sign of the offset. So the actual expression is:
`static_cast<HighsInt>(mipsolver.orig_model_->sense_) * static_cast<double>(mipsolver_quad_objective_value - (static_cast<HighsInt>(mipsolver.orig_model_->sense_) * mipsolver.model_->offset_));`
I'm happy to add that if you find it easier to read, but I personally prefer the conditional logic.

highs/mip/HighsDomain.cpp
Opt-Mucca: Comment added!
`// This case should only happen when a globally unbounded column is bounded`
`// in the local domain, e.g., by a branching choice in some heuristic.`

highs/mip/HighsMipSolverData.cpp
jajhall: I'm only multiplying `mipsolver_quad_objective_value` by `ObjSense`, not the whole expression

The original code is

>   double transformed_solobj;
  if (mipsolver.orig_model_->sense_ == ObjSense::kMaximize) {
    transformed_solobj = -static_cast<double>(mipsolver_quad_objective_value +
                                              mipsolver.model_->offset_);
  } else {
    transformed_solobj = static_cast<double>(mipsolver_quad_objective_value -
                                             mipsolver.model_->offset_);
  }

which is 

> s = int(sense) 
if (s == -1) { //maximize
$f = -(f_q+d) = -f_q -d$
} else { // s = 1: minimize
$f = f_q-d = f_q - d$
}

which simplifies to

$f = s * f_q - d$

Thus my expression in code is correct and simpler.

`mipsolver_quad_objective_value` is the problem's true objective value, $c^Tx+d$, and only used for storing objective values. The shift of offset is done because the MIP solver maintains bounds without the offset, and with the sense applied to $c^Tx$

Opt-Mucca: Nice spot! Will change it now.

fwesselm (approved):
Looks good, thanks @Opt-Mucca !

PR #2465, opened 2025-07-13
Fix 2463

### LINKED ISSUE #2463: Problems with warm start solutions. It identifies them as infeasible
I'm trying to solve a model with a warm solution because it's very huge, more than 600000 variables. I check that all the constraints defined in the model are passed in the warm solution before generate the file and used it to solve the model.

However, I'm getting infeasibilities that makes me set mip_feasibility_tolerance to 1, to get it work. 
Running HiGHS 1.11.0 (git hash: 364c83a51): Copyright (c) 2025 HiGHS under MIT licence terms
Set option presolve to "on"
Set option parallel to "on"
Set option primal_feasibility_tolerance to 0.001
Set option dual_feasibility_tolerance to 0.001
Set option simplex_scale_strategy to 1
Set option simplex_max_concurrency to 6
Set option log_file to "C:\Users\sgarcia\AppData\Local\Temp\highs-qS9j05\highs.log"
Set option write_solution_to_file to true
Set option read_solution_file to "C:\sgn\wspc\pocTsCompras\warm-start.sol"
Set option solution_file to "C:\Users\sgarcia\AppData\Local\Temp\highs-qS9j05\solution.txt"
Set option mip_max_start_nodes to 100
Set option mip_feasibility_tolerance to 0.0001
MIP  model has 12875 rows; 625028 cols; 5920991 nonzeros; 612154 integer variables (612154 binary)
Coefficient ranges:
  Matrix [2e-02, 1e+06]
  Cost   [1e+00, 1e+04]
  Bound  [1e-03, 2e+08]
  RHS    [1e+00, 1e+00]
Assessing feasibility of MIP using primal feasibility and integrality tolerance of      0.0001
WARNING: Col 625026 has         infeasibility of       0.001 from [lower, value, upper] = [          0.001;               0;   2.0367255e+08]
WARNING: Row   6434 has         infeasibility of           1 from [lower, value, upper] = [              1;               0;             inf]
Solution has               num          max          sum
Col     infeasibilities      1        0.001        0.001
Integer infeasibilities      0            0            0
Row     infeasibilities      1            1            1
Row     residuals            0            0            0
Attempting to find feasible solution by solving LP for user-supplied values of discrete variables

Playing with tolerances I'm able to continue, but highs set objective to 0.

Running HiGHS 1.11.0 (git hash: 364c83a51): Copyright (c) 2025 HiGHS under MIT licence terms
Set option presolve to "on"
Set option parallel to "on"
Set option primal_feasibility_tolerance to 0.001
Set option dual_feasibility_tolerance to 1
Set option simplex_scale_strategy to 1
Set option simplex_max_concurrency to 6
Set option log_file to "C:\Users\sgarcia\AppData\Local\Temp\highs-qS9j05\highs.log"
Set option write_solution_to_file to true
Set option read_solution_file to "C:\sgn\wspc\pocTsCompras\warm-start.sol"
Set option solution_file to "C:\Users\sgarcia\AppData\Local\Temp\highs-qS9j05\solution.txt"
Set option mip_max_start_nodes to 100
Set option mip_feasibility_tolerance to 1
MIP  model has 12875 rows; 625028 cols; 5920991 nonzeros; 612154 integer variables (612154 binary)
Coefficient ranges:
  Matrix [2e-02, 1e+06]
  Cost   [1e+00, 1e+04]
  Bound  [1e-03, 2e+08]
  RHS    [1e+00, 1e+00]
Assessing feasibility of MIP using primal feasibility and integrality tolerance of           1
Solution has               num          max          sum
Col     infeasibilities      0            0            0
Integer infeasibilities      0            0            0
Row     infeasibilities      0            0            0
Row     residuals            0            0            0
Presolving model
10965 rows, 623119 cols, 4686433 nonzeros  15s
4984 rows, 464609 cols, 2666057 nonzeros  57s
4984 rows, 463212 cols, 2658834 nonzeros  69s

MIP start solution is feasible, objective value is 0

Solving MIP model with:
   4984 rows
   463212 cols (0 binary, 458228 integer, 0 implied int., 4984 continuous, 0 domain fixed)
   2658834 nonzeros

I'm sure that the model is feasible, because if I eliminate 200000 variables of the set I'm testing the objective is feasible, the model is feasible from the begining but with some noise.

I've attached model, opts and warm start, so you can test it.

[highs.opt.txt](https://github.com/user-attachments/files/21180129/highs.opt.txt)
[warm-start.sol.txt](https://github.com/user-attachments/files/21180128/warm-start.sol.txt)

[model.part05.rar.txt](https://github.com/user-attachments/files/21180184/model.part05.rar.txt)
[model.part01.rar.txt](https://github.com/user-attachments/files/21180185/model.part01.rar.txt)

[model.part04.rar.txt](https://github.com/user-attachments/files/21180209/model.part04.rar.txt)
[model.part03.rar.txt](https://github.com/user-attachments/files/21180208/model.part03.rar.txt)
[model.part02.rar.txt](https://github.com/user-attachments/files/21180207/model.part02.rar.txt)

jajhall:
#2463 has exposed numerous issues!

Fundamentally, the problem is that in using a .lp file to store your model, when HiGHS loads the file the order of the variables is not the same as the order of the name-value pairs in a solution file written after solving the original model.

Matching the name in a solution name-value pair to the name in the incumbent model fixes the issue. However, when models have no names, the names used to write out the model and solution were inconsistent, due to them being "invented" in different ways in the .lp and MPS file writer and solution writer. 

Hence, when writing out a model or solution, in this PR HiGHS defines names in the model if none exist. Note that, blank names in a model are replaced by "invented names", and HiGHS returns an error rather than write out models or solutions if the user has defined a names containing a space.

The same issue with solution files also occurs with basis files, but these are only written out as a stream of HighsBasisStatus values cast to `int` - so no names to match. Hence, in this PR, the `v2` basis file format has a name-status pair for each row and column.

This PR still doesn't fix #2463, which is a large model. In testing with smaller models, it was discovered that the HiGHS .lp file reader fails when row names are simple integers (as in `rgn`). 

Fixing this was going to be the subject of a separate PR, but it was done by accident in this branch!

### DIFF
--- a/check/TestCheckSolution.cpp
+++ b/check/TestCheckSolution.cpp
@@ -699,3 +699,42 @@ void runSetLpSolution(const std::string model) {
 
   highs.resetGlobalScheduler(true);
 }
+
+TEST_CASE("miplib-sol-file", "[highs_filereader]") {
+  const std::string test_name = Catch::getResultCapture().getCurrentTestName();
+  std::string sol_file = test_name + ".sol";
+  std::string lp_file = test_name + ".lp";
+  FILE* file = fopen(lp_file.c_str(), "w");
+  std::string file_content =
+      "Minimize\n obj: 2 sel_2 + sel_3\nSubject To\nr0: sel_0 - sel_1 + sel_4 "
+      ">= "
+      "2\nEnd\n";
+  if (dev_run) printf("Using .lp file\n%s", file_content.c_str());
+  fprintf(file, "%s", file_content.c_str());
+  fclose(file);
+  Highs h;
+  h.setOptionValue("output_flag", dev_run);
+  REQUIRE(h.readModel(lp_file) == HighsStatus::kOk);
+
+  file = fopen(sol_file.c_str(), "w");
+  file_content =
+      "=obj= 203672547.1\nsel_0 1\nsel_1 0\nsel_2 0\nsel_3 0\nsel_4 1\n";
+  if (dev_run) printf("Using .sol file\n%s", file_content.c_str());
+  fprintf(file, "%s", file_content.c_str());
+  fclose(file);
+  REQUIRE(h.readSolution(sol_file) == HighsStatus::kOk);
+
+  std::vector<double> solution = h.getSolution().col_value;
+  REQUIRE(solution[0] == 0);
+  REQUIRE(solution[1] == 0);
+  REQUIRE(solution[2] == 1);
+  REQUIRE(solution[3] == 0);
+  REQUIRE(solution[4] == 1);
+
+  REQUIRE(h.run() == HighsStatus::kOk);
+
+  std::remove(lp_file.c_str());
+  std::remove(sol_file.c_str());
+
+  h.resetGlobalScheduler(true);
+}


jajhall:
So, I've fixed #2463, with another sting in the tail.

HiGHS expects its solution file to have the line (eg)

> \# Columns 625028

where 625028 is the number of columns in the model when the solution file was written (or a negative value to imply that only a subset of solution values are given - the "sparse" format).

However, the user in #2463 wrote his own solution file with 

> \# Columns undefined

and "undefined" was read a the integer 0. Hence the solution wasn't read at all! I've added code to return an error if this string contains characters other than "-01234546789".

When a solution or basis file is read in, the model isn't required to have column and row names - and it may well not if the model's been passed in as a HighsLp instance, rather than by reading a file. In this case, the solution and basis data are assumed to be in the order corresponding to the columns and rows in the model - as would naturally be the case.

odow:
I don't have any strong thoughts or suggestions here. I just think that the LP has a number of inherent issues.

On renaming, I have a single utility function:

https://github.com/jump-dev/MathOptInterface.jl/blob/d66c13dab9cbd76c3451faffafd7d828a4e04ee1/src/FileFormats/utils.jl#L11-L247

that gets called from the LP writer

https://github.com/jump-dev/MathOptInterface.jl/blob/d66c13dab9cbd76c3451faffafd7d828a4e04ee1/src/FileFormats/LP/LP.jl#L441-L449

and MPS writer

https://github.com/jump-dev/MathOptInterface.jl/blob/d66c13dab9cbd76c3451faffafd7d828a4e04ee1/src/FileFormats/MPS/MPS.jl#L230-L239

We don't deal with sol files.

jajhall:
> I don't have any strong thoughts or suggestions here. I just think that the LP has a number of inherent issues.
> 
> On renaming, I have a single utility function:
> 
> https://github.com/jump-dev/MathOptInterface.jl/blob/d66c13dab9cbd76c3451faffafd7d828a4e04ee1/src/FileFormats/utils.jl#L11-L247

I wish .lp files didn't exist! 

Solution files are useful for providing an initial primal bound for a MIP when people run HiGHS from the command line

I wondered about resolving duplicate file names, but decided not to modify anything supplied by a user other than blank file names. If there is a duplicate name or a name containing a space character I return error.

jajhall:
@galabovaa I've spotted that the Python test failures are due to me changing the version number in basis files.

galabovaa:
Seems to have fixed the python tests! Don't worry about the external ones failing: I will look at those when I get back but there is nothing major going on there at the moment, for now it is more of a scaffold

jajhall:
This closes #2463

PR #2470, opened 2025-07-16
Now summing duplicate matrix entries resulting from repeated terms in .lp constraints

### LINKED ISSUE #2469: Error reading a model due to variable duplication
When I'm constructing a model file, sometimes I need that the same variable appears more than once, and don't over engineering the solution to group all the occurences of the variable in the same restriction. Other solvers like scip don't provide an error and are able to solve the model.

Highs log:
Running HiGHS 1.11.0 (git hash: 364c83a): Copyright (c) 2025 HiGHS under MIT licence terms
ERROR:   LP matrix packed vector 0, entry 1, is duplicate index 0
Error loading file

[group-model-2025-07-16T11-21-56-249Z.lp.txt](https://github.com/user-attachments/files/21255854/group-model-2025-07-16T11-21-56-249Z.lp.txt)
[group-model-2025-07-16T11-21-56-249Z.sol.txt](https://github.com/user-attachments/files/21255855/group-model-2025-07-16T11-21-56-249Z.sol.txt)

jajhall:
Since constraints in a .lp file are an explicit sum, having more than one term involving the same variable is fine, as the corresponding single coefficient value is well defined. Previously such terms led to duplicate matrix entries, triggering an error return when the HighsLp instance was checked. 

A filter has been added after reading the .lp file, so that such repeated entries are summed as single entries.

This will close #2469

### DIFF
--- a/check/TestFilereader.cpp
+++ b/check/TestFilereader.cpp
@@ -485,7 +485,7 @@ TEST_CASE("lp-duplicate-variable", "[highs_filereader]") {
   std::string file_content =
       "Minimize\n obj: 2 x + y + z\nSubject To\nr0: 2 x + y - x + 0 z >= "
       "2\nr1: y + x - y >= 1\nEnd\n";
-  printf("Using .lp file\n%s", file_content.c_str());
+  if (dev_run) printf("Using .lp file\n%s", file_content.c_str());
   fprintf(file, "%s", file_content.c_str());
   fclose(file);
   Highs h;
--- a/check/TestFilereader.cpp
+++ b/check/TestFilereader.cpp
@@ -477,3 +477,20 @@ TEST_CASE("mps-silly-names", "[highs_filereader]") {
   HighsStatus return_status = h.readModel(model_file);
   REQUIRE(return_status == HighsStatus::kOk);
 }
+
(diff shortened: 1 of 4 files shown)


odow:
Same thing with MPS files?

jajhall:
> Same thing with MPS files?

Unclear: does a second instance of the same column entry replace or add to the existing entry? Of course I can see what commercial solvers do - or wait until a request comes up. I want to get back to the concurrent MIP solver...

odow:
We sum. It didn't occur to me to do anything else, so I didn't check what other people do.

But now you say: I don't actually know if we have explicit tests for this.

jajhall:
> We sum. It didn't occur to me to do anything else, so I didn't check what other people do.
> 
> But now you say: I don't actually know if we have explicit tests for this.

Gurobi (and HiGHS) use the first value encountered in an MPS file, ignoring any subsequent values for the same row in the column. Both warn about this - a limited number of times - but HiGHS says what numerical values are ignored, and gives a summary line for the file about duplicate values.

Gurobi and HiGHS both ignore column entries corresponding to undefined rows

odow:
TIL. There are so many potential ambiguities with how you parse these files.

PR #2474, opened 2025-07-20
Mip race

jajhall:
Some residual code from MIP race experience. 

`HighsTerminator` struct allows termination of the MIP solver to be communicated more readily - and across multiple threads 

`MipSolutionSource` now ordered alphabetically by key letters


jajhall:
Terminated MIP race instances take longer because they may be solving sub-mips when termination occurs.

Introduced `Highs::termination_status_` and `HighsTerminator` struct for MIP solver so that 

- sub-MIPs can be terminated
- the MIP solver can be terminated more promptly 
- the infinite loop when `HighsTermination` occurs in `HighsMipSolverData::performRestart` can be eliminated. This was _always_ a risk, since it's caused by `HighsMipSolverData::checkLimits()` returning true - as it might if the time limit were reached (for example).

highs/mip/HighsMipSolverData.cpp
```diff
@@ -2785,3 +2889,177 @@ void HighsMipSolverData::updatePrimalDualIntegral(const double from_lower_bound,
...
+
+void MipRaceRecord::clear() { this->incumbent.clear(); }
+
+void MipRaceRecord::initialise(const HighsInt mip_race_concurrency,
+                               const HighsInt num_col) {
+  this->clear();
+  MipRaceIncumbent incumbent_;
+  incumbent_.initialise(num_col);
+  for (HighsInt instance = 0; instance < mip_race_concurrency; instance++)
+    this->incumbent.push_back(incumbent_);
```
mathgeekcoder: Minor feedback.  Current code will create 1 extra memory allocation/deallocation of num_col (i.e., ```incumbent_```).  Perhaps consider:

```cpp
  incumbent_.initialise(num_col);
  for (HighsInt instance = 1; instance < mip_race_concurrency; instance++)
    this->incumbent.push_back(incumbent_);

  this->incumbent.push_back(std::move(incumbent_));
```

jajhall: Thanks @mathgeekcoder: I don't have the C++ to think of that!

jajhall:
Now that solution of sub-MIPs can be terminated, the MIP solver instances terminated in a race now take only a little longer than the winner.

Hence we have a 7X speedup on fiball 

Further experiments will follow once I've cleaned up the dev printf statements

highs/mip/HighsMipSolverData.cpp
mathgeekcoder: Apologies. The suggestion I made was correct - but not the best.  The following would be much better (and cleaner too).  Also, the memset in initialise is faster than a memcpy by the copy constructor.

```cpp
  this->clear();
  this->incumbent.resize(mip_race_concurrency);

  for (HighsInt instance = 0; instance < mip_race_concurrency; instance++)
    this->incumbent[instance].initialise(num_col);
```

highs/mip/HighsMipSolverData.cpp
```diff
@@ -2785,3 +2889,177 @@ void HighsMipSolverData::updatePrimalDualIntegral(const double from_lower_bound,
...
+  if (this->finish_write_incumbent < start_write_incumbent)
+    return kMipRaceNoSolution;
+  // finish_write_incumbent = start_write_incumbent so start reading
+  objective_ = this->objective;
+  solution_ = this->solution;
+  // Read is OK if no new write has started
+  return this->start_write_incumbent == start_write_incumbent
+             ? start_write_incumbent
+             : kMipRaceNoSolution;
+}
```
mathgeekcoder: Nice trick with the start/finish checks - but I believe there's still a chance where this could fail in a race condition... especially if some "smart" compiler optimizes away the final ```this->start_write_incumbent == start_write_incumbent``` check.

Perhaps consider using ```std::atomic<HighsInt>``` for your start/finish variables?  The atomic class is lock free and very fast on most systems.

```cpp
struct MipRaceIncumbent {
  std::atomic<HighsInt> start_write_incumbent = kMipRaceNoSolution;
  std::atomic<HighsInt> finish_write_incumbent = kMipRaceNoSolution;
  double objective = -kHighsInf;
  std::vector<double> solution;
  void clear();
  void initialise(const HighsInt num_col);
  void update(const double objective, const std::vector<double>& solution);
  HighsInt read(const HighsInt last_incumbent_read, double& objective_,
                std::vector<double>& solution_) const;

  MipRaceIncumbent() = default;

  MipRaceIncumbent(const MipRaceIncumbent& copy) {
    start_write_incumbent = copy.start_write_incumbent.load();
    finish_write_incumbent = copy.finish_write_incumbent.load();
    objective = copy.objective;
    solution = copy.solution;
  }

  MipRaceIncumbent(MipRaceIncumbent&& moving) {
    start_write_incumbent = moving.start_write_incumbent.load();
    finish_write_incumbent = moving.finish_write_incumbent.load();
    objective = moving.objective;
    solution = std::move(moving.solution);
  }
};
```

mathgeekcoder: That said, I'm guessing you're trying to avoid heavier ```std::mutex``` for thread synchronization.  As a lighter alternative, you can also use ```std::atomic_flag``` for a ```SpinLock``` implementation.  For example:

```cpp
class Spinlock {
private:
  std::atomic_flag lock_flag = ATOMIC_FLAG_INIT;

public:
  void lock() {
    while (lock_flag.test_and_set(std::memory_order_acquire)) {
      // Busy-wait (spin) until the lock is released
    }
  }
  
  bool try_lock() noexcept {
    return lock_flag.test_and_set(std::memory_order_acquire);
  }
  void unlock() { lock_flag.clear(std::memory_order_release); }
};
```

**Note: There is likely better SpinLock code available.**

highs/lp_data/Highs.cpp
```diff
@@ -4029,25 +4031,148 @@ HighsStatus Highs::callSolveMip() {
...
+        highsOpenLogFile(instance_options, worker_log_file);
+      worker_options.push_back(instance_options);
+      /*
+      HighsMipSolver worker_instance(worker_callback, worker_options[instance],
+      lp, solution_); worker.push_back(&worker_instance);
+      */
+    }
+    highs::parallel::for_each(
+        0, mip_race_concurrency, [&](HighsInt start, HighsInt end) {
+          for (HighsInt instance = start; instance < end; instance++) {
```
mathgeekcoder: BTW: AFAIK this inner for loop will execute only on one thread, so it might be a bit pointless if it loops more than once (i.e., that means a previous ```HighsMipSolver::run``` has already finished).

That said, the default grain size = 1, so it's unlikely to ever occur.  You could probably get rid of the inner for loop and just take ```instance = start``` instead.

Alternatively, you could try to use ```TaskGroup``` directly and spawn each task.  It would be more effort, but you might've been able to use the ```TaskGroup::cancel``` function to interrupt the other threads.

highs/mip/HighsMipSolverData.cpp
mathgeekcoder: Oh!  I forgot that we already have ```HighsSpinMutex```.  So ignore my SpinLock code above.

highs/mip/HighsMipSolver.h
```diff
@@ -19,6 +19,60 @@ struct HighsPseudocostInitialization;
...
+  HighsLogOptions log_options;
+  std::vector<HighsInt> last_incumbent_read;
+  void clear();
+  void initialise(const HighsInt mip_race_concurrency,
+                  const HighsInt my_instance_, MipRaceRecord* record_,
+                  const HighsLogOptions log_options_);
+  HighsInt concurrency() const;
+  void update(const double objective, const std::vector<double>& solution);
+  bool newSolution(const HighsInt instance, double objective,
+                   std::vector<double>& solution);
```
mathgeekcoder: The ```newSolution``` method is missing the **&** for the ```double objective```.  Without this, the solution sharing across threads isn't updating as frequently as it could.

```cpp
bool newSolution(const HighsInt instance, double& objective,
                   std::vector<double>& solution);
```

highs/lp_data/Highs.cpp
```diff
@@ -4029,25 +4031,148 @@ HighsStatus Highs::callSolveMip() {
...
+    for (HighsInt instance = 0; instance < mip_race_concurrency; instance++) {
+      const HighsMipSolverInfo& solver_info =
+          instance == 0 ? mip_solver_info : worker_info[instance];
+      HighsModelStatus instance_model_status = solver_info.modelstatus;
+      highsLogUser(options_.log_options, HighsLogType::kInfo,
+                   "   Solver %d has best objective %15.8g, gap %6.2f\% (time "
+                   "= %6.2f), and status %s\n",
+                   int(instance), solver_info.solution_objective,
+                   1e2 * solver_info.gap, mip_time[instance],
+                   modelStatusToString(instance_model_status).c_str());
```
mathgeekcoder: Minor fix:  ```\%``` should be ```%%```

i.e., `" Solver %d has best objective %15.8g, gap %6.2f`\\%` (time "`

highs/mip/HighsMipSolverData.cpp
jajhall: Implemented

highs/mip/HighsMipSolverData.cpp
jajhall: I tried implementing the new `struct MipRaceIncumbent`, but got the following compiler errors

> /home/jajhall/HiGHS/highs/mip/HighsMipSolver.h:38:25: error: field ‘start_write_incumbent’ has incomplete type ‘std::atomic<int>’
   38 |   std::atomic<HighsInt> start_write_incumbent = kMipRaceNoSolution;
      |                         ^~~~~~~~~~~~~~~~~~~~~

mathgeekcoder: Not entirely sure.  Did you ```#include <atomic>```?

highs/mip/HighsMipSolverData.cpp
```diff
@@ -2607,47 +2627,119 @@ bool HighsMipSolverData::interruptFromCallbackWithData(
...
+    if (!mip_race.newSolution(instance, instance_solution_objective_value,
+                              instance_solution))
+      continue;
+    // Have read a new incumbent
+    std::vector<double> reduced_instance_solution;
+    reduced_instance_solution =
+        postSolveStack.getReducedPrimalSolution(instance_solution);
+    addIncumbent(reduced_instance_solution, instance_solution_objective_value,
+                 kSolutionSourceHighsSolution);
+  }
```
mathgeekcoder: I believe there is a bug with the objective value when ```mipsolver.model_->offset_ != 0```.  This can prevent improved solutions from being accepted by other threads.

**Details:** 
```addIncumbent``` checks ```instance_solution_objective_value < upper_bound``` before updating, but ```upper_bound``` excludes the ```mipsolver.model_->offset_``` value and  ```instance_solution_objective_value``` includes it.

**Potential fix:**
```cpp
    // exclude offset from objective value
    instance_solution_objective_value -= mipsolver.model_->offset_;

    addIncumbent(reduced_instance_solution, instance_solution_objective_value,
                 kSolutionSourceHighsSolution);
```

highs/mip/HighsMipSolverData.cpp
mathgeekcoder: Ah, sorry.  That's a linux vs windows issue.  The following should work on both.

```cpp
std::atomic<HighsInt> start_write_incumbent {kMipRaceNoSolution}; 
std::atomic<HighsInt> finish_write_incumbent {kMipRaceNoSolution};
```

jajhall: > Not entirely sure. Did you `#include <atomic>`?

Yes, but doesn't fix the error

> Ah, sorry. That's a linux vs windows issue. The following should work on both.

Works on Linux! :-)

highs/mip/HighsMipSolver.h
jajhall: That's an important omission! It meant that the objective for the solution read from another thread wasn't being passed back, so the solution from the other thread never replaced the incumbent. Hence, sharing solutions was the same as a pure race. 

Making the correction, the performance on `fiball` when sharing solutions is slowed greatly, as none of the threads runs like the "lucky" `random_seed=1` case. This is what I'd expected to see. However, for problems where there isn't an extremely lucky `random_seed` value, maybe this will open the door to improved performance!

Naturally I'll experiment

highs/lp_data/Highs.cpp
jajhall: Corrected

highs/mip/HighsMipSolverData.cpp
jajhall: Thanks: the correction depends on whether the problem is a maximization or minimization

highs/mip/HighsMipSolver.h
mathgeekcoder: I noticed similar behaviour in my experiments.  Sharing solutions can help on other instances, but not really on `fiball`.  That said, when I performed the presolve once and shared that across the other threads, `fiball` was fast again.

BTW: I had to make many changes to ensure presolve is only performed once.  Do you know a better way?

highs/mip/HighsMipSolverData.cpp
mathgeekcoder: Good to know.  I was thinking the sign for `offset_` might've already accounted for min/max.

jajhall: Fixed

highs/mip/HighsMipSolver.h
mathgeekcoder: BTW: After more experiments with `fiball`, sharing the presolve doesn't necessarily solve it fast.  More investigation needed!

jajhall: Now I've thought a little, it's easy internally to perform presolve only once, and then race the solution of the presolved MIP

jajhall:
Can now perform presolve before the MIP race so that all participants are solving the same problem.

Controlled by option `mip_race_single_presolve` which is true by default.

jajhall:
The performance gain from the MIP race didn't justify introducing a non-deterministic solver into HiGHS, or the investment of time required to fix the segfaults and incorrect deductions of infeasibility. 

All the code relating to the MIP race has been deleted so that the one useful outcome - the `HighsTerminator` can be merged into latest

jajhall:
The performance gain from the MIP race didn't justify introducing a non-deterministic solver into HiGHS, or the investment of time required to fix the segfaults and incorrect deductions of infeasibility. 

All the code relating to the MIP race has been deleted so that the one useful outcome - the `HighsTerminator` can be merged into latest

PR #2482, opened 2025-07-31
Knapsack count

jajhall:
If the MIP solver is given a knapsack problem, it solves it by DP. Stats on the occurrence of knapsack problems are accumulated and logged. Tested by unit test `knapsack`.

### DIFF
--- a/highs/mip/HighsPrimalHeuristics.cpp
+++ b/highs/mip/HighsPrimalHeuristics.cpp
@@ -1786,7 +1786,8 @@ HighsStatus HighsPrimalHeuristics::solveMipKnapsackReturn(
 HighsStatus HighsPrimalHeuristics::solveMipKnapsack() {
   const HighsLp& lp = *(mipsolver.model_);
   const HighsLogOptions& log_options = mipsolver.options_mip_->log_options;
-  assert(mipsolver.mipdata_->mipIsKnapsack(false));
+  // Check silently that this really is a knapsack
+  assert(mipsolver.mipdata_->mipIsKnapsack(true));
 
   const bool upper = lp.row_upper_[0] < kHighsInf;
   const HighsInt constraint_sign = upper ? 1 : -1;
--- a/highs/presolve/HPresolve.cpp
+++ b/highs/presolve/HPresolve.cpp
@@ -4339,7 +4339,7 @@ HPresolve::Result HPresolve::presolve(HighsPostsolveStack& postsolve_stack) {
   analysis_.setup(this->model, this->options, this->numDeletedRows,
                   this->numDeletedCols);
 
-  if (options->presolve != kHighsOffString) {
+  if (options->presolve != kHighsOffString && reductionLimit > 0) {
     if (mipsolver) mipsolver->mipdata_->cliquetable.setPresolveFlag(true);
     if (!mipsolver || mipsolver->mipdata_->numRestarts == 0)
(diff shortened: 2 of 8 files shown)


jajhall:
Also tested by unit test `read-miplib-solution`

jajhall:
Now incorporated integral scaling into `mipIsKnapsack`, but still not found any knapsack sub-problems in 100 easy MIPLIB 2017

highs/presolve/HPresolve.cpp
```diff
@@ -4680,11 +4688,12 @@ HighsModelStatus HPresolve::run(HighsPostsolveStack& postsolve_stack) {
     if (options->presolve != kHighsOffString &&
         reductionLimit < kHighsSize_tInf) {
       highsLogUser(options->log_options, HighsLogType::kInfo,
-                   "Presolve performed %" PRId64 " of %" PRId64
-                   " permitted reductions\n",
-                   postsolve_stack.numReductions(), reductionLimit);
+                   "\nPresolve: Performed %" PRId64 " of %" PRId64
+                   " permitted true reductions\n",
```
Opt-Mucca, line 4692: Should it being worded differently to something like `x reductions (y from scaling)` or `x reductions and y scaling reductions`.
That would avoid using ambiguous like "true reductions".

highs/mip/HighsPrimalHeuristics.cpp
```diff
@@ -1645,3 +1647,212 @@ void HighsPrimalHeuristics::flushStatistics() {
...
+                    knapsack_optimal_objective_value, solution);
+  if (mipsolver.modelstatus_ != HighsModelStatus::kOptimal)
+    return solveMipKnapsackReturn(HighsStatus::kError);
+
+  // Get the objective value corresponding to the original problem
+  const double solution_objective =
+      lp.offset_ - sense * knapsack_optimal_objective_value;
+
+  // Compute the objective directly as a check
+  double check_objective = lp.offset_;
```
Opt-Mucca, line 1838: Would this check be removed if we merge this?

highs/mip/HighsPrimalHeuristics.cpp
```diff
@@ -1645,3 +1647,212 @@ void HighsPrimalHeuristics::flushStatistics() {
...
+                               std::vector<double>& solution) {
+  assert(capacity > 0);
+
+  // Set up the DP result array, indicating that no optimal objective
+  // values are known
+  std::vector<std::vector<double>> dp_result(
+      num_item + 1, std::vector<double>(capacity + 1, -1));
+  // Set up the item use array, indicating that items are not used
+  std::vector<std::vector<bool>> use_item(
+      num_item + 1, std::vector<bool>(capacity + 1, false));
```
Opt-Mucca, line 1719: I guess the downside here is that this vector can be huge?

Opt-Mucca (commented):
The changes look good! I've left a comment or two, but they're for minor things.

I think I screwed up the language when talking about knapsacks and sub-milps. I was referring to some higher dimensional generalisations (an example MIPLIB benchmark instance: https://miplib.zib.de/instance_details_mik-250-20-75-4.html).
If we want some instances of (what I believe are normal knapsacks) then we can use: https://matilda.unimelb.edu.au/matilda/problems/opt/knapsack#knapsack (although off memory they are all rather trivial for current solvers)

jajhall:
No value in having a knapsack solver in HiGHS

PR #2483, opened 2025-08-01
Fix warnings

fwesselm:
Fix some warnings concerned with type conversions / loss of precision reported [here](https://github.com/fwesselm/HiGHS/actions/runs/16650177883/job/47120526408).

### DIFF
--- a/highs/mip/HighsCliqueTable.cpp
+++ b/highs/mip/HighsCliqueTable.cpp
@@ -312,7 +312,7 @@ void HighsCliqueTable::doAddClique(const CliqueVar* cliquevars,
   cliques[cliqueid].equality = equality;
   cliques[cliqueid].origin = origin;
 
-  std::set<std::pair<HighsInt, HighsInt>>::iterator it;
+  decltype(freespaces)::iterator it;
   HighsInt maxEnd;
   if (freespaces.empty() ||
       (it = freespaces.lower_bound(


Opt-Mucca (approved):
Support for this change! Will make my IDE happier.

highs/lp_data/HighsLp.cpp
```diff
@@ -581,11 +581,11 @@ void HighsNameHash::form(const std::vector<std::string>& name) {
...
 
 bool HighsNameHash::hasDuplicate(const std::vector<std::string>& name) {
-  size_t num_name = name.size();
   this->clear();
   bool has_duplicate = false;
-  for (size_t index = 0; index < num_name; index++) {
-    has_duplicate = !this->name2index.emplace(name[index], index).second;
+  for (size_t index = 0; index < name.size(); index++) {
+    has_duplicate =
+        !this->name2index.emplace(name[index], static_cast<int>(index)).second;
```
jajhall, line 588: I see why this has to be `name2index.emplace(name[index], static_cast<int>(index))` - since  `HighsNameHash::name2index` is `std::unordered_map<std::string, int>`, but it seems odd to have the loop parameter `index` to be `size_t`. Yes, it fits with `index < name.size()`, but it looks confusing

I should have created `HighsNameHash::name2index` as `std::unordered_map<std::string, HighsInt>`, and then all "ints" would be `HighsInt`

Not a big issue

PR #2484, opened 2025-08-01
Clean up Feasibility Jump logging

BenChampion:
Minor change making Feasibility Jump (FJ) use the HiGHS logging calls rather than `printf`. Because HiGHS sets `log_options` correctly for submips, we no longer need FJ's internal `verbosity` parameter.

I also adjusted the logging levels for the different messages to make them more sensible.


Opt-Mucca (approved):
Good change!

jajhall (approved):
Looks good

PR #2490, opened 2025-08-08
Expand MIP debug solution capabilities

Opt-Mucca:
This expands how the MIP debug solution can be used. 

Previously: Presolve had to be disabled and the solution was removed if a restart occurred. The expected solution format was also SCIP's .sol format with anaonymised variable names. 

Now: The expected solution format is a .sol file produced by HiGHS (of the original problem). The solution is then saved both in the original and transformed space when loaded in, and transformed appropriately at each restart and after presolve. 

I guess this feature is not used by many people? It's not meant to public facing, and the code is hidden by a compiler flag. This makes is substantially easier to use, and has already been helpful. 

I'm still not too sure on C++ that isn't doing MIP stuff. Opinions on how the `.sol` file is parsed are welcome.

### DIFF
--- a/highs/mip/HighsDebugSol.cpp
+++ b/highs/mip/HighsDebugSol.cpp
@@ -35,30 +35,45 @@ void HighsDebugSol::activate() {
     if (file) {
       std::string varname;
       double varval;
+      std::string line;
+      bool incolsection;
       std::map<std::string, int> nametoidx;
 
-      for (HighsInt i = 0; i != mipsolver->model_->num_col_; ++i)
-        nametoidx["c" + std::to_string(i)] = i;
+      for (HighsInt i = 0; i != mipsolver->orig_model_->num_col_; ++i)
+        nametoidx[mipsolver->orig_model_->col_names_[i]] = i;
 
-      debugSolution.resize(mipsolver->model_->num_col_, 0.0);
-      while (!file.eof()) {
-        file >> varname;
+      debugOrigSolution.resize(mipsolver->orig_model_->num_col_, 0.0);
+      while (std::getline(file, line)) {
+        // Check for start and stop markers
+        if (line.find("# Columns") != std::string::npos) {
+          incolsection = true;
+          continue;
+        }
+        if (line.find("# Rows") != std::string::npos) {
+          break;
+        }
+        if (!incolsection)
+          continue;
+
+        std::istringstream linestream(line);
+        linestream >> varname;
         auto it = nametoidx.find(varname);
         if (it != nametoidx.end()) {
-          file >> varval;
+          linestream >> varval;
+          debugOrigSolution[it->second] = varval;
           highsLogDev(mipsolver->options_mip_->log_options, HighsLogType::kInfo,
                       "%s = %g\n", varname.c_str(), varval);
-          debugSolution[it->second] = varval;
+          debugOrigSolution[it->second] = varval;
         }
-
-        file.ignore(std::numeric_limits<std::streamsize>::max(), '\n');
       }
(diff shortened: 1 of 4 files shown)


galabovaa (commented):
LGTM

Parsing with stringstream seems fine

highs/mip/HighsDebugSol.cpp
```diff
@@ -35,30 +35,45 @@ void HighsDebugSol::activate() {
...
-        file.ignore(std::numeric_limits<std::streamsize>::max(), '\n');
       }
+      debugSolution = debugOrigSolution;
 
       HighsCDouble debugsolobj = 0.0;
-      for (HighsInt i = 0; i != mipsolver->model_->num_col_; ++i)
-        debugsolobj += mipsolver->model_->col_cost_[i] * debugSolution[i];
+      for (HighsInt i = 0; i != mipsolver->orig_model_->num_col_; ++i)
+        debugsolobj +=
+            mipsolver->orig_model_->col_cost_[i] * debugOrigSolution[i];
```
fwesselm: Is it worth having `static_cast<HighsCDouble>(mipsolver->orig_model_->col_cost_[i])` here? Otherwise, the multiplication would be done using double precision, but maybe that's OK...

highs/mip/HighsMipSolverData.cpp
```diff
@@ -1022,10 +1011,17 @@ void HighsMipSolverData::runSetup() {
...
-    assert(!debugSolution.debugSolActive ||
-           checkSolution(debugSolution.debugSolution));
+  if (debugSolution.debugSolActive) {
+    debugSolution.debugSolution.clear();
+    debugSolution.debugSolution = postSolveStack.getReducedPrimalSolution(
+        debugSolution.debugOrigSolution);
+    debugSolution.debugSolObjective = 0;
+    HighsCDouble debugsolobj = 0.0;
+    for (HighsInt i = 0; i != mipsolver.model_->num_col_; ++i)
+      debugsolobj += mipsolver.colCost(i) * debugSolution.debugSolution[i];
```
fwesselm: Same question as above: `static_cast<HighsCDouble>(mipsolver.colCost(i))`?

fwesselm (approved):
Thanks, @Opt-Mucca, this looks good.

Opt-Mucca:
I've put "draft" in the title because for some instance I'm debugging it brought up a potential bug in presolve. I thought the debug solution was only intended to work after presolve, so I quickly disabled it with a commit, and now I need to investigate if there's actually a bug In presolve or the debug solution should just be disabled for this bit of the solve process.....

Edit: Was just an oversight by me. The commit a few days ago should have fixed the problem (ignore the debug solution until post-presolve)

highs/mip/HighsDebugSol.cpp
Opt-Mucca: It does make sense here. I've added it. (It's now* made me second guess how many terms in cut generation are "HighsCDouble", but end up being the sum of products of doubles.

Opt-Mucca:
The issue with the existing `.sol` parser is that it never worked for me without a fair bit of manual editing, and that was not sustainable for larger instances. There is a good argument for expanding this reader though as that would help with convenience (the argument against this would be we could always just load in the solution of format `X` and convert it to HiGHS style)

jajhall:
> Shouldn't we reuse the existing parser code we have for .sol files? Or am I missing something that makes this problematic?

The existing parser code does (now) have a "sparse" option. The reason  that the SCIP .sol format was used was so that the optimal solution from XPress could be used (by Leona) to check for cut violations.

PR #2491, opened 2025-08-11
Small addition to row presolve

fwesselm:
- Add a simple test to `HPresolve::rowPresolve` that fixes a binary variable to one of its bounds if (tentatively) setting it to the opposite bound makes the problem infeasible.
- This is a special case of general bound tightening.
- Unfortunately, the effect on performance in terms of geometric means of running times and numbers of branching nodes is very small. However, it may still be useful for certain models.

### DIFF
--- a/highs/presolve/HPresolve.cpp
+++ b/highs/presolve/HPresolve.cpp
@@ -3276,10 +3276,12 @@ HPresolve::Result HPresolve::rowPresolve(HighsPostsolveStack& postsolve_stack,
           model->col_upper_[col] != model->col_lower_[col] + 1.0)
         continue;
 
-      // compute offset
-      HighsCDouble offset =
-          std::abs(val) * (static_cast<HighsCDouble>(model->col_upper_[col]) -
-                           static_cast<HighsCDouble>(model->col_lower_[col]));
+      // lambda for computing offset
+      auto computeOffset = [&](HighsInt col, double val) {
+        return std::abs(val) *
+               (static_cast<HighsCDouble>(model->col_upper_[col]) -
+                static_cast<HighsCDouble>(model->col_lower_[col]));
+      };
 
       auto degree1Tests = [&](HighsInt col, double val, HighsInt direction,
                               double rowActivityBound, double rowBound) {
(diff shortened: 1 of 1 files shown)


highs/presolve/HPresolve.cpp
```diff
@@ -3259,6 +3259,61 @@ HPresolve::Result HPresolve::rowPresolve(HighsPostsolveStack& postsolve_stack,
...
+      // skip continuous variables and non-binary variables
+      if (model->integrality_[col] == HighsVarType::kContinuous ||
+          model->col_upper_[col] != model->col_lower_[col] + 1.0)
+        continue;
+
+      // lambda for computing offset
+      auto computeOffset = [&](HighsInt col, double val) {
+        return std::abs(val) *
+               (static_cast<HighsCDouble>(model->col_upper_[col]) -
+                static_cast<HighsCDouble>(model->col_lower_[col]));
```
Opt-Mucca, line 3283: Shouldn't this always equal `abs(val)`?

Opt-Mucca (approved):
Had to play around with some pen and paper a bit to check everything (well done on keeping track of all the signs), but looks good!

I'd assume that presolve was already catching most of these cases, but I guess things like probing have a budget and wouldn't catch everything. This should definitely be merged if it is performance improving.

highs/presolve/HPresolve.cpp
fwesselm: The presolve code explicitly selects functions from the namespace `std`. I think overloading can create a mess if only `abs` is used (see [this discussion](https://stackoverflow.com/questions/21392627/abs-vs-stdabs-what-does-the-reference-say)).

Opt-Mucca: Nononononono. In the sense that we know that `col_upper_[col] - col_lower_[col] = 1` already?

fwesselm: Yes 💯 , sorry for the confusion. The code is written this way in order to handle general-integer variables, which is currently not used. I could change that if you like.

Opt-Mucca: If there's a plan to potentially extend it at some point then I'm fine with leaving it. It just seemed like overkill for what it was doing. Happy with everything then!

PR #2494, opened 2025-08-13
Minor changes to presolve

fwesselm:
- Add two utility methods `HPresolve::checkUpdateRowDualImpliedBounds` and `HPresolve::checkUpdateColImpliedBounds` to check whether bounds on row duals and columns can be updated.
- This allows for simplifying the presolve code a little.
- Testing showed that these change do not affect HiGHS behavior at all.


Opt-Mucca (approved):
Looks good!

PR #2495, opened 2025-08-13
Warnings g++ linux 

galabovaa:
Some warnings popping up with g++ on linux


highs/interfaces/highs_c_api.cpp
```diff
@@ -1309,32 +1309,32 @@ HighsInt Highs_getIis(void* highs, HighsInt* iis_num_col, HighsInt* iis_num_row,
...
     }
   }
   if (row_bound != nullptr) {
-    for (size_t i = 0; i < *iis_num_row; i++) {
+    for (size_t i = 0; i < (size_t)*iis_num_row; i++) {
       row_bound[i] = iis.row_bound_[i];
     }
   }
   if (col_status != nullptr) {
-    for (size_t i = 0; i < ((Highs*)highs)->getLp().num_col_; i++) {
```
jajhall, line 1332: Isn't this also going to flag up a warning (with some architectures/compilers) since `getLp().num_col_` is `HighsInt`, not `size_t`?

jajhall (approved):
@fwesselm seems to want `static_cast<HighsInt>(fred)` rather than `HighsInt(fred)` (or) (HighsInt)fred`: not that I know the difference between the three, other than `HighsInt(fred)` possibly being more modern than (HighsInt)fred`!

Otherwise OK

fwesselm:
> @fwesselm seems to want `static_cast<HighsInt>(fred)` rather than `HighsInt(fred)` (or) (HighsInt)fred`: not that I know the difference between the three, other than `HighsInt(fred)` possibly being more modern than (HighsInt)fred`!
> 
> Otherwise OK

As far as I understand, C++ style casts (`static_cast` etc.) are checked at compile-time. I think these also express the intent more clearly (compared to C style casts).

PR #2496, opened 2025-08-13
Flow cover cuts

Opt-Mucca:
This adds the lifted simple generalised flow cover cut from https://link.springer.com/content/pdf/10.1007/s101070050067.pdf

Currently a flow cover cut is only generated for `HighsPathSeparator`. I've been playing around with enabling it for `HighsTableauSeparator`, but can't decide if that has some performance loss. I've avoided doing anything for `mod-k` and `conflicts` because they already have a "dont-spend-much-time" cut flag attached to them. 

The general workflow of how the cut is generated (and which bits of the code does what):
- `preprocessSNFRelaxation` -> check whether the aggregated row can be used for a flow cover
- `initSNFRelaxation` -> initialise some data structures
- `transLp.transformSNFRelaxation` -> create the single-node flow relaxation (SNFR). That is, turn `\sum c_i x_i + \sum a_i y_i <= a_0` (x_i binary, y_i real non-neg) into `\sum_{j \in N+} y'_j - \sum_{j \in N-} y'_j <= b`, where `y'_j <= u_j x_j`.
- `computeFlowCover` -> Get sets C+ subset N+ and C- subset N- with `sum_{j in C+} u_j - sum_{j in C-} u_j = b + lambda, lambda > 0` (C+ and C- are then the "flow cover")
- `separateLiftedFlowCover` -> compute the actual cut (half of the code is creating data for the lifting function and functions for determining coefficients related to the lifting function)
- `transLp.cleanup` -> Remove slack, small coefficients, and calculate the efficacy

My smaller performance runs are showing a nice improvement on reducing the number of nodes, but it's still slightly slower than the baseline w.r.t. time. I'd say it's too heavily swayed by outliers currently. This needs to be more rigorously tested performance-wise before being merged. 

I opened the PR to get any feedback from a design perspective, and in case anyone wanted to contribute. I also need to test whether or not this is helpful for the energy instances specifically. 

Potential TODO: I am currently just taking the `bestVlb` or `bestVub` that is already stored for `transformSNFRelaxation`. Such a variable bound does not necessarily satisfy the requirements to be used in `transformSNFRelaxation` though, and could be blocking a candidate. It might be well worth it to either change the function that gets `bestVlb / bestVub` or to create a second search.    

Edit: And more checks for correctness.... Although they do pass locally!


Opt-Mucca:
Closing this because it's close to finished, and I don't want to spam everyone with notifications for minor performance changes as I experiment. I plan to open a non-draft PR in the next few days.

Issue #2501, opened 2025-08-16
Presolving the analytic centre LP


jajhall:
When @lgottwald wrote the MIP solver, the analytic centre LP was never presolved. For the Tulipa instances, the cost of the the analytic centre calculation without presolve is huge, particularly with IPX. Presolve reduces the problem size by factor of about 2.5 and with (also) using HiPO the instances are now solved vastly faster. 

In general, though, does presolve impair the quality of the interior solution that's obtained? Clearly postsolve starts from the analytic centre of the presolved problem - even if that's empty, in the case of `egout` (see #2500), but do the postsolve operations tend to introduce variables and constraints on/near their bounds?

jajhall:
Closed by #2507

PR #2507, opened 2025-08-20
Understand why presolve cannot be used for AC calculation

jajhall:
So, if the LP for the AC calculation is presolved, the solution obtained may well have integer variables at bounds. 

This is not just a "bad" interior point, it can lead to incorrect deductions about the feasible values of integer variables, and incorrect optimal solutions of the MIP being claimed. This was behind the errors observed with `egout`.

All that's come out of this is that when no basis exists, a basis is only constructed from an existing solution when it might be useful - ie when the problem is not a MIP, or a MIP relaxation is being solved.


jajhall:
This closes #2501

PR #2509, opened 2025-08-20
Refactor HPresolve::dominatedColumns

fwesselm:
- `HPresolve::dominatedColumns` contains lots of very similar / duplicate code.
- I added some lambdas to clean this up. This should improve code coverage.
- I ran 854 MIPs and HiGHS' behavior is not affected by these changes.


highs/presolve/HPresolve.cpp
```diff
@@ -1227,209 +1226,168 @@ HPresolve::Result HPresolve::dominatedColumns(
...
-            HighsInt k = nonz.index();
-            if (k == j || colDeleted[k]) continue;
-
-            double ak = nonz.value() * bestRowMinusScale;
-
-            if (-ajBestRowMinus <= -ak + options->small_matrix_value &&
-                (!isEqOrRangedRow ||
-                 -ajBestRowMinus >= -ak - options->small_matrix_value) &&
-                checkDomination(-1, j, -1, k)) {
-              // case (iii)  lb(x_j) = -inf, -x_j > -x_k: set x_k = ub(x_k)
```
Opt-Mucca, line 1262: Can some general comment that encapsulates the four cases be written?

highs/presolve/HPresolve.cpp
```diff
@@ -1227,209 +1226,168 @@ HPresolve::Result HPresolve::dominatedColumns(
...
-                                 HighsCliqueTable::CliqueVar(k, 0))) &&
-            checkDomination(-1, j, -1, k)) {
-          // case (iii)  lb(x_j) = -inf, -x_j > -x_k: set x_k = ub(x_k)
+        if (k == col || colDeleted[k]) continue;
+
+        double ak = nonz.value() * scale;
+
+        if (colCanBeFixed(col, k, bestVal, ak, direction, HighsInt{1},
+                          boundImplied, isEqOrRangedRow)) {
+          // direction =  1:
```
Opt-Mucca, line 1364: Something exactly like this but above as well! (for the binary case)

Opt-Mucca (approved):
Very nice change! 

Using `size_t` for `i` in a loop to later index stuff, e.g. `x[i]`, feels a bit wrong, but I've got no practical problem with it. Aside from the request to add an additional comment I'm happy with everything (it definitely improves the readability)

highs/presolve/HPresolve.cpp
fwesselm: @Opt-Mucca I have tried to add some general comments. I hope that these are helpful - the special code for binary variables was not documented at all in the old code.

Opt-Mucca: They're helpful (at least for me)! I'm happy for this to be merged whenever.

jajhall:
Sorry: missed the fact that these changes had been approved and tested

Issue #2478, opened 2025-07-24
Presolve times out for MILP problem - model solved in 4s when turning off presolver


severinh:
We're using HiGHS heavily in our applicaiton and it works well for us. Huge thanks to the authors! We've now encountered a model where the presolver times out after 20 minutes. If we turn off the presolve, the solver solves the problem in 4s.

## Steps to reproduce

1. Download the [model.lp.zip](https://github.com/user-attachments/files/21406406/model.lp.zip)
2. Run the model with default configuration: `highs model.lp`

Observed result: Presolve runs for about 20 minutes and eventually aborts with `Presolve: Time limit reached`. Full log (turning on verbose logging did not tell us anything new):
```
% highs model.lp --presolve=on 
Running HiGHS 1.11.0 (git hash: n/a): Copyright (c) 2025 HiGHS under MIT licence terms
Set option presolve to "on"
MIP  model has 132010 rows; 89748 cols; 320855 nonzeros; 42520 integer variables (42520 binary)
Coefficient ranges:
  Matrix [1e-01, 1e+00]
  Cost   [6e+02, 1e+08]
  Bound  [1e+00, 1e+00]
  RHS    [1e+00, 1e+00]
Presolving model
132010 rows, 89748 cols, 320855 nonzeros  0s
112115 rows, 69853 cols, 324946 nonzeros  0s
Presolve: Time limit reached
```

3. Now run the model with presolve turned off: `highs model.lp --presolve=off`.

Observed result: HiGHS successfully solves the model in 4s. Full log:

```
% highs model.lp --presolve=off         
Running HiGHS 1.11.0 (git hash: n/a): Copyright (c) 2025 HiGHS under MIT licence terms
Set option presolve to "off"
MIP  model has 132010 rows; 89748 cols; 320855 nonzeros; 42520 integer variables (42520 binary)
Coefficient ranges:
  Matrix [1e-01, 1e+00]
  Cost   [6e+02, 1e+08]
  Bound  [1e+00, 1e+00]
  RHS    [1e+00, 1e+00]

Presolve is switched off

Solving MIP model with:
   132010 rows
   89748 cols (42520 binary, 0 integer, 0 implied int., 47228 continuous, 0 domain fixed)
   320855 nonzeros

Src: B => Branching; C => Central rounding; F => Feasibility pump; J => Feasibility jump;
     H => Heuristic; L => Sub-MIP; P => Empty MIP; R => Randomized rounding; Z => ZI Round;
[...]

PR #2512, opened 2025-08-21
Decrease probing budget after some limit

Opt-Mucca:
This PR adds stricter limits on probing after 2500 columns have been probed on. Specifically, it decreases the amount of deleted columns used as a stopping criteria, and changes the weights used to calculate the dynamic probing budget. 
This fixes worst-case performance for #2478 where HiGHS was probing on everything for a huge amount of cliques / some deletions, despite the instance being easy to solve regardless. 
I've run some experiments on 60 or so instances, with most being unaffected, and some minor improvements that I'd say are noise. This change is only likely to affect large instances. The PR needs to go through more rigorous computational experiments before being merged.
I decided on the numbers from playing around a bit on some instances. They're unlikely to be perfect (as the original ones aren't), so am open to any changes / new approaches. I tried to be conservative because I don't want to ruin probings effectiveness on other instances just to avoid these fringe cases.

### DIFF
--- a/highs/presolve/HPresolve.cpp
+++ b/highs/presolve/HPresolve.cpp
@@ -1557,7 +1557,7 @@ HPresolve::Result HPresolve::runProbing(HighsPostsolveStack& postsolve_stack) {
             std::max(HighsInt{1000}, (model->num_row_ + model->num_col_) / 20);
       } else {
         probingEarlyAbort =
-            numDel > std::min(HighsInt{500}, model->num_col_ / 20);
+            numDel > std::min(HighsInt{1000}, (model->num_row_ + model->num_col_) / 40);
       }
       if (probingEarlyAbort) break;
 
--- a/highs/presolve/HPresolve.cpp
+++ b/highs/presolve/HPresolve.cpp
@@ -1556,7 +1556,8 @@ HPresolve::Result HPresolve::runProbing(HighsPostsolveStack& postsolve_stack) {
             std::max(HighsInt{1000}, (model->num_row_ + model->num_col_) / 20);
       } else {
         probingEarlyAbort =
-            numDel > std::min(HighsInt{1000}, (model->num_row_ + model->num_col_) / 20);
+            numDel >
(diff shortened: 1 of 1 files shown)


galabovaa:
Great!

galabovaa (approved):
LGTM

highs/presolve/HPresolve.cpp
```diff
@@ -1545,12 +1545,20 @@ HPresolve::Result HPresolve::runProbing(HighsPostsolveStack& postsolve_stack) {
...
 
+      bool tightenLimits = (numProbed - oldNumProbed) >= 2500;
+
       // when a large percentage of columns have been deleted, stop this round
       // of probing
       // if (numDel > std::max(model->num_col_ * 0.2, 1000.)) break;
-      probingEarlyAbort =
-          numDel >
-          std::max(HighsInt{1000}, (model->num_row_ + model->num_col_) / 20);
+      if (!tightenLimits) {
```
fwesselm, line 1553: This is a question of taste, but if you would want to avoid the if-else, you could have the following here:

 ```
      bool tightenLimits = (numProbed - oldNumProbed) >= 2500;
      HighsInt multiplier = tightenLimits ? -1 : 1;
      // when a large percentage of columns have been deleted, stop this round
      // of probing
      // if (numDel > std::max(model->num_col_ * 0.2, 1000.)) break;
      probingEarlyAbort =
          numDel >
          multiplier *
              std::max(multiplier * HighsInt{1000},
                       multiplier * (model->num_row_ + model->num_col_) / 20);
      if (probingEarlyAbort) break;
```

fwesselm (approved):
Looks good, thanks @Opt-Mucca. I am going to test this now.

highs/presolve/HPresolve.cpp
Opt-Mucca: Sorry for missing this comment! I'm happy with `if-else` logic. I find it easier to read than the double `multiplier`.

Issue #2508, opened 2025-08-20
Long time in reduced-cost fixing / Presolve failing to find good variable bounds


Opt-Mucca:
This issue was raised by @mathgeekcoder (I'd written it down in some notes and meant to get to it, but wanted to put it here so I don't forget about it as I haven't found the time to get to it). The instance [extensive_m15_s50.mps.txt](https://github.com/user-attachments/files/21899220/extensive_m15_s50.mps.txt) gets stuck in reduced cost fixing, where it tries the maximum (1024) different values for a huge amount of variables, which is just too much for such a small upside. This happens when there is a lot of integer variables with (`ub - lb >= 1024`). There would be two things to look into:
- Fix this worst-case scenario. I'd suggest something like decreasing the maximum value (+1024) every-time it's hit until it reaches something smaller like 16 / 32. Or scaling the amount at the beginning by the amount of integer variables with large domains.
- Improve `HPresolve` for this instance. Currently it doesn't find any upper bound for the integer variables. Their actually bound is supposed to be 1, so we should be able to at least infer something finite.

Opt-Mucca:
@fwesselm I've been working on fixing the missing presolve feature for this instance, i.e., why is HiGHS not finding the bound `[0, 1]` for `z` when `z` only features in the constraint `-z - x + y <= 0`, given that `x` and `y` are both binary and that `z` has a positive objective term (doesn't want to go towards infinity), we should be able to derive that `z \in [0, 1]` from the originally provided `z \in [0, inf]`.

The provided instance gives no upper bound for the `z` variables, and all such variables feature in exactly one single constraint of the form above. In another pull request I've fixed the instance from one direction, i.e., don't spend forever computing reduced cost fixings for variables with large domains, but it should be possible (for this instance) to also improve the presolver. I've added the following lines to `singletonCol` (In this case there should be exactly one row that the column has a non-zero in)

```
  // tighten column domain if possible
  if (impliedRowBounds.getResidualSumUpper(row, col, colCoef) <
          model->col_upper_[col] - primal_feastol &&
      model->col_cost_[col] >= 0) {
    changeColUpper(col,
                   impliedRowBounds.getResidualSumUpper(row, col, colCoef));
  }
  if (impliedRowBounds.getResidualSumLower(row, col, colCoef) >
          model->col_lower_[col] + primal_feastol &&
      model->col_cost_[col] <= 0) {
    changeColLower(col,
[...]

PR #2513, opened 2025-08-21
Add a dynamic redcost budget

Opt-Mucca:
Currently for each integral column with `ub - lb >= 1024`, HiGHS adds 1024 lurking bounds using reduced cost logic. In general this isn't an issue because most models don't have many integer columns with large domains. Additionally, if they do, presolve usually finds that most domains can be substantially tightened, and even if it doesn't, then it's unlikely for many of the integer columns to feature in the objective function. 
What happened in #2508 is that a huge amount of continuous columns got upgraded to implied integer, with most of them being unbounded and featuring in the objective. This led to a ridiculous amount of time being spent in adding lurking bounds derived from reduced cost, which is not that important for the solve process. 
The change: I've added a quick check to scale the max amount of values to query (instead of 1024 it's now some power of 2 in the range `[32, 1024]`). This is decided based on the amount of columns with non-zero reduced cost and domains wider than `ub - lb >= 512`.
The only downside: This adds a linear time check that is redundant for nearly all instances. I don't see it being a problem.

I've tested this on 60 or so instances, and as expected, there's no change in any of them. This should still be tested on a wider set of instances, just to see if the safety check does actually have an overhead, and if the change affect the solve path of any instance.

### DIFF
--- a/highs/mip/HighsRedcostFixing.cpp
+++ b/highs/mip/HighsRedcostFixing.cpp
@@ -200,6 +200,27 @@ void HighsRedcostFixing::addRootRedcost(const HighsMipSolver& mipsolver,
   mipsolver.mipdata_->lp.computeBasicDegenerateDuals(
       mipsolver.mipdata_->feastol);
 
+  HighsInt numRedcostLargeDomainCols = 0;
+  for (HighsInt col : mipsolver.mipdata_->integral_cols) {
+    if (mipsolver.mipdata_->domain.col_upper_[col] -
+            mipsolver.mipdata_->domain.col_lower_[col] >=
+        512) {
+      if (lpredcost[col] > mipsolver.mipdata_->feastol) {
+        numRedcostLargeDomainCols++;
+      } else if (lpredcost[col] < -mipsolver.mipdata_->feastol) {
+        numRedcostLargeDomainCols++;
+      }
+    }
+  }
+  HighsInt maxNumStepsExp = 10;
+  HighsInt expshift = 0;
+  std::frexp(numRedcostLargeDomainCols / 10, &expshift);
+  if (expshift > 5) {
+    expshift = std::min(expshift, maxNumStepsExp);
+    maxNumStepsExp = maxNumStepsExp - expshift + 5;
+  }
+  auto maxNumSteps = static_cast<HighsInt>(1ULL << maxNumStepsExp);
+
   for (HighsInt col : mipsolver.mipdata_->integral_cols) {
     if (lpredcost[col] > mipsolver.mipdata_->feastol) {
       // col <= (cutoffbound - lpobj)/redcost + lb
@@ -218,13 +239,14 @@ void HighsRedcostFixing::addRootRedcost(const HighsMipSolver& mipsolver,
 
       HighsInt maxub;
       if (mipsolver.mipdata_->domain.col_upper_[col] == kHighsInf)
-        maxub = lb + 1024;
+        maxub = lb + maxNumSteps;
       else
         maxub = (HighsInt)std::floor(
             mipsolver.mipdata_->domain.col_upper_[col] - 0.5);
 
       HighsInt step = 1;
-      if (maxub - lb > 1024) step = (maxub - lb + 1023) >> 10;
+      if (maxub - lb > maxNumSteps)
+        step = (maxub - lb + maxNumSteps - 1) >> maxNumStepsExp;
 
       for (HighsInt lurkub = lb; lurkub <= maxub; lurkub += step) {
         double fracbound = (lurkub - lb + 1) - 10 * mipsolver.mipdata_->feastol;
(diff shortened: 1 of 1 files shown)


jajhall (commented):
Please fix the code that fails the CI tests

Opt-Mucca:
@galabovaa I'm not sure how to handle code coverage here. This change should only handle the fringe case where there's many unbounded integer variables, and any such case is not going to be a quick solve that can be put into the tests.

highs/mip/HighsRedcostFixing.cpp
```diff
@@ -200,6 +200,31 @@ void HighsRedcostFixing::addRootRedcost(const HighsMipSolver& mipsolver,
...
+  // Compute maximum number of steps per column with large domain
+  // max_steps = 2 ** k, k = max(5, min(10 ,round(log(|D| / 10)))),
+  // D = {col : integral_cols | (ub - lb) >= 512}
+  // This is to avoid doing 2**10 steps when there's many unbounded columns
+  HighsInt numRedcostLargeDomainCols = 0;
+  for (HighsInt col : mipsolver.mipdata_->integral_cols) {
+    if (mipsolver.mipdata_->domain.col_upper_[col] -
+            mipsolver.mipdata_->domain.col_lower_[col] >=
+        512) {
+      if (lpredcost[col] > mipsolver.mipdata_->feastol) {
```
fwesselm: One could use the absolute here:

if (std::abs(lpredcost[col]) > mipsolver.mipdata_->feastol)

highs/mip/HighsRedcostFixing.cpp
```diff
@@ -200,6 +200,31 @@ void HighsRedcostFixing::addRootRedcost(const HighsMipSolver& mipsolver,
...
+    }
+  }
+  HighsInt maxNumStepsExp = 10;
+  int expshift = 0;
+  std::frexp(numRedcostLargeDomainCols / 10, &expshift);
+  if (expshift > 5) {
+    expshift = std::min(expshift, static_cast<int>(maxNumStepsExp));
+    maxNumStepsExp = maxNumStepsExp - expshift + 5;
+  }
+  auto maxNumSteps = static_cast<HighsInt>(1ULL << maxNumStepsExp);
```
fwesselm: Maybe use HighsInt here instead of auto?

highs/mip/HighsRedcostFixing.cpp
```diff
@@ -218,13 +243,14 @@ void HighsRedcostFixing::addRootRedcost(const HighsMipSolver& mipsolver,
...
-        maxub = lb + 1024;
+        maxub = lb + maxNumSteps;
       else
         maxub = (HighsInt)std::floor(
             mipsolver.mipdata_->domain.col_upper_[col] - 0.5);
 
       HighsInt step = 1;
-      if (maxub - lb > 1024) step = (maxub - lb + 1023) >> 10;
+      if (maxub - lb > maxNumSteps)
+        step = (maxub - lb + maxNumSteps - 1) >> maxNumStepsExp;
```
fwesselm, line 250: Is it worth having a lambda to compute step:

auto computeStep = [&](HighsInt lb, HighsInt ub){
      if (ub - lb > maxNumSteps)
        return (ub - lb + maxNumSteps - 1) >> maxNumStepsExp;
      return HighsInt{1};
};

fwesselm (approved):
Looks good, @Opt-Mucca . I am going to qualify this shortly.

highs/mip/HighsRedcostFixing.cpp
Opt-Mucca: Good suggestion! I'll add it.

highs/mip/HighsRedcostFixing.cpp
Opt-Mucca: I had this originally, but my IDE insists that you should use `auto` if everything is already wrapped in a `static_cast`. Will change it back, and look into changing my settings.

highs/mip/HighsRedcostFixing.cpp
jajhall: Possibly, but this PR is taking up rather a lot of time in the context of much higher priorities

Opt-Mucca: I don't think so. Is only used twice. 
At some point I cobbled together a lambda for the whole thing (as there's plenty of duplicate code for the positive vs negative reduced cost), but I wasn't convinced that my attempt was easier to read (even if it was a few lines shorter).

fwesselm: @Opt-Mucca , there was mention of code coverage in one comment, so I thought having this little piece of logic in a lambda would help to improve the coverage. But, it's up to you, of course.

fwesselm:
The test results look reasonable - only a few instances are affected - and "average" performance (in terms of ratios of shifted geometric means) looks good.

PR #2518, opened 2025-09-02
Flow cover cuts

Opt-Mucca:
This PR adds flow cover cuts to HiGHS. All logic related to the cut can be followed from the `HighsCutGeneration::tryGenerateFLowCoverCut` function. Currently the cuts are only generated based on rows in the LP, i.e., "aggregations" of size 1. 
The performance results are extremely finicky, but seem to be promising, and are especially so for some network based energy problems I've been testing on. This shouldn't be merged before any additional computational experiments are done.
@galabovaa I changed the two tests because (1) the instance now solved at the root node and therefore has the optimal solution but has status `interrupted` (2) the primal heuristics now jumped quicker to the optimal solution, so I had to up the objective limit to still catch the event.


highs/mip/HighsCutGeneration.cpp
```diff
@@ -1299,6 +1584,259 @@ bool HighsCutGeneration::generateConflict(HighsDomain& localdomain,
...
+    return (col < slackOffset ? domain.col_upper_[col]
+                              : lpRelaxation.slackUpper(col - slackOffset));
+  };
+
+  for (HighsInt i = 0; i < rowlen; ++i) {
+    HighsInt col = inds[i];
+    double lb = getLb(col);
+    double ub = getUb(col);
+    maxAbsVal = std::max(std::abs(vals[i]), maxAbsVal);
+    if (lb == -kHighsInf || ub == kHighsInf) {
```
fwesselm, line 1647: Both bounds have to be finite?

highs/mip/HighsCutGeneration.cpp
```diff
@@ -500,6 +500,255 @@ bool HighsCutGeneration::separateLiftedMixedIntegerCover() {
...
+  ld.m.resize(snfr.numNnzs);
+  HighsCDouble sumN2mC2LE = 0.0;
+  HighsCDouble sumC1LE = 0.0;
+  HighsCDouble sumC2 = 0.0;
+
+  for (HighsInt i = 0; i != snfr.numNnzs; ++i) {
+    // col is in N- \ C-
+    if (snfr.flowCoverStatus[i] == -1 && snfr.coef[i] == -1) {
+      assert(snfr.vubCoef[i] >= 0);
+      if (snfr.vubCoef[i] > snfr.lambda + 1e-8) {
```
fwesselm: Maybe add a const variable to store the hard-coded 1e-8?

highs/mip/HighsCutGeneration.cpp
```diff
@@ -500,6 +500,255 @@ bool HighsCutGeneration::separateLiftedMixedIntegerCover() {
...
+        ld.m[ld.r] = snfr.vubCoef[i];
```
fwesselm, line 532: Would it be worth adding a mini lambda to add elements to ld.m?

highs/mip/HighsCutGeneration.cpp
```diff
@@ -500,6 +500,255 @@ bool HighsCutGeneration::separateLiftedMixedIntegerCover() {
...
+
+    HighsInt i = 0;
+    while (i < ld.r && vubcoefpluslambda >= ld.M[i + 1] + 1e-8) {
+      ++i;
+    }
+
+    if (vubcoef <= ld.M[i] - 1e-8) {
+      assert(ld.M[i] < vubcoefpluslambda);
+      alpha = 1;
+      beta = -i * HighsCDouble(snfr.lambda) + ld.M[i];
```
fwesselm, line 596: It would be great to use `static_cast<HighsCDouble>(...)` instead of C-style casting.

highs/mip/HighsCutGeneration.cpp
```diff
@@ -500,6 +500,255 @@ bool HighsCutGeneration::separateLiftedMixedIntegerCover() {
...
+      if (ld.M[i] < vubcoef + 1e-8) {
+        return static_cast<double>(liftedcoef);
+      }
+      assert(i > 0 && ld.M[i] < vubcoef + snfr.lambda - 1e-8 &&
+             vubcoef <= ld.M[i]);
+      liftedcoef += vubcoef;
+      liftedcoef -= ld.M[i];
+      return static_cast<double>(liftedcoef);
+    }
+    if (i < ld.r) {
```
fwesselm: Is this actually an `else if`?

highs/mip/HighsCutGeneration.cpp
```diff
@@ -500,6 +500,255 @@ bool HighsCutGeneration::separateLiftedMixedIntegerCover() {
...
+  // L-- = N- \ (L- union C-)
+  HighsCDouble tmpRhs = ld.d1;
+  rowlen = 0;
+  for (HighsInt i = 0; i != snfr.numNnzs; ++i) {
+    // col is in N- \ C-
+    if (snfr.flowCoverStatus[i] == -1 && snfr.coef[i] == -1) {
+      if (snfr.vubCoef[i] > snfr.lambda + 1e-8) {
+        if (snfr.origBinCols[i] != -1) {
+          // col is in L-
+          vals[rowlen] = -snfr.lambda;
```
fwesselm: Suggestion: Add another mini-lambda to add to `vals, inds`, update `rowlen` and `tmpRhs`. Just wondering if this may shorten this loop...

fwesselm (commented):
@Opt-Mucca , thank you for adding the flow cover cuts.

One general question (- I have not thought this through -): Should new cut separators like flow cover cuts live in the `HighsCutGeneration` class or have its own (derived) class?

Opt-Mucca:
> One general question (- I have not thought this through -): Should new cut separators like flow cover cuts live in the HighsCutGeneration class or have its own (derived) class?

@fwesselm For the current design they shouldn't. Anything that is going to call `CutGeneration::generateCut()` should have its own file, e.g., if we create a new separator that identifies some network structure and aggregates certain LP rows we'd create something like `HighsNetworkSeparator.cpp`. The exception to this rule is something that is simply looping over some global data structure, e.g. `separateImpliedBounds` and `separateCliques`. 
Currently all the cut generators, e.g., `separateLiftedKnapsackCover`, `cmirCutGenerationHeuristic`, and now `separateLiftedFlowCover`, use the same data structures in `HighsCutGeneration` so are just functions in the single file. We could make some derived classes...... (After writing out my argument against I don't think the logic holds, so I'm on the fence). I'd still not push for putting them all into separate files, but it definitely makes more sense each time we add a new cut.

highs/mip/HighsCutGeneration.cpp
Opt-Mucca: Yup! Let's imagine we have some continuous variable `y \in [l ,u]` and some binary `x` with the VUB `y <= u'x + d` where the coefficient of `y` in the row is `a`. We're always going to try and create a new variable with the substitution `y' = a(y - d)` (sometimes also involve the binary term in the row, but we'll consider the easier case)
(1) `l = -inf`
We can do the substitution, but the lb of `y'` is `-inf`, and therefore violates the assumption `y' >= 0` in the flow cover. We can't just create a shift for it either as we'd be introducing an infinite RHS.
(2) `u = inf`
A finite VUB cannot exist. The only potential VUB we can have is `y <= inf x + d` (so some finite bound for `y` if `x=0`). I don't think HiGHS stores such VUBs though, and it'd later introduce an `inf` into the cut later anyway.

highs/mip/HighsCutGeneration.cpp
Opt-Mucca: I've added `const double vubEpsilon = 1e-8;`

highs/mip/HighsCutGeneration.cpp
Opt-Mucca: As in turn the entire loop into a lambda?

highs/mip/HighsCutGeneration.cpp
Opt-Mucca: Is this C-style casting? My IDE at least doesn't complain like it normally does for other conversions. It's rather a call to the constructor:
`HighsCDouble(double val) : hi(val), lo(0.0) {}` in `HighsCDouble.h`.

I'm happy to change it if this is still counted as bad practice.

highs/mip/HighsCutGeneration.cpp
Opt-Mucca: Yes, but as all the cases always return something there's no functional difference.
I've added the `else` to help with readability.

highs/mip/HighsCutGeneration.cpp
Opt-Mucca: It would definitely shorten the loop. I should have added this earlier.... Good suggestion!

highs/mip/HighsCutGeneration.cpp
jajhall: The C style of cast - even (type)foo, rather than type(foo) - is everywhere in HiGHS and has never caused an issue.

Hence the static_cast (which is excessively verbose) is not necessary. We have greater priorities

jajhall (approved):
I can't comment on the technical details as they're way beyond me. However, it's the sort of stuff you understand well, and @fwesselm  has been with you in the process. 

My only observation is that

>   struct SNFRelaxation {
    HighsInt numNnzs;                    // |N-| + |N+|
    std::vector<HighsInt> coef;          // (+-1) coefficient of col in SNFR
    std::vector<double> vubCoef;         // u_j in y'_j <= u_j x_j in SNFR
    std::vector<double> binSolval;       // lp[x_j], y'_j <= u_j x_j in SNFR
    std::vector<HighsInt> origBinCols;   // orig x_i, y'_j <= u_j x_j in SNFR
    std::vector<HighsInt> origContCols;  // orig y_i used to make y'_j in SNFR
    std::vector<double> aggrBinCoef;     // coef of x_i in y'_j aggregation
    std::vector<double> aggrContCoef;    // coef of y_i in y'_j aggrregation
    std::vector<double> aggrConstant;    // constant shift in y'_j aggregation
    std::vector<bool> complementation; 

with the HiGHS naming convention would be

>   struct SnfRelaxation {
    HighsInt num_n_nz;                    // |N-| + |N+|
    std::vector<HighsInt> coef;          // (+-1) coefficient of col in SNFR
    std::vector<double> vub_coef;         // u_j in y'_j <= u_j x_j in SNFR
    std::vector<double> bin_sol_val;       // lp[x_j], y'_j <= u_j x_j in SNFR
    std::vector<HighsInt> orig_bin_cols;   // orig x_i, y'_j <= u_j x_j in SNFR
    std::vector<HighsInt> orig_cont_cols;  // orig y_i used to make y'_j in SNFR
    std::vector<double> aggr_bin_coef;     // coef of x_i in y'_j aggregation
    std::vector<double> aggr_cont_coef;    // coef of y_i in y'_j aggrregation
    std::vector<double> aggr_constant;    // constant shift in y'_j aggregation
    std::vector<bool> complementation; 

Don't fee you need to change it, though. I also acknowledge that Leona wasn't great at following the naming convention....

Opt-Mucca:
@jajhall Currently the big hurdle is getting this to be performance improving on @fwesselm cluster. (Last time it was measured it was performance neutral).
There's also the issue of some instances from those tests having slightly off results, but I either can't recreate them (something to do with the testing environment) or they've disappeared with more recent changes, and in some cases it's a deviation from a previously incorrect solution path..... Still looking into all of them.
For the naming convention: I've been trying to mostly use underscores when I write new code, but the entire cut generation code uses camel case, and I'd rather fix that all at once instead of mix up the standard.

jajhall:
Hmm... Can you easily switch off the flow cover constraints by default? That way the code is merged, possibly avoiding conflicts that might occur if merged later. 

Probably a useless comment, but I have a memory of Philip Christophel saying that flow cover constraints weren't so useful until he revisited them...

Opt-Mucca:
I can add a parameter to turn them off before they get merged. In the long term I'd like to add some high level cut parameters anyways (something like an emphasis setting) because it's clear some users want to test settings for their application. Currently I'm just trying to make it so the behaviour of when they're turned off is identical to before.
For the second point: I think you're misremembering that with path mixing inequalities (he always presents them together). They're the cuts he thought weren't that helpful, and it took more instances being available and Leona trying them out to actually see that the cuts were good.

jajhall:
Is this PR still active @Opt-Mucca ?

Opt-Mucca:
@jajhall It's still active, but I've been failing to make them anything but performance neutral on-and-off for months now. @fwesselm has been benchmarking them whenever I try something new and I think the change is significant enough. 
I'm happy to close the PR if you don't want clutter. I'm hesitant to merge them with the parameter off by default because I don't want to introduce dead code. I also firmly believe this PR is a couple of lines away from improving general MILP performance by a couple of percent and being properly helpful for certain problem classes. I just can't seem to tune them correctly...... (or there's some logic error and I'm weakening them by accident)

jajhall:
> @jajhall It's still active, but I've been failing to make them anything but performance neutral on-and-off for months now. @fwesselm has been benchmarking them whenever I try something new and I think the change is significant enough. I'm happy to close the PR if you don't want clutter. I'm hesitant to merge them with the parameter off by default because I don't want to introduce dead code. I also firmly believe this PR is a couple of lines away from improving general MILP performance by a couple of percent and being properly helpful for certain problem classes. I just can't seem to tune them correctly...... (or there's some logic error and I'm weakening them by accident)

I'm happy to keep it open, but it would be good to merge latest into it so that conflicts don't build up and have CI tests pass

Opt-Mucca:
I'm closing this PR. I don't see this being merged anytime soon. For some reason (likely bugs or all the lifted cover cuts), HiGHS is not getting the expected performance improvement that other solvers experience. The branch should still very much stay alive, and this should be looked at again after some months.

jajhall:
Shame, but worth a try

PR #2522, opened 2025-09-04
Chvátal-Gomory strengthening

fwesselm:
- I am proposing to add Chvátal-Gomory strengthening for inequality constraints to `HPresolve::rowPresolve` (as described by Achterberg et al., Presolve Reductions in Mixed Integer Programming). 
- I tested on more than 800 MIPs and, unfortunately, the impact on performance is small. Only 97 instances are affected in terms of numbers of branching nodes, and the ratios between the shifted geometric means of running times and numbers of branching nodes are very close to 1.0.
- These results are in line with what Achterberg et al. report. But, the reduction still is turned on by default in Gurobi. Therefore, it may be good to have it in HiGHS as well.

### DIFF
--- a/highs/presolve/HPresolve.cpp
+++ b/highs/presolve/HPresolve.cpp
@@ -3998,7 +3998,7 @@ HPresolve::Result HPresolve::rowPresolve(HighsPostsolveStack& postsolve_stack,
             }
           };
 
-          // try different scalars and return an improving one
+          // try different scalars and return if an improving one was found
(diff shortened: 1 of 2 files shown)


fwesselm:
> Changes looks good! The implementation is directly out of the paper and easy to follow. I've got zero intuition why those scalar values would be the best to check and if that's too many scalars to check, but I guess it "works"?

Same here, I did not try using other scalar values.

PR #2531, opened 2025-09-10
Dual fixing

fwesselm:
- I am proposing to add dual fixing (fixing, substitutions, bound strengthening) to `HPresolve::colPresolve` and `HPresolve::singletonCol` (as described by Achterberg et al., Presolve Reductions in Mixed Integer Programming, see section 4.4).
- This fixes the performance bug #2508, i.e. HiGHS not being able to identify tight bounds for some variables. 
- Testing shows that dual fixing increases the "average" running time by around 2 percent while also increasing the number of nodes by around 2 percent. However, 3 additional models can be solved within the time limit.

### DIFF
--- a/highs/presolve/HPresolve.cpp
+++ b/highs/presolve/HPresolve.cpp
@@ -4558,6 +4558,13 @@ HPresolve::Result HPresolve::dualFixing(HighsPostsolveStack& postsolve_stack,
   std::vector<nonZeros> nzs;
   nzs.reserve(colsize[col]);
 
+  // lambda for checking whether a row provides an implied lower bound
+  // (direction = 1) or implied upper bound (direction = -1)
+  auto hasImpliedBound = [&](HighsInt row, HighsInt direction, double val) {
+    return ((direction * val < 0 && model->row_upper_[row] != kHighsInf) ||
+            (direction * val > 0 && model->row_lower_[row] != -kHighsInf));
+  };
+
   // lambda for computing locks
   auto computeLocks = [&](HighsInt col, HighsInt& numDownLocks,
                           HighsInt& numUpLocks, HighsInt& downLockRow,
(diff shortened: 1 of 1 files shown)


highs/presolve/HPresolve.cpp
```diff
@@ -4523,6 +4538,100 @@ HPresolve::Result HPresolve::detectDominatedCol(
...
+
+        // compute residual
+        residual = impliedRowBounds.getResidualSumLowerOrig(row, col, val);
+        if (residual == -kHighsInf) return false;
+      }
+
+      // compute bound
+      double candidateBound =
+          direction *
+          static_cast<double>((static_cast<HighsCDouble>(rhs) -
```
Opt-Mucca, line 4753: Was this the bit I was getting wrong in my code? I thought the residual calculation took into account `val`?

highs/presolve/HPresolve.cpp
```diff
@@ -4523,6 +4538,100 @@ HPresolve::Result HPresolve::detectDominatedCol(
...
+void HPresolve::dualBoundTightening(HighsPostsolveStack& postsolve_stack,
+                                    HighsInt col) {
+  // tighten bounds using dual arguments
+  // see section 4.4 "Dual fixing, substitution and bound strengthening",
+  // Achterberg et al., Presolve Reductions in Mixed Integer Programming,
+  // INFORMS Journal on Computing 32(2):473-506.
+  assert(!colDeleted[col]);
+
+  // only tighten bounds for integer-constrained variables
+  if (model->integrality_[col] == HighsVarType::kContinuous) return;
```
Opt-Mucca: I don't understand why this has to be restricted to integer columns. (I'm probably missing something instead of it being wrong)

highs/presolve/HPresolve.cpp
```diff
@@ -4523,6 +4538,100 @@ HPresolve::Result HPresolve::detectDominatedCol(
...
+
+    // flip sign
+    newBound *= direction;
+    return true;
+  };
+
+  double newBound = 0.0;
+  if (hasTighterBound(col, HighsInt{1}, model->col_upper_[col], newBound)) {
+    // update upper bound
+    changeColUpper(col, std::max(newBound, model->col_lower_[col]));
```
Opt-Mucca: Should we be checking if this fixes the column? If so I'd then call `fixColToUpperOrUnbounded(postsolve_stack, col)`. Or will this just be handled directly after anyways?

Opt-Mucca (commented):
@fwesselm This is a great change! Thanks for generalising the argument to more than singletons (is the current presolve code completely missing these reductions?)
My gut is telling me this should have some nice performance improvements, especially over anything that is poorly modelled (unnecessarily large bounds).

fwesselm:
> @fwesselm This is a great change! Thanks for generalising the argument to more than singletons (is the current presolve code completely missing these reductions?) My gut is telling me this should have some nice performance improvements, especially over anything that is poorly modelled (unnecessarily large bounds).

Thanks, @Opt-Mucca . Correct, the code was missing this reduction. The substitutions described by Achterberg et al. would be another thing that is missing, I think. I will put it on my TODO list.

highs/presolve/HPresolve.cpp
fwesselm: Presolve does not modify bounds of continuous variables in order to produce correct dual information (for LPs).

highs/presolve/HPresolve.cpp
fwesselm: I thought about that, but the column will be marked due to the bound change and should be fixed subsequently. I did not want to put duplicate code.

highs/presolve/HPresolve.cpp
fwesselm: Consider this equation from the paper:

Bildschirmfoto 2025-09-10 um 11 03 50

Here `impliedRowBounds.getResidualSumUpperOrig(row, col, val)` computes the `sup{...}`. To get a valid upper bound we need to divide by the coefficient.

I hope this helps.

Opt-Mucca: That was definitely what I was doing wrong then..... That and not skipping the continuous columns. Thanks for the explanation!

Opt-Mucca:
Super happy with this change! We should do a performance check before merging though (incredibly interested to see the difference)

fwesselm:
I updated the pull request with all the techniques described in section 4.4. (dual fixing, substitution and bound strengthening) of Achterberg et al.'s presolve paper. The results look reasonable (similar rations of shifted geometric means and 3 more problems solved within the time limit).

highs/presolve/HPresolve.cpp
```diff
@@ -4523,6 +4538,315 @@ HPresolve::Result HPresolve::detectDominatedCol(
...
+        // substitute variable
+        double offset = otherColBound;
+        double scale = colBound - otherColBound;
+        postsolve_stack.doubletonEquation(
+            -1, col, rowNz.index(), 1.0, -scale, offset, model->col_lower_[col],
+            model->col_upper_[col], 0.0, false, false,
+            HighsPostsolveStack::RowType::kEq, HighsEmptySlice());
+        markColDeleted(col);
+        substitute(col, rowNz.index(), offset, scale);
+        HPRESOLVE_CHECKED_CALL(checkLimits(postsolve_stack));
```
Opt-Mucca, line 4700: Amazing job at putting that all into one short function! That took me a fair while to read, but it's proper neat.

Opt-Mucca:
Code looks good! I'd be happy to merge this whenever.

PR #2533, opened 2025-09-10
Minor feasibility jump enhancements (fix #2487)

### LINKED ISSUE #2487: Feasibility Jump running order and use of existing incumbent solution
In a recent discussion, @jajhall brought to my attention that Feasibility Jump (FJ) runs before the trivial heuristics. It may not make much difference, but it seems more sensible for it to run afterwards.

On a related note, when running FJ we should check if we already have an incumbent (e.g. from the user). If so, we should use this as a starting point for FJ.

BenChampion:
This small PR implements the enhancements in #2487.

* The trivial heuristics now run before feasibility jump
* Feasibility jump now uses incumbent solutions, with some guardrails, since incumbents need not always be feasible **(!)**


BenChampion:
> Looks good! Have you seen a change in the found solution for any runs?

No (apart from when I hardcoded the solution in just to test). This is really just housekeeping.

Anecdotally it's pretty rare for the trivial heuristics to find anything. I imagine it's even rarer for them to find something that FJ doesn't find anyway.

PR #2535, opened 2025-09-15
Ipx cupldlp logging

jajhall:
HiPO logs the values that are compared with tolerances used to determine feasibility (relative primal and dual infeasibilities) and optimality (relative duality gap).

With these modifications cuPDLP-C and IPX do the same


jajhall:
> This looks fine. However, you have defined bounds_measure_ and costs_measure_ to already include "1.0 +", so it should not be added again in the denominator of logging_presidual and logging_dresidual.

Oops, well spotted!

PR #2539, opened 2025-09-19
Added/updated HiGHS header and eliminated some trivial MIP logging

jajhall:
Added/updated HiGHS header for qpsolver, model and parallel, and removed trivial breakdown of 0 LP iterations in final MIP log

Only affects HiGHS logging


galabovaa (approved):
Looks good to me!

PR #2541, opened 2025-09-21
Minor change to dual fixing

fwesselm:
- Skip redundant rows when computing locks in `HPresolve::dualFixing`.
- Unfortunately, this change was lost during a last-minute cleanup of the dual fixing code. My test runs included this piece of code, so the change is fully qualified.


jajhall (approved):
Fine

PR #2546, opened 2025-09-25
Standardised factorization logging for IPX, and added time stamp

jajhall:
IPX now gives logging between main logging lines if factorization takes too long, but it's better than being silent for a _long_ time. Perhaps more logging could be added - such as after each system solve


filikat (commented):
Looks fine.
I wonder if HiPO should have the same when the factorisation takes very long.

jajhall:
I think it should, as we've already seen one user who assumed it had stalled.

Issue #2542, opened 2025-09-24
HiPO and GKlib


odow:
Success:

```
sandbox:${WORKSPACE}/srcdir/HiGHS/build # bin/highs --solver=hipo ../check/instances/afiro.mps 
Running HiGHS 1.11.0 (git hash: 9f7beeab5): Copyright (c) 2025 HiGHS under MIT licence terms
Set option solver to "hipo"
Set option log_file to "HiGHS.log"
LP afiro has 27 rows; 32 cols; 83 nonzeros
Coefficient ranges:
  Matrix [1e-01, 2e+00]
  Cost   [3e-01, 1e+01]
  Bound  [0e+00, 0e+00]
  RHS    [4e+01, 5e+02]
Presolving model
22 rows, 29 cols, 74 nonzeros  0s
15 rows, 22 cols, 60 nonzeros  0s
9 rows, 12 cols, 32 nonzeros  0s
7 rows, 10 cols, 28 nonzeros  0s
7 rows, 10 cols, 28 nonzeros  0s
Presolve reductions: rows 7(-20); columns 10(-22); nonzeros 28(-55) 
Solving the presolved LP
HiPO model has 7 rows, 10 columns and 28 nonzeros

Running HiPO
Threads:                        1
Rows:                           7.0e+00
Cols:                           1.6e+01
Nnz:                            3.4e+01
Range of A:                     [5.0e-01, 2.0e+00], ratio 4.0e+00
Range of b:                     [8.0e+01, 3.0e+02], ratio 3.8e+00
Range of c:                     [3.2e-01, 1.6e+00], ratio 5.0e+00
Range of bounds:                [2.0e+01, 1.2e+02], ratio 6.2e+00
Scaling coefficients:           [5.0e-01, 4.0e+00], ratio 8.0e+00
Newton system:                  NE preferred
Parallelism:                    Node preferred

Factorisation statistics
Size:                           7.00e+00
Nnz:                            2.10e+01
Fill-in:                        1.00
Serial memory:                  2.8e+02 B
Flops:                          3.8e+01

 iter       primal obj         dual obj       pinf       dinf       gap    time
    0  -4.34668376e+01  -4.52399084e+02   5.29e-01   1.15e+00  1.64e+00     0.0
    1  -9.65758864e+01  -6.32355868e+02   1.12e-01   2.32e-01  1.47e+00     0.0
[...]

PR #2550, opened 2025-09-30
Refactor the HiPO build system to support cross compilation

odow:
x-ref #2542


odow:
@amontoison: I see that you forked METIS for Julia. What's the status of upstream? Should I just grab your binaries instead?

amontoison:
@odow The original website is not working anymore but it seems that they did the new web page now:
https://karypis.github.io/glaros/software/metis/overview.html

I suggest to grap my binaries because it contains a few patches, particularly for Windows.

odow:
@filikat do you know about these segfaults on Windows?

filikat:
@odow Ivet told me about them, but I don't know why they happen and I don't have a machine with Windows. 
I had a look at the log files of the checks that failed. It looks like the debug build works, but the release build gives segfault.
The debug build with tests on fails because one of the checks in TestHipo.cpp fails. This is the test that checks if running HiPO twice with the same problem produces the exact same solution. Could there be sometihng non deterministic when running on Windows? I am not sure of exactly what BLAS library gets linked on Windows and whether it is fully deterministic.

odow:
What about the JuliaCompileAndTest failure? Why is HiPO being run in the HiGHS.jl tests? Isn't it opt-in? HiGHS.jl doesn't know about HiPO yet, so it should never be called.

odow:
@filikat here's a bug in HiPO. It doesn't detect primal unboundedness:

```
NAME issue-2550
ROWS
 N  OBJ
 L  c
COLUMNS
    x         OBJ       2         c         1
    y         OBJ       1         c         1
RHS
    RHS_V     c         1
BOUNDS
 FR BOUND     x
 FR BOUND     y
ENDATA
```

```
% ./bin/highs --solver=hipo ../check/instances/issue-2550.mps --presolve=off
Running HiGHS 1.11.0 (git hash: e5c14c6a7): Copyright (c) 2025 HiGHS under MIT licence terms
Set option presolve to "off"
Set option solver to "hipo"
LP issue-2550 has 1 row; 2 cols; 2 nonzeros
Coefficient ranges:
  Matrix [1e+00, 1e+00]
  Cost   [1e+00, 2e+00]
  Bound  [0e+00, 0e+00]
  RHS    [1e+00, 1e+00]
Not presolved: solving the LP
HiPO model has 1 rows, 2 columns and 2 nonzeros

Running HiPO
Threads:                        5
Rows:                           1.0e+00
Cols:                           3.0e+00
Nnz:                            3.0e+00
Range of A:                     [1.0e+00, 1.0e+00], ratio 1.0e+00
Range of b:                     [1.0e+00, 1.0e+00], ratio 1.0e+00
Range of c:                     [1.0e+00, 2.0e+00], ratio 2.0e+00
Range of bounds:                [0.0e+00, 0.0e+00], ratio -
Scaling coefficients:           [0.0e+00, 0.0e+00], ratio -
Newton system:                  NE preferred
Parallelism:                    Tree preferred

Factorisation statistics
Size:                           1.00e+00
Nnz:                            1.00e+00
Fill-in:                        1.00
Serial memory:                  8.0e+00 B
Flops:                          0.0e+00

 iter       primal obj         dual obj       pinf       dinf       gap    time
    0   1.00000000e+00   1.00000000e+00   8.00e-01   1.05e+00  0.00e+00     0.0
    1  -4.99950000e+11   1.29411765e+00   8.00e-05   4.31e-01  2.00e+00     0.0
    2  -2.49975000e+12  -1.38628714e+10   6.10e-05   4.62e+09  1.98e+00     0.0
    3  -3.01469026e+12  -1.38628565e+06   2.25e+06   4.62e+05  2.00e+00     0.0
    4  -3.52979558e+12  -1.37133766e+02   3.67e+02   4.64e+01  2.00e+00     0.0
    5  -4.06142313e+12   2.10037437e-02   5.51e-02   6.60e-01  2.00e+00     0.0
    6  -6.06122313e+12   9.99902100e-01   1.22e-04   3.33e-01  2.00e+00     0.0
    7  -8.06102313e+12   9.99999990e-01   2.44e-04   3.33e-01  2.00e+00     0.0
    8  -1.00608231e+13   1.00000000e+00   0.00e+00   3.33e-01  2.00e+00     0.0
    9  -1.20606231e+13   1.00000000e+00   0.00e+00   3.33e-01  2.00e+00     0.0
```

odow:
So this happens in Julia because we force `solver = ipm` in one of the tests. I think we should use HiPO only if we force `solver = hipo`.

odow:
The related issue is that we have re-interpreted `solver = ipm` to previously mean IPX and now mean `IPX | HiPO`, and there is a new `solver = "ipx"` option for IPX.

https://github.com/ERGO-Code/HiGHS/blob/364c83a51e44ba6c27def9c8fc1a49b1daf5ad5c/highs/lp_data/HighsOptions.h#L260-L262

This is potentially a breaking change for folks like GenX which force the solver to `ipm`. People with the default or `choose` should expect the algorithm to change between HIGHS versions. But not if they are explicitly choosing the algorithm.

odow:
I can reproduce the segfault
```
(base) odow@Mac build % ./bin/highs ../check/instances/p01.mps 
Running HiGHS 1.11.0 (git hash: aee998245): Copyright (c) 2025 HiGHS under MIT licence terms
Number of BV entries in BOUNDS section is 210
MIP p01 has 30 rows; 210 cols; 420 nonzeros; 210 integer variables (210 binary)
Coefficient ranges:
  Matrix [1e+00, 1e+00]
  Cost   [1e+01, 8e+01]
  Bound  [1e+00, 1e+00]
  RHS    [1e+00, 1e+00]
Presolving model
30 rows, 210 cols, 420 nonzeros  0.001479s
30 rows, 210 cols, 420 nonzeros  0.006358s
Presolve reductions: rows 30(-0); columns 210(-0); nonzeros 420(-0) - Not reduced

Solving MIP model with:
   30 rows
   210 cols (210 binary, 0 integer, 0 implied int., 0 continuous, 0 domain fixed)
   420 nonzeros

 J       0       0         0   0.00%   -inf            633                Large        0      0      0         0     0.0s
 T       0       0         0   0.00%   0               263              100.00%        0      0      0        30     0.0s
highs(70556,0x20b75e0c0) malloc: *** error for object 0x16b46e670: pointer being freed was not allocated
highs(70556,0x20b75e0c0) malloc: *** set a breakpoint in malloc_error_break to debug
zsh: abort      ./bin/highs ../check/instances/p01.mps
```

odow:
But of course, when I hit it with lldb it now works...

odow:
> The debug build with tests on fails because one of the checks in TestHipo.cpp fails. This is the test that checks if running HiPO twice with the same problem produces the exact same solution. Could there be sometihng non deterministic when running on Windows? 

I don't think non-determinism is the problem. By the looks of it, you set `openblas_set_num_threads(1)` and you fix the METIS_OPTION_SEED. 

Isn't it just a case that you've solved the problem, then you're warm-starting from a different solution? It's not the same as solving two models independently.

filikat:
> @filikat here's a bug in HiPO. It doesn't detect primal unboundedness:

Thanks, I will have a look. The detection of infeasibility/unboundedness is very simplistic for now and it needs some tuning.

filikat:
> The related issue is that we have re-interpreted `solver = ipm` to previously mean IPX and now mean `IPX | HiPO`, and there is a new `solver = "ipx"` option for IPX.
> 
> https://github.com/ERGO-Code/HiGHS/blob/364c83a51e44ba6c27def9c8fc1a49b1daf5ad5c/highs/lp_data/HighsOptions.h#L260-L262
> 
> This is potentially a breaking change for folks like GenX which force the solver to `ipm`. People with the default or `choose` should expect the algorithm to change between HIGHS versions. But not if they are explicitly choosing the algorithm.

I agree with you, we should maintain compatibility for people who were selecting "ipm" thinking that they got IPX.
I thought that it was already the case honestly.

jajhall:
I made this change deliberately as it aligns with what will happen in future. The ipm string can no longer be interpreted as a specific implementation.

jajhall:
> > @filikat here's a bug in HiPO. It doesn't detect primal unboundedness:
> 
> Thanks, I will have a look. The detection of infeasibility/unboundedness is very simplistic for now and it needs some tuning.
> 

Don't worry about this for now. The priority is performance for feasible LPs

filikat:
> But of course, when I hit it with lldb it now works...

@odow  Could you try to run the same problem, setting output_flag to true for the ipm in the analytic centre computation and log_dev_level to 1. Maybe we can see at what point the ipm fails.

filikat:
> > The debug build with tests on fails because one of the checks in TestHipo.cpp fails. This is the test that checks if running HiPO twice with the same problem produces the exact same solution. Could there be sometihng non deterministic when running on Windows?
> 
> I don't think non-determinism is the problem. By the looks of it, you set `openblas_set_num_threads(1)` and you fix the METIS_OPTION_SEED.
> 
> Isn't it just a case that you've solved the problem, then you're warm-starting from a different solution? It's not the same as solving two models independently.

@odow I see. I will change the test so that it uses two separate Highs objects, just to be sure.

highs/lp_data/HighsOptions.h
```diff
@@ -262,9 +262,9 @@ void reportOption(FILE* file, const HighsLogOptions& report_log_options,
                   const HighsFileType file_type);
 
 const string kSimplexString = "simplex";
-const string kIpmString = "ipm";
```
jajhall, line 265: @odow Revert this please

highs/lp_data/HighsOptions.h
```diff
@@ -262,9 +262,9 @@ void reportOption(FILE* file, const HighsLogOptions& report_log_options,
...
+const string kIpmString = "interior-point";
 const string kHipoString = "hipo";
-const string kIpxString = "ipx";
```
jajhall, line 267: @odow Revert this please

check/TestSpecialLps.cpp
```diff
@@ -215,9 +215,9 @@ void issue425(Highs& highs) {
       false;
 #endif
   if (use_hipo) {
-    solve(highs, "off", "ipm", require_model_status, 0, -15);
```
jajhall, line 218: @odow Revert this please

check/TestSpecialLps.cpp
```diff
@@ -533,9 +533,9 @@ void almostNotUnbounded(Highs& highs) {
     //
     // Prevent infinite loop in HiPO
     highs.setOptionValue("ipm_iteration_limit", 200);
```
jajhall, line 535: @odow Revert this please

jajhall (changes requested):
@odow Please revert the changes of

> const string kIpmString = "ipm";
const string kHipoString = "hipo";
const string kIpxString = "ipx";

to 

> const string kIpmString = "interior-point";
const string kHipoString = "hipo";
const string kIpxString = "ipm";

since IPM should not be synonymous with IPX, and the corresponding changes in `TestSpecialLps`.

jajhall:
> The related issue is that we have re-interpreted `solver = ipm` to previously mean IPX and now mean `IPX | HiPO`, and there is a new `solver = "ipx"` option for IPX.
> 
> https://github.com/ERGO-Code/HiGHS/blob/364c83a51e44ba6c27def9c8fc1a49b1daf5ad5c/highs/lp_data/HighsOptions.h#L260-L262
> 
> This is potentially a breaking change for folks like GenX which force the solver to `ipm`. People with the default or `choose` should expect the algorithm to change between HIGHS versions. But not if they are explicitly choosing the algorithm.

@odow How can `solver=ipm` mean anything other than "use interior point to solve the LP"?

With my changes that's still happening - except that HiPO (when available) is being used by default. Anyone who is setting `solver=ipm`  rather than the default value `solver=choose` (which implies simplex) is, most likely, solving big enough LPs that they want IPM rather than simplex. Hence it seems reasonable that the better IPM solver for large LPs - ie HiPO - is used by default.

All that's changed is that `solver=ipm` is now a "choice" setting, with HiGHS choosing HiPO. 

In due course HiGHS will choose between IPX and HiPO, using the former for small LPs, or on the basis of other characteristics. 

Similarly, in due course, with `solver=choose` HiGHS will choose between simplex, IPX and HiPO

If people wanting IPM are finding that HiPO is slower or less reliable than IPX, it's not difficult for them to change a single parameter to force the use of IPX.

odow:
This is a draft WIP PR. My goal is just to get the tests passing. 

Re ipm: This just shows the difficulty of naming things with backward compatibility. Fully agree that `ipm` should mean "solve with the interior point method of your choice" but people currently think means "use IPX". (I don't think people know that the "ipm" solver in HiGHS is called "ipx".) 

Until HiPO is mature, I think we should be making HiPO explicitly opt-in. 

IMO: the priority should be: fixing Windows segfaults, fixing failing unit tests, making it so we can actually compile and distribute HiPO. Once it's in `latest` and people are using it, then we can improve performance.

jajhall:
> This is a draft WIP PR. My goal is just to get the tests passing.

Agreed, which is why I was surprised at you making changes to solver code.

> Re ipm: This just shows the difficulty of naming things with backward compatibility. Fully agree that `ipm` should mean "solve with the interior point method of your choice" but people currently think means "use IPX". (I don't think people know that the "ipm" solver in HiGHS is called "ipx".)

So `solver=ipm` was correct. The error would have been to have the option `solver=ipx` for IPM when there was only one

If people don't know that the "ipm" solver in HiGHS is called "ipx", I fail to see the problem with switching the default IPM solver

> Until HiPO is mature, I think we should be making HiPO explicitly opt-in.

I note your opinion. However, by making HiPO default when `solver=ipm` (which few users will do) will expose it to a lot more models, and most users will get a meaningful performance enhancement. Those who experience slow-down or failure can contribute their instances and help @filikat improve the performance and robustness of HiPO. This is already happening.

> IMO: the priority should be: fixing Windows segfaults, fixing failing unit tests, making it so we can actually compile and distribute HiPO. Once it's in `latest` and people are using it, then we can improve performance.

I note your opinion, but @filikat has only so much to contribute in this area. His priority is performance and robustness. At this stage, infeasibility detection is not important.

odow:
@filikat here's now a segfault in the Julia build on linux:

image

This was the Julia test:
```julia
function test_callback_ctrl_c()
    model = HiGHS.Optimizer()
    MOI.set(model, MOI.RawOptimizerAttribute("solver"), "ipm")
    MOI.set(model, MOI.RawOptimizerAttribute("presolve"), "off")
    x = MOI.add_variables(model, 3)
    c = MOI.add_constraint.(model, 1.0 .* x, MOI.EqualTo.(1.0:3.0))
    MOI.set(model, MOI.ObjectiveSense(), MOI.MAX_SENSE)
    f = 1.0 * x[1] + x[2] + x[3]
    MOI.set(model, MOI.ObjectiveFunction{typeof(f)}(), f)
    user_callback = (args...) -> throw(InterruptException())
    MOI.set(model, HiGHS.CallbackFunction(), user_callback)
    MOI.optimize!(model)
    @test MOI.get(model, MOI.TerminationStatus()) == MOI.INTERRUPTED
    return
end
```
We terminate from a callback with `ipm`. Does HiPO support interrupts?

```c
#include <stdlib.h>
#include "interfaces/highs_c_api.h"

static void callback_cfn(
    const int callback_type, const char* message,
    const HighsCallbackDataOut* data_out, HighsCallbackDataIn* data_in,
    void* user_callback_data) {
    data_in->user_interrupt = 1;
    return;
}

int main() {
    void* highs = Highs_create();
    int ret = Highs_setStringOptionValue(highs, "solver", "ipm");
    ret = Highs_setStringOptionValue(highs, "presolve", "off");
    double INF = Highs_getInfinity(highs);
    ret = Highs_addCol(highs, 1.0, -INF, INF, 0, NULL, NULL);
    ret = Highs_addCol(highs, 1.0, -INF, INF, 0, NULL, NULL);
    ret = Highs_addCol(highs, 1.0, -INF, INF, 0, NULL, NULL);
    int a_index[3] = {0, 1, 2};
    double a_value[3] = {1.0, 1.0, 1.0};
    ret = Highs_addRow(highs, 1.0, 1.0, 3, a_index, a_value);
    ret = Highs_changeObjectiveSense(highs, kHighsObjSenseMaximize);
    ret = Highs_setCallback(highs, callback_cfn, NULL);
    ret = Highs_startCallback(highs, kHighsCallbackIpmInterrupt);
    ret = Highs_run(highs);
    Highs_destroy(highs);
    return 0;
}

(base) odow@Mac build % clang ../check/TestHiPOInterrupt.c -I../highs -I. -L./lib -lhighs -Wl,-rpath,@executable_path/lib
(base) odow@Mac build % ./a.out                                                                                          
Running HiGHS 1.11.0 (git hash: 6b65db15d): Copyright (c) 2025 HiGHS under MIT licence terms
LP has 1 row; 3 cols; 3 nonzeros
Coefficient ranges:
  Matrix [1e+00, 1e+00]
  Cost   [1e+00, 1e+00]
  Bound  [0e+00, 0e+00]
  RHS    [1e+00, 1e+00]
Not presolved: solving the LP
HiPO model has 1 rows, 3 columns and 3 nonzeros
WARNING: Hipo: User interrupt
WARNING: Hipo: Crossover not run
zsh: segmentation fault  ./a.out
(base) odow@Mac build % lldb a.out
(lldb) target create "a.out"
Current executable set to '/Users/odow/git/ERGO-Code/HiGHS/build/a.out' (arm64).
(lldb) run
Process 84381 launched: '/Users/odow/git/ERGO-Code/HiGHS/build/a.out' (arm64)
Running HiGHS 1.11.0 (git hash: 6b65db15d): Copyright (c) 2025 HiGHS under MIT licence terms
LP has 1 row; 3 cols; 3 nonzeros
Coefficient ranges:
  Matrix [1e+00, 1e+00]
  Cost   [1e+00, 1e+00]
  Bound  [0e+00, 0e+00]
  RHS    [1e+00, 1e+00]
Not presolved: solving the LP
HiPO model has 1 rows, 3 columns and 3 nonzeros
WARNING: Hipo: User interrupt
WARNING: Hipo: Crossover not run
Process 84381 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x20)
    frame #0: 0x0000000100e1e674 libhighs.1.dylib`std::__1::vector<double, std::__1::allocator<double>>::begin[abi:ne190102]() const + 20
libhighs.1.dylib`std::__1::vector<double, std::__1::allocator<double>>::begin[abi:ne190102]:
->  0x100e1e674 <+20>: ldr    x1, [x0]
    0x100e1e678 <+24>: bl     0x100e1e8c8    ; std::__1::vector<double, std::__1::allocator<double>>::__make_iter[abi:ne190102](double const*) const
    0x100e1e67c <+28>: str    x0, [sp, #0x8]
    0x100e1e680 <+32>: ldr    x0, [sp, #0x8]
Target 0: (a.out) stopped.
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x20)
  * frame #0: 0x0000000100e1e674 libhighs.1.dylib`std::__1::vector<double, std::__1::allocator<double>>::begin[abi:ne190102]() const + 20
    frame #1: 0x0000000101177a14 libhighs.1.dylib`hipo::Iterate::extract(std::__1::vector<double, std::__1::allocator<double>>&, std::__1::vector<double, std::__1::allocator<double>>&, std::__1::vector<double, std::__1::allocator<double>>&, std::__1::vector<double, std::__1::allocator<double>>&, std::__1::vector<double, std::__1::allocator<double>>&, std::__1::vector<double, std::__1::allocator<double>>&, std::__1::vector<double, std::__1::allocator<double>>&) const + 68
    frame #2: 0x0000000101183ce4 libhighs.1.dylib`hipo::Solver::getInteriorSolution(std::__1::vector<double, std::__1::allocator<double>>&, std::__1::vector<double, std::__1::allocator<double>>&, std::__1::vector<double, std::__1::allocator<double>>&, std::__1::vector<double, std::__1::allocator<double>>&, std::__1::vector<double, std::__1::allocator<double>>&, std::__1::vector<double, std::__1::allocator<double>>&, std::__1::vector<double, std::__1::allocator<double>>&) const + 112
    frame #3: 0x0000000100d9e914 libhighs.1.dylib`getHipoNonVertexSolution(HighsOptions const&, HighsLp const&, int, int, std::__1::vector<double, std::__1::allocator<double>> const&, std::__1::vector<char, std::__1::allocator<char>> const&, hipo::Solver const&, HighsModelStatus, HighsSolution&) + 192
    frame #4: 0x0000000100d9dbec libhighs.1.dylib`solveLpHipo(HighsOptions const&, HighsTimer&, HighsLp const&, HighsBasis&, HighsSolution&, HighsModelStatus&, HighsInfo&, HighsCallback&) + 2244
    frame #5: 0x0000000100d9d31c libhighs.1.dylib`solveLpHipo(HighsLpSolverObject&) + 84
    frame #6: 0x0000000100e4e1e4 libhighs.1.dylib`solveLp(HighsLpSolverObject&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>) + 936
    frame #7: 0x0000000100db70a4 libhighs.1.dylib`Highs::callSolveLp(HighsLp&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>) + 184
    frame #8: 0x0000000100dae38c libhighs.1.dylib`Highs::optimizeModel()::$_0::operator()(HighsLp&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, double&) const + 116
    frame #9: 0x0000000100daa938 libhighs.1.dylib`Highs::optimizeModel() + 5448
    frame #10: 0x0000000100da8f28 libhighs.1.dylib`Highs::run() + 336
    frame #11: 0x0000000100d4cb1c libhighs.1.dylib`Highs_run + 24
    frame #12: 0x00000001000006e4 a.out`main + 388
    frame #13: 0x000000019d42ab98 dyld`start + 6076
```

odow:
For Windows `Release`, there's a segfault with `check/instances/flugpl.mps`. It's not there in `Debug`, but there's one later on.

filikat:
> @filikat here's now a segfault in the Julia build on linux:
>
> We terminate from a callback with `ipm`. Does HiPO support interrupts?

Thanks for spotting this. I had one interrupt check before the data was fully initialised. If triggered immediately, some pointers were not set and this was causing the segfault.

filikat:
> For Windows `Release`, there's a segfault with `check/instances/flugpl.mps`. It's not there in `Debug`, but there's one later on.

I cannot replicate this for now, because it works fine on Mac and Linux.
If I can get access to a Windows machine I will have a look.
If you manage to obtain a crash report for this problem, please share it with me.

odow:
@filikat I managed to trap a backtrace:
```
* thread #4, stop reason = EXC_BAD_ACCESS (code=1, address=0x48)
  * frame #0: 0x00000001017ad9a8 libhighs.1.dylib`hipo::Factorise::processSupernode(int) + 752
    frame #1: 0x00000001017ae2b8 libhighs.1.dylib`hipo::Factorise::processSupernode(int)::$_0::operator()() const + 32
    frame #2: 0x00000001017ae408 libhighs.1.dylib`HighsTask::Callable<hipo::Factorise::processSupernode(int)::$_0>::operator()() + 44
    frame #3: 0x0000000101498b34 libhighs.1.dylib`HighsTask::run() + 60
    frame #4: 0x00000001014989d0 libhighs.1.dylib`highs::parallel::sync(HighsSplitDeque*) + 140
    frame #5: 0x00000001016a7664 libhighs.1.dylib`highs::parallel::sync() + 16
    frame #6: 0x00000001017ad970 libhighs.1.dylib`hipo::Factorise::processSupernode(int) + 696
    frame #7: 0x00000001017ae2b8 libhighs.1.dylib`hipo::Factorise::processSupernode(int)::$_0::operator()() const + 32
    frame #8: 0x00000001017ae408 libhighs.1.dylib`HighsTask::Callable<hipo::Factorise::processSupernode(int)::$_0>::operator()() + 44
    frame #9: 0x00000001000ae420 unit_tests`HighsTask::run(HighsSplitDeque*) + 76
    frame #10: 0x00000001000ae048 unit_tests`HighsSplitDeque::runStolenTask(HighsTask*) + 64
    frame #11: 0x00000001000afd80 unit_tests`HighsTaskExecutor::run_worker(int, HighsTaskExecutor*) + 140
    frame #12: 0x00000001000b4fe4 unit_tests`decltype(std::declval<void (*)(int, HighsTaskExecutor*)>()(std::declval<int>(), std::declval<HighsTaskExecutor*>())) std::__1::__invoke[abi:ne190102]<void (*)(int, HighsTaskExecutor*), int, HighsTaskExecutor*>(void (*&&)(int, HighsTaskExecutor*), int&&, HighsTaskExecutor*&&) + 52
    frame #13: 0x00000001000b4f60 unit_tests`void std::__1::__thread_execute[abi:ne190102]<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct>>, void (*)(int, HighsTaskExecutor*), int, HighsTaskExecutor*, 2ul, 3ul>(std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct>>, void (*)(int, HighsTaskExecutor*), int, HighsTaskExecutor*>&, std::__1::__tuple_indices<2ul, 3ul>) + 68
    frame #14: 0x00000001000b47c4 unit_tests`void* std::__1::__thread_proxy[abi:ne190102]<std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct>>, void (*)(int, HighsTaskExecutor*), int, HighsTaskExecutor*>>(void*) + 88
    frame #15: 0x000000019d7cbc0c libsystem_pthread.dylib`_pthread_start + 136
```

I also get occasional stalls somewhere in the MIP solver.

Is HiPO threadsafe?

Edit: here's another one

```
(lldb) bt
* thread #3, stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
  * frame #0: 0x00000001017b0728 libhighs.1.dylib`hipo::HybridHybridFormatHandler::extremeEntries() + 336
    frame #1: 0x00000001017adc88 libhighs.1.dylib`hipo::Factorise::processSupernode(int) + 1488
    frame #2: 0x00000001017af0d4 libhighs.1.dylib`hipo::Factorise::run(hipo::Numeric&)::$_0::operator()() const + 32
    frame #3: 0x00000001017af224 libhighs.1.dylib`HighsTask::Callable<hipo::Factorise::run(hipo::Numeric&)::$_0>::operator()() + 44
    frame #4: 0x00000001000ae420 unit_tests`HighsTask::run(HighsSplitDeque*) + 76
    frame #5: 0x00000001000ae048 unit_tests`HighsSplitDeque::runStolenTask(HighsTask*) + 64
    frame #6: 0x00000001000afd80 unit_tests`HighsTaskExecutor::run_worker(int, HighsTaskExecutor*) + 140
    frame #7: 0x00000001000b4fe4 unit_tests`decltype(std::declval<void (*)(int, HighsTaskExecutor*)>()(std::declval<int>(), std::declval<HighsTaskExecutor*>())) std::__1::__invoke[abi:ne190102]<void (*)(int, HighsTaskExecutor*), int, HighsTaskExecutor*>(void (*&&)(int, HighsTaskExecutor*), int&&, HighsTaskExecutor*&&) + 52
    frame #8: 0x00000001000b4f60 unit_tests`void std::__1::__thread_execute[abi:ne190102]<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct>>, void (*)(int, HighsTaskExecutor*), int, HighsTaskExecutor*, 2ul, 3ul>(std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct>>, void (*)(int, HighsTaskExecutor*), int, HighsTaskExecutor*>&, std::__1::__tuple_indices<2ul, 3ul>) + 68
    frame #9: 0x00000001000b47c4 unit_tests`void* std::__1::__thread_proxy[abi:ne190102]<std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct>>, void (*)(int, HighsTaskExecutor*), int, HighsTaskExecutor*>>(void*) + 88
    frame #10: 0x000000019d7cbc0c libsystem_pthread.dylib`_pthread_start + 136
```
```
(lldb) bt
* thread #3, stop reason = signal SIGABRT
  * frame #0: 0x000000019d792388 libsystem_kernel.dylib`__pthread_kill + 8
    frame #1: 0x000000019d7cb88c libsystem_pthread.dylib`pthread_kill + 296
    frame #2: 0x000000019d6d4a3c libsystem_c.dylib`abort + 124
    frame #3: 0x000000019d5d8174 libsystem_malloc.dylib`malloc_vreport + 892
    frame #4: 0x000000019d5dbc90 libsystem_malloc.dylib`malloc_report + 64
    frame #5: 0x000000019d5e021c libsystem_malloc.dylib`___BUG_IN_CLIENT_OF_LIBMALLOC_POINTER_BEING_FREED_WAS_NOT_ALLOCATED + 32
    frame #6: 0x0000000101341968 libhighs.1.dylib`void std::__1::__libcpp_operator_delete[abi:ne190102]<void*>(void*) + 24
    frame #7: 0x0000000101341944 libhighs.1.dylib`void std::__1::__do_deallocate_handle_size[abi:ne190102]<>(void*, unsigned long) + 28
    frame #8: 0x000000010134191c libhighs.1.dylib`std::__1::__libcpp_deallocate[abi:ne190102](void*, unsigned long, unsigned long) + 36
    frame #9: 0x0000000101368258 libhighs.1.dylib`std::__1::allocator<double>::deallocate[abi:ne190102](double*, unsigned long) + 44
    frame #10: 0x00000001013680e4 libhighs.1.dylib`std::__1::allocator_traits<std::__1::allocator<double>>::deallocate[abi:ne190102](std::__1::allocator<double>&, double*, unsigned long) + 40
    frame #11: 0x0000000101376130 libhighs.1.dylib`std::__1::vector<double, std::__1::allocator<double>>::__vdeallocate() + 92
    frame #12: 0x0000000101382d28 libhighs.1.dylib`std::__1::vector<double, std::__1::allocator<double>>::__move_assign(std::__1::vector<double, std::__1::allocator<double>>&, std::__1::integral_constant<bool, true>) + 36
    frame #13: 0x000000010137a9ec libhighs.1.dylib`std::__1::vector<double, std::__1::allocator<double>>::operator=[abi:ne190102](std::__1::vector<double, std::__1::allocator<double>>&&) + 40
    frame #14: 0x00000001017af808 libhighs.1.dylib`hipo::FormatHandler::terminate(std::__1::vector<double, std::__1::allocator<double>>&, std::__1::vector<double, std::__1::allocator<double>>&, std::__1::vector<int, std::__1::allocator<int>>&, std::__1::vector<double, std::__1::allocator<double>>&) + 56
    frame #15: 0x00000001017adce0 libhighs.1.dylib`hipo::Factorise::processSupernode(int) + 1576
    frame #16: 0x00000001017af0d4 libhighs.1.dylib`hipo::Factorise::run(hipo::Numeric&)::$_0::operator()() const + 32
    frame #17: 0x00000001017af224 libhighs.1.dylib`HighsTask::Callable<hipo::Factorise::run(hipo::Numeric&)::$_0>::operator()() + 44
    frame #18: 0x00000001000ae420 unit_tests`HighsTask::run(HighsSplitDeque*) + 76
    frame #19: 0x00000001000ae048 unit_tests`HighsSplitDeque::runStolenTask(HighsTask*) + 64
    frame #20: 0x00000001000afd80 unit_tests`HighsTaskExecutor::run_worker(int, HighsTaskExecutor*) + 140
    frame #21: 0x00000001000b4fe4 unit_tests`decltype(std::declval<void (*)(int, HighsTaskExecutor*)>()(std::declval<int>(), std::declval<HighsTaskExecutor*>())) std::__1::__invoke[abi:ne190102]<void (*)(int, HighsTaskExecutor*), int, HighsTaskExecutor*>(void (*&&)(int, HighsTaskExecutor*), int&&, HighsTaskExecutor*&&) + 52
    frame #22: 0x00000001000b4f60 unit_tests`void std::__1::__thread_execute[abi:ne190102]<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct>>, void (*)(int, HighsTaskExecutor*), int, HighsTaskExecutor*, 2ul, 3ul>(std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct>>, void (*)(int, HighsTaskExecutor*), int, HighsTaskExecutor*>&, std::__1::__tuple_indices<2ul, 3ul>) + 68
    frame #23: 0x00000001000b47c4 unit_tests`void* std::__1::__thread_proxy[abi:ne190102]<std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct>>, void (*)(int, HighsTaskExecutor*), int, HighsTaskExecutor*>>(void*) + 88
    frame #24: 0x000000019d7cbc0c libsystem_pthread.dylib`_pthread_start + 136
```

filikat:
> @filikat I managed to trap a backtrace:

@odow Thanks! Are these all on Windows? Does it just give different errors running on the same problem?

> Is HiPO threadsafe?

I think it is. I never had any indication that it wasn't, running on Mac and Linux.
It would be interesting to see if the same errors happear when running with parallel=off.

I have an idea of something that may fix the third crash that you reported. The second happens in a function that is used only for debugging, so for now I can just comment it out. The first one is a bit more cryptic, because it's not clear where exactly it happens.

BTW, the test to check for determinism on Windows still fails, even if now it definitely uses two separate Highs objects. There's something weird going on.

filikat:
@odow I pushed some changes to hipo-dev. I would be curious to know if any of the errors that you got go away.

odow:
> Are these all on Windows

On aarch64 macos with ` lldb -- ./bin/unit_tests`

>  I pushed some changes to hipo-dev. I would be curious to know if any of the errors that you got go away.

I've rebased. But just push to this branch? Or open a PR for hipo-dev. You can see the CI runs for it as still failing as well.

filikat:
> On aarch64 macos with ` lldb -- ./bin/unit_tests`

Ok, I have the same. I will try to replicate.

> But just push to this branch? 

I didn't want to push to your branch. If you prefer it's fine.

> You can see the CI runs for it as still failing as well.

Yes, I can see them. I have a separate branch only to run specific workflows and tests on Windows. Ivet has a PC with Windows, so hopefully we'll find the issue soon.

odow:
For the Julia tests, it's okay if `test_continuous_objective_bound` fails (it looks like you're now getting `0` gap, which is good).

But `test_infeasible_point` should pass:

https://github.com/jump-dev/HiGHS.jl/blob/99792277fa7b950e88a45b47f34162a003cd045a/test/MOI_wrapper.jl#L862-L876

filikat:
@odow I did manage to replicate the issue that you observed on Mac. I think I managed to fix it. 

I will push this change to hipo, since it's a critical fix that should immediately be on the main hipo branch, so that everyone who uses it can benefit. Please disregard the changes that I made to hipo-dev yesterday, as these were only made to try to find the issue (sorry for all the changes during these days).

The issue of segfaults and stalling on Mac should be resolved. Please let me know if the issue persists. 
I noticed that another test fails now (tsp-p01, only in Debug). This seems to be related to the mip solver starting the AC computation but not completing it. I will try to fix this as well.

I leave an explanation here for future reference, since this was an important issue:
The issue arises only when calling HiPO and then cancelling the task (either due to an interrupt, or because the AC computation is terminated early). This should clear the parallel deque of all tasks that HiPO spawned, but this was not the case, because I was using the scheduler improperly. I now use only TaskGroup, which guarantees that all tasks are either synced or cancelled in its destructor. Before, there were tasks left in the queue after the HiPO object was destroyed, producing some dangling references and potentially segfaults or deadlocks.

filikat:
@odow Could you help me understand something in the Julia tests? I don't know if I'm reading the output correctly.

> For the Julia tests, it's okay if `test_continuous_objective_bound` fails (it looks like you're now getting `0` gap, which is good).

Is this test failing because the gap is not strictly larger than zero?


> But `test_infeasible_point` should pass:

Above "test_infeasible_point: Test Failed", I see a problem with 0 rows and 1 col, which is not the same as the problem described at the link that you sent.

odow:
@filikat Mac is still failing

> Is this test failing because the gap is not strictly larger than zero?

Yes. I've relaxed the test in HiGHS.jl. If I make a new release it will flow through to here.

> which is not the same as the problem described at the link that you sent.

Printing was unrelated. Most of the tests have printing turned off.

Here's a C example
```c
#include <stdlib.h>
#include "interfaces/highs_c_api.h"

int main() {
    void* highs = Highs_create();
    int ret = Highs_setStringOptionValue(highs, "solver", "ipm");
    ret = Highs_setStringOptionValue(highs, "presolve", "off");
    double INF = Highs_getInfinity(highs);
    ret = Highs_addCol(highs, 0.0, 0.0, INF, 0, NULL, NULL);
    ret = Highs_addCol(highs, 0.0, 0.0, INF, 0, NULL, NULL);
    int a_index[2] = {0, 1};
    double a_value[2] = {1.0, 2.0};
    ret = Highs_addRow(highs, -INF, -1.0, 2, a_index, a_value);
    ret = Highs_run(highs);
    Highs_destroy(highs);
    return 0;
}
```

```
Running HiGHS 1.11.0 (git hash: 01b730692): Copyright (c) 2025 HiGHS under MIT licence terms
LP has 1 row; 2 cols; 2 nonzeros
Coefficient ranges:
  Matrix [1e+00, 2e+00]
  Cost   [0e+00, 0e+00]
  Bound  [0e+00, 0e+00]
  RHS    [1e+00, 1e+00]
Not presolved: solving the LP
HiPO model has 1 rows, 2 columns and 2 nonzeros

Running HiPO
Threads:                        5
Rows:                           1.0e+00
Cols:                           3.0e+00
Nnz:                            3.0e+00
Range of A:                     [1.0e+00, 1.0e+00], ratio 1.0e+00
Range of b:                     [1.0e+00, 1.0e+00], ratio 1.0e+00
Range of c:                     [0.0e+00, 0.0e+00], ratio -
Range of bounds:                [0.0e+00, 0.0e+00], ratio -
Scaling coefficients:           [5.0e-01, 1.0e+00], ratio 2.0e+00
Newton system:                  NE preferred
Parallelism:                    Tree preferred

Factorisation statistics
Size:                           1.00e+00
Nnz:                            1.00e+00
Fill-in:                        1.00
Serial memory:                  8.0e+00 B
Flops:                          0.0e+00

 iter       primal obj         dual obj       pinf       dinf       gap    time
    0   0.00000000e+00   0.00000000e+00   1.03e+00   3.00e+00  0.00e+00     0.0
    1   0.00000000e+00   1.80397621e+00   1.67e-01   3.00e-04  9.48e-01     0.0

Summary
HiPO runtime:                   0.00
Status:                         internal error
HiPO iterations:                2
ERROR:   Hipo: Internal error
Model status        : Solve error
HiGHS run time      :          0.00

Sub-solver timing
Solver                 Calls    Time       Time/call
HiPO                       1  3.1324e-03  3.1324e-03
```

filikat:
> @filikat Mac is still failing

Still segfault? 
I know that there is still a unit test that fails with an uncaught exception, but I will work on that. 
If you noticed other segfaults though, it means that my fix is not good.

> Here's a C example

Thanks. This is probably some edge case that I didn't consider. I will look at it tomorrow.

odow:
Uncaught exception

filikat:
The test for determinism should pass now. The issue was with a missing compile definition in the CMake file for Windows. I also had to install openblas using 'vcpkg install openblas[threads]' to get deterministic results.

The uncaught exception on Mac should also be fixed, but this is a temporary ugly fix. I need to understand something more about the inner mechanism of the parallel scheduler to fix it properly.

The Julia tests should all pass.

We managed to track the segfault on Windows Release and it is related to Metis. Avoiding the call to Metis, the code runs fine. There doesn't seem to be any issue with the data that is passed to Metis and Ivet cannot reproduce this on her local Windows machine, as it runs fine. 
@odow Is there any other way that we can install Metis on Windows? I see that for Mac and Linux you have used binaries coming from Julia.

odow:
What's up with all of this?

https://github.com/ERGO-Code/HiGHS/blob/5ffa824d135c68dc78b75c89188bd6aa10a3e6df/cmake/FindHipoDeps.cmake#L5-L49

Why do we ignore `METIS_ROOT` and `BLAS_ROOT` on Windows, and not search for BLAS on Mac etc?

amontoison:
@odow @filikat 
If you are not aware, you can ssh directly the CI runner of GitHub Actions with this:
```shell
# Uncomment this section to obtain ssh access to VM
# - name: Setup tmate session
#   if: inputs.os == 'windows-latest'
#   uses: mxschmitt/action-tmate@v3
```
I used it a few time to debug Windows issues that I was not able to reproduce locally.

odow:
Or we can just get @galabovaa to take a look with her Windows machine :smile: I have forgotten everything there is to know about Windows.

odow:
@filikat I've also just noticed:

https://github.com/ERGO-Code/HiGHS/blob/561058c589763214fcf8542c9778c301411b047a/highs/CMakeLists.txt#L189-L202

Why are we hard-coding AppleAccelerate for Mac? For Julia we want to be able to bring our own BLAS if specified by `BLAS_ROOT`

odow:
Oh d'oh. The Julia files are mingw, so they have only the `.dll` and not the `.lib`.

filikat:
> Why do we ignore `METIS_ROOT` and `BLAS_ROOT` on Windows

This was done by Ivet to make the Windows build simpler by using vcpkg. In light of the recent issues, we probably need to change this.

> Why are we hard-coding AppleAccelerate for Mac? For Julia we want to be able to bring our own BLAS if specified by `BLAS_ROOT`

This was done by me since Apple Blas is very fast and is available on all Macs. I understand though that it is good to offer an alternative if the user wants it.

> Or we can just get @galabovaa to take a look with her Windows machine 😄 I have forgotten everything there is to know about Windows.

Unfortunately, she cannot repicate the Metis segfault. On her PC it works just fine.
Alexis mentioned that the issue may be that the precompiled Metis in vcpkg is not fully compatible with the rest, so we may need to compile Metis directly on Windows as well.

odow:
> In light of the recent issues, we probably need to change this.

Yeah, when cross compiling we need to be able to specify the paths. We won't be using vcpkg.

> I understand though that it is good to offer an alternative if the user wants it.

Yeah, our plan would be to ship HiGHS with https://github.com/JuliaLinearAlgebra/libblastrampoline, so that we can change out the BLAS backend at runtime: https://github.com/jump-dev/Ipopt.jl?tab=readme-ov-file#blas-and-lapack

odow:
Soooo. We now have the three Julia builds for Mac, windows, and linux, that all cross-compile and then test. They pass tests because they don't really use HiPO (because HiGHS.jl doesn't now about HiPO)

odow:
@galabovaa can you reproduce these segfaults now?

galabovaa:
Yes, and making progress locally!

odow:
So the fact that we have a segfault with the vcpkg on Windows, but not in the Julia build does make it look like there's something funky with the vcpkg Metis. Is it compiled with the right `Int` type, etc?

galabovaa:
No, that segfault happened with the latest metis version from vcpkg, but not the older one I was using locally. Seems to be an issue with the latest metis only

odow:
@amontoison any ideas?

highs/CMakeLists.txt
```diff
@@ -178,48 +178,21 @@ else()
...
-        else()
-            message(FATAL_ERROR "No BLAS library available")
-        endif(BLAS_LIB)
-    endif(APPLE)
-
-    if (NOT WIN32)
-      # Find_package works if deps are installed with vcpkg.
-
-      # Metis
+    target_link_libraries(highs BLAS::BLAS)
```
odow, line 187: @filikat I've removed the special handing for BLAS/OpenBLAS/Accelerate. Is that really necessary?

amontoison:
> @amontoison any ideas?

- They compiled METIS without the patches for Windows
- They compiled a different version than 5.1.0
- Incompatibility of C / C++ libraries between METIS compiled in vcpkg and other librairies in mingw.

highs/CMakeLists.txt
filikat: It's important that HiPO knows if openblas is used, because the function to set the number of threads to 1 is only defined in the openblas library. If it's better, we can probably achieve that with try_compile?

odow:
@galabovaa can we just pull in the older version from vcpkg?

highs/CMakeLists.txt
filikat: It's also good if HiPO knows when Accelerate is used. Apple BLAS is already parallelised, if I try to make it more parallel by hand, I actually make it worse. So if Accelerate is used, I turn off some of the parallelisation.
I guess there will be other specific details of BLAS implementations that the code will need to be aware of, to achieve the best performance, so it would be useful if there is a way to communicate to HiPO which BLAS is being used.

galabovaa:
We can, but there is a better solution that I'm working on. Will let you know in due course

highs/CMakeLists.txt
odow: I've added back a check based on the string name of the library (for `accelerate` or `openblas`). That's not perfect, it's probably okay for now. 

If we're using libblastrampoline and the BLAS is chosen at runtime, then I guess we can force the number of BLAS threads in HiGHS.jl.

odow:
Any progress? Is there something I can try?

galabovaa:
Yes: this is the latest version of my edits with the latest updates from @filikat 

https://github.com/ERGO-Code/HiGHS/tree/hipo-tt-dev-gk

The runner tests pass with a heavily patched up version 5.1.0 of metis. The version used in Julia with several extra fixes that were required for MSVC. Also, in another repo, I have initial tests passing with the older version from vcpkg on windows and the latest metis from GitHub on ubuntu and macos. 

There are still more checks I would like to make, like whether cmake install works and including HiGHS in other CMake projects but this branch is a good place to start. So, it would be great if you could pull the changes from there and double check that the Julia build you have here still works

highs/ipm/hipo/factorhighs/Factorise.cpp
```diff
@@ -20,7 +20,12 @@ Factorise::Factorise(const Symbolic& S, const std::vector<Int>& rowsA,
                      const std::vector<double>& valA, const Regul& regul,
                      const Log* log, DataCollector& data,
                      std::vector<std::vector<double>>& sn_columns)
-    : S_{S}, sn_columns_{sn_columns}, regul_{regul}, log_{log}, data_{data} {
+    : S_{S},
+      child_left_(S_.sn()),
+      sn_columns_{sn_columns},
+      regul_{regul},
+      log_{log},
+      data_{data} {
```
odow: What about these changes? I don't see them in your branch. Do we still need them?

odow:
> So, it would be great if you could pull the changes from there and double check that the Julia build you have here still works

Easier if you pull the Julia actions into your branch?

odow:
Trying here: https://github.com/ERGO-Code/HiGHS/pull/2572

odow:
If we're patching METIS, we could just carry the patch so that glib doesn't need to be linked explicitly? There really should be no need for HiGHS to know about gklib

highs/ipm/hipo/factorhighs/Factorise.cpp
filikat: You can disregard the commits that I made in this PR, since I updated them and applied to hipo-dev, from where Ivet started to produce hipo-tt-dev-gk

filikat:
@galabovaa it seems like hipo-tt-dev-gk doesn't include the latest changes from hipo-dev. Maybe you made the branch starting from hipo instead?

odow:
What if I just use your patched version of METIS here? (Oh, because of the gklib stuff)

filikat:
> If we're patching METIS, we could just carry the patch so that glib doesn't need to be linked explicitly? There really should be no need for HiGHS to know about gklib

I think that's one of the patches that Ivet applied to her version of Metis (https://github.com/galabovaa/METIS/tree/510-w)

galabovaa:
@filikat 

> @galabovaa it seems like hipo-tt-dev-gk doesn't include the latest changes from hipo-dev. Maybe you made the branch starting from hipo instead?

I did start from hipo, I thought I merged hipo-dev but must have forgotten to pull. I will merge it tomorrow when we meet

@odow 

> If we're patching METIS, we could just carry the patch so that glib doesn't need to be linked explicitly? There really should be no need for HiGHS to know about gklib

The GKlib path is only optional as we discussed, with the patched up metis or vcpkg you just don't specify it. Sure, I can add the julia workflows from this PR too

filikat:
> I did start from hipo, I thought I merged hipo-dev but must have forgotten to pull. I will merge it tomorrow when we meet

Looks like you merged hipo-dev into hipo-tt-dev but not hipo-tt-dev-gk

highs/ipm/hipo/factorhighs/Factorise.cpp
odow: So it's okay to remove, even though they aren't in `hipo-dev`?

https://github.com/ERGO-Code/HiGHS/blob/91f7d7635a93b763cbdb8d89b87b709767201792/highs/ipm/hipo/factorhighs/Factorise.cpp#L23

odow:
I'm seeing lots of green! Looks like Ivet's patched METIS did the trick.

odow:
Why do we have `hipo` and `hipo-dev`?

highs/ipm/IpxWrapper.cpp
```diff
@@ -83,6 +83,7 @@ HighsStatus solveLpIpx(const HighsOptions& options, HighsTimer& timer,
   } else if (options.log_dev_level == kHighsLogDevLevelVerbose) {
     parameters.debug = 4;
   }
+  parameters.debug = 4;
```
odow:
```suggestion
```

filikat:
> Why do we have `hipo` and `hipo-dev`?

Some people are already using the code in `hipo`. I use `hipo-dev` for stuff that I'm working on, before moving it to `hipo` when it works. It's just for now, once `hipo` gets merged in latest, `hipo` will become the working branch for me.

highs/ipm/hipo/factorhighs/Factorise.cpp
filikat: Yes, I found a better way of dealing with the uncaught exception, that is in `hipo-dev`.

odow:
I've rebased onto `hipo-dev`. Fingers crossed I didn't make a mistake...

odow:
I've rebased this onto the latest `hipo-dev`. 

What's the status of this?

If we don't like detecting openblas/accelerate from the string, another option for setting `HIPO_USES_OPENBLAS|ACCELERATE` is for them to be options that are explicitly set by user when compiling?

filikat:
@odow Ivet is making progress in`hipo-tt`. Once she finishes, we will merge that into `hipo-dev`. She's working also on BLAS detection.

galabovaa:
I added the julia build and workflows to hipo-tt and the tests are passing now. @odow, can you please have a look and double check that things there work as expected? I looked up how blas trampoline would work and found that a suggested way was to use BLA_VENDOR so I added this to our CMake for the BLAS, is that OK?

odow:
See the builds https://github.com/ERGO-Code/HiGHS/actions/runs/18570757342/job/52943747516

Mac and linux are finding the host BLAS: 

image

We need them to find the BLAS in `$prefix`, like Windows does:

image

We cannot assume that the machines we deploy the binaries on have a BLAS in the same location.

odow:
I don't really see the need for

https://github.com/ERGO-Code/HiGHS/blob/fb5e7cf040b2a9e10d25fa1febcc6ff9cbe54e53/cmake/FindHipoDeps.cmake#L5-L109

What's the problem if we just do

https://github.com/ERGO-Code/HiGHS/blob/fb5e7cf040b2a9e10d25fa1febcc6ff9cbe54e53/cmake/FindHipoDeps.cmake#L102C5-L108

and then

https://github.com/ERGO-Code/HiGHS/blob/fb5e7cf040b2a9e10d25fa1febcc6ff9cbe54e53/highs/CMakeLists.txt#L219-L227

Is there some issue if a downstream project uses HiGHS in its CMake?

.github/julia/build_tarballs.jl
```diff
@@ -33,7 +33,10 @@ cmake -DCMAKE_INSTALL_PREFIX=${prefix} \
     -DCMAKE_BUILD_TYPE=Release \
     -DBUILD_SHARED_LIBS=${BUILD_SHARED} \
     -DZLIB_USE_STATIC_LIBS=${BUILD_STATIC} \
-    -DFAST_BUILD=ON ..
+    -DHIPO=ON \
+    -DBLAS_LIBRARIES="${libdir}/libopenblas.${dlext}" \
```
odow, line 37: @galabovaa in the Julia build I just forced the BLAS I wanted

odow:
Stepping back, what are the use cases we're trying to support?

It seems you want the defaults to work if:

1. Someone on linux who has done `sudo apt install libopenblas-dev`
2. Someone is on macOS and has accelerate
3. Someone on Windows if they have vcpkg install openblas and provided the vcpkg CMake thing

We also want to support:

4. cross-compiling from linux to any platform (e.g., Julia)
5. someone on Mac and using a BLAS other than accelerate
6. someone on linux and using MKL?

The last three can be supported by someone explicitly setting the BLAS library, like I'm doing for Julia in this PR?

And 1. and 2. and 3. are supported by `find_package(BLAS, required)`?

galabovaa:
Sure, find_package(BLAS) is probably enough. Will leave the other code as optional, for now. 

It would be great if you could pull the updates from hipo-dev and double check that the julia side is OK

odow:
CI runs look okay to me

PR #2551, opened 2025-10-01
Simplify test-python CI workflow

odow:
Opening this PR for @galabovaa to take a look at.

The current .github/workflows files are very verbose. We can use build strategies instead of separate files for every possible job. Here's a simple one.

If there is interest, I'd propose that we do the others. 

Also: do we need to run jobs on every push and pull_request? We end up with stuff like https://github.com/ERGO-Code/HiGHS/pull/2518 where we're running 309 jobs!!! (Most of which are duplicates)

WIth this change, CI would run:

 * when you push to latest
 * when you push to master
 * when you open or update a pull request

CI would not run:

 * when you push to a branch that does not have an open PR

This might be a downside, but it does nudge people to working in public PRs, rather than feature branches.

If we wanted to, we could add `- "ci-*"` which would trigger CI if a branch was named `ci-foo`.

### DIFF
--- a/.github/workflows/test-fortran.yml
+++ b/.github/workflows/test-fortran.yml
@@ -26,6 +26,10 @@ jobs:
       - run: |
           cmake -E make_directory ${{runner.workspace}}/build
           cd ${{runner.workspace}}
-          cmake $GITHUB_WORKSPACE -DFORTRAN=ON
+          if [[ "${{ matrix.os }}" == "macos-15-intel" ]]; then
+            cmake $GITHUB_WORKSPACE -DFORTRAN=ON -DCMAKE_Fortran_COMPILER=gfortran-14
+          else
+            cmake $GITHUB_WORKSPACE -DFORTRAN=ON
+          fi
           cmake --build . --parallel
           ./bin/fortrantest


.github/workflows/cmake-cpp.yml
```diff
@@ -0,0 +1,34 @@
...
+jobs:
+  test:
+    name: ${{ matrix.os }} CMAKE_BUILD_TYPE=${{ matrix.build_type }} ALL_TESTS=${{ matrix.all_tests }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ['ubuntu-latest', 'macos-latest', 'windows-latest']
+        build_type: ['Release', 'Debug']
+        all_tests: ['OFF', 'ON']
```
odow, line 12: @galabovaa is it necessary to test both `ALL_TESTS` ON and OFF?

.github/workflows/test-fortran.yml
```diff
@@ -0,0 +1,37 @@
...
+        include:
+          - os: ubuntu-latest
+            compiler: gcc
+            version: 12
+          - os: ubuntu-latest
+            compiler: gcc
+            version: 13
+          - os: macos-13
+            compiler: gcc
+            version: 11
```
odow: Is it necessary to test this mix of compilers? Why 11 on Mac but 12 and 13 on linux?

.github/workflows/test-python.yml
```diff
@@ -0,0 +1,36 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: macos-latest
+            python: 3.11
+          - os: ubuntu-latest
+            python: 3.11
+          - os: windows-2022
+            python: 3.12
```
odow, line 16: Why 3.12 on Windows but 3.11 on Mac and linux?

.github/workflows/cmake-cpp.yml
```diff
@@ -0,0 +1,34 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - shell: bash
+        run: |
+          cmake -E make_directory ${{runner.workspace}}/build
```
odow:
```suggestion
          cmake -E make_directory "$GITHUB_WORKSPACE/build"
```

.github/workflows/cmake-cpp.yml
```diff
@@ -0,0 +1,34 @@
...
+          cd ${{runner.workspace}}/build
```
odow:
```suggestion
          cd build
```

odow:
So @galabovaa, thoughts on simplifying some of the actions into a single file and using the build matrix?

galabovaa:
As we discussed on our last Tuesday meeting, we will stick to development branches. 

Build matrix yes, single file not for a while

PR #2552, opened 2025-10-01
Similar code in dual fixing and dominated columns reductions

fwesselm:
- This PR adds some utilities to the presolve code to remove similar / duplicate code from `HPresolve::dominatedColumns` and `HPresolve::dualFixing`.
- These changes are needed to make additional improvements to `HPresolve::dominatedColumns`.
- I ran 850 MIPs and the results look reasonable: the ratios of shifted geometric means are 0.99 for solution times and 0.97 for numbers of branching nodes.

### DIFF
--- a/highs/presolve/HPresolve.cpp
+++ b/highs/presolve/HPresolve.cpp
@@ -4626,14 +4626,14 @@ HPresolve::Result HPresolve::dualFixing(HighsPostsolveStack& postsolve_stack,
         double residual = 0.0;
         if (direction * triplet.jval < 0) {
           rhs = model->row_upper_[triplet.row];
-          residual = impliedRowBounds.getResidualSumLowerOrig(triplet.row, col,
-                                                              triplet.jval) +
-                     std::max(triplet.kval, 0.0);
+          residual = impliedRowBounds.getResidualSumLowerOrig(
+              triplet.row, col, triplet.jval, rowNz.index(), triplet.kval,
+              true);
         } else {
           rhs = model->row_lower_[triplet.row];
-          residual = impliedRowBounds.getResidualSumUpperOrig(triplet.row, col,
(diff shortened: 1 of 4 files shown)


highs/presolve/HPresolve.cpp
```diff
@@ -4594,57 +4545,13 @@ HPresolve::Result HPresolve::dualFixing(HighsPostsolveStack& postsolve_stack,
...
-        bestBound = std::max(bestBound, candidateBound);
-      }
-
-      // round bound
-      if (model->integrality_[col] != HighsVarType::kContinuous)
-        bestBound = std::ceil(bestBound - primal_feastol);
+      // upper bound.
+      double bestBound =
+          direction > 0
+              ? computeImpliedLowerBound(col, rowNz.index(),
```
Opt-Mucca, line 4551: Just to make sure: Is this supposed to be `col_upper` instead of `col_lower`?

Opt-Mucca (approved):
Well done on keeping track of everything..... I'm not sure I find the restructure of `getResidualSum{Upper,Lower}{Orig}` easier to read, but it's definitely more compact and seems to be performance improving. I guess the big advantage is that it offloads the complexity from `HPresolve` and makes it easier to expand / reuse later.
Everything looks good to me! (I've raised one question, which I believe isn't an error, but I'm not 100% sure).

highs/presolve/HPresolve.cpp
fwesselm: We would like to compute an implied bound for column `col` given that another (binary) variable `rowNz.index()` is at its upper bound. Therefore, `col_upper_` should be used in both cases here.

fwesselm:
I am not sure why, but `windows_debug64_fb_off` seems to be stuck `Looking for a Fortran compiler`.

PR #2553, opened 2025-10-01
Tests: fix bazel tests on Windows

TendTo:
For some reason I did not think of this before. I was told the tests failed because of lack of access to the MPS files, and the reason is that, by default, bazel tests running on Windows do not have access to data files, and must be [enabled explicitly](https://bazel.build/configure/windows#symlink). It's a easy fix.


galabovaa:
Great, thank you!

PR #2554, opened 2025-10-02
Allow CUDA build with bazel

TendTo:
Add the option to build Highs (more precisely cupdlp) with CUDA using Bazel as the build system.
For the CUDA build to be selected, the following 3 must be true:

- a CUDA toolchain is detected on the system
- the flag `@rules_cuda//cuda:enable=False` is NOT used (true by default)
- the flag `--//:cupdlp_gpu` is used

I hadn't the chance to test it on many systems with different configurations, but the little testing I did seem to confirm that this small change works as expected. 
Let me know if there are any issues with it.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -13,6 +13,7 @@ copy_file(
 bool_flag(
     name = "cupdlp_gpu",
     build_setting_default = False,
+    visibility = ["//visibility:public"],
 )
 
 config_setting(


galabovaa:
My local tests passed as well! Thank you, @TendTo, for this small but important update!

galabovaa:
@TendTo this code breaks the highs update in bazel-central-registry. I will have to modify it somehow, so 

```
bazel_dep(
    name = "rules_cuda",
    version = "0.2.5",
)
```
is only optional in MODULE.bazel or moved to a separate file.

Could you tell me a little bit more about how you are using HiGHS with bazel in your project?

TendTo:
Interesting, could you share some more details on how the dependency breaks the update? I would assume that the dependency (rules_cuda) is only pulled if you actually need it, and it should not impact anything else, but there is clearly some flaws in my assumption.

I was planning on using Highs mostly as a configurable LP solver, since it supports Simplex, IPM and pdlp. Moreover, being open source, it would not require the user to have a licence and I was also considering statically linking it if needed.

galabovaa:
Actually, I think it's fine. The failure was my mistake, I needed to update one more file in bcr. I will test this a bit more locally with and without NVvidia, but I think it will go OK. 

I was curious if you are just using bazel to build the highs library or binary or if your parent project is also bazel-based. So far, the bazel build has been used mostly for OR-Tools, at least as far as we know. It is great to hear that you have decided to try HiGHS and thank you again for your contribution!

Issue #2534, opened 2025-09-13
Ensure log file is closed when running executable


jajhall:
It's been reported that "the log file is not closed when HiGHS times out". This is only an issue when running the executable, as the log file is never closed when running HiGHS via the C++ API. That said, perhaps there could be a `Highs::closeLogFile()` method.

jajhall:
Closed by #2558

PR #2558, opened 2025-10-04
Added `Highs::closeLogFile()`, and calling this before returning from the command line executable

jajhall:
This ensures that the log file is closed, but surely that happens in the destructor of the `Highs` class. 

Hence I'm puzzled by the issue raised in #2534 

Still `Highs::closeLogFile()` may be useful to someone, and now there's an explicit flush of `Highs::options_.log_options.log_stream` in the destructor.


galabovaa (approved):
Looks good to me!

Issue #2557, opened 2025-10-04
Add method to solve the LP corresponding to discrete variables fixed at the optimal solution of a MIP


jajhall:
In other words, add the equivalent of [GRBfixmodel](https://docs.gurobi.com/projects/optimizer/en/current/reference/c/solving.html#c.GRBfixmodel) to HiGHS. Needs a decision on whether to try to hot start from the MIP solution, or whether to use the LP presolve, which may reduce the LP that has to be solved "from scratch" by simplex to such an extent that it's faster than hot starting from a solution.

Motivated by #2556 , but something that was planned when discovering that some folk want dual information and/or a basis at the solution of a MIP.

peno64:
Note that semi-continious variables are a special case. For integer/binary variables, the fixed model is fixing the integer/variable value to its obtained (integer) value and remove the integer attribute.
However for semi-cont variables you have to look at the two possible states the variable can obtain. If its value becomes 0, then also a min/max of 0 must be put on this variable (ie fixed to 0). However if the value is non-zero and thus >= minimum, the variable may not be fixed on its value but instead the min/max on this variable must be set. Otherwise wrong dual information can be returned.

peno64:
Except if the variable is semi-cont *and* integer in which I guess it must be treated as integer and this fixed on its obtained value.

jajhall:
Indeed, this is what I've done with semi-continous and semi-integer variables

jajhall:
Closed by #2559

PR #2559, opened 2025-10-04
Added `Highs::getFixedLp(HighsLp& lp)`

jajhall:
As with Gurobi, only generates the fixed LP. Users will have to pass the LP to HiGHS (losing the corresponding MIP) and then solve it.


jajhall:
Now to add C API

jajhall:
This is now ready for review @galabovaa

jajhall:
Closes #2557

galabovaa:
Looks good to me! Just one conflict but it seems that we should just accept both tests

galabovaa (approved):

PR #2561, opened 2025-10-07
Update README.md

Somesh-Kennedy:
Add link to CONTRIBUTING.md for Meson build methods.


jajhall:
Since it's only a documentation change, I've left this as a PR to `master` [@Somesh-Kennedy, we normally only accept PRs to `latest`.]

If @galabovaa is happy, then we can merge this without running the CI tests

PR #2575, opened 2025-10-16
Predictive bound analysis

fwesselm:
- This PR adds predictive bound analysis to `HPresolve::dominatedColumns` as described by Gamrath et al., Progress in presolving for mixed integer programming. Math. Prog. Comp. 7, 367–398 (2015).
- Several test runs on 850+ MIPs showed that predictive bound analysis only yields a small performance improvement: the average running times and numbers of branching nodes decrease by 1-2 percent (in terms of shifted geometric means).
- Side note: While testing I realized that, irrespective of these changes, HiGHS does sometimes not find the reported optimal solution for instance `map18` from `MIPLIB2010`. I plan to debug this in due course.

### DIFF
--- a/highs/presolve/HPresolve.cpp
+++ b/highs/presolve/HPresolve.cpp
@@ -1306,6 +1306,9 @@ HPresolve::Result HPresolve::dominatedColumns(
       bool tryToStrengthenBounds =
           (isDominatingBoundFinite || isDominatedBoundFinite) && sameVarType;
       if (tryToFix || tryToStrengthenBounds) {
+        // increment counter for number of domination checks due to predictive
+        // bound analysis
+        if (!tryToFix) numDomChecksPredBndAnalysis++;
         // check for domination
         if (checkDomination(direction, col, direction_k, k)) {
           if (tryToFix &&
@@ -1335,9 +1338,6 @@ HPresolve::Result HPresolve::dominatedColumns(
                                 dominatingBound, direction * direction_k));
           }
         }
-        // increment counter for number of domination checks due to predictive
-        // bound analysis
-        if (!tryToFix) numDomChecksPredBndAnalysis++;
       }
       return Result::kOk;
     };


highs/presolve/HPresolve.cpp
```diff
@@ -1083,9 +1089,12 @@ HPresolve::Result HPresolve::dominatedColumns(
 
   auto checkDomination = [&](HighsInt scalj, HighsInt j, HighsInt scalk,
                              HighsInt k) {
+    // increment counter
+    numDomChecks++;
+
     // rule out domination from integers to continuous variables
     if (model->integrality_[j] == HighsVarType::kInteger &&
-        model->integrality_[k] != HighsVarType::kInteger)
+        model->integrality_[k] == HighsVarType::kContinuous)
```
Opt-Mucca, line 1097: This now includes implicit integers then?

highs/presolve/HPresolve.cpp
```diff
@@ -1146,232 +1155,254 @@ HPresolve::Result HPresolve::dominatedColumns(
...
+                                dominatedBound, direction * direction_k));
+            if (!colDeleted[col] && isDominatingBoundFinite)
+              HPRESOLVE_CHECKED_CALL(
+                  tightenBounds(k, dominatedBound, direction_k < 0, col,
+                                dominatingBound, direction * direction_k));
+          }
+        }
+        // increment counter for number of domination checks due to predictive
+        // bound analysis
+        if (!tryToFix) numDomChecksPredBndAnalysis++;
```
Opt-Mucca: Should this counter be moved a few lines up?

highs/presolve/HPresolve.cpp
```diff
@@ -5148,6 +5152,17 @@ void HPresolve::computeColBounds(HighsInt col, HighsInt boundCol,
   std::vector<nonZeros> nzs;
   nzs.reserve(colsize[col]);
 
+  auto storeTriplet = [&](HighsInt row, double jval, double kval) {
+    // are coefficients required to have the same or opposite signs?
+    // otherColCoeffPattern =  0: accept any coefficient pair
+    // otherColCoeffPattern =  1: accept coefficients with same signs
+    // otherColCoeffPattern = -1: accept coefficients with opposite signs
+    if (boundColCoeffPattern != 0 &&
+        std::signbit(jval) != std::signbit(boundColCoeffPattern * kval))
```
Opt-Mucca, line 5161: I'm not following the logic for accepting the triplet. Can you shortly explain?

highs/presolve/HPresolve.cpp
fwesselm: Yes, I thought that treating implicit integer variables as integer variables here would make sense.

highs/presolve/HPresolve.cpp
fwesselm: Well spotted. Since `HPRESOLVE_CHECKED_CALL` may return from the lambda, I moved this code segment to the beginning of the `if (tryToFix || tryToStrengthenBounds)` block.

highs/presolve/HPresolve.cpp
fwesselm: Sure. The predictive bound functions defined by Gamrath et al. (see Definition 3 in the paper) require the coefficients to have the same signs, i.e. either both positive or both negative. 

The HiGHS code also checks whether scaling the columns by (-1) results in a dominance relationship (e.g. -x_j dominates x_k). In this case, the coefficients in the predictive bound functions should have opposite signs.

PR #2576, opened 2025-10-16
Minor correction to nqueens.py

BenChampion:
Preparing to use this for a demo, I noticed a minor inconsistency.

The `axis` kwarg specifies the axis over which to sum, with 0 being rows and 1 being columns. But to get row totals you have to sum over columns, and vice versa.

Although the symmetry in this problem means it makes no difference, I thought it'd be better to avoid confusion.

Tagging @mathgeekcoder since I can't add you as a reviewer!

### DIFF
--- a/examples/nqueens.py
+++ b/examples/nqueens.py
@@ -13,8 +13,8 @@ h.silent()
 
 x = h.addBinaries(N, N)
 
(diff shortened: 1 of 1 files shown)


mathgeekcoder:
Great, thanks!  Yeah, that was totally my mistake for those comments...  good spotting.

As very minor feedback on the fix, I would probably have changed the comment rather than the axis.

Issue #2579, opened 2025-10-17
highs_c_api.h defines globals, leading to multiple definition across TUs


mwelt:
Hi, and thanks for HiGHS!

Including `<highs/interfaces/highs_c_api.h>` in multiple translation units triggers linker errors because the header defines global objects (not just declares them), e.g.:

``` c
const HighsInt kHighsMaximumStringLength = 512;
const HighsInt kHighsStatusOk = 0;
/* ... */
```

**Minimal repro:**

``` c
/* a.c */ #include <highs/interfaces/highs_c_api.h> int main(void){return 0;}
/* b.c */ #include <highs/interfaces/highs_c_api.h> int f(void){return 0;}
/* build */ gcc -I/usr/local/include a.c b.c -L/usr/local/lib -lhighs
/* multiple definition of `kHighs...` */
```

**Proposed fix (header-only, stb-style):**

- Default: only extern declarations.
- If `HIGHS_C_API_DEFINE_GLOBALS` is defined before including the header in exactly one TU, emit the definitions.

*Sketch:*

``` c
/* highs_c_api.h */

#ifndef HIGHS_C_API_H
#define HIGHS_C_API_H
/* ... existing includes ... */

/* --- Declarations (visible everywhere) --- */
extern const HighsInt kHighsMaximumStringLength;
extern const HighsInt kHighsStatusOk;
/* ... all other kHighs* ... */

/* --- Optional one-TU definitions --- */
#ifdef HIGHS_C_API_DEFINE_GLOBALS
  const HighsInt kHighsMaximumStringLength = 512;
  const HighsInt kHighsStatusOk = 0;
  /* ... all other kHighs* with initializers ... */
#endif

/* ... rest of API ... */
#endif /* HIGHS_C_API_H */
```

**Usage:**

- Library build (or any single TU in an app):

``` c
#define HIGHS_C_API_DEFINE_GLOBALS
#include <highs/interfaces/highs_c_api.h>
```

- All other TUs: just `#include <highs/interfaces/highs_c_api.h>` (no macro).

I’m happy to open a PR with this change if you agree.

Env: HiGHS v1.11.0, gcc 13.3.0, Ubuntu/Kubuntu 24.04.1, installed to /usr/local.

Thanks!

-- Michael

jajhall:
My only immediate problem with this is that unless

HIGHS_C_API_DEFINE_GLOBALS

[...]

PR #2582, opened 2025-10-20
Make C API constants static const to prevent multiple definition linker errors

mwelt:
**Description:**

This PR fixes linker errors in HiGHS C-API [#2579](https://github.com/ERGO-Code/HiGHS/issues/2579) because the header defined non-static global constants.

This patch marks all constants in the C API header as `static const`,
which ensures internal linkage and eliminates the issue without affecting ABI or API behavior.

**I verified the fix by:**

Building HiGHS locally (cmake -S . -B build && cmake --build build)
Running ctest --output-on-failure.

### DIFF
--- a/highs/interfaces/highs_c_api.h
+++ b/highs/interfaces/highs_c_api.h
@@ -23,115 +23,115 @@
 //
 #include "lp_data/HighsCallbackStruct.h"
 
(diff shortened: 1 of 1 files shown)


jajhall:
Closes #2579

PR #2583, opened 2025-10-21
Hipo-dev into latest

galabovaa:


docs/src/options/definitions.md
filikat, line 1: I'm not sure why pdlp_native_termination option is added back here. It could be that I had an old version of the documentation. Should it be there?

jajhall: I'll sort out the options documentation

readme-hipo-deps.md
filikat, line 1: These instructions should be updated now, with the new options for Metis and with instructions for Windows as well maybe.

galabovaa: I've got these already in hipo-tt which I'll merge afterwards in another PR

galabovaa: I see there are conflicts so I may as well merge it now

fwesselm:
> Clearly I can't comment on all the build changes, or the HiPO source, so there's not much left beyond what I wrote to accommodate HiPO in options, Highs::optimizeModel and unit tests.
> 
> Has the (default) use of HiPO for the analytic centre calculation been tested by @fwesselm? I'm not sure I've tested it on large numbers of MIPs - just the energy MIPs. At this stage pre-release, I think it's safer to keep IPX as the default

No, I have not tested this. Let me know and I can try this.

jajhall:
> No, I have not tested this. Let me know and I can try this.

Thanks. For now the MIP solver won't use HiPO by default. There's no reason why it should fail - and it's backed up by a call to IPX, but using HiPO will lead to a random change in MIP solver behaviour

PR #2584, opened 2025-10-21
Reintroduce getPresolveOrigRowsIndex

fwesselm:
Method `getPresolveOrigRowsIndex` was recently deleted from `Highs.h`. I am not sure if it was removed intentionally, but it would be great if it could be added again.

### DIFF
--- a/highs/Highs.h
+++ b/highs/Highs.h
@@ -1273,14 +1273,14 @@ class Highs {
   }
 
   /**
-   * @Brief Put a copy of the current iterate - basis; invertible
+   * @brief Put a copy of the current iterate - basis; invertible
    * representation and dual edge weights - into storage within
    * HSimplexNla. Advanced method: for HiGHS MIP solver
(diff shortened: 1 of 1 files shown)


galabovaa:
Sure, thank you for spotting this @fwesselm!

BenChampion:
[This](https://github.com/ERGO-Code/HiGHS/commit/e2d44e907f3821ef4d24cab6113f05296fdc8f4e#diff-ff9f93943d87372658ed39dde87f44b979191f3f2a099f2134ce76d5f2c7829e) commit removed it, FYI @jajhall 

(See `git log -S getPresolveOrigRowsIndex highs/Highs.h`)

jajhall:
I must have cut and not pasted twice when creating the header for ` Highs::getFixedLp(HighsLp& lp) const;`, sorry

In an ideal world, we'd have a unit test that uses it

PR #2587, opened 2025-10-22
Added `HighsSolution::print` with prefix for grep, and added prefix to `HighsBasis::print`

jajhall:
Handy development utility that goes along with an enhanced version of the basis printing utility


jajhall:
Only a dev utility so never mind code coverage

PR #2596, opened 2025-10-23
Merging fix-2460 into latest

jajhall:
Merging fix-2460 into latest with Git yielded incomprehensible merge conflicts, so I created latest-2460 from latest and merged ix-2460 into it by hand


galabovaa (approved):
Great!

PR #2598, opened 2025-10-24
Metis interface back to normal

filikat:
Accomodate small change to Metis 510-ts interface.


filikat:
@galabovaa Not sure if the failing test is an issue. Could you check please?

galabovaa:
Probably unrelated! HiPO is not included in the nuget yet anyway

Let's see if it fails consistently

PR #2600, opened 2025-10-24
Documentation updates

jajhall:
Updated two minor dev logging cases, updated `gup.md` and `kkt.md`, and added `numerics.md`

Also updated two minor dev logging cases


jajhall:
@galabovaa: I should add that I've no way of previewing changes to the documentation - julia make.jl fails locally - so the documenter CI is the only way I can see whether I've broken anything, and pushing to latest is the only way I can preview it.

galabovaa:
what error do you get when you try to build locally? I may be able to help

jajhall:
> what error do you get when you try to build locally? I may be able to help

I'd mis-formed a link @id so the @ref didn't work. OK now!

galabovaa:
Great!

Issue #1590, opened 2024-01-24
Return solution object properties as numpy arrays.


timellemeet:
When using the python interface, calling `getSolution()` gives an object with native python lists instead of numpy arrays:
- col_value
- col_dual
- row_value
- row_dual

I expect most users will after convert these to numpy arrays for further processing. In my use cases, column generation, this actually results in a small but noticeable performance decrease.

Considering during model construction we also pass numpy arrays, I think it would be both more consistent and desired to also get numpy arrays from these outputs. 

Not sure if implementing it in highspy you can actually do it in a more efficient way then just doing np.array() though.

jajhall:
In short, I don't know whether we can return anything other than native python lists when using `pybind11` to create a Python interface. We're not `pybind11` experts.

We're talking to the person who created the prototype of `highspy` tomorrow, so I'll flag it up

michaelbynum:
This could be done on the c++ side by creating a new solution object for the python interface, but it would still require copying the std::vectors into arrays. That is probably faster than creating a python list and then converting it to a numpy array, but I'd be surprised if it was significantly faster. I can give it a go (probably in a couple of weeks).

timellemeet:
Thank you! Even if it's not noticeably faster I would say it is a usability improvement.

PR #2603, opened 2025-10-25
Enhance getSolution(): convert lists to NumPy arrays

ParthPore10:
This PR updates the Python interface of getSolution() to return NumPy arrays instead of native Python lists for the following attributes:

1. col_value
2. col_dual
3. row_value
4. row_dual

Returning NumPy arrays improves numerical performance and consistency, since most users already work with NumPy-based data when constructing models or performing post-processing. This change avoids redundant np.array() conversions in downstream applications and aligns input and output data types.

### **Implementation Details** 

Overrides getSolution() to convert solution lists into NumPy float arrays using _as_float_array().
Ensures compatibility by preserving the original structure of the solution object.

**References**
Fixes [#1590](https://github.com/ERGO-Code/HiGHS/issues/1590)

### DIFF
--- a/.github/workflows/build-bazel.yml
+++ b/.github/workflows/build-bazel.yml
@@ -1,6 +1,11 @@
 name: build-bazel
 
-on: [push, pull_request]
+on:
+  push :
+    branches: [ master, main, Parthpore10_1590issue_fix ]
+  pull_request:
+    branches: [master,main]
+  workflow_dispatch: 
 
 jobs:
   bazel:


jajhall:
What do you reckon @mathgeekcoder ?

mathgeekcoder:
@jajhall: I wouldn't accept this change.  I have thought about returning these as numpy arrays (instead of lists), but this is not the best way to do this - it will cause an unnecessary copy and breaks the typing information.

@ParthPore10: The code isn't bad, but could be better.  Per the documentation, `np.asarray(a)` will copy the memory (unless `a` is already a numpy array of same type). Your check inside `float_array` is basically redundant, and since we have a list, it will perform a copy.

That said, if we modify `highs_bindings.cpp` similar to what I did for `HighsLp::col_cost_`, e.g.,:
```cpp
  py::class_<HighsLp>(m, "HighsLp", py::module_local())
      .def(py::init<>())
      .def_property("col_cost_", make_readonly_ptr(&HighsLp::col_cost_),
                    make_setter_ptr(&HighsLp::col_cost_))
```

This should avoid unnecessary copies and return a numpy array on the python side.  If this is changed, the corresponding types in `_core/__init__.pyi` needs to be updated too, e.g.,:

```python
class HighsSolution:
    col_dual: list[float]  # list should be changed to numpy.ndarray[typing.Any, numpy.dtype[numpy.float64]]
    col_value: list[float]
    dual_valid: bool
    row_dual: list[float]
    row_value: list[float]
    value_valid: bool

    def __init__(self) -> None: ...
```

I'm hesitant to use numpy arrays everywhere (in some cases lists are actually faster), but for this it should be fine.

BTW: if you use the `highs.vals` function, it will already return a numpy array (or dictionary if appropriate).  Although it requires a copy by design.

jajhall:
> @jajhall: I wouldn't accept this change. I have thought about returning these as numpy arrays (instead of lists), but this is not the best way to do this - it will cause an unnecessary copy and breaks the typing information.

Thanks, I guess that returning these as numpy arrays (instead of lists) means that we avoid users having to avoid [this issue](https://ergo-code.github.io/HiGHS/stable/interfaces/python/#Extracting-values-efficiently)?

mathgeekcoder:
> Thanks, I guess that returning these as numpy arrays (instead of lists) means that we avoid users having to avoid [this issue](https://ergo-code.github.io/HiGHS/stable/interfaces/python/#Extracting-values-efficiently)?

Great question!  I didn't think it would help - but I gave it a quick test, and you are correct - that issue doesn't seem to occur with a numpy array from pybind (using my suggested change above).  e.g.,:
```python
lp = h.getLp()

# fast (uses numpy)
value = [lp.col_cost_[icol] for icol in range(num_vars)]

# slow (uses list)
value = [lp.col_lower_[icol] for icol in range(num_vars)]
```

mathgeekcoder:
> @ParthPore10 Please edit the PR as suggested by @mathgeekcoder

@ParthPore10: Since this is a larger change, and modifying the C++ code - please let me know if you'd prefer me to make these changes, or if you'd like to make an attempt.

ParthPore10:
Hii @mathgeekcoder , @jajhall  tthanks for pointing that out. I’ll review the problem and take another crack at it.

mathgeekcoder:
@ParthPore10: there's something odd about your PR.  The code-diff looks like it's comparing to an older branch, e.g., it doesn't recognize the changes for `col_cost_` have already been made.

Also, BTW: I wouldn't change the bazel build; there's already a setter function (so you should use that instead of writing your own); and you didn't actually change the getter/setter for `HighsSolution` in `highs_bindings.cpp`.

jajhall:
No progress, but happy to reopen if @ParthPore10 addresses concerns of @mathgeekcoder

Issue #2141, opened 2025-01-18
HighsHashTree.h:855:32: runtime error


FlorianSchwendinger:
I get the 
```
/tmp/RtmpW6hzfN/R.INSTALL1defe9d24e2/highs/inst/HiGHS/src/util/HighsHashTree.h:855:32: runtime error: index 1 out of bounds for type 'NodePtr [1]'
    #0 0x7fb9218ae7f9 in HighsHashTree<int, int>::insert_recurse(HighsHashTree<int, int>::NodePtr*, unsigned long, int, HighsHashTableEntry<int, int>&) /tmp/RtmpW6hzfN/R.INSTALL1defe9d24e2/highs/inst/HiGHS/src/util/HighsHashTree.h:855
    #1 0x7fb921771f25 in bool HighsHashTree<int, int>::insert<int&, int&>(int&, int&) /tmp/RtmpW6hzfN/R.INSTALL1defe9d24e2/highs/inst/HiGHS/src/util/HighsHashTree.h:1371
    #2 0x7fb921771f25 in HighsCliqueTable::doAddClique(HighsCliqueTable::CliqueVar const*, int, bool, int) /tmp/RtmpW6hzfN/R.INSTALL1defe9d24e2/highs/inst/HiGHS/src/mip/HighsCliqueTable.cpp:374
    #3 0x7fb9217c0805 in HighsCliqueTable::addClique(HighsMipSolver const&, HighsCliqueTable::CliqueVar*, int, bool, int) /tmp/RtmpW6hzfN/R.INSTALL1defe9d24e2/highs/inst/HiGHS/src/mip/HighsCliqueTable.cpp:810
    #4 0x7fb9217edca2 in HighsCliqueTable::extractObjCliques(HighsMipSolver&) /tmp/RtmpW6hzfN/R.INSTALL1defe9d24e2/highs/inst/HiGHS/src/mip/HighsCliqueTable.cpp:1508
    #5 0x7fb920719977 in HighsMipSolverData::addIncumbent(std::vector<double, std::allocator<double> > const&, double, int, bool) /tmp/RtmpW6hzfN/R.INSTALL1defe9d24e2/highs/inst/HiGHS/src/mip/HighsMipSolverData.cpp:1467
[...]

PR #2606, opened 2025-10-25
Hashtree uses trailing storage instead of array member hack

ben-schwen:
Follow up to #2141

This PR refactors the `BranchNode` to use a trailing storage instead of C-style flexible array member idiom, eliminating undefined behavior and passing UBSAN checks

The previous implementation used a flexible array member hack where:
* `sizeof(BranchNode)` was assumed to include one `NodePtr`
* size calculations used `numChilds - 1` to account for this
* direct array indexing (`child[i]`) accessed memory beyond the declared array
* this triggers undefined behavior that UBSAN detects

Note the change from `offsetof(BranchNode, child)` to `sizeof(BranchNode)` because:
* `child` member no longer exists
* consistent with the `children()` accessor: `this + 1` = `this + sizeof(BranchNode)`

Other changes are most simple change of accessors


jajhall:
Thanks for this. I had to run clang formatting on it, but #2608 corresponds to your original PR

FlorianSchwendinger:
Thanks Benjamin!

PR #2608, opened 2025-10-26
Formatted 2606

jajhall:
Replaces #2606 using clang formatted code

### DIFF
--- a/highs/util/HighsHashTree.h
+++ b/highs/util/HighsHashTree.h
@@ -392,9 +392,7 @@ class HighsHashTree {
   struct BranchNode {
     Occupation occupation;
 
-    NodePtr* children() {
-      return reinterpret_cast<NodePtr*>(this + 1);
-    }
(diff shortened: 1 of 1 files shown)


fwesselm:
I think this looks reasonable. It would be nice to improve the code coverage. Maybe some of the long `switch` blocks could be simplified using templates?

ben-schwen:
I didn't dare to touch the other lines. 

What I noticed is that there are multiple unreachable lines (756, 762, 767, 772), since these are breaks after return.

Since you are using C++11, the code dedoubling from templates would also not be that much, since you will still need functor structs because return type deduction is not available.

fwesselm:
I am planning to do some testing with these changes and will report back.

fwesselm:
> I am planning to do some testing with these changes and will report back.

I tested on 400 MIPs and (as expected) these changes do not affect HiGHS' behavior at all.

PR #2613, opened 2025-10-28
fix some compiler warnings and removes a debug printf from 1.12.0

svigerske:
- removes a printf that is going to stdout, probably debug leftover
- initialize `below_lower` for compiler: avoids a maybe uninitialized warning
- make sure `data_` is "used" also if HIPO_TIMING_LEVEL is 0: avoids an unused private field warning
- add missing `#include` statements: avoids missing-declaration warnings and ensures `DataCollector` is defined when compiling `DenseFact.h`
- make local functions static: avoids missing-declaration warnings, and allows inlining
- bring back comment line that was probably removed by accident


jajhall:
Thanks @svigerske. These can merge to master, and we'll have to consider a patch release

PR #2616, opened 2025-10-29
Clarify HiPO option feedback when the solver is unavailable

NightBaRron1412:
## Summary
- hide the HiPO entry from the command-line solver help when the build lacks HiPO support
- emit dedicated error messages for the LP, MIP LP, and MIP IPM solver options when HiPO is requested but not compiled in


jajhall:
I did this originally, but felt that it was unnecessary. It's no problem to reinstate it if users find it confusing

NightBaRron1412:
Yeah, that makes sense! I actually got confused myself and spent a while checking if I’d misconfigured the build ... having the CLI indicate when HiPO isn’t compiled in would really help avoid that. Thanks for being open to bringing it back!

PR #2624, opened 2025-10-30
Hipo fixes

filikat:
Various fixes to HiPO:
- Fix some overflows when computing statistics of analyse phase.
- Free memory used for normal equations, if augmented system is preferred.
- Fix bug in supernode amalgamation.
- Printing of BLAS library moved to HiGHS header, so it is printed when using HiPO without logging.
- Recommend to install Metis branch 521-ts due to better oredring quality on many problems. Update workflows and documentation accordingly.
- Add option hipo_metis_no2hop to control option no2hop of Metis and add warning if the fill-in is large.


filikat:
@odow One of the Julia tests if failing. Do you have a C version of this test that I can look at?

odow:
I think it's a good thing. You now solve the problem at the root node, where previously it required branching. I'll update the Julia test.

filikat:
Oh, I see. 
I thought that the MIP solver still used IPX by default though @jajhall.
Or did I understand wrong?

odow:
Then it might be because of unrelated changes? Either way, I'll relax the test: https://github.com/jump-dev/HiGHS.jl/pull/306

jajhall:
> Oh, I see. I thought that the MIP solver still used IPX by default though @jajhall. Or did I understand wrong?

Was using IPX by default in two main cases where IPM is (might be) used in the MIP solver: analytic centre (LP solve without basis), but now using IPX by default in the other: when switching to IPM after simplex reaches iteration limit in `HighsLpRelaxation::run`. This use my have tweaked the run of @odow.

See #2626

PR #2625, opened 2025-10-30
Name callback data structs

jschueller:
Rationale: this allows to forward-declare callback data types in case a callback needs to be declared in a public header:

struct _HighsCallbackDataOut;
struct _HighsCallbackDataIn;
typedef _HighsCallbackDataOut HighsCallbackDataOut; typedef _HighsCallbackDataIn HighsCallbackDataIn;

### DIFF
--- a/highs/lp_data/HighsCallbackStruct.h
+++ b/highs/lp_data/HighsCallbackStruct.h
@@ -21,7 +21,7 @@ extern "C" {
  * Struct to handle callback output data
  *
  */
-typedef struct _HighsCallbackDataOut {
+typedef struct HighsCallbackDataOut {
   void* cbdata;  // cast of HighsCallbackOutput
   int log_type;  // cast of HighsLogType
   double running_time;
@@ -50,7 +50,7 @@ typedef struct _HighsCallbackDataOut {
 // Some external packages (e.g., jump) currently assume that the first 2 fields
 // of this struct are interrupt and solution. Rearranging the struct may be a
 // breaking change.
-typedef struct _HighsCallbackDataIn {
+typedef struct HighsCallbackDataIn {
   int user_interrupt;
   double* user_solution;
   void* cbdata;  // cast of HighsCallbackInput (for internal use)


galabovaa:
@jschueller Would

```
typedef struct HighsCallbackDataOut {
...
```
without
```
typedef _HighsCallbackDataOut HighsCallbackDataOut; 
```
allow forward-declaring as well? Could you give us an example of the code you need to build, if possible? If you define another HighsCallbackDataOut would that not be an issue?

jschueller:
no, that wouldnt work (error: conflicting declaration ‘typedef struct HighsCallbackDataOut HighsCallbackDataOut’)

here is what I'm trying to do in a nutshell:

Solver.hxx:
```
#include <iostream>

// predeclaration (public header, so we cannot use highs.h here)
struct _HighsCallbackDataOut;
struct _HighsCallbackDataIn;
typedef _HighsCallbackDataOut HighsCallbackDataOut;
typedef _HighsCallbackDataIn HighsCallbackDataIn;


class Solver
{
public:
  Solver() {};
  void run() {};

protected:
  // we need to declare this here because we want to access protected member of the Solver class
  static void Callback(const int callback_type,
                       const char* message,
                       const HighsCallbackDataOut* data_out,
                       HighsCallbackDataIn* data_in,
                       void* user_callback_data);
};
```

Solver.cxx:
```
#include "Solver.hxx"

// actual include of highs
#include <Highs.h>


void Solver::Callback(const int callback_type,
                       const char* message,
                       const HighsCallbackDataOut* data_out,
                       HighsCallbackDataIn* data_in,
                       void* user_callback_data) {}

```

galabovaa:
I tried this minimal example locally, and, with the HiGHS code from latest, simply declaring the struct compiled as well:

Solver.hxx
```
#include <iostream>

struct HighsCallbackDataOut;
struct HighsCallbackDataIn;

class Solver
{
public:
  Solver() {};
  void run() {};

protected:
  static void Callback(const int callback_type,
                       const char* message,
                       const HighsCallbackDataOut* data_out,
                       HighsCallbackDataIn* data_in,
                       void* user_callback_data);
};
```
and 
Solver.cxx
```
#include "Solver.hxx"

// actual include of highs
#include <Highs.h>


void Solver::Callback(const int callback_type,
                       const char* message,
                       const HighsCallbackDataOut* data_out,
                       HighsCallbackDataIn* data_in,
                       void* user_callback_data) {}
                       
```
This way you still include `Highs.h` only in your implementation file. Do you get a different error?

jschueller:
yes, I still get a conflicting declaration with your solution

I prepared a repo to replicate the issue, see the github actions logs:
https://github.com/jschueller/hello-highs.git

jschueller:
hello @galabovaa, any new thoughts on this ?

galabovaa:
I have not had a chance to investigate properly yet. Some of us were at a conference and we have some new developments to complete. We will likely accept this, I just need to try it out myself first

PR #2626, opened 2025-10-30
MIP IPX default

jajhall:
Was using IPX by default in two main cases where IPM is/might be used in the MIP solver, but now using IPX by default in the other: when switching to IPM after simplex reaches iteration limit in `HighsLpRelaxation::run`


filikat (commented):
Looks fine.
The failing test is weird. It's happened before, always on Windows, that the test completes immediately, but then hangs until it times out.
Is it the issue of hanging on Windows happening again?

jajhall:
Who can say

PR #2628, opened 2025-10-31
Add missing includes

fontanf:
Some includes of `#include <functional>` are not explicit. This makes the compilation fail for some compilers.


jajhall:
Thanks. 

For what compilers does this fail?

jajhall:
Had to move #include <functional> to satisfy clang-format

fontanf:
icl (I'm not sure of which version)

PR #2629, opened 2025-10-31
Fix min and max for compilation on Windows

fontanf:
When compiling on Windows with icl, I also had issues with the min/max macros from `windows.h`.

I tried to define `NOMINMAX` but that didn't change anything.

The only way I managed to make the compilation work is by adding parentheses

Feel free to keep these changes or not.

### DIFF
--- a/highs/lp_data/HConst.h
+++ b/highs/lp_data/HConst.h
@@ -20,9 +20,9 @@
 const std::string kHighsCopyrightStatement =
     "Copyright (c) 2025 HiGHS under MIT licence terms";
 
-const size_t kHighsSize_tInf = std::numeric_limits<size_t>::max();
-const HighsInt kHighsIInf = std::numeric_limits<HighsInt>::max();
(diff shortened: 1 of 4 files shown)


jajhall:
Very strange. I wonder what's special about

> max_percent_sum_clock_times = std::max(percent_sum_clock_times[i], max_percent_sum_clock_times);

as there are vast numbers of cases where HiGHS uses `std::max` or `std::min`

fontanf:
Could it be the only min/max in header files?

filikat (commented):
This is a known issue:
https://cplusplus.com/forum/general/13468/
https://stackoverflow.com/questions/11544073/how-do-i-deal-with-the-max-macro-in-windows-h-colliding-with-max-in-std

Have you tried #undef-ining min and max?

fontanf:
No, I haven't. I didn't find this workaround. If it works, would you prefer this solution over parentheses? It seems more risky to me since it has a more global impact

PR #2632, opened 2025-11-02
Better IPX logging

jajhall:
After HiPO switches  to IPX, the IPX logging is displayed

The IPX logging header is (re-)printed after "Constructing starting basis..." so that it appears after a switch from HiPO.

IPX timer is very much simpler, and includes an offset so that it gives the HiGHS timer values. Hence it can use the HiGHS time limit without a correction. Interval logging is now only every 60s.

IPX now logs times to 1dp and without the "s" character - like HiPO 

Simplex now logs times to 1dp but retains the "s" character, since it's not obviously a time

### DIFF
--- a/highs/ipm/ipx/timer.cc
+++ b/highs/ipm/ipx/timer.cc
@@ -7,26 +7,10 @@ Timer::Timer(const double offset)
     Reset(true);
 }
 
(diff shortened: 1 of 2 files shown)


filikat (commented):
All good

PR #2634, opened 2025-11-03
#2633 pass-by-ref to avoid expensive copying of matrix for every constraint…

### LINKED ISSUE #2633: Slow LP model writing - 200 seconds for 20 million non-zeros
LP model writing to file is taking more than 200 seconds to write a model of size:
10000 variables
5000 constraints
~20 million non-zeros

sk-surya:
Fixes #2633 
… while writing LP Model - gives 10x faster LP model writes for large models (20M nz)


jajhall (approved):
Well spotted

Issue #2635, opened 2025-11-04
Highs_getIis segfaults


odow:
Am I calling something wrong?

```Julia
julia> using HiGHS

julia> unsafe_string(Highs_version())
"1.12.0"

julia> highs = Highs_create()
Ptr{Nothing}(0x000000013f53b800)

julia> Highs_readModel(highs, "/Users/odow/git/jump-dev/MathOptIIS/test/small.lp.gz")
Running HiGHS 1.12.0 (git hash: 755a8e027): Copyright (c) 2025 HiGHS under MIT licence terms
0

julia> m, n = Highs_getNumCol(highs), Highs_getNumRow(highs)
(396, 936)

julia> iis_num_col, iis_num_row = Ref{Cint}(m), Ref{Cint}(n)
(Base.RefValue{Int32}(396), Base.RefValue{Int32}(936))

julia> col_index, col_bound, col_status = zeros(Cint, m), zeros(Cint, m), zeros(Cint, m)
(Int32[0, 0, 0, 0, 0, 0, 0, 0, 0, 0  …  0, 0, 0, 0, 0, 0, 0, 0, 0, 0], Int32[0, 0, 0, 0, 0, 0, 0, 0, 0, 0  …  0, 0, 0, 0, 0, 0, 0, 0, 0, 0], Int32[0, 0, 0, 0, 0, 0, 0, 0, 0, 0  …  0, 0, 0, 0, 0, 0, 0, 0, 0, 0])

julia> row_index, row_bound, row_status = zeros(Cint, n), zeros(Cint, n), zeros(Cint, n)
(Int32[0, 0, 0, 0, 0, 0, 0, 0, 0, 0  …  0, 0, 0, 0, 0, 0, 0, 0, 0, 0], Int32[0, 0, 0, 0, 0, 0, 0, 0, 0, 0  …  0, 0, 0, 0, 0, 0, 0, 0, 0, 0], Int32[0, 0, 0, 0, 0, 0, 0, 0, 0, 0  …  0, 0, 0, 0, 0, 0, 0, 0, 0, 0])

julia> Highs_getIis(
           highs,
           iis_num_col,
           iis_num_row,
           col_index,
           row_index,
           col_bound,
           row_bound,
           col_status,
           row_status,
       )

[9037] signal 11 (2): Segmentation fault: 11
in expression starting at REPL[18]:1
[...]

PR #2636, opened 2025-11-04
Fix 2635

jajhall:
Now setting `HighsIis::col_status_` and `HighsIis::row_status_` to appropriate values when LP is feasible


jajhall:
This will close #2635

galabovaa (approved):
Looks good to me, more weird timeouts on windows

PR #2638, opened 2025-11-05
Corrected IIS LP generation and checking when incumbent LP is stored rowwise

jajhall:
Corrected IIS LP generation and checking when incumbent LP is stored rowwise

Corrected some HiGHS typos elsewhere


galabovaa (approved):
Great!

PR #2642, opened 2025-11-10
Better refinement for HiPO

filikat:
- Apply iterative refinement on the full 6x6 system (more robust and faster for many problems)
- Small changes to the printing, to allow simpler collection of data from log files
- Add missing line breaks in the instructions to install Metis


jajhall (approved):
Excellent!

PR #2649, opened 2025-11-14
Fix time limit in HiPO

filikat:
Time limit was not passed properly to HiPO


filikat:
Suggesting to compile with 64-bit integers may not fix the issue, depending on which BLAS is linked. I need to fix something else before.

jajhall (approved):
Thanks

PR #2660, opened 2025-11-21
Refactor HighsRedcostFixing::addRootRedcost

fwesselm:
- This PR is a by-product of debugging and removes duplicate code from `HighsRedcostFixing::addRootRedcost()`.
- The new code behaves slightly different due to the fact that the lower bound check now uses a tolerance for both directions (i.e., `mipsolver.mipdata_->lower_bound + mipsolver.mipdata_->feastol`).
- I ran 850+ MIPs and the relative geometric means for solution times and numbers of branching nodes are 0.99 and 1.01 respectively, indicating that the impact on HiGHS' performance is very small.

### DIFF
--- a/highs/mip/HighsRedcostFixing.cpp
+++ b/highs/mip/HighsRedcostFixing.cpp
@@ -252,9 +252,10 @@ void HighsRedcostFixing::addRootRedcost(const HighsMipSolver& mipsolver,
              direction * lurkingBound <= direction * lastBound;
              lurkingBound += step) {
           double fracBound = lurkingBound - bound + shift;
-          double requiredCutoffBound =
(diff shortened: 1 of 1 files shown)


fwesselm:
> What do you mean by the second point? Isn't `lower_bound` already accounting for `feastol`? From what I understood you're now doubling the tolerance in your check.

Thanks for checking the code. I do not think `lower_bound` includes a tolerance. I was trying to explain that the former if-part did not use a tolerance when comparing `lower_bound` and cutoff bound while the else-part did.

Opt-Mucca:
I'm blind. All good! The `lower_bound` in `HighsPrimalHeuristics` is not `mipsolver->mipdata_`.......

PR #2665, opened 2025-11-25
Fix compiler warning

fwesselm:
Fix compiler warning in `HighsIis::indexStatusOk` issued by `VS 2022` on `win64`:

```
1>HighsIis.cpp
1>L:\HiGHS\myHiGHS\highs\lp_data\HighsIis.cpp(867,30): warning C4805: '!=': unsafe mix of type 'const _Ty' and type 'bool' in operation
1>        with
1>        [
1>            _Ty=HighsInt
1>        ]
1>L:\HiGHS\myHiGHS\highs\lp_data\HighsIis.cpp(874,30): warning C4805: '!=': unsafe mix of type 'const _Ty' and type 'bool' in operation
1>        with
1>        [
1>            _Ty=HighsInt
1>        ]
```


jajhall:
Thanks @BenChampion, this looks as if I originally tested against a bool, which I then removed by copying the bool definition

PR #2666, opened 2025-11-25
HiPO factorisation uses 64-bit integers

filikat:
- Update to the factorisation of HiPO so that it uses 64-bit integers where needed, regardless of the size of HighsInt. The matrix (NE or AS) is still computed using HighsInt, Metis runs using HighsInt; the only thing that uses int64_t is the factorisation, to keep track of the number of nonzero entries of the factor. This should get rid of the "Error in analyse phase" that was reported when the factorisation didn't fit in 32-bit integers.
- The declarations of BLAS functions use int32_t regardless of the size of HighsInt, since 32-bit BLAS should be linked. It is very unlikely that the arrays used by BLAS will overflow 32-bit integers. The analyse phase checks for this and returns an error in case.
- The analyse phase detects integer overflow while forming the matrix and reports an error. It also warns the user if a large amount of memory is required during the factorisation.
- Added a custom stack data structure to handle the Schur complements during the factorisation, when the tree is not parallelised. Updated the heuristic to choose the type of parallelisation accordingly. If the serial memory required is large (>20GB) and parallel is set to choose, tree parallelism is switched off to reduce the chance of running out of memory.
- Added checks to detect the size of integer used by BLAS and Metis at runtime. These are weird tricks that may not work on all CPUs. They are inspired by Jump's libblastrampoline autodetection. HiGHS now prints also the integer model used by BLAS (LP64 or ILP64) when printing the BLAS library used. HiPO now checks that Metis is using the same integer type as HiGHS before starting the factorisation.
- Fixed small bugs

### DIFF
--- a/highs/ipm/hipo/factorhighs/FormatHandler.h
+++ b/highs/ipm/hipo/factorhighs/FormatHandler.h
@@ -88,8 +88,6 @@ class FormatHandler {
   virtual void extremeEntries() {}
 };
 
-const Int64 extra_space_frontal = 1;
-
 }  // namespace hipo
 
 #endif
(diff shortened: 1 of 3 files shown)


galabovaa (approved):
Looks great!

PR #2669, opened 2025-11-27
Add utility for lower bound change

fwesselm:
- Add `HighsMipSolverData::updateLowerBound` to remove some duplicate code. 
- Testing confirmed that the change does not affect HiGHS' behavior at all (as expected).


highs/mip/HighsMipSolverData.cpp
```diff
@@ -2691,6 +2636,15 @@ void HighsMipSolverData::limitsToBounds(double& dual_bound,
   }
 }
 
+void HighsMipSolverData::updateLowerBound(double new_lower_bound) {
+  // Update lower bound
+  double prev_lower_bound = lower_bound;
+  lower_bound = new_lower_bound;
```
Opt-Mucca, line 2642: I guess this is not needed, but thoughts on changing this to `lower_bound = std::max(prev_lower_bound, new_lower_bound)`?

Edit: `max`

Opt-Mucca (approved):
I'm a huge fan of this PR! Great quality of life change

highs/mip/HighsMipSolverData.cpp
fwesselm: I have thought about something similar (checking against `upper_bound`), but was not sure if there would be a case where this could yield a behavior change somehow. I can have another look.

Opt-Mucca: I'm also not 100% sure if there's maybe a case where you purposely want to decrease* the lower bound....

Issue #2664, opened 2025-11-24
Inconsistent optimal solution for map10


Opt-Mucca:
While testing some changes I've been finding different objective values for `map10` (https://miplib.zib.de/instance_details_map10.html) Given this is a MIPLIB 2017 benchmark instance we can't blame the instance
On `c4012f564` (some recent `latest` commit)  a valid optimal solution [test.sol.txt](https://github.com/user-attachments/files/23725095/test.sol.txt) gets separated by `HighsTableauSeparator` when computing a lifted binary cover cut. I can't seem to find any numerical error that would cause this from the cut logic. The tableau row (aggregation) before separation is violated by some tiny amount `1e-17`, but that shouldn't necessarily cause the bug..... I'm putting it here to track the issue and because I've failed to fix it today.
Warning: It's a bit annoying to recreate because the size of the problem and wait time.

romz-pl:
I confirm the inconsistent optimal solution for problem `map10`.

Using the **HiGHS version 1.12**, and solving the non scaled problem [map10](https://miplib.zib.de/instance_details_map10.html) the objective value `-494.86468647` was obtained.  The reference value reported in MIPLIB is `-495`.

However, the output provided by **HiGHS version 1.12** contains the important WARNING messages:
> WARNING: Problem has some excessively large row bounds
> WARNING:    Consider scaling the    bounds by 1e-2, or setting the user_bound_scale option to -7

The full output listing is provided below:
```
Running HiGHS 1.12.0 (git hash: n/a): Copyright (c) 2025 HiGHS under MIT licence terms
Set option solver to "simplex"
Set option parallel to "off"
Set option time_limit to 1200
Set option log_file to "HiGHS.log"
Set option mip_rel_gap to 0
MIP map10 has 328818 rows; 164547 cols; 549920 nonzeros; 146 integer variables (146 binary)
Coefficient ranges:
  Matrix  [1e+00, 1e+07]
  Cost    [1e+00, 1e+00]
  Bound   [1e+00, 1e+00]
  RHS     [1e+00, 9e+07]
WARNING: Problem has some excessively large row bounds
WARNING:    Consider scaling the    bounds by 1e-2, or setting the user_bound_scale option to -7
Presolving model
71338 rows, 37549 cols, 151723 nonzeros  0s
42027 rows, 18750 cols, 90532 nonzeros  0s
29265 rows, 14387 cols, 65005 nonzeros  0s
[...]

PR #2671, opened 2025-11-28
Apply implications when performing a bound change

Opt-Mucca:
We gather implications, e.g., `x = 1 -> y <= 0`, during the solve process (mainly from probing). We never actually apply them when making the original bound change later though. They should all be globally valid, so this change shouldn't need any conditional checks to be correct (It did result in me finding the issue #2664) 
The positives and negatives of this change:
(+1) Can apply bound changes that can no longer be found from propagation, e.g., some cut is no longer propagated.
(+2) Can propagate more in a single round because more changes have been made
vs
(-1) The reason for the bound change will now default to the clique table and lose the original reason. This might cause some performance loss for conflicts.
(-2) Is propagating more going to be performance positive?

There's also a nice bit of text in the PR explaining why not all bound changes are applied. This should be tested w.r.t. performance before being merged (or not merged)

### DIFF
--- a/highs/mip/HighsImplications.cpp
+++ b/highs/mip/HighsImplications.cpp
@@ -831,22 +831,33 @@ void HighsImplications::cleanupVub(HighsInt col, HighsInt vubCol,
 }
 
 void HighsImplications::applyImplications(HighsDomain& domain,
-                                          const HighsInt col, const bool val) {
+                                          const HighsInt col,
+                                          const HighsInt val) {
   assert(domain.isFixed(col));
 
-  auto checkImplication = [&](HighsDomainChange& domchg) -> bool {
+  auto checkImplication = [&](const HighsDomainChange& domchg) -> bool {
     assert(!domain.infeasible());
     if (domain.isFixed(domchg.column)) return false;
+    const bool isint =
+        domain.variableType(domchg.column) != HighsVarType::kContinuous;
+    // Directly change bounds on all integer columns. Only change continuous
+    // columns that fix the column, as changing their domains risks
+    // suppressing further bound changes found in propagation, e.g.,
+    // change [0, 100] -> [0, 50], propagation could tighten to [0, 48], but
+    // such a tightening would not be applied due to min boundRange improvement.
     if (domchg.boundtype == HighsBoundType::kLower) {
-      if (domchg.boundval >= domain.col_upper_[domchg.column]) {
-        // TODO: Long term: Should there be an implication reason?
-        HighsDomainChange domchgcpy = domchg;
-        domain.changeBound(domchgcpy, HighsDomain::Reason::unspecified());
+      if ((!isint && domchg.boundval >
+                         domain.col_upper_[domchg.column] - domain.feastol()) ||
+          (isint && domchg.boundval >
+                        domain.col_lower_[domchg.column] + domain.feastol())) {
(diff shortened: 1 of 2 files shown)


highs/mip/HighsImplications.cpp
```diff
@@ -829,3 +829,44 @@ void HighsImplications::cleanupVub(HighsInt col, HighsInt vubCol,
...
+                         domain.col_lower_[domchg.column] + domain.feastol()) ||
+          (isint && domchg.boundval <
+                        domain.col_upper_[domchg.column] - domain.feastol())) {
+        domain.changeBound(domchg, HighsDomain::Reason::cliqueTable(col, val));
+      }
+    }
+    return domain.infeasible();
+  };
+
+  HighsInt loc = 2 * col + val;
```
fwesselm, line 866: Feel free to ignore this, but I see the calculation of "loc" in different places. I guess it's not possible to make this a mini method somewhere?

highs/mip/HighsImplications.cpp
```diff
@@ -829,3 +829,44 @@ void HighsImplications::cleanupVub(HighsInt col, HighsInt vubCol,
...
+
+void HighsImplications::applyImplications(HighsDomain& domain,
+                                          const HighsInt col,
+                                          const HighsInt val) {
+  assert(domain.isFixed(col));
+
+  auto checkImplication = [&](const HighsDomainChange& domchg) -> bool {
+    assert(!domain.infeasible());
+    if (domain.isFixed(domchg.column)) return false;
+    const bool isint =
```
fwesselm, line 841: Not important, but maybe having an "isInteger" method in HighsDomain would be good.

highs/mip/HighsImplications.cpp
Opt-Mucca: I guess this could be a `static` method attached to `HighsMipSolver`? I think the majority of uses of `2 * i + val` are in implications and the cliquetable (and the code that queries them). 
I'll put it on the to-do list.

highs/mip/HighsImplications.cpp
Opt-Mucca: That would make sense. Something like:

`bool isIntegral(HighsInt col) const { return variableType(col) != HighsVarType::kContinuous }`?

(I'd use integral as in other bits of the HiGHS differentiates between integer and integral (integer + implied integer))

fwesselm: Sure, I noticed that similar checks are already done in the `HighsDomain` class.

highs/mip/HighsImplications.cpp
fwesselm: No problem. Just noticed it.

highs/mip/HighsImplications.cpp
Opt-Mucca: Created a PR in #2680 (Covers nearly all cases)

highs/mip/HighsImplications.cpp
Opt-Mucca: Have put on the to-do list. Won't do this one now

Opt-Mucca:
Experiments found this be have a 2.2% performance improvement over all instances. Merging now.

BenChampion:
Nice!

PR #2673, opened 2025-12-01
Fix #2641: avoid circular references in highspy

### LINKED ISSUE #2641: Use weak references in highspy to allow garbage collection
As @mathgeekcoder [notes](https://github.com/ERGO-Code/HiGHS/issues/2080#issuecomment-3504983535) in #2080

>Instead of pointing directly to the relevant `highs` object, [`HighsCallback.highs`] probably should use `weakref.ref(highs)`.  There's also `HighspyArray.highs`, though only the callback has the cyclic dependency.

BenChampion:
As per #2641. With `weakref.proxy` (a more seamless way to use `weakref.ref`) we can avoid circular references that prevent garbage collection. (This doesn't seem to be causing any major issues but it also seems easy to fix!)

@mathgeekcoder


mathgeekcoder:
@BenChampion looks great!

PR #2674, opened 2025-12-01
Order path separator

Opt-Mucca:
This PR sorts rows via their `fractionalActivity` in `HighsPathSeparator` when considering which row to add to the aggregation.
High level overview: The separator tries to find a continuous column to aggregate out, and to do so finds an additional row that also contains that continuous column. Which row to use is non-obvious.

The old method:
- No sorting. Given the array of rows, each time you want to select a new row, scan through the array starting at a random point.

The new method:
- Sort rows via their fractional activities. Scan through them in order each time.

Pros:
- More informed ordering. Should prioritise rows that can generate good cMIR cuts.
- Randomly selecting a starting point to scan through the array already kind of enforced a priority. That is, `array[i-1]` is checked before `array[i]` in all cases aside from starting the scan at `i`.

Cons:
- Losing some inherent diversity that a random algorithm provides. It's now more likely that we'd aggregate similar rows each time, which could lead to many duplicate cuts.

This should only be merged after doing a performance run.

### DIFF
--- a/highs/mip/HighsPathSeparator.cpp
+++ b/highs/mip/HighsPathSeparator.cpp
@@ -216,13 +216,13 @@ void HighsPathSeparator::separateLpSolution(HighsLpRelaxation& lpRelaxation,
             inArcRows.begin() + colInArcs[col].second,
             [&](const std::pair<HighsInt, double>& i,
                 const std::pair<HighsInt, double>& j) {
-              return rowScore[i.first] < rowScore[j.first];
+              return rowScore[i.first] > rowScore[j.first];
             });
     pdqsort(outArcRows.begin() + colOutArcs[col].first,
             outArcRows.begin() + colOutArcs[col].second,
             [&](const std::pair<HighsInt, double>& i,
                 const std::pair<HighsInt, double>& j) {
-              return rowScore[i.first] < rowScore[j.first];
+              return rowScore[i.first] > rowScore[j.first];
             });
   }
 


highs/mip/HighsPathSeparator.cpp
```diff
@@ -106,6 +108,27 @@ void HighsPathSeparator::separateLpSolution(HighsLpRelaxation& lpRelaxation,
...
 
+  // Score will be used for deciding order in which rows get aggregated
+  std::vector<std::pair<double, double>> rowscore(lp.num_row_,
+                                                  std::make_pair(0.0, 0.0));
+  for (HighsInt col = 0; col != lp.num_col_; ++col) {
+    for (HighsInt i = lp.a_matrix_.start_[col];
+         i != lp.a_matrix_.start_[col + 1]; ++i) {
+      HighsInt row = lp.a_matrix_.index_[i];
+      if (rowtype[row] == RowType::kUnusuable) continue;
+      double val = lp.a_matrix_.value_[i];
```
fwesselm: Could you have

`double absval = std::abs(lp.a_matrix_.value_[i]);`

here and use this in the following two lines (instead of having the `std::abs` twice)?

highs/mip/HighsTransformedLp.cpp
```diff
@@ -32,6 +32,7 @@ HighsTransformedLp::HighsTransformedLp(const HighsLpRelaxation& lprelaxation,
                  std::make_pair(-1, HighsImplications::VarBound()));
   boundTypes.resize(numTransformedCol);
   vectorsum.setDimension(numTransformedCol);
+  colFractionality.resize(lprelaxation.numCols());
```
fwesselm, line 35: This vector does not have to be re-initialized (with zeros), correct?

highs/mip/HighsTransformedLp.cpp
```diff
@@ -57,9 +58,26 @@ HighsTransformedLp::HighsTransformedLp(const HighsLpRelaxation& lprelaxation,
     if (ubDist[col] <= mipsolver.mipdata_->feastol) ubDist[col] = 0.0;
 
     boundDist[col] = std::min(lbDist[col], ubDist[col]);
+
+    if (lbDist[col] <= ubDist[col] && lbDist[col] == 0 &&
+        bestVlb[col].first != -1) {
+      const double frac =
```
fwesselm, line 64: There is an utility 

`inline T fractionality(T input, T* intval = nullptr)`

in `HighsLpUtils.h` that you could maybe use here.

jajhall:
It would be better to maintain the use of `fractionality(T input, T* intval = nullptr)` in `HighsLpUtils.h`

@Opt-Mucca Please make this modification so we can merge this PR

Opt-Mucca:
@jajhall Can do. 
As a heads up: @fwesselm tested and the current change is slightly performance negative (does solve an instance or two more but pretty sure that's noise). This is on my list to look into again as the current approach is noisy and can't be impossible to beat. (TLDR: I'd like this to be performance positive before it's merged)

highs/mip/HighsTransformedLp.cpp
Opt-Mucca: Correct. It will be populated completely with the code below. Would you code it differently because of this?

fwesselm: No, just wanted to double-check, thanks.

highs/mip/HighsPathSeparator.cpp
Opt-Mucca: Done.

highs/mip/HighsTransformedLp.cpp
Opt-Mucca: Done

Opt-Mucca:
I'm going to close this PR for now. The fixed ordering of rows per column seems to always lead to negative performance and I don't think the trade-off for stability is worth it. 
There cannot not be some improvement that's possible here though. I'd like to explore "strong" aggregation, where you tentatively use a row to aggregate out a column and see the effects before committing to the (column, row) pair. Not sure on how to efficiently implement this and I want to prioritise working on other things for now, so I'm going to close the PR.

PR #2678, opened 2025-12-03
Add option to disable cut separation at nodes

fwesselm:
- I have a model where a considerable fraction of the time is spent on separating cuts at branching nodes, but either no cuts are found or they are not effective.
- I am proposing to add a boolean option `mip_allow_cut_separation_at_nodes` (default: `true`) to control cut separation at nodes.
- I am happy to change the option name if there are concerns.

### DIFF
--- a/highs/mip/HighsMipSolver.cpp
+++ b/highs/mip/HighsMipSolver.cpp
@@ -607,6 +607,9 @@ restart:
         analysis_.mipTimerStart(kMipClockNodeSearchSeparation);
         sepa.separate(search.getLocalDomain());
         analysis_.mipTimerStop(kMipClockNodeSearchSeparation);
+      } else {
+        // perform aging
+        mipdata_->cutpool.performAging();
       }
 
       if (mipdata_->domain.infeasible()) {


highs/mip/HighsMipSolver.cpp
```diff
@@ -648,9 +648,11 @@ void HighsMipSolver::run() {
...
 
       // the node is still not fathomed, so perform separation
-      analysis_.mipTimerStart(kMipClockNodeSearchSeparation);
-      sepa.separate(search.getLocalDomain());
-      analysis_.mipTimerStop(kMipClockNodeSearchSeparation);
+      if (options_mip_->mip_allow_cut_separation_at_nodes) {
+        analysis_.mipTimerStart(kMipClockNodeSearchSeparation);
+        sepa.separate(search.getLocalDomain());
+        analysis_.mipTimerStop(kMipClockNodeSearchSeparation);
+      }
```
Opt-Mucca, line 613: I don't think this change is needed, but it may be worth adding: Add a `cutpool->performAging()` in an `else` statement?
There is a worst case where cuts keep getting added from sources that aren't separators, e.g., conflicts, and clog up propagation. As we're suppressing all `performAging` and `separate` calls, the cut pool is never potentially shrinking.

fwesselm:
> Happy with the correctness of the change! (and adding the parameter as a feature) Not sure I love the parameter name, but I'm not coming up with anything better.... Thoughts on `mip_separate_cuts_at_nodes` or `mip_separate_cuts_in_search`?

Thanks @Opt-Mucca - I am not married to the name. Since there is `mip_allow_restart`, I tried to come up with something similar.

jajhall:
Since this is a somewhat niche option, and ere is already a 'mip_allow' option, I don't mind the long name

fwesselm:
Thank you both, @jajhall and @Opt-Mucca. If you both think that `mip_separate_cuts_at_nodes` would be better, I am happy to change the name. Please let me know.

jajhall:
Sorry, I meant to send imply that the original 'mip_allow_cut_separation_at_nodes' is fine by me

Opt-Mucca:
Also happy with the name (don't think my suggestions are actually better).

highs/mip/HighsMipSolver.cpp
fwesselm: Added, thanks!

svigerske:
So, to clarify for me, this option is disable separation on all nodes other than root nodes?

Or does have HiGHS separation that isn't at any node, like presolve or MPS parsing?

Maybe at least the description of the option could clarify this.

Opt-Mucca:
@svigerske your first sentence is correct. 

>  this option is disable separation on all nodes other than root nodes?
> 

Would "Whether cut separation at nodes (after root) is permitted" be better?

svigerske:
Yes. After root, or other than the root node.

Some other solvers work avoid the naming issue by having options to specify the maximal depth of the node at which to do separation, or a multiple and offset for the node depth when to do separation.

PR #2680, opened 2025-12-04
Add col integrality helpers

Opt-Mucca:
This was inspired from a comment by @fwesselm from PR #2671 
It is a small refactor that adds some helper functions for determining the variable type of a column.

### DIFF
--- a/highs/mip/HighsMipSolver.h
+++ b/highs/mip/HighsMipSolver.h
@@ -100,6 +100,18 @@ class HighsMipSolver {
     return model_->integrality_[col];
   }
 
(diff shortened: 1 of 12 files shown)


highs/mip/HighsMipSolver.h
```diff
@@ -100,6 +100,18 @@ class HighsMipSolver {
     return model_->integrality_[col];
   }
 
+  bool isColIntegral(const HighsInt col) const {
+    return variableType(col) != HighsVarType::kContinuous;
```
fwesselm, line 104: :) You could even say

`return !isColContinuous(col);`

PR #2682, opened 2025-12-08
Fix 2681

### LINKED ISSUE #2681: [C API] variable names lost in the presolved model
I notice that the presolved MILP model built with C API methods loses the variable names unless I use the File I/O approach.

In particular, the presolved model is built with `Highs_passMip()`, from the infromation extracted from `Highs_getPresolvedLp()`. Naturally, variables in the presolved model built in this way have no name, since the information about names is not extracted in `Highs_getPresolvedLp()`. The problem is that I could not find any method in the C API  to help me assign the variable names to the presolved model, as I don't know which column in the original model corresponds to a column in the presolved model. 

However, if I use the File I/O approach, that is, using `Highs_writePresolvedModel()` and `Highs_readModel()` to write the mps file of the presolved model to the disk and then read the mps file from the disk, the variable names in the original model are preserved in the presolved model.

Just wondering if a method like `Highs_getPresolvedColNames(`) can be added to the C API, which simply extracts column names in the presolved model in an array.

jajhall:
- Added Highs::getColOrRowName as Generalisation of getColName and getRowName. 
- Using `Highs_getHighsLpColOrRowName` in `Highs_getRowName` and `Highs_getColName`
- Using `getColOrRowName` in `getColName` and `getRowName`
- Created `Highs_getPresolvedColName` and `Highs_getPresolvedRowName` in the C API to get a column or row name from the presolved model. Returns `kHighsStatusError` if the index is out-of-range or there are no names.

This closes #2681

### DIFF
--- a/check/TestCAPI.c
+++ b/check/TestCAPI.c
@@ -9,7 +9,7 @@
 #include <math.h>
 #include <string.h>
 
-const HighsInt dev_run = 1;
+const HighsInt dev_run = 0;
 const double double_equal_tolerance = 1e-5;
 
 void checkGetCallbackDataOutPointer(const HighsCallbackDataOut* data_out,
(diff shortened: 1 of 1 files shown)


BenChampion:
What happens to the col names if presolve substitutes variables, or, more generally, a single col/row in the presolved model corresponds to multiple cols/rows in the original model?

jajhall:
Currently there's a map from the original column and row indices to the column and row indices of the reduced problem, so that's how the names are mapped. Clearly some reductions will add multiples of removed columns and rows into retained columns and rows, changing what they represent in the reduced problem, but that for the user to deal with. 

We don't yet have presolve operations (like Gurobi does) that create columns in order to reduce the number of rows, in which case the new columns would have artificial names

Issue #2676, opened 2025-12-02
Minimal python example of IIS


braaannigan:
I want to use the IIS analysis on infeasible problems in highspy.  I've been going through the code and docs, but I can't get my head around how to use the IIS features. 

Here is a minimal example of an infeasible LP:
```python
import numpy as np
from highspy import Highs, HighsLp, HighsModelStatus,HighsIis

# Create HiGHS instance
highs = Highs()

# Build an infeasible LP problem
# Problem: minimize x + y
# Subject to: x + y <= 0.5 and x + y >= 2.0 (contradictory constraints)
# Bounds: 0 <= x <= 1, 0 <= y <= 1

lp = HighsLp()
lp.num_col_ = 2
lp.num_row_ = 2

# Objective: minimize x + y
lp.col_cost_ = [1.0, 1.0]

# Variable bounds: 0 <= x <= 1, 0 <= y <= 1
lp.col_lower_ = [0.0, 0.0]
lp.col_upper_ = [1.0, 1.0]

# Constraints:
# Row 0: x + y <= 0.5
# Row 1: -x - y <= -2.0  (equivalent to x + y >= 2.0)
lp.row_lower_ = [-np.inf, -np.inf]
lp.row_upper_ = [0.5, -2.0]

# Constraint matrix (CSC format)
lp.a_matrix_.start_ = [0, 2, 4]
lp.a_matrix_.index_ = [0, 1, 0, 1]  # Row indices
lp.a_matrix_.value_ = [1.0, 1.0, -1.0, -1.0]  # Coefficients

# Pass model to HiGHS
highs.passModel(lp)

# Solve the problem
highs.run()

model_status = highs.getModelStatus()
``` 
It looks like I need to create a HighsIis object and pass this to highs.getIis, is this correct? What info do I pass to HighsIis? Where is the IIS strategy set?

Thanks for any help you can offer!

Liam

jajhall:
Sorry, this is a mistake on my part. The HighsIis instance should be passed back to the user in the return from getIis, since parameters can't be passed by reference in Python. 

I'll get this fixed when I return from holiday - next week

jajhall:
Closed by #2689

PR #2689, opened 2025-12-12
`getIis` now correct in `highspy`

jajhall:
- Corrected `highspy` method for `getIis`
- Added IIS test to `test_highspy.py`
- Added example of `getIis` to `call_highs_from_python.py


jajhall:
This will close #2676

braaannigan:
Great work @jajhall!

Any idea when the next python release will be?

jajhall:
It depends on progress integrating our new IPM solver into the Python build. If that's slow, we may make an interim release in the New Year.

PR #2694, opened 2025-12-16
Remove redundant copy operations in FactorHiGHSSolver

### LINKED ISSUE #2683: Unnecessary copies at each iteration in HiPO
Hi,

@ZiruiZhou and I have identified some redundant matrix copying within the HiPO implementation that could be optimized, particularly for Linear Programming (LP) problems. I would be happy to do some refactoring if you are open to accepting this improvement.

Specifically, the methods `FactorHiGHSSolver::factorAS` and `FactorHiGHSSolver::factorNE` currently perform a copy of the constraint matrix $A$ every time they are called (i.e., each iteration).

While this copying mechanism may be required for Quadratic Constraint Programs or Non-Linear Programs, it appears unnecessary for standard LP problems, based on the current implementation (since the constraint matrix is not modified between iterations).

Removing this redundant copy for LPs should improve performance for all instances, but will be particularly visible for large-scale LP problems.

Would you like me to proceed with the refactoring and submit a PR for review?

aasivas:
Add Zed editor caches to .gitignore
Fix a compiler warning about potential infinite loops

Fixes #2683

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,10 @@ local.properties
 # PDT-specific
 .buildpath
 
+#################
+## Zed Editor
+#################
+.cache
 
 #################
 ## Visual Studio
--- a/highs/ipm/hipo/ipm/FactorHiGHSSolver.cpp
+++ b/highs/ipm/hipo/ipm/FactorHiGHSSolver.cpp
@@ -1,7 +1,5 @@
 #include "FactorHiGHSSolver.h"
 
-#include <limits>
-
 #include "Status.h"
 #include "ipm/hipo/auxiliary/Auxiliary.h"
 #include "ipm/hipo/auxiliary/Log.h"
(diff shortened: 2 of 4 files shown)


jajhall:
Please look at this @filikat, and see whether it extends what you've already implemented

filikat:
After you opened the issue, I had a better look at FactorHiGHSSolver.cpp and noticed a few other things that needed changing. The whole file became quite a mess when I added the automatic selection of certain options, and it may get even worse with further options added. So I decided to rewrite some parts and organise it better. In doing so, I also got rid of the redundant copies. There is some overlap between your PR and what I did, but I prefer to keep my version, as it is now tidier. Sorry about this, I hope it didn't take you too much time.

jajhall:
Thanks @filikat

aasivas:
No problem at all @filikat . I am used to contributing to open source projects (though mostly on other platforms) :) I tried to make the code changes minimal to avoid interfering with your work, the most time consuming part was testing and it was passive anyways.

Thanks for considering my code.

PR #2695, opened 2025-12-16
Solution enumeration

fwesselm:
This PR adds `HPresolve::enumerateSolutions` as a new presolve reduction:
- The new reduction enumerates all solutions to pure binary constraints with up to 8 variables. It analyses up to 400 of those constraints per call to the reduction.
- The reduction may fix variables, tighten bounds or substitute variables.
- Computational results show that the reduction improves average performance in terms of shifted geometric means by 1 percent. Solution enumeration allows HiGHS to solve 6 additional models to optimality within the time limit (of two hours).
- Solution enumeration is run before probing to avoid (expensive) probing on binary variables that could be fixed.

To do:
- It may still be possible to further tune the code to obtain a general performance improvement in terms of the shifted geometric means (e.g., reduce number of rows that are checked, or reduce the maximum number of binary variables that these rows may contain).

### DIFF
--- a/highs/presolve/HPresolve.cpp
+++ b/highs/presolve/HPresolve.cpp
@@ -4927,6 +4927,8 @@ HPresolve::Result HPresolve::enumerateSolutions(
   std::vector<HighsInt> worstCaseBounds(model->num_col_);
   std::vector<double> worstCaseLowerBound(model->num_col_, kHighsInf);
   std::vector<double> worstCaseUpperBound(model->num_col_, -kHighsInf);
+  std::vector<double> col_lower(domain.col_lower_);
+  std::vector<double> col_upper(domain.col_upper_);
 
   // lambda for branching (just performs initial lower branch)
   auto doBranch = [&](HighsInt& numBranches) {
@@ -5004,7 +5006,15 @@ HPresolve::Result HPresolve::enumerateSolutions(
     return Result::kOk;
   };
 
-  auto handleSolution = [&](size_t& numWorstCaseBounds) {
+  auto removeWorstCaseBounds = [&](size_t pos, size_t& numWorstCaseBounds) {
+    worstCaseLowerBound[worstCaseBounds[pos]] = kHighsInf;
+    worstCaseUpperBound[worstCaseBounds[pos]] = -kHighsInf;
+    worstCaseBounds[pos] = worstCaseBounds[numWorstCaseBounds - 1];
+    worstCaseBounds[numWorstCaseBounds - 1] = 0;
+    numWorstCaseBounds--;
+  };
+
+  auto handleSolution = [&](size_t& numWorstCaseBounds, bool& noReductions) {
     // propagate
     domain.propagate();
     if (domain.infeasible()) return;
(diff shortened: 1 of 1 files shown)


jajhall:
Thanks @fwesselm, this looks like a lot of work!

highs/presolve/HPresolve.cpp
```diff
@@ -4830,6 +4821,316 @@ HPresolve::Result HPresolve::singletonColStuffing(
...
+    size_t numDomainChanges;
+    size_t numChangedCols;
+  };
+  std::vector<std::vector<HighsInt>> solutions;
+  std::vector<HighsInt> vars;
+  std::vector<branch> branches;
+  std::set<HighsInt> worstCaseBounds;
+  std::vector<double> worstCaseLowerBound;
+  std::vector<double> worstCaseUpperBound;
+  worstCaseLowerBound.resize(model->num_col_, kHighsInf);
```
Opt-Mucca: Minor style thing: Initialise this on a single line?
`std::vector<double> worstCaseLowerBound(model_->num_col_, kHighsInf);`?

highs/presolve/HPresolve.cpp
```diff
@@ -4830,6 +4821,316 @@ HPresolve::Result HPresolve::singletonColStuffing(
...
+
+    // main loop
+    HighsInt numBranches = -1;
+    while (true) {
+      bool backtrack = domain.infeasible();
+      if (!backtrack) {
+        backtrack = solutionFound(domain, vars);
+        if (backtrack) {
+          // propagate
+          domain.propagate();
```
Opt-Mucca: Is this what is eating up the compute time (compared to the first draft)? Or is the scan over worst case bounds adding up?

highs/presolve/HPresolve.cpp
fwesselm: Fixed, thank you.

fwesselm:
@jajhall, @Opt-Mucca - I have made some additional changes to the code (mainly use a common code base for parts of probing and enumeration). The new version is still able to solve 6 more models to optimality within the time limit, but is now 1% faster on average. I will update this PR shortly.

highs/presolve/HPresolve.cpp
fwesselm: I think the code that handled the worst-case bounds was not implemented in an optimal way. I have made some additional changes.

Opt-Mucca: Super intelligent change! I like the use of `changedCols`

fwesselm:
> I like the change and the extended probing refactor. It's now performance positive, so I'm happy to merge this whenever @fwesselm says it's ready.
> 
> Note for the future: Revisit some non-static working limits, e.g. `splayContingent` from probing, if we ever find an applications where this ends up taking too long (the `if` does seem decently heavy lifting here)

Thanks, @Opt-Mucca . I would like to try out one more potential improvement (stop search early if there are no reductions). I will report back here.

fwesselm:
I added code to remove worst-case bounds that are not tighter than global bounds while iterating. Tests showed that this additional change it is not improving performance very much (since the vectors only have relatively few elements) for MIPLIB etc. I think the code is ready to go (- I hope @Opt-Mucca and @jajhall agree -) and I plan to experiment a little more with other work limits in the future.

Issue #2696, opened 2025-12-16
Set basic duals to zero


jajhall:
Duals of basic variables are currently set to the residual of $B^T\pi-c_B$

This was done in the old days as a measure of numerical accuracy

However
- it can lead to complementarity violations, as observed in #2643
- the residual of $B^T\pi-c_B$ is computed directly now.

Would be better to set duals of basic variables to be set to zero so that the residual in the dual equations for basic variables is nonzero (so cannot lead to complementarity violations).

Since it's not sensible to use duals of basic variables for anything, solver behaviour should not be affected

jajhall:
Closed by #2697

PR #2697, opened 2025-12-16
Set basic duals to zero

jajhall:
Now that complementarity violations are being computed, they can result from nonzero dual values for basic variables being computed as $B^T\pi-c_B$. Since these are zero using exact arithmetic, they are now zeroed. The numerical error corresponding to nonzero values of $B^T\pi-c_B$ is now exhibited by nonzero residuals in the dual equations.


jajhall:
Closes #2696

PR #2698, opened 2025-12-16
Fix incorrect decrement value

Opt-Mucca:
Small fix I found while looking at something else. I was decrementing `numVarBounds` by the size of the tree structure instead of the number of variable bounds attached to the column.
I couldn't find any way to query the number of nodes cleanly, so I've calculated it myself.

### DIFF
--- a/highs/mip/HighsImplications.cpp
+++ b/highs/mip/HighsImplications.cpp
@@ -718,7 +718,7 @@ void HighsImplications::cleanupVarbounds(HighsInt col) {
     HighsInt numVubs = 0;
     vubs[col].for_each([&](HighsInt vubCol, VarBound& vub) { numVubs++; });
     HighsInt numVlbs = 0;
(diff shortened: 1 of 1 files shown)


fwesselm (approved):
I guess there is no `size()` method, so what you did is fine. Thanks!

jajhall:
> I couldn't find any way to query the number of nodes cleanly, so I've calculated it myself.

In  `HighsMipSolverData::printDisplayLine` there are the following lines

>   auto print_nodes = convertToPrintString(num_nodes);
  auto queue_nodes = convertToPrintString(nodequeue.numActiveNodes());
  auto print_leaves = convertToPrintString(num_leaves - num_leaves_before_run);

Doesn't `print_nodes` or `queue_nodes` give you the number of nodes?

Opt-Mucca:
@jajhall "nodes" here was a terrible way to refer to the size of `HighsHashTree`.
The number of branch-and-bound nodes is stored properly by and can be easily queried.

PR #2699, opened 2025-12-16
Add Metis, AMD and RCM to HiGHS

filikat:
- The source code of Metis, AMD and RCM is now part of HiGHS. Only the routines which are essential for HiPO are included, and they are compiled with the same integer type as HiGHS.
- A new option `hipo_ordering` allows to select which ordering to use between "metis", "amd", "rcm". The default is "choose", for which HiPO selects the best ordering between "metis" and "amd".
- A new unit test checks that the three orderings run correctly.
- Option `hipo_metis_no2hop` has been removed, since setting no2hop=true in Metis seems to always be beneficial.
- A file "LicenseNote.txt" has been added specifying that building HiGHS with HiPO, the resulting library and executables are not MIT licensed.

### DIFF
--- a/check/TestHipo.cpp
+++ b/check/TestHipo.cpp
@@ -16,7 +16,7 @@
 #include <iostream>
 #include <vector>
 
-const bool dev_run = true;
+const bool dev_run = false;
 
 TEST_CASE("test-hipo-afiro", "[highs_hipo]") {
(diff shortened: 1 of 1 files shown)


galabovaa (approved):
Great work! 🚀 ⭐

PR #2702, opened 2025-12-16
Remove METIS and GKlib from GitHub actions

odow:
Follow-up to https://github.com/ERGO-Code/HiGHS/pull/2699

We need to remove the manually installed versions of metis and gklib in CI to ensure we're actually using the vendored copies.

### DIFF
--- a/.github/julia/build_tarballs.jl
+++ b/.github/julia/build_tarballs.jl
@@ -35,7 +35,6 @@ cmake -DCMAKE_INSTALL_PREFIX=${prefix} \
     -DZLIB_USE_STATIC_LIBS=${BUILD_STATIC} \
     -DHIPO=ON \
     -DBLAS_LIBRARIES="${libdir}/libopenblas.${dlext}" \
-    -DMETIS_ROOT=${prefix} \
     ..
 
 if [[ "${target}" == *-linux-* ]]; then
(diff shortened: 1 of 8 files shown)


odow:
There's a segfault in the unit tests which seems suspicious.

filikat:
> There's a segfault in the unit tests which seems suspicious.

That test is not compiled with HiPO, so it's a bit weird.
However, it made me realise two things that we are doing wrong @galabovaa:
1 - TestHipo.cpp should only be included in the unit tests if HIPO=ON.
2 - if solver=hipo is used in a build that does not include HiPO, at the moment IPX is selected, because of how lp_data/HighsSolve.cpp:useHipo is set up (line 724). It would be better to give the error in lp_data/HighsSolve.cpp:93, which I don't think is reachable at the moment.

galabovaa:
Noted, @filikat 

I've got updates to the workflows on a WIP branch. I have updated build_tarballs there too, so I'm closing this, @odow

odow:
Can we please consider moving to a PR-based workflow over branches? It's hard to tell what people are working on.

odow:
I think this is also a good motivation to simplify some of the CI workflows with a build matrix. We shouldn't need to copy-paste code in so many places. (e.g., where I was headed with https://github.com/ERGO-Code/HiGHS/pull/2551)

galabovaa:
Thanks for the suggestion, but we will retain them for now

PR #2706, opened 2025-12-17
Various minor improvements to HiPO

filikat:
- Adjust some variable names that were not following the naming convention.
- Simplify some of the internal interfaces between objects in HiPO.
- Simplify the timers that are switched on/off at compile time.
- Other various minor changes to make the code cleaner.
- Reorganize FactorHiGHSSolver.cpp and remove some redundant copies of matrix A.


filikat:
Closes #2683

Issue #2705, opened 2025-12-17
[C API] error with Highs_passMip()


JunchengLiamLi:
I noticed that `Highs_passMip()` does not work on certain problems, on which the file I/O approach works.

Basically I run the same code containing `Highs_passMip()` on different MILP problems in the MPS format. `Highs_passMip()` works on some, but fails on the others. Meanwhile, the file I/O approach (writing the presolved model to disk with `Highs_writePresolvedModel()` and then reading the mps file from the disk with `Highs_readModel()`) works on all the .mps files. 

The code with `Highs_passMip()` is like the following:

```
    presolved_num_col = HiGHS.Highs_getPresolvedNumCol(model)
    presolved_num_row = HiGHS.Highs_getPresolvedNumRow(model)
    presolved_num_nz = HiGHS.Highs_getPresolvedNumNz(model)

    col_cost = zeros(Cdouble, presolved_num_col)
    col_lower = zeros(Cdouble, presolved_num_col)
    col_upper = zeros(Cdouble, presolved_num_col)
    row_lower = zeros(Cdouble, presolved_num_row)
    row_upper = zeros(Cdouble, presolved_num_row)
    a_start = zeros(Cint, presolved_num_col)
    a_index = zeros(Cint, presolved_num_nz)
    a_value = zeros(Cdouble, presolved_num_nz)
    integrality = zeros(Cint, presolved_num_col)
    a_format = HiGHS.kHighsMatrixFormatColwise
    num_col_ref = Ref{Cint}(presolved_num_col)
    num_row_ref = Ref{Cint}(presolved_num_row)
    num_nz_ref = Ref{Cint}(presolved_num_nz)
    sense_ref = Ref{Cint}(0)
    offset_ref = Ref{Cdouble}(0.0)

    ret_get = HiGHS.Highs_getPresolvedLp(
        model,
        a_format,
        num_col_ref,
        num_row_ref,
        num_nz_ref,
        sense_ref,
        offset_ref,
        col_cost,
        col_lower,
        col_upper,
        row_lower,
        row_upper,
        a_start,
        a_index,
        a_value,
        integrality
    )

    reduced_model = HiGHS.Highs_create()
    
    ret_pass = HiGHS.Highs_passMip(
        reduced_model,
        presolved_num_col,
        presolved_num_row,
        presolved_num_nz,
        a_format,
        sense_ref[],
        offset_ref[],
        col_cost,
        col_lower,
        col_upper,
        row_lower,
[...]

PR #2707, opened 2025-12-17
Can now pass a presolved MIP from C if it contains implied integers

jajhall:
When a MIP is passed to HiGHS via arrays (ie from Fortran/Julia/C), any integrality array was checked to see whether it contained only the four types continuous, integer, semi-continuous and semi-integer. 

When the presolved MIP is extracted, it may contain implied integers, so passing it via arrays led to it being rejected.

Values of kImplicitInteger are now permitted

Added highsVarTypeToString methods to HighsLpUtils for HighsVarType and HighsInt; allowing implied integers in models passed via arrays

### DIFF
--- a/check/TestMipSolver.cpp
+++ b/check/TestMipSolver.cpp
@@ -1228,6 +1228,42 @@ TEST_CASE("get-fixed-lp", "[highs_test_mip_solver]") {
   h.resetGlobalScheduler(true);
 }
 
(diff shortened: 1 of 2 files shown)


jajhall:
This will close #2705

JunchengLiamLi:
Here to confirm the problem that I mentioned in https://github.com/ERGO-Code/HiGHS/issues/2705 has been solved.

Thank you very much @jajhall!

Issue #2643, opened 2025-11-10
Avoid primal simplex when unscaled problem has primal infeasibilities but is dual feasible: use dual simplex to clean up


shwhmmmam:
[fail.mps.txt](https://github.com/user-attachments/files/23463271/fail.mps.txt)

triggers an assertion failure in HiGHS:

```
Running HiGHS 1.11.0 (git hash: 26b632af1): Copyright (c) 2025 HiGHS under MIT licence terms
Set option log_file to "HiGHS.log"
MIP  small has 11 rows; 9 cols; 84 nonzeros; 6 integer variables (0 binary)
Coefficient ranges:
  Matrix [3e-02, 5e+02]
  Cost   [2e+01, 3e+02]
  Bound  [1e+02, 1e+02]
  RHS    [2e+01, 7e+04]
Presolving model
10 rows, 9 cols, 83 nonzeros  0.001548s
10 rows, 9 cols, 83 nonzeros  0.003174s

Solving MIP model with:
   10 rows
   9 cols (0 binary, 6 integer, 0 implied int., 3 continuous, 0 domain fixed)
   83 nonzeros

Src: B => Branching; C => Central rounding; F => Feasibility pump; J => Feasibility jump;
     H => Heuristic; L => Sub-MIP; P => Empty MIP; R => Randomized rounding; Z => ZI Round;
     I => Shifting; S => Solve LP; T => Evaluate node; U => Unbounded; X => User solution;
     z => Trivial zero; l => Trivial lower; u => Trivial upper; p => Trivial point

        Nodes      |    B&B Tree     |            Objective Bounds              |  Dynamic Constraints |       Work      
Src  Proc. InQueue |  Leaves   Expl. | BestBound       BestSol              Gap |   Cuts   InLp Confl. | LpIters     Time

         0       0         0   0.00%   -102811         inf                  inf        0      0      0         0     0.1s
         0       0         0   0.00%   21580.316208    inf                  inf        0      0      0        11     0.1s
highs: /tmp/HiGHS/highs/lp_data/HighsInterface.cpp:2709: HighsStatus Highs::lpKktCheck(const std::string &): Assertion `info.num_complementarity_violations == 0' failed.
Aborted (core dumped)
```

fwesselm:
The solution of an LP solved by the feasibility pump heuristic has one complementarity violation.

jajhall:
Thanks, I'll look at this.

jajhall:
[...]

PR #2709, opened 2025-12-18
Avoid primal simplex when unscaled problem has primal infeasibilities but is dual feasible: use dual simplex to clean up

jajhall:
Minor merge conflict fixed by [0993d73](https://github.com/ERGO-Code/HiGHS/pull/2709/commits/0993d7352966ce2e761c37422e0241e89e4993fd)


jajhall:
Closes #2643

PR #2728, opened 2026-01-06
Binary Packages and BUILD_OPENBLAS

galabovaa:
Add release binary packages including HiPO and OpenBLAS on Windows and Linux. MacOS packages depend on Apple Accelerate for now. If compatibility issues pop up, we could restrict the Accelerate version to an older one after investigating performance variations.

The following binary packages are generated: 

Windows 64bit AMD and ARM 
Linux 64bit AMD and ARM 
MacOS Universal 

The packages are generated using CPack from CMake.
 
Added a CMake option BUILD_OPENBLAS. If enabled, HiGHS downloads and configures OpenBLAS v0.3.30 as a subproject, following the OR-Tools CMake dependency management approach. 

Supported platforms 

Windows 64bit AMD and ARM
Windows 32bit AMD
Linux 64bit AMD and ARM
Linux 32bit AMD 

 TestHipo.cpp is only included in the unit tests if HIPO=ON.

### DIFF
--- a/.github/workflows/release-cpack.yml
+++ b/.github/workflows/release-cpack.yml
@@ -23,7 +23,7 @@ jobs:
       - name: Upload artifacts
         uses: actions/upload-artifact@v4
         with:
-          name: linux-${{ matrix.arch }}
+          name: linux-x64-apache
           path: build/*.tar.gz
 
       - name: Release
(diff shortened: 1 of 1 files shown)


jajhall (approved):
I have to trust you!

galabovaa:
Thank you @filikat!

I can edit the OPENBLAS_32

I only added it for the autodetect because it was failing on 32bit systems. If you think we should remove it I am happy to do that. Alternatively, should we only call it if BLAS is not using 64bit integers? This I am not sure how to check.

filikat:
@galabovaa Please remove AutoDetect.cpp

Issue #2734, opened 2026-01-07
Outputs to stdout even when log_to_console option is false


SanjayVas:
I'm seeing messages like the following in stdout:

```
degeneracy? not possible to find non-active constraint to leave basis. max: log(d[0]) = -inf
```

This is despite passing `verbose=False` when calling qpsolvers, [which in turn](https://github.com/qpsolvers/qpsolvers/blob/1829864f4a2a20b310d5ded57c26824c8c84f9b6/qpsolvers/solvers/highs_.py#L192) sets the `log_to_console` option to `false` when calling highspy.

This causes problems with programs that rely on stdout being parseable. (In our case, having highs fail is sometimes expected and should result in the program still having the appropriate output via stdout).

Using qpsolvers 4.3.3 with highspy 1.7.2.

jajhall:
This is, presumably, a rare event in general for the QP solver, and the logging comes from a simple `printf` statement.

I am working on a fairly easy fix

jajhall:
Closed by #2735

jajhall:
> Using qpsolvers 4.3.3 with highspy 1.7.2.

We're up to  highspy 1.12 now, so qpsolvers will have to update to highspy 1.13 (once released!) to pick up this fix.

keithbriggs:
What is the correct way in highspy to suppress solver output?  It seems to be undocumented. I've got 

```
h.output_flag=False
h.log_to_console=False
```

but I still get lots of very verbose stuff stuff like

```
Running HiGHS 1.12.0 (git hash: 755a8e0): Copyright (c) 2025 HiGHS under MIT licence terms
MIP has 1 row; 210 cols; 2 nonzeros; 210 integer variables (210 binary)
Coefficient ranges:
  Matrix  [1e+00, 1e+00]
  Cost    [8e-04, 4e+00]
  Bound   [1e+00, 1e+00]
  RHS     [0e+00, 0e+00]
Presolving model
1 rows, 0 cols, 0 nonzeros  0s
0 rows, 0 cols, 0 nonzeros  0s
Presolve reductions: rows 0(-1); columns 0(-210); nonzeros 0(-2) - Reduced to empty
Presolve: Optimal
,,,
```

PR #2735, opened 2026-01-07
Now handling rare `printf` in QP solver properly!

jajhall:
Added another of the awfully cumbersome logging methods into the QP solver to avoid a `printf`


jajhall:
This will close #2734

PR #2737, opened 2026-01-09
Fix 2722

### LINKED ISSUE #2722: Performance regression in default HiGHS LP solver (v1.11.0 → v1.12.0)
### Discussed in https://github.com/ERGO-Code/HiGHS/discussions/2720

<div type='discussions-op-text'>

<sup>Originally posted by **maniok098** December 31, 2025</sup>
Dear HiGHS team,

after updating to v1.12.0, for my problem, I observed a significant performance regression of the default LP solver compared to v1.11.0. Both in C++ and highspy. For this LP problem [sampleLP.mps.gz](https://github.com/user-attachments/files/24397246/sampleLP.mps.gz), the solution time increased from 0.85s (v1.11.0) to 1.84s (v1.12.0).  The number of simplex iterations increased from 19292 to 60660.

After git bisection, this regression was tracked to the commit 20186bf352ea914351e64c23c64b85ba450e3dfd . (The solver was built from C++ source on an Archlinux machine.) Please let me know if any additional information would be helpful.

Best and happy new year.</div>

fwesselm:
Fix #2722 
1. Use a tolerance (including the column's dynamism) to decide whether a column is forcing or not.
3. With the change, presolve achieves the same reduction numbers as v1.11.0 on the reproduction example and running times are also similar.
4. Testing on 850+ MIPs showed that only 11 models are affected by this change in terms of node counts. There is a tiny reduction in the running times and numbers of nodes (-0.5 to -1 percent).

### DIFF
--- a/highs/presolve/HPresolve.cpp
+++ b/highs/presolve/HPresolve.cpp
@@ -4463,7 +4463,7 @@ HPresolve::Result HPresolve::detectDominatedCol(
                                         col, -model->col_cost_[col])
                                   : -impliedDualRowBounds.getSumLowerOrig(
                                         col, -model->col_cost_[col]);
-      if ((-direction) * boundOnColDual <=
+      if (std::abs(boundOnColDual) <=
           options->dual_feasibility_tolerance / dynamism) {
         // 1. column dual's upper bound is zero (since the column's lower bound
         // is infinite) and column dual's lower bound is zero as well


fwesselm:
> Happy with the change! It has some nice cleaning up too.
> 
> @fwesselm To check I understand correctly: Does this in theory introduce a new incrediblyyyy tiny chance of cutting off an optimal solution? Am fine if it does because if it would then it'd clearly be from numerics, but want to check I haven't missed something.

Thanks, @Opt-Mucca. Correct, the check now depends on `dual_feasibility_tolerance`, so there may be a numerically challenging model where we cut off a feasible solution. But, this should be very unlikely.

Issue #2738, opened 2026-01-09
Refactor `Highs::run()`


jajhall:
`Highs::run()` has too much in it, and needs refactoring into a set of nested methods. 

1. The "outer" layer should contain just the actions to (possibly) read from a solution or basis file, write out the model and, once it's been solved, write out the IIS model, presolved model, and solution or basis file. These file options are cached in the `Highs` class and cleared from `options_` so that they aren't applied at lower levels. Since all this was moved here so that a call to `run()` can perform what was previously only possible by running the `exe`, it's suggested that this method is called `Highs::runFromExe()`.
2. The next layer should apply any user scaling and then assess excessive objective and bound scaling. These scaling options are cached in the `Highs` class and cleared from `options_` so that they aren't applied at lower levels. It's suggested that this method is called `Highs::runUserScaling()`. 
3. The next layer optimizes the problem defined in the `Highs` class with respect to the (remaining) options. Currently this will be either a single call to optimize what's in the `HighsModel`, or a call to `Highs::multiobjectiveSolve()` if there are multiple objectives. It's suggested that this method is called `Highs::optimizeHighs()`
4. The next layer just optimizes what's in the `HighsModel`, and is the current `Highs::optimizeModel()`

- `Highs::run()` is (of course) retained, but should just return the value of `Highs::runFromExe()`
- Although there is no overhead within the nest of methods when solving an LP without file or user scaling options, calling `Highs::run()` from the MIP solver is opaque, so maybe these calls should be to  `Highs::optimizeModel()`

jajhall:
Closed by #2740

PR #2740, opened 2026-01-11
Refactoring `Highs::run()`

jajhall:
Sufficient refactoring of `Highs::run()` has been done to enable #2635 to be fixed.

Split `Highs::run()` into a set of nested methods according to [HighsRun.md](https://github.com/ERGO-Code/HiGHS/blob/fix-2738/highs/HighsRun.md), which includes chat about what still needs to be done. This will take time, so this PR avoids a long-lasting dev branch.

### DIFF
--- a/highs/Highs.h
+++ b/highs/Highs.h
@@ -1296,7 +1296,6 @@ class Highs {
   // Nested methods below Highs::run()
   //
   // See highs/HighsRun.md
-  HighsStatus runFromUserScaling();
   HighsStatus optimizeHighs();
   HighsStatus optimizeModel();
   HighsStatus calledOptimizeModel();
--- a/highs/HighsRun.md
+++ b/highs/HighsRun.md
@@ -95,14 +95,10 @@ method where any "mods" are made, so that at the level below the
 problem defined by the `HighsModel` class (without semi-variables) is
 solved.
 
-### `Highs::runFromExe()`
+### `Highs::run`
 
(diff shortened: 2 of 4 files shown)


jajhall:
Closes #2738

jajhall:
> I'm a big fan of refactoring this! Some current thoughts:
> 
> * I dislike calling `optimizeHighs` inside of `runFromUserScaling`. I'd rather a `scale` function and then an `undoscale` function on either side of the call.

If I understand correctly, you're suggesting putting a `scale` function and then an `undoscale` function either side of `optimizeHighs` within `runFromExe`? 

This does reduce the number of nested methods, and I can't think of circumstances where user scaling would need to be applied without the "Highs files" methods, so fair enough

> * Why do we actually need `runFromExe`? Could this not just be `run`?

I find `run` to be so opaque, but if we eliminate `runFromUserScaling` then it's the only `run` method, so there's no distinction to be made.

> * I like the general breakdown (run-time options -> user scaling -> "mods")

Thanks

Opt-Mucca:
> 
> If I understand correctly, you're suggesting putting a `scale` function and then an `undoscale` function either side of `optimizeHighs` within `runFromExe`?
> 
Yup!

jajhall:
Thanks, I've simplified the calling structure.

Are you OK with this PR?

Opt-Mucca (approved):
Happy with the change!

PR #2747, opened 2026-01-14
BUG: initialize save_value field to fix uninitialized memory use

stratakis:
When a HighsSimplexBadBasisChangeRecord is created, save_value is not set until later in applyTabooRowOut/applyTabooVariableIn. Pushing the struct with an uninitialized field copies garbage memory.

Found via Coverity static analysis

### DIFF
--- a/highs/simplex/SimplexStruct.h
+++ b/highs/simplex/SimplexStruct.h
@@ -249,7 +249,7 @@ struct HighsSimplexBadBasisChangeRecord {
   HighsInt variable_out;
   HighsInt variable_in;
   BadBasisChangeReason reason;
-  double save_value;
+  double save_value = 0.0;
 };
 
 struct HighsRayRecord {


stratakis:
Coverity message:
```
 17. scipy-1.16.2/subprojects/highs/src/simplex/HEkk.cpp:4022:5: var_decl: Declaring variable "record" without initializer.
18. scipy-1.16.2/subprojects/highs/src/simplex/HEkk.cpp:4028:5: uninit_use_in_call: Using uninitialized value "record". Field "record.save_value" is uninitialized when calling "push_back". [Note: The source code implementation of the function has been overridden by a builtin model.]
  4026|       record.variable_in = variable_in;
  4027|       record.reason = reason;
  4028|->     bad_basis_change_.push_back(record);
  4029|       bad_basis_change_num = bad_basis_change_.size() - 1;
  4030|     } else {
```

Issue #2748, opened 2026-01-14
Segfault in `highsSparseTranspose`


jajhall:
A segfault  in `highsSparseTranspose` on AWS Lambdas using ARM64 has been reported "when we submit problems with ridiculously high precision".

The only oddity in `highsSparseTranspose` is that it uses `AcountX = Aindex.size();` to determine the number of nonzeros in the matrix - and hence resize `ARindex` and `ARvalue`, rather than `Astart[numCol]` - although both values should be the same. So, if the size of `Aindex` is incorrect, the size of  `ARindex` and `ARvalue` is incorrect. However, it's only conceivable that  `AcountX` is larger than `Astart[numCol]` - otherwise I'd have expected a segfault earlier - so size of  `ARindex` and `ARvalue` would be larger than necessary, which wouldn't lead to a segfault.

Issue #2742, opened 2026-01-11
Refactor `Highs::optimizeModel()`


jajhall:
Continue with "ToDo" list outlined in [HighsRun.md](https://github.com/ERGO-Code/HiGHS/blob/fix-2738/highs/HighsRun.md).

PR #2750, opened 2026-01-15
Added checks to identify cause of segfault exposed in #2748

jajhall:
In `highsSparseTranspose`, now using `Astart[numCol]` rather than `Aindex.size()` as number of nonzeros.

In debug, calling `HPresolve::checkCSC` from `HPresolve::toCSC` to make sure that CSC index and value are of the correct size, that the entries of index are in bounds, and that there there are no explicit zeros in value arrays. No reason to suspect that anything is wrong, but what might be causing the segfault exposed in #2748?


jajhall:
Includes changes relating to #2742, so make this a draft PR until the PR corresponding to #2742 is merged

PR #2761, opened 2026-01-21
Enhancement to `HPresolve::enumerateSolutions`

fwesselm:
- This PR adds an enhancement to `HPresolve::enumerateSolutions` suggested by @Opt-Mucca (thank you!)
- The code now checks if all binary variables (from a constraint) form a clique.
- The average solution times and numbers of branching nodes increase by around 1% with this code change. However, 7 additional problems can be solved to optimality within the time limit of 2 hours.


jajhall:
That's great, thanks, @fwesselm !

PR #2768, opened 2026-01-23
Remove copies of LP in HiPO wrapper

filikat:
Currently, when calling HiPO, in IpxWrapper.cpp, a copy of the LP is made using fillInIpxData. This copy is then passed to hipo::Solver::load, which makes another copy to store internally. The copy made in IpxWrapper.cpp is unused, but remains in scope for the whole time that the solver runs.

This PR changes the interface of hipo::Solver::load to accept a HighsLp instance. Then, fillInIpxData is called from hipo::Model::init and a copy of the LP is made directly in the HiPO internal storage, eliminating the need for an intermediate copy. A pointer to the original HighsLp is stored, in case it needs to be loaded into IPX later.

To achieve this, I had to separate the construction of rhs and constraints into a separate function fillInRhsAndConstraints, because a copy of these is needed in IpxWrapper.

While doing this, I noticed that at the end of fillInIpxData, the objective vector is augmented with some slacks, but num_var is not changed. These slacks are useless, since both HiPO and IPX copy only the first num_var entries of the objective vector. 

I also fixed the way that option timeless_log is passed to IPX from HiPO and suppressed the printing of some times in the HiPO summary if timeless_log is true.

### DIFF
--- a/highs/ipm/hipo/ipm/Model.cpp
+++ b/highs/ipm/hipo/ipm/Model.cpp
@@ -518,6 +518,8 @@ Int Model::loadIntoIpx(ipx::LpSolver& lps) const {
   std::vector<char> ipx_constraints;
   double ipx_offset;
 
+  if (!lp_orig_) return kStatusError;
+
   fillInIpxData(*lp_orig_, ipx_n, ipx_m, ipx_offset, ipx_c, ipx_lower,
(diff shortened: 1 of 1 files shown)


filikat:
@jajhall I can do the same for IPX if we want. There is no real performance gain though.

PR #2770, opened 2026-01-23
Now initialising `sense` in `struct Model`  of LP file reader

### LINKED ISSUE #2769: Unassigned variable used in `test_garbage_lp` with `filereader-edge-cases`
`m.sense` needs to be initialised

jajhall:
This closes #2769


jajhall:
It's a trivial change, @galabovaa, so the only issue seems to be the Julia CI tests

galabovaa:
I think I know how to fix the julia tests

Issue #2781, opened 2026-01-28
'solver' documentation 


jajhall:
Interpretation of 'solver' option is becoming complicated, so add a documentation page for it

filikat:
I have added some clarification to the solvers page of the documentation in the branch hipo-qp.

jajhall:
OK I'll merge your documentation changes into mine.

Issue #2752, opened 2026-01-16
Docs: Add alternative section for parallel instance solving


L-Bodi:
Add a section to docs/src/parallel.md

In this chapter of the docs you state, that parallelization is limited. However, it might be valuable for the community to read in the docs, that there are alternative ways to reduce computation time of multiple instances due to parallelization.

## Alternative

If you have multiple instances to solve, you might have an alternative to overcome the mentioned limits. The idea is to assign one instance per worker (cpu core or thread) so that multiple instances are solved individually, but parallel. Reading tip: "10 times faster, running cases in parallel" by Solver Max.

jajhall:
Thanks. This sounds too obvious to mention, but does no harm.

jajhall:
Closed by #2787

PR #2787, opened 2026-01-29
Update solver documentation - for v1.14

jajhall:
This page will be correct when cuPDLP-C is replaced by HiPDLP, and the QP IPM solver is added, so this is just a draft PR


jajhall:
Closes #2781

jajhall:
Closes #2752

jajhall:
> In definitions.md for option solver, should there be both "pdlp" and "hipdlp"? The option for "pdlp" should also coincide with what is used in the Summary of solvers.md.

Yes, the "hipdlp" value for __solver__ is still in Yanyu's branch, so that she can run cuPDLPc (by setting "pdlp") for sanity checking, but it will disappear when we have only her PDLP solver 

> It seems like you removed the statements "Setting the solver option to ... forces the ... solver to be used" from the LP section, since they are now present in the summary. But they still appear in the QP section. For consistency, it may be better to remove them from there as well.

Indeed, I've updated solvers.md

jajhall:
Documentation is fixed

filikat:
@jajhall Maybe we should specify that options `qp_iteration_limit`, `qp_nullspace_limit` and `qp_regularization_value` refer to the active-set solver, to avoid confusion.

Issue #2786, opened 2026-01-29
HiGHS Infinite Loop 32 bit RelWithDebInfo


Opt-Mucca:
The issue can be seen here: https://github.com/ERGO-Code/HiGHS/actions/runs/21218197653/job/61045323342 and was brought up by @galabovaa 

I've traced the bug to this line:
https://github.com/ERGO-Code/HiGHS/blob/3c26f3306c53c31a16b54a32610ac551fe31cd40/highs/mip/HighsDomain.cpp#L1233

That seems to be evaluating to `false` for this very specific build despite the value (645.37999000000002) repeatedly not changing. So it's something like:
```
a = double(val)
b = double(val)
if (a == b) break; // somehow evaluating to false
```
Because I can't build in full debug mode nor print stuff on either side without the bug disappearing I'm not entirely sure why that's the case, but I'd guess it's some strange interaction for 32 bit code. A potential fix could be changing the line check to something like:
```
if (std::abs(a - b) < 1e-14 break; // or something else suitably small`
```

Edit: I made the issue in case someone with more knowledge had a stronger opinion on how to best handle the issue.

galabovaa:
This looks fine to me, we can do 1e-08. What do you think, @jajhall?

aasivas:
I know this is unsolicited, I just saw the issue and wanted to share my knowledge/guess.

With 32-bit version of ubuntu, the compiler maybe trying to use (x87) 80-bit registers. If `a` is 80-bits but `b` has done a roundtrip e.g. 80 to 64 to 80 again (losing 16 bits of accuracy), `a` and `b` would be different. 

One way to confirm this hypothesis is to make `a` and `b` volatile. If the bug disappears, it is highly likely that was the issue.

Proper solution would be $\epsilon$ comparison as proposed, but I would suggest $\tfrac{|a-b|}{max(|a|,|b|) + 1} \leq Cu$, where $u$ is the machine epsilon in double precision and $C$ is constant, maybe 2 or 4.

Edit to add more context:
This is a "Heisenbug" often encountered on 32-bit x86 systems and a side effect of how older Intel processors handle floating-point math versus data is stored in memory.

If `a` is currently in an 80-bit register (just calculated) and `b` has been moved to memory and then reloaded losing 16 bits of precision (roundtrip mentioned above), they are not bit-for-bit identical in 80-bit comparison. The number itself (645.37999000000002) can cause this.

[...]

PR #2788, opened 2026-01-29
Check for zero bug

galabovaa:
Closes https://github.com/ERGO-Code/HiGHS/issues/2786

### DIFF
--- a/.github/julia/build_tarballs.jl
+++ b/.github/julia/build_tarballs.jl
@@ -75,6 +75,6 @@ build_tarballs(
     platforms,
     products,
     dependencies;
-    preferred_gcc_version = v"6",
+    preferred_gcc_version = v"11",
     julia_compat = "1.6",
 )
--- a/.github/workflows/valgrind-hipo.yml
+++ b/.github/workflows/valgrind-hipo.yml
@@ -30,65 +30,17 @@ jobs:
         run: |
           cmake --build . -j2
 
-      - name: Test AVX-512
+      - name: Test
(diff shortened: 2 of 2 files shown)


fwesselm:
I have not read all the comments in the issue (sorry!), but here are my comments:

1. I do not like hard-coding values. Is there an existing tolerance that could be used?
2. On the other hand, why does the current code check for strict equality? Checking floating point numbers for strict equality rarely makes sense.
3. Would it be possible to count the number of bound changes found in the current iteration and stop if no new bounds are found? I can look into this if you like.

@Opt-Mucca , what do you think (in particular, regarding using a different criterion for exiting the loop)?

jajhall:
I don't know what the current code does, but I was concerned that the test was being relaxed just to eliminate an infinite loop that occurs when using 32-bit doubles. 

I'm concerned that the modification might compromise solver behaviour with 64-bit doubles, and it's important that this doesn't happen.

jajhall:
Looking at the code, I see that it compares objective bounds. Since these can be very large, a relative difference seems to be necessary. 

The existing tolerance that seems relevant to me is 'mip_rel_gap', but any change needs to understand the context of the code, and why @lgottwald coded it originally as an equality.

Opt-Mucca:
@fwesselm Sorry for the late reply. I think the idea of counting bound changes is obvious and genius. I'm not seeing a way that the objective can change in that loop without making a bound change.

jajhall:
Closed by #2807

PR #2796, opened 2026-02-02
Guarding `printf` in `HighsInterface.cpp` with `kIisDevReport`

jajhall:


galabovaa (approved):
Great! Not sure why the CI got cancelled but should all be fine

PR #2807, opened 2026-02-04
Fix issue 2786

fwesselm:
- Count number of bound changes instead of checking for change in objective lower bound in `HighsDomain::ObjectivePropagation::propagate` (which should fix the stall reported in #2786).
- Added utility `HighsDomain::checkChangeBound` to simplify the code.
- Testing showed that the code change does not affect behavior very much. The new code is slightly faster and computes less nodes (time ratio is 0.98 and node ratio 0.99), but this may just be a random effect.


fwesselm:
> Code looks good! I'm a big fan of the new function and checking the number of bounds vs comparing the objective value directly. Are the affected instances happening because new bound changes can now be made in a second round, i.e., the objective doesn't change when `x_j` was updated and the old code would've stopped but the new change allows you to change `x_i` in the next round of the loop? Or is the code just different and I'm missing something?

Thanks, @Opt-Mucca. Yes, I think this is what is happening. For example, on `stein45` the new code finds additional bound changes although `objectiveLower` does not change.

PR #2812, opened 2026-02-05
Modified `1451.lp` to test code to permit keywords as constraint names

jajhall:
Model supplied by @khantute in #1345 failed to load - since it contains a constraint named "end" - which was interpreted as an LP keyword. However, because it's followed by ":", this is illegal in LP files. 

Added code to interpret an LP keyword followed by a colon as a constraint name followed by a colon!

### DIFF
--- a/highs/io/filereaderlp/reader.cpp
+++ b/highs/io/filereaderlp/reader.cpp
@@ -55,6 +55,8 @@ enum class RawTokenType {
   ASTERISK
 };
 
+/*
+  // Useful for debugging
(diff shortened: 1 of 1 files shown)


jajhall:
To clarify, @khantute, it's the keyword followed by colon that's illegal, not your use of "end" as a constraint name!

odow:
I wrote a proper LP parser last year. Some of the grammar can be horrific!

https://jump.dev/MathOptInterface.jl/stable/submodules/FileFormats/LP/#Whether-variables-can-have-the-same-name-as-a-keyword

PR #2813, opened 2026-02-05
Performance when changing LP data

smharwood:
Potential performance improvements when changing LP data (costs, row and column bounds)

In particular, this avoids re-allocating local copies of data and re-sorting index sets

### DIFF
--- a/check/TestLpModification.cpp
+++ b/check/TestLpModification.cpp
@@ -809,7 +809,10 @@ TEST_CASE("LP-modification", "[highs_data]") {
   col1357_upper[2] = 0;
   col1357_upper[3] = 0;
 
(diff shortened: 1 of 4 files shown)


jajhall (commented):
I'll look at this again, and thanks for submitting the PR, but my instinct is that the overhead you're avoiding doesn't justify the additional code complexity: solving the modified problem is very much more expensive than performing the modification with the local copy

smharwood:
Thanks for taking a look - the "repeated sort" is what caught my eye when I was profiling one of my problems.

I'll see if I can come up with an example that I can share that captures the behavior I'm seeing.

smharwood:
Been meaning to say for the past month or two - this is not a priority for me anymore. I still think it _could_ be a bottleneck, but I agree overall its probably not worth the effort. I'll close the PR, but who knows - might revisit.

PR #2815, opened 2026-02-06
LP KKT check

jajhall:
The code to perform the LP KKT check is now in `lpKktCheck` in `HighsSolution.cpp`. When no basis is available, `lpNoBasisKktCheck` can be called.

Method computes the KKT measures in a `HighsInfo` instance, where the relative measures are calculated using the norm of "active" costs and bounds: these correspond to 
- costs where the squared dual is less than the dual feasibility tolerance
- lower/upper bounds where the residual with respect to a bound is, when squared, less than the primal feasibility tolerance

On entry, if `model_status` is `HighsModelStatus::kOptimal` then 
- any KKT failures are logged
- if the relative KKT errors are more than `max_allowed_tolerance_relative_violation` (currently 100) times the corresponding tolerance, then `model_status` is set to `HighsModelStatus::kUnknown` 

Finish this later!


filikat:
Thanks, this looks fine. I may need to switch off logging if I call this function to check the solution within HiPO, but we'll figure that out later.

> 
> Finish this later!

Should we leave it as a draft, or merge it already?

PR #2819, opened 2026-02-06
Expand param description

Opt-Mucca:
This comes via a request from @svigerske in #2678 

@svigerske I don't think there's any desire to expand this parameter to something like "don't separate for nodes after a certain depth". Do you think values outside default and 0 are commonly used?

@galabovaa Does this automatically get updated in the documentation?

### DIFF
--- a/highs/lp_data/HighsOptions.h
+++ b/highs/lp_data/HighsOptions.h
@@ -1219,8 +1219,8 @@ class HighsOptions : public HighsOptionsStruct {
 
     record_bool = new OptionRecordBool(
         "mip_allow_cut_separation_at_nodes",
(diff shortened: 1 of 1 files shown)


svigerske:
If one wants to limit separation, but not only to the root node, then something like a depth limit seems suitable. But no, I don't think this is commonly done.

jajhall:
> @galabovaa Does this automatically get updated in the documentation?

Not currently, as some options in `docs/src/options/definitions.md` have an `@id` tag so they can be referenced elsewhere in the documentation. Simplest if they all have such a tag, so `definitions.md` can be entirely auto-generated

PR #2825, opened 2026-02-08
Change names to Metis api

### LINKED ISSUE #2823: Conflict with METIS symbols
The release v1.13.0 of `HiGHS` contains two common symbols of METIS, `METIS_NodeND` and `METIS_SetDefaultOptions`.
I don't know if you are using exactly the same signature but with a static linking they can collapse with symbols of the original METIS 5.1.0.
For the nonlinear optimization solver `Uno`, we need to link it with `libdmumps.a`, `libmetis.a`, `libhighs.a`, `libhsl.a`, etc...
I suggest to add a suffix or prefix for these two symbols.
All platforms are impacted.

Reference: https://github.com/JuliaPackaging/Yggdrasil/pull/13067

filikat:
Change names to Metis API within HiGHS, to avoid conflicting with external Metis when HiGHS is linked in some other project.

Closes #2823.

### DIFF
--- a/extern/metis/libmetis/auxapi.c
+++ b/extern/metis/libmetis/auxapi.c
@@ -17,7 +17,7 @@
     \param options points to an array of size at least METIS_NOPTIONS.
 */
 /*************************************************************************/
-int METIS_SetDefaultOptions(idx_t *options)
+int Highs_METIS_SetDefaultOptions(idx_t *options)
 {
   iset(METIS_NOPTIONS, -1, options);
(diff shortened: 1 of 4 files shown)


filikat:
@amontoison Is this ok? Let me know if anything else needs to be changed.
If the name conflict is a big problem, we can make a patch, but the next release should come shortly anyway.

jajhall (approved):
All fine by me

Issue #2826, opened 2026-02-09
Timer fields in `simplex/HApp.ts` can cause a buffer underflow


Bubblyworld:
I've been playing with a WASM build of HiGHS v1.13.0 as part of a [new project](https://github.com/Bubblyworld/scip-ts) I'm working on, and hit a little snag when running it in Chrome. I _think_ the root cause is as follows:

The code at [`highs/simplex/HApp.h:154-155`](https://github.com/ERGO-Code/HiGHS/blob/1bce6d5c801398dab6d2e6f98ac8935f3d4eec9c/highs/simplex/HApp.h#L154-L155) and [`highs/simplex/HApp.h:76-77`](https://github.com/ERGO-Code/HiGHS/blob/1bce6d5c801398dab6d2e6f98ac8935f3d4eec9c/highs/simplex/HApp.h#L76-L77) tracks the time spent in a particular sub-solver by setting `run_time` to `-timer.read()` at the start of the solve, and adding `timer.read()` to `run_time` at the end of the solve. Logically this makes sense, but depending on the speed of the solve and the precision of the timer you can have a situation where `timer.read()` returns the same value for both calls. This results in `run_time` being set to `-0.0` at the end of the solve.

I can trigger this code path in Chrome, for instance, as Chrome [deliberately truncates high-precision timer resolution](https://developer.chrome.com/blog/cross-origin-isolated-hr-timers) (to mitigate side-channel attacks like spectre, I believe).

The bug is that the code checks whether `run_time < 0` to determine the active sub-solver - see [`highs/simplex/HApp.h:47-58`](https://github.com/ERGO-Code/HiGHS/blob/1bce6d5c801398dab6d2e6f98ac8935f3d4eec9c/highs/simplex/HApp.h#L47-L58). According to the IEEE standard `-0.0` is _not_ less than `0`, so `sub_solver_ix` never gets set to the correct index and eventually leads to out-of-bounds read/writes at [`highs/simplex/HApp.h:75-77`](https://github.com/ERGO-Code/HiGHS/blob/1bce6d5c801398dab6d2e6f98ac8935f3d4eec9c/highs/simplex/HApp.h#L75-L77).

[...]

PR #2827, opened 2026-02-09
fix: replace less-than-zero check with std::signbit

Bubblyworld:
See #2826


jajhall:
PRs should be to latest, leaving master for releases and simple, critical patches

jajhall:
test-clang-format fails, so this will have to be merged into an interim branch and fixed

Bubblyworld:
Ah, my apologies for the formatting/branch - thanks for your time.

PR #2832, opened 2026-02-09
Avoid conflict with amd and rcm api

filikat:
Same as #2825, but for amd and rcm orderings.


filikat:
The functions that call printf within amd are actually not used, and I don't see them ever being useful for us. I will just remove them so we can get rid of all mentions of printf within amd.

filikat:
Actually, metis and rcm also have many printf and they may be useful to report errors. 
I will have to define a local print function to be used in the ordering heuristics.
I will do this as a separate pr though, and leave this to be only the name changes.

jajhall:
> The functions that call printf within amd are actually not used, and I don't see them ever being useful for us. I will just remove them so we can get rid of all mentions of printf within amd.

Better still

PR #2835, opened 2026-02-09
Remove printf from ordering heuristics

filikat:
- Define a macro `HIGHS_ORDERING_PRINT` that does nothing in release, and uses `printf` to print a message in debug.
- Replace all instances of `printf` in metis, amd and rcm with the new macro.
- The word "printf" now does not appear anywhere in the folders `extern/metis`, `extern/amd`, `extern/rcm`.


jajhall:
Is this still draft @filikat?

Issue #2824, opened 2026-02-08
How to not compile binaries / examples of HiGHS with CMake?


amontoison:
I prepared a script to cross-compile all dependencies of Uno as static libraries and I would like to avoid any call to the linker.
Do you an option to do that with `CMake`?
I use the following script:
```shell

mkdir build
cd build
if [[ "${target}" == *-mingw* ]]; then
    LIBGFORTRAN=libgfortran-5
else
    LIBGFORTRAN=libgfortran
fi
cmake .. \
    -DCMAKE_INSTALL_PREFIX=${prefix} \
    -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
    -DCMAKE_BUILD_TYPE=Release \
    -DBUILD_SHARED_LIBS=OFF \
    -DHIPO=ON \
    -DBUILD_TESTING=OFF \
    -DUILD_CXX_EXE=OFF \
    -DBLAS_LIBRARIES="${prefix}/lib/libcblas.a;${prefix}/lib/libblas.a;${libdir}/${LIBGFORTRAN}.${dlext}"

make -j ${nproc}
make install
```
The last line `-DBLAS_LIBRARIES="${prefix}/lib/libcblas.a;${prefix}/lib/libblas.a;${libdir}/${LIBGFORTRAN}.${dlext}"` was required for the linker even if don't want the binary `highs` for `libuno`.

@odow It can be also relevant for `HiGHS_jll,jl`.
We probably don't need to compile the examples.
For example, the option `-DBUILD_TESTING=OFF` could be already used.
It will reduce the size of the tarballs.

jajhall:
Can this be accommodated @galabovaa ?

galabovaa:
In your script there is a typo @amontoison, to switch the executable off it should be 

```
    -DBUILD_CXX_EXE=OFF \
```

amontoison:
Thanks @galabovaa !
The bug was between the chair and the screen :)

But now I have another issue:
```shell
[...]

PR #2836, opened 2026-02-09
Allow static build of HiGHS with HiPO without specifying BLAS

galabovaa:


jajhall:
Closes #2824

jajhall:
Unfortunately the build-macos failure looks genuine

galabovaa:
Should be good to merge now

PR #2838, opened 2026-02-10
Independent logging to console and/or file, and use of logging callback

### LINKED ISSUE #2816: Make "log to stdout" and "logging callback" independent for 1.14
https://github.com/google/or-tools/pull/5023

jajhall:
Logging to console and/or file still performed (according to options), even if logging callback is enabled.

This closes #2816

### DIFF
--- a/check/TestCallbacks.cpp
+++ b/check/TestCallbacks.cpp
@@ -8,7 +8,7 @@
 #include "lp_data/HConst.h"
 #include "lp_data/HighsCallback.h"
 
-const bool dev_run = false;
+const bool dev_run = true;  // false;
 
 const double egout_optimal_objective = 568.1007;
 const double egout_objective_target = 610;
--- a/highs/io/HighsIO.cpp
+++ b/highs/io/HighsIO.cpp
@@ -124,23 +124,24 @@ void highsLogUser(const HighsLogOptions& log_options_, const HighsLogType type,
       log_options_.user_log_callback ||
       (log_options_.user_callback && log_options_.user_callback_active);
 
-  if (!use_log_callback) {
-    // Write to log file stream unless it is NULL
-    if (log_options_.log_stream) {
-      if (prefix)
-        fprintf(log_options_.log_stream, "%-9s", HighsLogTypeTag[(int)type]);
-      vfprintf(log_options_.log_stream, format, argptr);
-      if (flush_streams) fflush(log_options_.log_stream);
-      va_end(argptr);
-      va_start(argptr, format);
-    }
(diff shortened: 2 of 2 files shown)


filikat:
Was this supposed to go to master?

jajhall:
No, my error. Fortunately @galabovaa will merge latest into master to create the patch

svigerske:
So I suppose it is this change, that with the move from 1.13.0 to 1.13.1 I now get a lot of output twice or not at all, and the HiGHS header to stdout.
Can one summarize what users will now need to change in their interfaces to adapt to this change in logging in HiGHS?

I'm setting a callback and ask it to be called for `HighsCallbackType::kCallbackLogging`, among other things.
It seems to be called twice for each line of output from HiGHS. Is it because it is a C-style callback?

Further, the header
```
Running HiGHS 1.13.1 (git hash: 1d267d97c): Copyright (c) 2026 under Apache 2.0 license terms
Using BLAS: ...
```
goes to stdout instead of the callback.

If I set one or both of `log_to_console` and `output_flag` to false, the HiGHS log completely disappears, but I don't manage to get things only through my log callback and only once per output.

svigerske:
If I repair
```cpp
void highsLogUser(const HighsLogOptions& log_options_, const HighsLogType type,
                  const char* format, ...) {
  if (!*log_options_.output_flag ||
      (log_options_.log_stream == NULL && !*log_options_.log_to_console))
    return;
...
```
to also not just return when there are user callbacks, and set log_to_console=false, I seem to get back to the old behavior. Will open a PR.

jajhall:
> So I suppose it is this change, that with the move from 1.13.0 to 1.13.1 I now get a lot of output twice or not at all, and the HiGHS header to stdout. Can one summarize what users will now need to change in their interfaces to adapt to this change in logging in HiGHS?
> 
> I'm setting a callback and ask it to be called for `HighsCallbackType::kCallbackLogging`, among other things. It seems to be called twice for each line of output from HiGHS. Is it because it is a C-style callback?
> 
> Further, the header
> 
> ```
> Running HiGHS 1.13.1 (git hash: 1d267d97c): Copyright (c) 2026 under Apache 2.0 license terms
> Using BLAS: ...
> ```
> 
> goes to stdout instead of the callback.
> 
> If I set one or both of `log_to_console` and `output_flag` to false, the HiGHS log completely disappears, but I don't manage to get things only through my log callback and only once per output.

Indeed, you're right, and I immediately realised what I'd failed to do!

If you've already started preparing your PR, go ahead, otherwise I'll fix it.

galabovaa:
Just FYI @Mizux

PR #2849, opened 2026-02-12
do not return early in highsLogUser if there are (log) callbacks

svigerske:
With 1.13.1, HiGHS stopped printing output through log callbacks when one disables log_to_console; https://github.com/ERGO-Code/HiGHS/pull/2838#issuecomment-3891286547.

This fixes that by not returning early in highsLogUser if callbacks are around.

One may still want to fix that with log_to_console enabled and callbacks present, the callbacks are called twice, instead of sending the log once to the callback and once to the console.

### DIFF
--- a/highs/io/HighsIO.cpp
+++ b/highs/io/HighsIO.cpp
@@ -108,8 +108,12 @@ std::array<char, 32> highsDoubleToString(const double val,
 
 void highsLogUser(const HighsLogOptions& log_options_, const HighsLogType type,
                   const char* format, ...) {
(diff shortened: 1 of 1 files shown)


jajhall:
Thanks, but the desired behaviour is to have console/file logging and callback logging completely independent. Only if both are off does HighsLogUser (and HighsLogDev) return immediately.

I'll fix this, and create a unit test

svigerske:
OK. I didn't read until highsLogDev. :)

I'm happy if you just take over this PR, or start a new one.

I've a fix that works for me now.

PR #2854, opened 2026-02-15
Hipo QP solver

filikat:
- Added logic to solve QPs with HiPO.
- Introduced a new solver option "qpasm", which is the default QP solver for now. Selecting "hipo" or "ipm" instead uses HiPO.
- The HiPO preprocessor has been refactored and made more powerful, since HiGHS does not have presolve for QPs.
- The tests in TestQpSolver.cpp now run both with qpasm and hipo.
- Since now there are options to choose both LP and QP solvers, give a warning if the selected solver is not compatible with the type of the problem, and ignore the option.
- Added some functions to HighHessian to access its diagonal and do matrix-vector products.
- Introduced some logic to better detect stagnation for problems that would otherwise keep iterating forever.
- Fixed an assertion in HighsSparseMatrix.

### DIFF
--- a/highs/ipm/hipo/ipm/Model.cpp
+++ b/highs/ipm/hipo/ipm/Model.cpp
@@ -262,7 +262,7 @@ void Model::print(const LogHighs& log) const {
   if (qp()) {
     log_stream << textline("Range of Q:") << "[" << sci(Qmin, 5, 1) << ", "
                << sci(Qmax, 5, 1) << "], ratio ";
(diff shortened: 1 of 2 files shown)


jajhall:
Please update FEATURES.md with respect to this PR, @filikat Just update latest...

jajhall:
> I'll now see what happens if I merge in fix-2781 and fix-2821

Not as bad as I feared! 😅

PR #2858, opened 2026-02-17
Fix warnings

filikat:
Get rid of compilation warnings in HiPO, due to undeclared functions in extern/rcm.cpp and due to comparisons of HighsInt and size_t.


odow:
Is it worth turning warnings into errors in CI so we can avoid things like this in the future?

filikat:
> Is it worth turning warnings into errors in CI so we can avoid things like this in the future?

The CI tests didn't show warnings, at least the ones that I checked. I think it only happens for specific machine, OS and compiler.

PR #2859, opened 2026-02-18
Console/file/callback logging now independent

### LINKED ISSUE #2816: Make "log to stdout" and "logging callback" independent for 1.14

jajhall:
This closes #2816  (correctly!)

It also fixes compiler warnings in cuPDLP-C by making local functions static, and prevents the IPX logging from appearing when IPX is used by HiPO and `output_flag` is false. Hence @filikat as reviewer!

### DIFF
--- a/check/TestIO.cpp
+++ b/check/TestIO.cpp
@@ -1,5 +1,6 @@
 #include <cstdio>
 #include <cstring>
+#include <fstream>
 
 #include "HCheckConfig.h"
 #include "Highs.h"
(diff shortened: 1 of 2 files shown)


highs/ipm/hipo/ipm/Solver.cpp
```diff
@@ -193,6 +193,8 @@ bool Solver::prepareIpx() {
 
   ipx::Parameters ipx_param;
   ipx_param.display = options_.display_ipx;
```
filikat, line 195: I think it's better to set `ipx_param.display = options_.display_ipx && options_.display;`
I have it fixed in another branch anyway.

filikat: The option display_ipx was there mostly for development really. I can just remove it and use the same diplay options for HiPO and IPX.

jajhall: Go ahead. I'm finished with this branch so you can pull and modify it

filikat: I will do it in another PR soon. You can merge this for now.

Issue #2806, opened 2026-02-04
HiGHS 1.13 Terminates MIP Prematurely Similar to Issue #1865 Using Same Problem


jeffreydeankelly2:
Dear HiGHS;

Below is the log output for same issue reported in #1865.

The correct objective function is 937000.0.

In addition, 1.13 takes 319-seconds to solve prematurely and 1.12 solves to 937000.0 in 34-seconds.

All the best - Jeff

Running HiGHS 1.13.0 (git hash: 1bce6d5): Copyright (c) 2026 under MIT licence terms
Cols:           2 lower bounds    less than or equal to       -1e+20 are treated as -Infinity
Cols:           2 upper bounds greater than or equal to        1e+20 are treated as +Infinity
Rows:       25170 lower bounds    less than or equal to       -1e+20 are treated as -Infinity
SOLVE
MIP has 33392 rows; 15608 cols; 318754 nonzeros; 5538 integer variables (5538 binary)
Coefficient ranges:
  Matrix  [1e-04, 2e+04]
  Cost    [1e+00, 1e+00]
  Bound   [5e-01, 1e+04]
  RHS     [1e+00, 2e+04]
Presolving model
27851 rows, 10067 cols, 306660 nonzeros  0s
19594 rows, 9535 cols, 288379 nonzeros  0s
15495 rows, 7822 cols, 279873 nonzeros  0s
8821 rows, 5153 cols, 265826 nonzeros  0s
4916 rows, 3037 cols, 64501 nonzeros  75s
4483 rows, 2366 cols, 61441 nonzeros  106s
Presolve reductions: rows 4483(-28909); columns 2366(-13242); nonzeros 61441(-257313)

Src: B => Branching; C => Central rounding; F => Feasibility pump; H => Heuristic;
     I => Shifting; J => Feasibility jump; L => Sub-MIP; P => Empty MIP; R => Randomized rounding;
     S => Solve LP; T => Evaluate node; U => Unbounded; X => User solution; Y => HiGHS solution;
     Z => ZI Round; l => Trivial lower; p => Trivial point; u => Trivial upper; z => Trivial zero
         0       0         0   0.00%   1573466.666667  -inf                 inf        0      0      0         0   106.7s
         0       0         0   0.00%   1083591.573449  -inf                 inf        0      0      4      2244   107.0s
         0       0         0   0.00%   1082738.9208    -inf                 inf     4208    306    204      2885   113.0s
[...]

PR #2861, opened 2026-02-19
Speed up enumeration presolve

fwesselm:
These changes resolve the slowdown reported in #2806 
- Reuse probing scores to sort candidate rows
- Avoid enumerating solutions for similar rows (overlapping binary variables)
- Stop if enumeration did not produce any reductions after a number of consecutive tries
- In my experiments the new code behaves very similar in terms of relative shifted geometric means (1.0 for running times and 0.99 for branching nodes). However, it is much faster on cases like the one reported in #2806.
- I cleaned up some code in the `HighsCliqueTable` class along the way.


jajhall:
Thanks: looking good on CI. I'll see what it looks like on the problems in #2478 ([model.mps](https://github.com/user-attachments/files/25399303/model.mps.txt)) and #2806 ([gaswellrig.mps](https://github.com/user-attachments/files/25099956/gaswellrig.mps.txt)) for the probing logging and timeout facility I've just written!

PR #2862, opened 2026-02-19
Incorporate kktCheck into HiPO

filikat:
HiPO now checks if the kktCheck passes before returning a solution, if crossover is not run.
If HiPO finds a primal-dual feasible solution, but kktCheck disagrees, more iterations are performed until either a solution is found that satisfies the check, or the solver stagnates, or IPX is called. The check is not performed on the solution returned by IPX. 
If crossover is requested, no check is performed, because a tighter tolerance is required by the solver already.

Option parsing has been moved inside of hipo::Solver rather than in the wrapper. Final residuals are no longer printed in the summary.

### DIFF
--- a/highs/ipm/hipo/ipm/Solver.cpp
+++ b/highs/ipm/hipo/ipm/Solver.cpp
@@ -19,7 +19,7 @@ Int Solver::load(const HighsLp& lp, const HighsHessian& Q) {
   return kStatusOk;
 }
 
-Int Solver::setOptions(const HighsOptions& highs_options) {
+void Solver::setOptions(const HighsOptions& highs_options) {
   options_.display = true;
   if (!highs_options.output_flag | !highs_options.log_to_console)
     options_.display = false;
@@ -56,8 +56,6 @@ Int Solver::setOptions(const HighsOptions& highs_options) {
   options_orig_ = options_;
   Hoptions_ = highs_options;
   resetOptions();
-
-  return kStatusOk;
 }
 
 void Solver::resetOptions() {
--- a/highs/ipm/IpxWrapper.cpp
+++ b/highs/ipm/IpxWrapper.cpp
@@ -476,10 +476,7 @@ HighsStatus solveHipo(const HighsOptions& options, HighsTimer& timer,
                  1e-1 * options.kkt_tolerance);
   }
 
-  if (hipo.setOptions(options)) {
-    model_status = HighsModelStatus::kSolveError;
-    return HighsStatus::kError;
-  }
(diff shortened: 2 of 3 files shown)


highs/ipm/hipo/ipm/Solver.cpp
```diff
@@ -17,10 +19,101 @@ Int Solver::load(const HighsLp& lp, const HighsHessian& Q) {
...
+    options_.feasibility_tol = highs_options.kkt_tolerance;
+    options_.optimality_tol = 1e-1 * highs_options.kkt_tolerance;
+    options_.crossover_tol = 1e-1 * highs_options.kkt_tolerance;
+  }
+
+  // hipo uses same timer as highs, so it is fine to pass the same time limit
+  options_.time_limit = highs_options.time_limit;
+
+  options_.max_iter =
+      highs_options.ipm_iteration_limit - highs_info.ipm_iteration_count;
```
jajhall: IPM iteration limit is limit for a particular solve, not total "all time" iterations of IPM, so subtracting `highs_info.ipm_iteration_count` shouldn't be done. Hence `HighsInfo& highs_info` isn't needed in this method

highs/ipm/hipo/ipm/Solver.cpp
```diff
@@ -17,10 +19,101 @@ Int Solver::load(const HighsLp& lp, const HighsHessian& Q) {
...
+  } else if (highs_options.hipo_system == kHighsChooseString) {
+    options_.nla = hipo::kOptionNlaChoose;
+  } else {
+    highsLogUser(highs_options.log_options, HighsLogType::kError,
+                 "Unknown value of option %s\n", kHipoSystemString.c_str());
+    return kStatusError;
+  }
+
+  // Reordering heuristic
+  if (highs_options.hipo_ordering != kHipoMetisString &&
```
jajhall: Checking the validity of string options is done by `Highs::setOptionValue`, since this must not return `HighsStatus::kOk` if the option value is not valid.

I'll leave you to add the necessary calls and methods to `lp_data/HighsOptions.h/cpp`, along the lines of what's in the version of `checkOptionValue` for strings.

For sanity you can then leave an assert in `Solver::setOptions`, but it can return void, hence you don't have to handle an error return below the call.

filikat:
I changed all the options in HiPO from enum to string, so that they can be passed directly from the highs options. I added functions to check the validity of hipo_parallel_type, hipo_system, hipo_ordering. The asserts are present when the options are actually used, in FactorHiGHSSolver.cpp.

Issue #2860, opened 2026-02-18
Add logging and timeout to `HPresolve::enumerateSolutions` and `HPresolve::runProbing`


jajhall:
`HPresolve::enumerateSolutions` and (in particular) `HPresolve::runProbing` can take a very long time. 

There must be some logging and a timeout check

fwesselm:
@jajhall, FYI, I am working on speeding up the solution enumeration.

jajhall:
Closed by #2865

PR #2865, opened 2026-02-21
Add logging and time budget for probing in presolve, and increase visibility of presolve rule switch-off

jajhall:
The potentially prohibitive high cost of probing means that there should be some logging so that HiGHS doesn't go "quiet", and this has been added.

The high cost is mitigated by giving probing a budget of 10% of `time_limit`. It may make the MIP solver non-deterministic if `time_limit` is finite, but that's better than it "never" returning.

Since this may lead users to want to switch off probing, the ability to switch off more advanced/expensive presolve rules is now visible.

### DIFF
--- a/highs/presolve/HPresolve.cpp
+++ b/highs/presolve/HPresolve.cpp
@@ -1683,69 +1683,73 @@ HPresolve::Result HPresolve::runProbing(HighsPostsolveStack& postsolve_stack) {
       };
     }
 
-    assert(mipsolver);
-    // Don't log probing for presolve before restart
+    // Set up logging for probing
     const bool silent = silentLog();
     HighsInt iBin = -1;
     HighsInt iBin_probed = -1;
     HighsInt num_binary = binaries.size();
-    double log_tt_interval = 5.0;
-    double tt0 = this->timer->read();
+    double tt = this->timer->read();
+    double tt0 = tt;
     double log_tt = tt0;
     HighsInt log_iBin_probed = iBin_probed;
-    double time_remaining = options->time_limit - tt0;
-    double probing_time_limit = 0.1 * time_remaining;
-    bool logged_probing_time_limit = false;
-    for (const auto& binvar : binaries) {
-      iBin++;
-      HighsInt i = std::get<3>(binvar);
+    auto probingLog = [&]() {
+      if (silent || options->timeless_log) return;
+      // Ensure that enough time has elapsed, and at least on binary
+      // has been probed
+      const double log_tt_interval = 5.0;
+      if (tt > log_tt + log_tt_interval && iBin_probed > log_iBin_probed) {
+        // Get the average rate from the start of probing
+        assert(iBin_probed > 0);
+        double rate0 = (tt - tt0) / double(iBin_probed);
+        // Get the average rate since last logging
+        HighsInt dl_iBin_probed = iBin_probed - log_iBin_probed;
+        assert(dl_iBin_probed > 0);
+        double rate1 = (tt - log_tt) / double(dl_iBin_probed);
+        // Assess the time for probing based on the greater rate
+        double rate = std::max(rate0, rate1);
+        std::string rate_str =
+            " (rate " + highsTimeToString(1e3 * rate) + "/ms";
+        double expected_probing_finish_time =
+            tt + rate * (num_binary - iBin_probed);
+        std::string expected_probing_finish_time_str =
+            " => expected probing finish time " +
+            highsTimeSecondToString(expected_probing_finish_time) + ")";
+        std::string time_str = highsTimeSecondToString(tt);
+        highsLogUser(options->log_options, HighsLogType::kInfo,
+                     "   Considered %d / %d binaries; %d probed %s%s %s\n",
+                     int(iBin), int(num_binary), int(iBin_probed),
+                     rate_str.c_str(), expected_probing_finish_time_str.c_str(),
+                     time_str.c_str());
+        log_tt = tt;
+        log_iBin_probed = iBin_probed;
+      }
+    };
 
+    for (const auto& binvar : binaries) {
+      // Count the binaries considered
+      iBin++;
+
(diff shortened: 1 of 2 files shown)


jajhall:
Closes #2860

highs/presolve/HPresolve.cpp
```diff
@@ -1683,12 +1683,70 @@ HPresolve::Result HPresolve::runProbing(HighsPostsolveStack& postsolve_stack) {
...
     for (const auto& binvar : binaries) {
+      iBin++;
       HighsInt i = std::get<3>(binvar);
 
       if (cliquetable.getSubstitution(i) != nullptr || !domain.isBinary(i))
         continue;
 
+      iBin_probed++;
+      double tt = this->timer->read();
+      if (tt > log_tt + log_tt_interval && iBin_probed > log_iBin_probed) {
```
fwesselm: It seems that nothing in this `if` block is done when `silent = true`, so maybe the outer if-check could look like this:

`if (!silent && tt > log_tt + log_tt_interval && iBin_probed > log_iBin_probed)`

and then the other checks for `!silent` could be removed.

highs/presolve/HPresolve.cpp
```diff
@@ -1683,12 +1683,70 @@ HPresolve::Result HPresolve::runProbing(HighsPostsolveStack& postsolve_stack) {
...
+    const bool silent = mipsolver->mipdata_->numRestarts > 0;
+    HighsInt iBin = -1;
+    HighsInt iBin_probed = -1;
+    HighsInt num_binary = binaries.size();
+    double log_tt_interval = 5.0;
+    double tt0 = this->timer->read();
+    double log_tt = tt0;
+    HighsInt log_iBin_probed = iBin_probed;
+    double time_remaining = options->time_limit - tt0;
+    double probing_time_limit = 0.1 * time_remaining;
```
fwesselm: I understand that a time limit may be needed, but I also do not really like hard-coding things. However, having an option may also be too much.

highs/presolve/HPresolve.cpp
```diff
@@ -5468,35 +5523,33 @@ HPresolve::Result HPresolve::presolve(HighsPostsolveStack& postsolve_stack) {
     model->sense_ = ObjSense::kMinimize;
   }
 
+  const bool silent = mipsolver && mipsolver->mipdata_->numRestarts > 0;
+  if (options->presolve != kHighsOffString) {
```
fwesselm, line 5531: `if (!silent && options->presolve != kHighsOffString) `?

highs/presolve/HPresolve.cpp
```diff
@@ -1683,12 +1683,70 @@ HPresolve::Result HPresolve::runProbing(HighsPostsolveStack& postsolve_stack) {
...
+    double probing_time_limit = 0.1 * time_remaining;
+    bool logged_probing_time_limit = false;
     for (const auto& binvar : binaries) {
+      iBin++;
       HighsInt i = std::get<3>(binvar);
 
       if (cliquetable.getSubstitution(i) != nullptr || !domain.isBinary(i))
         continue;
 
+      iBin_probed++;
```
fwesselm, line 1737: Would it be worth putting the new code that checks the time limit into a lambda? This could also be done later - it would make reading the probing code easier for me.

highs/presolve/HPresolve.cpp
```diff
@@ -5468,35 +5523,33 @@ HPresolve::Result HPresolve::presolve(HighsPostsolveStack& postsolve_stack) {
...
-#else
-        std::string time_str =
-            " " + std::to_string(static_cast<int>(run_time)) + "s";
-#endif
+        std::string time_str = highsTimeSecondToString(run_time);
         if (options->timeless_log) time_str = "";
         highsLogUser(options->log_options, HighsLogType::kInfo,
                      "%" HIGHSINT_FORMAT " rows, %" HIGHSINT_FORMAT
-                     " cols, %" HIGHSINT_FORMAT " nonzeros %s\n",
+                     " cols, %" HIGHSINT_FORMAT " nonzeros  %s\n",
```
fwesselm: Is the extra space needed?

highs/presolve/HPresolve.cpp
```diff
@@ -5468,35 +5523,33 @@ HPresolve::Result HPresolve::presolve(HighsPostsolveStack& postsolve_stack) {
     model->sense_ = ObjSense::kMinimize;
   }
 
+  const bool silent = mipsolver && mipsolver->mipdata_->numRestarts > 0;
```
fwesselm: Would it be worth making this a method in `HPresolve.h`? Maybe not, but anyway:

  ```
bool isLogSilent() const {
    return mipsolver && mipsolver->mipdata_->numRestarts > 0;
  }
```

highs/presolve/HPresolve.cpp
jajhall: No, the block contains the return with `Result::kStopped;` if the probing time limit is reached. Since that is "lost" in what appears to be just logging tells me that I should refactor this section!

highs/presolve/HPresolve.cpp
jajhall: Agreed, but I felt that something was needed so that users for whom probing is very expensive could get some advantage from it.

Maybe it's better to let probing run to `options->time_limit` and print a message suggesting that probing in presolve is switched off.

highs/presolve/HPresolve.cpp
jajhall: Will do

highs/presolve/HPresolve.cpp
jajhall: No, I don't know how that's crept in

highs/presolve/HPresolve.cpp
jajhall: Would be more elegant

jajhall:
Thanks @fwesselm !

Issue #2773, opened 2026-01-26
Potential overhead reduction for dual fixing


ZhaoWeiWang0319:
Hi, I am reading the presolve code added in https://github.com/ERGO-Code/HiGHS/pull/2531
https://github.com/ERGO-Code/HiGHS/blob/3c26f3306c53c31a16b54a32610ac551fe31cd40/highs/presolve/HPresolve.cpp#L4616-L4626
In particular, in lines 4617 and 4623, for variables whose lock number is one, we check if they can be substituted by a binary variable.

One interesting case is when the single row marked by ``downLockRow`` or ``upLockRow`` is an equation. In this case, if fixing a binary variable x_k in this equation could deduce the equation redundant, then all variables in this equation must be fixed (since we use constraint activities to deduce redundancy). Substitution in this case, however, can be accessed by standard probing on x_k. Moreover, another technique called "simple probing" may also be able to deal with this (see 3.6. "Simple Probing on a Single Equation", Achterberg et al., Presolve Reductions in Mixed Integer Programming, INFORMS Journal on Computing 32(2):473-506). 
Therefore, I wonder if equations can be skipped here by 
```c++
if (mipsolver != nullptr && model->col_lower_[col] != -kHighsInf && 
     model->col_upper_[col] != kHighsInf) { 
   // try substitution 
   if (numDownLocks == 1 && downLockRow != -1 && model->row_lower_[downLockRow] != model->row_upper_[downLockRow]) { 
     HPRESOLVE_CHECKED_CALL(substituteCol(col, downLockRow, HighsInt{1}, 
                                          model->col_upper_[col], 
                                          model->col_lower_[col])); 
     if (colDeleted[col]) return Result::kOk; 
   } 
   if (numUpLocks == 1 && upLockRow != -1 && model->row_lower_[upLockRow] != model->row_upper_[upLockRow]) { 
     HPRESOLVE_CHECKED_CALL(substituteCol(col, upLockRow, HighsInt{-1}, 
                                          model->col_lower_[col], 
```
to reduce some computation overhead.

I know that probing generally would not probing on all binary variables, but dual fixing here will check all the variables.
Maybe some substitutions are lost after skipping equations, so this is just an advice for you to consider.

Finally, I am happy to commit a PR for this if needed! (●'◡'●)

jajhall:
Please would you have a look at this, @fwesselm

fwesselm:
Hi @ZhaoWeiWang0319!

[...]

PR #2869, opened 2026-02-24
Extension of dual fixing for single equations

fwesselm:
- Following a discussion in #2773, I am proposing to extend `HPresolve::dualFixing` to handle single equations (as described by Achterberg et al., Presolve Reductions in Mixed Integer Programming, see section 6.1).
- My test runs showed that the extension is not very effective (ratios of shifted geometric means of running times and numbers of nodes are very close to 1.0). However,  the extension is relatively cheap, so I would be in favor of adding it anyway.
- I also modified `HPresolve::enumerateSolutions` to compute the [overlap coefficient](https://en.wikipedia.org/wiki/Overlap_coefficient) when comparing two sets of binary variables.

### DIFF
--- a/highs/presolve/HPresolve.cpp
+++ b/highs/presolve/HPresolve.cpp
@@ -4725,8 +4725,8 @@ HPresolve::Result HPresolve::dualFixing(HighsPostsolveStack& postsolve_stack,
         computeActivity(rowNz.index(), rowNz.value(), activitySCMinus,
                         activitySCMinusFinite, HighsInt{1});
       // break if activities are not finite
-      if ((!activityTPlusFinite || !activitySCPlusFinite) &&
-          (!activityTMinusFinite || !activitySCMinusFinite))
+      if ((sMinus.empty() || !activityTPlusFinite || !activitySCPlusFinite) &&
+          (sPlus.empty() || !activityTMinusFinite || !activitySCMinusFinite))
         break;
     }
     // fix variables


highs/presolve/HPresolve.cpp
```diff
@@ -5222,7 +5350,7 @@ HPresolve::Result HPresolve::enumerateSolutions(
     while (true) {
       bool backtrack = domain.infeasible();
       if (!backtrack) {
-        backtrack = solutionFound(numVars);
+        backtrack = findBranchVar(numVars) < 0;
```
Opt-Mucca, line 5353: Nice spot on duplicate code!

highs/presolve/HPresolve.cpp
```diff
@@ -4684,20 +4802,34 @@ HPresolve::Result HPresolve::dualFixing(HighsPostsolveStack& postsolve_stack,
...
+          hasSingleDownLock && isEquation(downLockRow)
+              ? downLockRow
+              : (hasSingleUpLock && isEquation(upLockRow) ? upLockRow : -1);
+      if (equationRow != -1) {
+        // see section 6.1 "Extension of dual fixing for single equations",
+        // Achterberg et al., Presolve Reductions in Mixed Integer
+        // Programming, INFORMS Journal on Computing 32(2):473-506.
+        HPRESOLVE_CHECKED_CALL(handleSingleEquation(equationRow));
         if (colDeleted[col]) return Result::kOk;
+      } else if (mipsolver != nullptr && model->col_lower_[col] != -kHighsInf &&
```
Opt-Mucca, line 4818: A quick sanity check: Is there a solution path where the single equation would not fix any cols, e.g. due to infinite activities, and we are now skipping the substitution that we could have done in the original code?

Opt-Mucca:
FYI (in case notifications for non-direct tags are turned off) @ZhaoWeiWang0319

highs/presolve/HPresolve.cpp
fwesselm: To substitute the variable, we need to find a binary variable that, when set to zero, makes the row redundant. As far as I understand, setting a single binary variable to zero will not make an equation redundant. Therefore, to answer your question, the code should fail in finding a substitution in an equation anyway.

fwesselm:
> Code looks good! I've got the one open question and if that's simply me being paranoid then this should be good to merge.

Thank you, @Opt-Mucca!

highs/presolve/HPresolve.cpp
Opt-Mucca: Makes sense! Thanks for the answer.

ZhaoWeiWang0319:
Thanks, @Opt-Mucca! 
I wasn't receiving notifications before you mentioned it, much appreciated!

I am really happy to see that the discussion being helpful.
Huge thanks to @fwesselm  for the implementation and evaluation work!

fwesselm:
Let me know if there are other questions or things to address here. Otherwise, I think it's good to merge.

PR #2870, opened 2026-02-24
Add timeout support for IIS search

haman80:
Add timeout support to getIIS
- Introduced a new iis_time_limit option to distinguish IIS search limits from the time_limit option
- Updated deletion filtering to return a partial IIS when the time limit is reached
- Added new model status kIisModelStatusTimeLimit

Improve handling of unexpected model statuses

- Temporarily modify options that may interfere with IIS search, including: simplex_iteration_limit, objective_bound, allow_unbounded_or_infeasible, Callbacks
- Updated getIisInterface to first establish infeasibility prior to elastic filtering or deletion filtering; abort IIS computation if the model is not confirmed infeasible
- After confirming infeasibility, attempt elastic filtering if requested; if it fails due to an unknown status, fall back to deletion filtering over the full set of row and column bounds
- Enhanced deletion filtering to skip bounds whose removal results in an unknown model status and return a partial IIS if possible

Enhancements to HighsIis

- Extended HighsIisInfo to track min, max, and mean iteration counts and LP solve times
- Centralize responsibility for updating these metrics
- Added helper methods determineBoundStatus, and processBoundRelaxation to reduce duplicated logic and improve maintainability

### DIFF
--- a/check/TestIis.cpp
+++ b/check/TestIis.cpp
@@ -27,8 +27,8 @@ void reportIisInfo(const HighsLogOptions& log_options,
                int(iis_info.max_simplex_iteration_count));
   highsLogUser(log_options, HighsLogType::kInfo,
                "   Simplex time (min / sum / max) = %g / %g / %g\n",
-               int(iis_info.min_simplex_time), int(iis_info.sum_simplex_times),
-               int(iis_info.max_simplex_time));
+               iis_info.min_simplex_time, iis_info.sum_simplex_times,
+               iis_info.max_simplex_time);
 }
 void testMps(std::string& model, const HighsInt iis_strategy,
              const HighsModelStatus require_model_status =


jajhall:
@haman80 I've cleared the merge conflicts and pushed to Git

haman80:
Hi @jajhall ,
Apologies for the additional commits. I added some unit tests to improve coverage, including a couple of new infeasible LP instances (cplex1 and vol1). The vol1 case was initially exhibiting inconsistent behavior across machines due to ill conditioning, which required some iteration to stabilize. I look forward to hearing your thoughts.

highs/lp_data/HighsInterface.cpp
```diff
@@ -1895,75 +1916,137 @@ HighsStatus Highs::getIisInterface() {
...
+    double simplex_time = -this->getRunTime();
+    HighsInt simplex_iterations = -info_.simplex_iteration_count;
     HighsStatus run_status = this->optimizeModel();
     options_.presolve = presolve;
-    if (run_status != HighsStatus::kOk) return run_status;
-    iis_info.simplex_time += this->getRunTime();
-    iis_info.simplex_iterations += -info_.simplex_iteration_count;
-    this->iis_.info_.push_back(iis_info);
+    simplex_time += this->getRunTime();
+    simplex_iterations += -info_.simplex_iteration_count;
```
jajhall: This should be as follows: it's a cut-and-paste error of mine that you inherited

simplex_iterations += info_.simplex_iteration_count;

jajhall: Fixed in latest commit

haman80:
> Partly to understand the flow of your modified code, I've added a couple more unit tests to `TestIis.cpp`. These expose a few minor issues (tagged with "fix-2870") that need resolving

@jajhall Thank you so much for the review. I pushed another commit addressing the issues tagged with "fix-2870". Please feel free to take a look and let me know your thoughts.

jajhall (approved):
This addresses my original request for changes, so is good to be merged once the CI tests pass.

haman80:
> This addresses my original request for changes, so is good to be merged once the CI tests pass.

@jajhall Thanks again for the review. I fixed the CI failure, which was due to a format-type mismatch in `reportIisInfo` in `TestIis.cpp` (passing int to %g instead of double). I will go ahead and merge.

PR #2871, opened 2026-02-25
Fix failing HiPO CI tests

filikat:
We still had some CI tests using GKlib and Metis, which where fetching from the original repository. These are now failing due to small changes to their build system.

### DIFF
--- a/.github/workflows/hipo.yml
+++ b/.github/workflows/hipo.yml
@@ -14,24 +14,6 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Checkout metis overlay ports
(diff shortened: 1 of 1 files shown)


galabovaa:
Thank you, @filikat!

filikat:
I think the failure of the wheels is unrelated. It's failing in another PR as well.

galabovaa:
Yes, I just re-ran them and they are still failing, I will look into those separately

jajhall:
Can this be merged now, @galabovaa / @filikat ?

filikat:
I think so, the wheels test failures is unrelated. Maybe wait for @galabovaa to confirm

PR #2878, opened 2026-02-27
Add indicator constraints

mcoughlin:
This PR adds indicator constraints.

### DIFF
--- a/check/TestIndicatorConstraints.cpp
+++ b/check/TestIndicatorConstraints.cpp
@@ -222,3 +222,100 @@ TEST_CASE("indicator-mps-read", "[highs_test_indicator]") {
   REQUIRE(fabs(solution.col_value[1] - 0.0) < double_equal_tolerance);
   REQUIRE(fabs(info.objective_function_value - 0.0) < double_equal_tolerance);
 }
+
+TEST_CASE("indicator-infinite-bounds", "[highs_test_indicator]") {
+  // Test indicator with unbounded variables (would produce incorrect
+  // results under big-M with M=1e8 for infinite bounds)
+  // min x
+  // s.t. z=1 -> x >= 5
+  //      z binary, x >= 0 (no upper bound)
+  // Optimal: z=0, x=0, obj=0
+  Highs highs;
+  const HighsInfo& info = highs.getInfo();
+  highs.setOptionValue("output_flag", dev_run);
+
+  highs.addVar(0.0, inf);   // x (col 0) - no upper bound
+  highs.addVar(0.0, 1.0);   // z (col 1)
+
+  highs.changeColCost(0, 1.0);
+  highs.changeColCost(1, 0.0);
+  highs.changeColIntegrality(1, HighsVarType::kInteger);
+
+  // Indicator: z=1 -> x >= 5
+  HighsInt indices[] = {0};
+  double values[] = {1.0};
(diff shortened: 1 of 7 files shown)


Opt-Mucca:
Thanks for the well structured PR! As this is a major change I'd like to address high level stuff first:
- The general design looks good and follows the existing structure. I like how easy it is to following the semi-continuous design.
- I'm hesitant to add the automated linear reformulation. I see this inevitably ending up with incorrect results for users (from assuming some infinite value and from bad numerics).
- I've always planned to add indicators that transform to SOS constraints under the hood. That way we could reformulate all indicators, linearise those that are numerically friendly, and then easily handle the unfriendly ones via branching (and also potentially add some SOS specific cuts). This would also allow us to potentially generalise to `disjunctive constraints` if the need ever arises.

jajhall:
I've skimmed through the code changes on my phone, but not looked in general at the reformulation, and echo the comments of @Opt-Mucca.

I've changed the base to latest since master is only for releases, and we should also get the view of @fwesselm. 

Otherwise, I'll leave the PR in the hands of @Opt-Mucca

mcoughlin:
@Opt-Mucca Best I can tell SOS constraints aren't implemented, so does this need to be a big-M reformulation? Or what do you imagine?

jajhall:
> @Opt-Mucca Best I can tell SOS constraints aren't implemented, so does this need to be a big-M reformulation? Or what do you imagine?

Indeed we can't handle SOS in any way. There is a plan to handle them as a special case of disjunctive constraints, and linearise the latter. However, as @Opt-Mucca says, there is value in handling SOS constraints explicitly in the MIP solver.

mcoughlin:
@Opt-Mucca I think this is more aligned with your vision now.

jajhall (commented):
There's no point in formulating indicator constraints as SOS. I know that's what @Opt-Mucca said, but we can't handle SOS explicitly in the MIP solver, and won't be able to do so for a long time.

We discussed your original PR today, and will accept it

mcoughlin:
@jajhall Yes I was trying to work through @Opt-Mucca's suggestion. 513 worked on SOS and then 202 indicators. Thanks for looking at the original.

jajhall:
I've just tried to create a local branch from latest with [513ad06](https://github.com/ERGO-Code/HiGHS/pull/2878/commits/513ad06f6f9f4c2448f024089f694be0fcf6598a) merged into it, but it picked up [2023f89](https://github.com/ERGO-Code/HiGHS/pull/2878/commits/2023f89009cb70e73502289f7e0e77ce5bd0e659). I also see that this has closed #2878. Since I deleted the local branch, I can't revert the merge.

We can probably merge [513ad06](https://github.com/ERGO-Code/HiGHS/pull/2878/commits/513ad06f6f9f4c2448f024089f694be0fcf6598a) from your fork into a local branch of ours, but maybe you could recreate your original PR, thanks!

Opt-Mucca:
@mcoughlin Sorry for the confusion (I was just trying to give the long term vision I have for indicator constraints).

The general flow you started with was good, and is exactly the design one should go with for neatly adding indicator constraints with minimal code changes. Long term we'd like to add "native" handling of such constraints, but this will take ages.
I personally use SOS constraints a fair bit, and from a solver perspective believe they're super important for generalising constraints, but indicator usage from practitioners dwarfs that of SOS (and if anything is growing in popularity).

mcoughlin:
@Opt-Mucca yeah that makes sense.!

PR #2881, opened 2026-03-02
Piecewise Linear Constraints

mcoughlin:
This PR adds Piecewise Linear Constraints

### DIFF
--- a/check/CMakeLists.txt
+++ b/check/CMakeLists.txt
@@ -73,6 +73,7 @@ if ((NOT FAST_BUILD OR ALL_TESTS) AND NOT (BUILD_EXTRA_UNIT_ONLY))
       TestRanging.cpp
       TestRays.cpp
       TestSemiVariables.cpp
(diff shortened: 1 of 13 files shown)


jajhall:
I've discussed this with @opt-mucca, and PWL constraints are clearly beyond what we can incorporate at this time

mcoughlin:
@jajhall Ok thanks, indicator constraints is the big one for my use case, piecewise linear can be done in the formulation. Thanks for considering.

jajhall:
> @jajhall Ok thanks, indicator constraints is the big one for my use case, piecewise linear can be done in the formulation. Thanks for considering.

We discussed your PRs last night at a HiGHS-JuMP meeting, and reiterated the view that HiGHS shouldn't handle piecewise linear constraints. The argument was that they are not defined for solvers to handle, aren't covered in any major file format, and should be modelled using more fundamental objects.

PR #2886, opened 2026-03-02
MIP Worker Refactor

Opt-Mucca:


Opt-Mucca:
@fwesselm Sorry for not being able to break this up into smaller parts. Please read `parallel.md` before diving into the changes.
There will for sure be some bugs in a change this large (probably going to be disabled by default on the first version regardless of whether or not we can find them). What I'm most looking for is whether or not you agree with the design decisions.

Opt-Mucca:
@jajhall The two issues with the timers:
- In the MIP code there's plenty of timers that can't be called because they are now potentially being called multiple times (It's usually those that measure one heuristic or cut generation etc. I've commented those out for now). I didn't want to remove them as I regularly use the timers for development.
- There is some LP timer buried fairly deep that causes sanitizer issues from solving so many LPs on different threads.

highs/mip/HighsCutGeneration.cpp
```diff
@@ -755,7 +756,7 @@ double HighsCutGeneration::scale(double val) {
   return std::ldexp(1.0, expshift);
 }
 
-bool HighsCutGeneration::postprocessCut() {
+bool HighsCutGeneration::postprocessCut(HighsDomain& globaldom) {
```
fwesselm: `const HighsDomain& globaldom`?

highs/mip/HighsCutGeneration.cpp
```diff
@@ -1206,6 +1205,7 @@ bool HighsCutGeneration::generateCut(HighsTransformedLp& transLp,
 }
 
 bool HighsCutGeneration::generateConflict(HighsDomain& localdomain,
```
fwesselm: `const HighsDomain& localdomain`?

highs/mip/HighsCutGeneration.cpp
```diff
@@ -1206,6 +1205,7 @@ bool HighsCutGeneration::generateCut(HighsTransformedLp& transLp,
...
+                                          HighsDomain& globaldom,
```
fwesselm: `const HighsDomain& globaldom`?

highs/mip/HighsCutGeneration.h
```diff
@@ -67,7 +67,7 @@ class HighsCutGeneration {
 
   double scale(double val);
 
-  bool postprocessCut();
+  bool postprocessCut(HighsDomain& globaldom);
```
fwesselm: `const HighsDomain& globaldom`?

highs/mip/HighsCutGeneration.h
```diff
@@ -96,13 +96,14 @@ class HighsCutGeneration {
 
   /// generate a conflict from the given proof constraint which cuts of the
   /// given local domain
-  bool generateConflict(HighsDomain& localdom, std::vector<HighsInt>& proofinds,
+  bool generateConflict(HighsDomain& localdom, HighsDomain& globaldom,
```
fwesselm: `const HighsDomain& localdom, const HighsDomain& globaldom,` ?

highs/mip/HighsCutGeneration.h
```diff
@@ -96,13 +96,14 @@ class HighsCutGeneration {
...
+                        std::vector<HighsInt>& proofinds,
                         std::vector<double>& proofvals, double& proofrhs);
 
   /// applies postprocessing to an externally generated cut and adds it to the
   /// cutpool if it is violated enough
-  bool finalizeAndAddCut(std::vector<HighsInt>& inds, std::vector<double>& vals,
-                         double& rhs);
+  bool finalizeAndAddCut(HighsDomain& globaldom, std::vector<HighsInt>& inds,
```
fwesselm: `const HighsDomain& globaldom` ?

highs/mip/HighsCutPool.cpp
```diff
@@ -115,9 +115,43 @@ double HighsCutPool::getParallelism(HighsInt row1, HighsInt row2) const {
...
+
+  HighsInt i2 = pool2.matrix_.getRowStart(row2);
+  const HighsInt end2 = pool2.matrix_.getRowEnd(row2);
+
+  const HighsInt* ARindex1 = matrix_.getARindex();
+  const double* ARvalue1 = matrix_.getARvalue();
+  const HighsInt* ARindex2 = pool2.matrix_.getARindex();
+  const double* ARvalue2 = pool2.matrix_.getARvalue();
+
+  double dotprod = 0.0;
```
fwesselm, line 131: Stupid question (and not relevant for the parallelism): Would it be worthwhile to compute the dot product using quad precision (HighsCDouble) and cast to double on returning? Maybe not...

highs/mip/HighsCutPool.cpp
```diff
@@ -166,13 +226,13 @@ void HighsCutPool::performAging() {
 }
 
 void HighsCutPool::separate(const std::vector<double>& sol, HighsDomain& domain,
```
fwesselm: `const HighsDomain& domain`?

highs/mip/HighsCutPool.h
```diff
@@ -128,8 +145,15 @@ class HighsCutPool {
     ageDistribution.resize(agelim_ + 1);
   }
 
+  void increaseNumLps(HighsInt cut, HighsInt n) {
+    assert(numLps_[cut] >= 1);
+    numLps_[cut].fetch_add(n, std::memory_order_relaxed);
+  };
+
   void separate(const std::vector<double>& sol, HighsDomain& domprop,
```
fwesselm: `const HighsDomain& domprop`?

highs/mip/HighsCliqueTable.cpp
```diff
@@ -1698,9 +1701,11 @@ void HighsCliqueTable::separateCliques(const HighsMipSolver& mipsolver,
                    false, false);
   }
 
-  numNeighbourhoodQueries += data.numNeighbourhoodQueries;
+  localNumNeighbourhoodQueries += data.numNeighbourhoodQueries;
+
+  if (runcliquesubsumption && &randgen == &this->randgen) {
+    if (cliquehits.size() < cliques.size()) cliquehits.resize(cliques.size());
```
fwesselm: `if (cliquehits.size() < cliques.size()) cliquehits.resize(cliques.size()); `

I think this resize is now done by the utility `HighsCliqueTable::collectCliques`. Maybe a merge issue?

fwesselm (commented):
Wow, @Opt-Mucca, looks like a lot of work! Good stuff.

I have added some minor comments (for the first 10 files). I will continue soon.

highs/mip/HighsCutPool.cpp
Opt-Mucca: Not stupid. I'd say that it's not worthwhile because small changes in the result aren't too important here (the cuts at this point should also be scaled so magnitudes can't differ too crazily), but I imagine there's some extreme cases where we are potentially mislabelling cuts.

highs/mip/HighsCliqueTable.cpp
Opt-Mucca: Yup! Very nice spot. I think this got swept up in me mistakenly looking at `master` instead of `latest` for some files when doing the merge. I've deleted the line

Opt-Mucca:
@fwesselm I deleted the line in `HighsCliqueTable` and added `const` following all your suggestions

highs/mip/HighsLpRelaxation.cpp
```diff
@@ -245,14 +261,15 @@ void HighsLpRelaxation::loadModel() {
   colUbBuffer.resize(num_col);
 }
 
-void HighsLpRelaxation::resetToGlobalDomain() {
+void HighsLpRelaxation::resetToGlobalDomain(HighsDomain& globaldom) {
```
fwesselm: `const HighsDomain& globaldom`?

highs/mip/HighsLpRelaxation.h
```diff
@@ -41,6 +43,7 @@ class HighsLpRelaxation {
     Origin origin;
     HighsInt index;
     HighsInt age;
+    HighsInt cutpool;
```
fwesselm: I am not a fan of naming discussions, but here we go: The name `cutpool` is often used for `HighsCutPool` instances. Therefore, I wonder whether another name would be better here. Maybe `cutpoolindex` or something better?

highs/mip/HighsLpRelaxation.cpp
```diff
@@ -87,7 +88,9 @@ void HighsLpRelaxation::LpRow::get(const HighsMipSolver& mipsolver,
                                    const double*& vals) const {
   switch (origin) {
     case kCutPool:
-      mipsolver.mipdata_->cutpool.getCut(index, len, inds, vals);
+      assert(cutpool <=
```
fwesselm: Could this assertion be moved out of the `switch-case`? Can it only be asserted once on entering this method maybe?

Also, I am sure I am missing something, but why is the assertion checking for less-than-or-equal and not

`assert(cutpool < static_cast<HighsInt>(mipsolver.mipdata_->cutpools.size()));`

highs/mip/HighsLpRelaxation.cpp
```diff
@@ -938,7 +980,10 @@ void HighsLpRelaxation::storeDualInfProof() {
     for (HighsInt j = 0; j < len; ++j) row_ap.add(inds[j], weight * vals[j]);
   }
 
-  const HighsDomain& globaldomain = mipsolver.mipdata_->domain;
+  HighsDomain& globaldomain =
```
fwesselm: Not `const` anymore?

highs/mip/HighsLpRelaxation.cpp
```diff
@@ -1002,12 +1048,18 @@ void HighsLpRelaxation::storeDualUBProof() {
...
   dualproofvals.clear();
 
-  if (lpsolver.getSolution().dual_valid)
-    hasdualproof = computeDualProof(mipsolver.mipdata_->domain,
-                                    mipsolver.mipdata_->upper_limit,
-                                    dualproofinds, dualproofvals, dualproofrhs);
-  else
+  if (lpsolver.getSolution().dual_valid) {
+    hasdualproof =
+        computeDualProof((worker_ && mipsolver.mipdata_->parallelLockActive())
```
fwesselm: Sorry, this may just be nonsense, but here we go again:

`worker_ && mipsolver.mipdata_->parallelLockActive()`

is evaluated twice here (unless the compiler does something clever).

Could it happen that it is `true` for the first evaluation and `false` for the second (change to lock), so that mixed information (one from worker, one global) is passed?

fwesselm (commented):
Some more unqualified comments from me. I will continue in the afternoon.

highs/mip/HighsLpRelaxation.cpp
Opt-Mucca: Added back. Not sure why this was removed....

highs/mip/HighsLpRelaxation.cpp
Opt-Mucca: Why should it be moved outside of the switch case? It's irrelevant in other cases? I am worried this assert in general is a bit overboard (it is more there because I was experiencing some bug at some point and added them as a precaution).

You're not missing anything. This seems to purely be an oversight on my end. I've changed it to your suggestion!

highs/mip/HighsLpRelaxation.h
Opt-Mucca: Lololololololol. I'm happy you dived in! Strong support on changing the name. I've followed your suggestion because I've got no strong opinion on something different, so it should now be `cutpoolindex` everywhere.

highs/mip/HighsLpRelaxation.cpp
Opt-Mucca: Added!

highs/mip/HighsLpRelaxation.cpp
Opt-Mucca: It's not nonsense. The current code is just set up in such a way that I never imagined this being possible. This isn't safe against future changes though.

I've now created a boolean `use_worker_info` so the switch is only evaluated once. Is this now alright?

highs/mip/HighsLpRelaxation.cpp
fwesselm: Ok, I was just wondering because the assertion is repeated in different `case` sections.

highs/mip/HighsLpRelaxation.cpp
fwesselm: Ok, thanks.

cmatraki:
I can't reproduce it with the latest version, but commit 4e29a0d with default options, finds app1-2 from miplib infeasible:
```
Running HiGHS 1.12.0 (git hash: 4e29a0d95): Copyright (c) 2025 HiGHS under MIT licence terms
Number of PL entries in BOUNDS section is 13571
Number of BV entries in BOUNDS section is 13300
MIP app1-2 has 53467 rows; 26871 cols; 199175 nonzeros; 13300 integer variables (13300 binary)
Coefficient ranges:
  Matrix  [1e-05, 1e+00]
  Cost    [1e+00, 1e+00]
  Bound   [1e+00, 1e+00]
  RHS     [1e-05, 1e+00]
WARNING: Problem has some excessively small row bounds
Presolving model
53461 rows, 26859 cols, 199133 nonzeros  0s
39849 rows, 13559 cols, 275633 nonzeros  0s
34782 rows, 11265 cols, 206079 nonzeros  18s
Presolve reductions: rows 34782(-18685); columns 11265(-15606); nonzeros 206079(+6904) 

Solving MIP model with:
   34782 rows
   11265 cols (11000 binary, 0 integer, 0 implied int., 265 continuous, 0 domain fixed)
   206079 nonzeros

         0       0         0   0.00%   -238            inf                  inf        0      0      0         0    19.0s
         0       0         0   0.00%   -231.7553454    inf                  inf        0      0      2       817    20.2s
         0       0         0   0.00%   -215.3802548    inf                  inf    12422    709     99      3956    25.6s
         0       0         0   0.00%   -206.5307028    inf                  inf    11112    924    141      6686    30.7s
         0       0         0   0.00%   -198.3939772    inf                  inf    11604   1131    194      9142    36.0s
         0       0         0   0.00%   -186.7867751    inf                  inf    12614   1501    242     11571    41.1s
         0       0         0   0.00%   -182.4413041    inf                  inf    11793   1307    396     97428   233.9s

0.3% inactive integer columns, restarting
Model after restart has 34702 rows, 11228 cols (10963 bin., 0 int., 0 impl., 265 cont., 0 dom.fix.), and 205562 nonzeros

         0       0         0   0.00%   -182.4413041    inf                  inf     1307      0      0     97428   237.3s
         0       0         0   0.00%   -182.4368136    inf                  inf     1307   1152      4     99837   239.6s
         0       0         0   0.00%   -178.2681812    inf                  inf    13238   2059     69    101864   244.8s
         0       0         0   0.00%   -172.3725509    inf                  inf    19141   2120    129    105295   249.9s
         0       0         0   0.00%   -161.889637     inf                  inf    19369   2349    186    107863   254.9s
         0       0         0   0.00%   -156.2504424    inf                  inf    24243   2471    255    109553   260.2s
         0       0         0   0.00%   -156.2504424    inf                  inf    24257   2239    328    161721   368.1s
        31       5        13  99.77%   -156.2504424    inf                  inf    24015   2239    341    165970   373.3s
        48       6        21  99.82%   -156.2504424    inf                  inf    24015   2239    349    169966   378.6s
        69       7        31  99.82%   -156.2504424    inf                  inf    24018   2239    359    174356   384.2s
       100       0        46 100.00%   inf             inf                  inf     2326   2421    376    183365   395.2s
       100       0        46 100.00%   inf             inf                  inf     2326   2421    376    183365   395.2s

Solving report
  Model             app1-2
  Status            Infeasible
  Primal bound      inf
  Dual bound        inf
  Gap               inf
  P-D integral      0
  Solution status   -
  Timing            395.23
  Max sub-MIP depth 4
  Nodes             100
  Repair LPs        0
  LP iterations     183365
                    0 (strong br.)
                    25562 (separation)
                    136761 (heuristics)
```
Note the the presolved model is different than what is obtained with the latest code, but is consistent with the numbers shown by 1.12.0.

Opt-Mucca:
> I can't reproduce it with the latest version, but commit [4e29a0d](https://github.com/ERGO-Code/HiGHS/commit/4e29a0d9585617d72c4ce254ade08f990eb37c6d) with default options, finds app1-2 from miplib infeasible:
> Note the the presolved model is different than what is obtained with the latest code, but is consistent with the numbers shown by 1.12.0.

@cmatraki christ you are on it. Did you test over the entire MIPLIB? If so can you share any observations of the change? (I haven't really focused on performance, only focusing on correctness, but am properly interested in any opinions and user insights)

For app1-2: I've got that instance written down and can reproduce the incorrect status on an older commit locally (before the v1.13.1 merge). It was causing me enough headaches that I've just decided to push ahead for now and hope it pops up on a smaller instance. If you're thinking of looking into it: There's an incorrect bound change from propagating a cut (cut is completely valid), which I suspect is coming from stale activities, although I've failed to prove this (history of issues with this and multiple cutpools, but I thought I'd finally managed to fix it).

cmatraki:
> Did you test over the entire MIPLIB? If so can you share any observations of the change? (I haven't really focused on performance, only focusing on correctness, but am properly interested in any opinions and user insights)

Unfortunately no, I was just following along during development doing random tests for correctness. At some point everything started working but then I found this issue (and only this one) - just didn't know how to report it until now.

> If you're thinking of looking into it

I'm not familiar enough with the code yet, I can only offer my own suspicion that the presolve changes since 1.12.0 mask the issue.

highs/mip/HighsMipSolver.cpp
```diff
@@ -138,6 +160,18 @@ void HighsMipSolver::run() {
...
+  if (mipdata_->domain.infeasible()) {
+    cleanupSolve();
+    return;
+  }
+  // Initialise master worker.
+  mipdata_->workers.emplace_back(*this, &mipdata_->lp, &mipdata_->domain,
+                                 &mipdata_->cutpool, &mipdata_->conflictPool,
+                                 &mipdata_->pseudocost);
+
+  HighsMipWorker& master_worker = mipdata_->workers.at(0);
```
fwesselm: I think `at` is not used very much in the HiGHS code. Do you need the extra bound checking?

highs/mip/HighsMipSolver.cpp
```diff
@@ -212,17 +248,220 @@ void HighsMipSolver::run() {
...
+      highs::parallel::num_threads() == 1 || mip_search_concurrency <= 0 ||
+              submip
+          ? 1
+          : mip_search_concurrency * highs::parallel::num_threads();
+  HighsInt num_workers = 1;
+  highs::parallel::TaskGroup tg;
+
+  auto destroyOldWorkers = [&]() {
+    if (mipdata_->workers.size() <= 1) return;
+    while (mipdata_->domains.size() > 1) {
```
fwesselm, line 252: What is the reason for iteratively removing elements using `pop_back`?

Would this work?

```
mipdata_->domains.erase(mipdata_->domains.begin() + 1,
                            mipdata_->domains.end());
```

highs/mip/HighsMipSolver.cpp
```diff
@@ -233,15 +472,410 @@ void HighsMipSolver::run() {
...
+  };
+
+  auto infeasibleWorkerGlobalDomain = [&]() -> bool {
+    for (HighsMipWorker& worker : mipdata_->workers) {
+      if (worker.getGlobalDomain().infeasible()) return true;
+    }
+    return false;
+  };
+
+  auto getSearchIndicesWithNoNodes = [&]() -> std::vector<HighsInt> {
```
fwesselm: Will this involve copying? Could the lambda return a reference? (Sorry, C++ noob here...)

highs/mip/HighsMipSolver.cpp
```diff
@@ -233,15 +472,410 @@ void HighsMipSolver::run() {
...
+    }
+    if (static_cast<HighsInt>(indices.size()) >
+        mipdata_->nodequeue.numActiveNodes()) {
+      indices.resize(mipdata_->nodequeue.numActiveNodes());
+    }
+    return indices;
+  };
+
+  auto getSearchIndicesWithNodes = [&]() -> std::vector<HighsInt> {
+    std::vector<HighsInt> indices;
```
fwesselm: Same question about return a reference....

highs/mip/HighsMipSolver.cpp
```diff
@@ -233,15 +472,410 @@ void HighsMipSolver::run() {
...
+    }
+  };
+
+  auto handlePrunedNodes = [&](std::vector<HighsInt>& indices) -> bool {
+    std::vector<uint8_t> infeasible(num_workers, 0);
+    std::vector<uint8_t> flush(num_workers, 0);
+    std::vector<uint8_t> prune(num_workers, 0);
+    bool multiple_workers = num_workers > 1;
+    auto doHandlePrunedNodes = [&](HighsInt i) {
+      if (!mipdata_->workers[i].search_ptr_->currentNodePruned()) return;
```
fwesselm: Return a `bool`?

highs/mip/HighsMipSolver.cpp
```diff
@@ -233,15 +472,410 @@ void HighsMipSolver::run() {
...
+
+      if (globaldom.infeasible()) {
+        infeasible[i] = 1;
+        if (!multiple_workers) {
+          mipdata_->nodequeue.clear();
+          mipdata_->pruned_treeweight = 1.0;
+          mipdata_->updateLowerBound(
+              std::min(kHighsInf, mipdata_->upper_bound));
+        }
+        return;
```
fwesselm: Need to return a `bool`?

highs/mip/HighsMipSolver.cpp
```diff
@@ -233,15 +472,410 @@ void HighsMipSolver::run() {
...
+      }
+
+      prune[i] = 1;
+
+      if (multiple_workers || mipdata_->checkLimits()) {
+        return;
```
fwesselm: Sorry, again, return a `bool`? :)

highs/mip/HighsMipSolverData.cpp
```diff
@@ -2755,6 +2847,11 @@ bool HighsMipSolverData::terminatorTerminated() const {
   return mipsolver.termination_status_ != HighsModelStatus::kNotset;
 }
 
+bool HighsMipSolverData::terminatorTerminatedWorker(
+    HighsMipWorker& worker) const {
```
fwesselm: `const HighsMipWorker& worker`?

highs/mip/HighsMipSolver.cpp
Opt-Mucca: Nope. I can change it to `[0]`.

highs/mip/HighsMipSolver.cpp
Opt-Mucca: I do this because the compiler complains for `workers` (something about unique pointers and operator assignments) and (potentially) for `cutpools` if you don't delete them one at a time from the end (even `resize(1)` causes an error).
I copied over the style that worked for all of them so the code looked similar. I've tested your change, and it seems to work on everything but `workers` (My gut is telling me there might be problems on some systems with `cutpools` because of `std::atomic` too).
Would you like me to change it for `domains`, `pseudocosts`, `conflictpools`, and `lps`?

highs/mip/HighsMipSolver.cpp
Opt-Mucca: I trust your C++ knowledge more than mine. It would have involved copying the `std::vector<HighsInt>`. I've changed it to receive a reference (also removed the resize by adding an extra condition to the first loop)

highs/mip/HighsMipSolver.cpp
Opt-Mucca: This lambda was never used, so I've removed it.

highs/mip/HighsMipSolver.cpp
Opt-Mucca: This is in the lambda inside the lambda (so no need to return anything).

highs/mip/HighsMipSolver.cpp
Opt-Mucca: Same as above

highs/mip/HighsMipSolver.cpp
Opt-Mucca: Here too

highs/mip/HighsMipSolverData.cpp
Opt-Mucca: Can change!

highs/mip/HighsMipSolver.cpp
fwesselm: Ok, on second thought, it may be better to keep the while loops that you had and find out how to resolve the issue with the references at another time.

highs/mip/HighsMipSolver.cpp
fwesselm: Please don't trust me. :) I was just wondering if this might affect performance. It may be perfectly fine to copy.

highs/mip/HighsMipWorker.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+  const double transformed_solobj = static_cast<double>(
+      static_cast<HighsInt>(mipsolver_.orig_model_->sense_) *
+          mipsolver_quad_objective_value -
+      mipsolver_.model_->offset_);
+
+  return std::make_pair(feasible, transformed_solobj);
+}
+
+bool HighsMipWorker::trySolution(const std::vector<double>& solution,
```
fwesselm, line 113: Is this a copy of `HighsMipSolverData::trySolution`?

highs/mip/HighsMipWorker.cpp
```diff
@@ -0,0 +1,145 @@
...
+  search_ptr_->setLpRelaxation(lp_);
+}
+
+void HighsMipWorker::resetSepa() {
+  sepa_ptr_.reset();
+  sepa_ptr_ = std::unique_ptr<HighsSeparation>(new HighsSeparation(*this));
+  sepa_ptr_->setLpRelaxation(lp_);
+}
+
+bool HighsMipWorker::addIncumbent(const std::vector<double>& sol, double solobj,
```
fwesselm, line 54: This looks like a modified version of `HighsMipSolverData::addIncumbent`. I guess it is not possible to re-use the existing code? Should this live in `HighsMipSolverData`?

highs/mip/HighsMipWorker.cpp
```diff
@@ -0,0 +1,145 @@
...
+            mipsolver_.options_mip_->mip_rel_gap);
+      }
+    }
+    // Can't repair solutions locally, so also buffer infeasible ones
+    solutions_.emplace_back(sol, solobj, solution_source);
+  }
+  return true;
+}
+
+std::pair<bool, double> HighsMipWorker::transformNewIntegerFeasibleSolution(
```
fwesselm, line 77: This looks like a modified version of `HighsMipSolverData::transformNewIntegerFeasibleSolution`. I guess it is not possible to re-use the existing code? Should this live in `HighsMipSolverData`?

highs/mip/HighsPrimalHeuristics.cpp
```diff
@@ -80,9 +75,10 @@ void HighsPrimalHeuristics::setupIntCols() {
 }
 
 bool HighsPrimalHeuristics::solveSubMip(
-    const HighsLp& lp, const HighsBasis& basis, double fixingRate,
-    std::vector<double> colLower, std::vector<double> colUpper,
-    HighsInt maxleaves, HighsInt maxnodes, HighsInt stallnodes) {
+    HighsMipWorker& worker, const HighsLp& lp, const HighsBasis& basis,
+    double fixingRate, std::vector<double> colLower,
```
fwesselm, line 79: Vectors `colLower` and `colUpper` are not passed by reference. Is this correct?

highs/mip/HighsPrimalHeuristics.cpp
```diff
@@ -292,7 +295,7 @@ void HighsPrimalHeuristics::rootReducedCost() {
             return a.first > b.first;
           });
 
-  auto localdom = mipsolver.mipdata_->domain;
+  HighsDomain localdom = worker.getGlobalDomain();
```
fwesselm, line 326: Creating a new domain is correct? Not a reference?

highs/mip/HighsPrimalHeuristics.cpp
```diff
@@ -333,8 +338,8 @@ void HighsPrimalHeuristics::rootReducedCost() {
   double fixingRate = neighbourhood.getFixingRate();
   if (fixingRate < 0.3) return;
 
-  solveSubMip(*mipsolver.model_, mipsolver.mipdata_->firstrootbasis, fixingRate,
-              localdom.col_lower_, localdom.col_upper_,
+  solveSubMip(worker, *mipsolver.model_, mipsolver.mipdata_->firstrootbasis,
```
fwesselm, line 369: Just out of curiosity: Wouldn't is be possible to access `mipsolver.model_` and `mipsolver.mipdata_->firstrootbasis` from the `worker`? Maybe it's better to pass them separately...

highs/mip/HighsPrimalHeuristics.cpp
```diff
@@ -358,24 +363,33 @@ static double calcFixVal(double rootchange, double fracval, double cost) {
...
+
+  HighsPseudocost pscost(worker.getPseudocost());
+  HighsSearch heur(worker, pscost);
 
-  HighsPseudocost pscost(mipsolver.mipdata_->pseudocost);
-  HighsSearch heur(mipsolver, pscost);
   HighsDomain& localdom = heur.getLocalDomain();
   heur.setHeuristic(true);
 
+  std::vector<HighsInt> intcols_;
```
fwesselm, line 405: I don't understand this code segment (`intcols_`, `intcols`, `this->intcols`). What is it doing?

highs/mip/HighsPrimalHeuristics.cpp
```diff
@@ -404,7 +418,7 @@ void HighsPrimalHeuristics::RENS(const std::vector<double>& tmp) {
   //        heur.getCurrentDepth(), targetdepth);
   if (heur.getCurrentDepth() > targetdepth) {
     if (!heur.backtrackUntilDepth(targetdepth)) {
-      lp_iterations += heur.getLocalLpIterations();
+      worker.heur_stats.lp_iterations += heur.getLocalLpIterations();
```
fwesselm: Not for now, but maybe worth having a `getHeurLpIterations` method at some point.

highs/mip/HighsPseudocost.cpp
```diff
@@ -22,6 +22,7 @@ HighsPseudocost::HighsPseudocost(const HighsMipSolver& mipsolver)
       ncutoffsdown(mipsolver.numCol()),
       conflictscoreup(mipsolver.numCol()),
       conflictscoredown(mipsolver.numCol()),
+      changed(mipsolver.numCol()),
```
fwesselm: Sorry, I should maybe know by now, but what is `changed`?

highs/mip/HighsPseudocost.h
```diff
@@ -361,6 +383,153 @@ class HighsPseudocost {
...
+                       std::vector<HighsInt>& ninferencesup,
+                       std::vector<HighsInt>& ninferencesdown,
+                       std::vector<HighsInt>& ncutoffsup,
+                       std::vector<HighsInt>& ncutoffsdown) {
+    int64_t orig_nsamplestotal = this->nsamplestotal;
+    int64_t orig_ninferencestotal = this->ninferencestotal;
+    for (HighsInt col : pseudocost.indschanged) {
+      assert(col >= 0 &&
+             col < static_cast<HighsInt>(pseudocost.ncutoffsup.size()) &&
+             pseudocost.ncutoffsup.size() == ncutoffsup.size() &&
```
fwesselm: Not necessary, but could the 

`assert(pseudocost.ncutoffsup.size() == ncutoffsup.size() && pseudocost.ncutoffsup.size() == this->ncutoffsup.size());`

be done before the loop?

fwesselm (commented):
Some more comments from me.

highs/mip/HighsMipSolver.cpp
Opt-Mucca: No difference to me from a readability perspective. My gut is also telling me the change can't be less performant. So I'll keep to what the first recommendation suggested.

highs/mip/HighsMipWorker.cpp
Opt-Mucca: Yes...... It is just calling a different `addIncumbent` at the end of feasibility check (which requires the `HighsMipWorker` object). I could pass `HighsMipWorker* worker = nullptr` as a default argument, and call the global `HighsMipSolverData::addIncumbent` vs `HighsMipWorker::addIncumbent` depending on whether or not it's a `nullptr`. Any strong opinion for or against (or a different approach)?

highs/mip/HighsMipWorker.cpp
Opt-Mucca: See the above suggestion (pass the worker as a pointer)

highs/mip/HighsMipWorker.cpp
Opt-Mucca: After looking into this again, I guess it is now possible to reuse some of the existing code (I think I may have gotten lazy at this point). It would need:
- Some `nullptr` default argument like in the above point
- Some new statistics added to `HighsMipWorker` (repair LP iterations)
- Some timer magic (delaying fixing all of this for now)

Would you want this to be unified into one function?

highs/mip/HighsPrimalHeuristics.cpp
Opt-Mucca: This was done before. The data is immediately moved into the `submip` object, which I imagine can then be changed during the solving process. If the `submip` is infeasible then that domain is later reused, so I think we need to copy it.

highs/mip/HighsMipWorker.cpp
fwesselm: Maybe another thing to look at after the initial submission.

highs/mip/HighsMipWorker.cpp
fwesselm: We can make a note and look at it later.

highs/mip/HighsPrimalHeuristics.cpp
Opt-Mucca: As mentioned above

highs/mip/HighsPrimalHeuristics.cpp
Opt-Mucca: I guess they would lose there `const` specifiers? Otherwise I don't see an issue.

highs/mip/HighsPrimalHeuristics.cpp
Opt-Mucca: I have now done this a few times: I don't want to add boolean switches at all points in the code, i.e., wherever `intcols` is accessed, so I simply define it once as a reference, which involves creating the thing it potentially is referencing, e.g., `intcols_`. 
There might be some obvious C++ mechanic I could be using here.

highs/mip/HighsPrimalHeuristics.cpp
Opt-Mucca: That would return a reference to `heur_stats.lp_iterations`?

highs/mip/HighsPseudocost.cpp
Opt-Mucca: `changed` contains the boolean flag on whether each column has had a pseudo-cost update or not.
`indschanged` contains the vector of columns that have a `true` value in `changed`.
I need both so I don't continue populating `indschanged` with the same index, and so I don't have to scan through all columns at each sync.

highs/mip/HighsPseudocost.h
Opt-Mucca: Of course. I've moved it.

highs/mip/HighsPrimalHeuristics.cpp
fwesselm: Yes, I just saw similar updates in different places and wanted to ask.

highs/mip/HighsTransformedLp.h
```diff
@@ -29,6 +29,7 @@ class HighsLpRelaxation;
 class HighsTransformedLp {
  private:
   const HighsLpRelaxation& lprelaxation;
+  HighsDomain& globaldom_;
```
fwesselm: Could this be `const`?

highs/mip/HighsTransformedLp.h
```diff
@@ -58,6 +59,8 @@ class HighsTransformedLp {
 
   bool untransform(std::vector<double>& vals, std::vector<HighsInt>& inds,
                    double& rhs, bool integral = false);
+
+  HighsDomain& getGlobaldom() const { return globaldom_; }
```
fwesselm: Return `const` reference?

highs/mip/HighsTransformedLp.h
```diff
@@ -48,7 +49,7 @@ class HighsTransformedLp {
 
  public:
   HighsTransformedLp(const HighsLpRelaxation& lprelaxation,
-                     HighsImplications& implications);
+                     HighsImplications& implications, HighsDomain& globaldom);
```
fwesselm: `const HighsDomain& globaldom`?

highs/mip/HighsTransformedLp.cpp
```diff
@@ -13,8 +13,9 @@
 #include "util/HighsIntegers.h"
 
 HighsTransformedLp::HighsTransformedLp(const HighsLpRelaxation& lprelaxation,
-                                       HighsImplications& implications)
-    : lprelaxation(lprelaxation) {
+                                       HighsImplications& implications,
+                                       HighsDomain& globaldom)
```
fwesselm: `const HighsDomain& globaldom`?

highs/presolve/HighsPostsolveStack.h
```diff
@@ -609,8 +609,23 @@ class HighsPostsolveStack {
 
   /// undo presolve steps for primal dual solution and basis
   void undo(const HighsOptions& options, HighsSolution& solution,
-            HighsBasis& basis, const HighsInt report_col = -1) {
-    reductionValues.resetPosition();
+            HighsBasis& basis, const HighsInt report_col = -1,
```
fwesselm, line 612: Why is this needed? Don't the workers have separate postsolve stacks?

highs/presolve/HPresolve.cpp
```diff
@@ -1010,17 +1010,20 @@ void HPresolve::shrinkProblem(HighsPostsolveStack& postsolve_stack) {
...
     mipsolver->mipdata_->objectiveFunction = HighsObjectiveFunction(*mipsolver);
-    mipsolver->mipdata_->domain = HighsDomain(*mipsolver);
+    mipsolver->mipdata_->domains[0] = HighsDomain(*mipsolver);
     mipsolver->mipdata_->cliquetable.rebuild(model->num_col_, postsolve_stack,
                                              mipsolver->mipdata_->domain,
                                              newColIndex, newRowIndex);
     mipsolver->mipdata_->implications.rebuild(model->num_col_, newColIndex,
                                               newRowIndex);
-    mipsolver->mipdata_->cutpool =
+    // TODO: Find a sensible way to do this
```
fwesselm: Are you deleting the cut pool?

highs/mip/HighsTransformedLp.h
Opt-Mucca: Yup!

highs/mip/HighsTransformedLp.h
Opt-Mucca: Done

highs/mip/HighsTransformedLp.h
Opt-Mucca: Done. Nice spot on being able to make this `const`

highs/presolve/HighsPostsolveStack.h
Opt-Mucca: The workers don't have separate postsolve stacks. The current `undo` function is changing some global data structures, so you can't call `undo` on two different threads simultaneously.
This change is adding an argument so it does everything on copied data.

highs/presolve/HPresolve.cpp
Opt-Mucca: Yup! This was done before and I didn't see a need to change the workflow.

 I had to get creative here because the `std::atomic` in `HighsCutPool`. I didn't want to change `cutpool` to a `ptr` in `HighsMipData` either (wanted to keep it as a reference). I'm happy with any ideas here as I think the current thing looks ugly (even if it works).

jajhall:
@Opt-Mucca I've merged latest into this branch to see whether all the CI tests pass

Opt-Mucca:
@jajhall : Would this be the timer issue you mentioned on Tuesday (when the same highs instance is run multiple times)?

>> intlinprog(prob);
Running HiGHS 1.14.0: Copyright (c) 2026 under MIT licence terms
MIP has 212 rows; 260 cols; 8619 nonzeros; 259 integer variables (259 binary)
Coefficient ranges:
  Matrix  [1e+00, 7e+02]
  Cost    [1e+00, 1e+00]
  Bound   [1e+00, 1e+00]
  RHS     [1e+00, 3e+02]
ERROR:   Option 'threads' is set to 4 but global scheduler has already been initialized to use 2 threads. The previous scheduler instance can be destroyed by calling Highs::resetGlobalScheduler().
Error using intlinprog (line 108)
INTLINPROG has stopped because it encountered an internal error. We are sorry for the inconvenience.

That's some output from the test of @fwesselm of this branch with `mip_search_concurrency > 0`

jajhall:
No, this is due to someone running a second Highs instance with threads set to a different value

Opt-Mucca:
@fwesselm I've updated the branch so the interrupt callback is not flagged when `parallelLockActive`, and I've made most of `HighsMipWorker` private (a point you raised in an earlier comment).
I believe this is on track to be merged next week (after I change `mip_search_concurrency=0` to default)

Opt-Mucca:
@jajhall I am getting an error from an assert in the timers when running https://miplib.zib.de/instance_details_lectsched-4-obj.html with `mip_search_concurrency=2`. Could you look into this? (I will also try and see if I can debug it)

```
HiGHS/highs/lp_data/HighsInterface.cpp:4346: void HighsSubSolverCallTime::start(HighsInt): Assertion `this->submip_clock_running[thread] < 0' failed.
Signal: SIGABRT (Aborted)
```

Edit: Would also need `threads=8`

jajhall:
> @jajhall I am getting an error from an assert in the timers when running https://miplib.zib.de/instance_details_lectsched-4-obj.html with `mip_search_concurrency=2`. Could you look into this? (I will also try and see if I can debug it)
> 
> ```
> HiGHS/highs/lp_data/HighsInterface.cpp:4346: void HighsSubSolverCallTime::start(HighsInt): Assertion `this->submip_clock_running[thread] < 0' failed.
> Signal: SIGABRT (Aborted)
> ```
> 
> Edit: Would also need `threads=8`

The sub-solver profiling has developed since, so I'll try merging my most recent work to see what happens

galabovaa (approved):

highs/lp_data/Highs.cpp
```diff
@@ -1051,29 +1053,10 @@ HighsStatus Highs::calledOptimizeModel() {
     }
   }
 
+  if (!options_.use_warm_start) this->clearSolver();
```
Opt-Mucca, line 1074: @jajhall Is this something necessary for the timers, or something that got caught up in merge conflicts?

jajhall: No, this is a vanilla HiGHS option that forces Highs::run() to ignore any information from previous solves

Opt-Mucca: Ahhhhhhhh. I missed the deleted call earlier. All good

highs/mip/HighsCutPool.h
```diff
@@ -39,6 +41,7 @@ struct HighsCutSet {
 
   void clear() {
     cutindices.clear();
+    cutpools.clear();
     upper_.clear();
```
fwesselm, line 45: `lower_.clear()` missing?

highs/mip/HighsPrimalHeuristics.cpp
```diff
@@ -939,33 +1034,33 @@ bool HighsPrimalHeuristics::tryRoundedPoint(const std::vector<double>& point,
...
       }
       return false;
     } else if (lprelax.unscaledPrimalFeasible(st)) {
       const auto& lpsol = lprelax.getLpSolver().getSolution().col_value;
       if (!integerFeasible) {
         // there may be fractional integer variables -> try ziRound heuristic
-        ziRound(lpsol);
-        return mipsolver.mipdata_->trySolution(lpsol, solution_source);
+        ziRound(worker, lpsol);
+        trySolution(lpsol, solution_source, worker);
```
fwesselm: `return trySolution(lpsol, solution_source, worker);` ?

highs/mip/HighsCutPool.cpp
```diff
@@ -287,8 +360,13 @@ void HighsCutPool::separate(const std::vector<double>& sol, HighsDomain& domain,
...
 
-  bestObservedScore = std::max(efficacious_cuts[0].first, bestObservedScore);
-  double minScore = minScoreFactor * bestObservedScore;
+  double bestObservedScoreCopy = bestObservedScore;
+  double& bestObservedScore_ =
+      thread_safe ? bestObservedScoreCopy : bestObservedScore;
+  bestObservedScore_ = std::max(efficacious_cuts[0].first, bestObservedScore);
+  double minScoreFactorCopy = minScoreFactor;
+  double& minScoreFactor_ = thread_safe ? minScoreFactorCopy : minScoreFactor;
+  double minScore = minScoreFactor_ * bestObservedScore;
```
fwesselm: `bestObservedScore_` ?

highs/mip/HighsCutPool.cpp
```diff
@@ -303,38 +381,55 @@ void HighsCutPool::separate(const std::vector<double>& sol, HighsDomain& domain,
 
   if (numefficacious <= lowerThreshold) {
     numefficacious = std::max(efficacious_cuts.size() / 2, size_t{1});
-    minScoreFactor =
+    minScoreFactor_ =
         efficacious_cuts[numefficacious - 1].first / bestObservedScore;
```
fwesselm: `bestObservedScore_` ?

highs/mip/HighsCutPool.cpp
```diff
@@ -303,38 +381,55 @@ void HighsCutPool::separate(const std::vector<double>& sol, HighsDomain& domain,
...
   } else if (numefficacious > upperThreshold) {
-    minScoreFactor = efficacious_cuts[upperThreshold].first / bestObservedScore;
+    minScoreFactor_ =
+        efficacious_cuts[upperThreshold].first / bestObservedScore;
```
fwesselm: `bestObservedScore_` ?

highs/mip/HighsMipWorker.cpp
```diff
@@ -0,0 +1,202 @@
...
+        upper_limit = new_upper_limit;
+        optimality_limit = mipdata_.computeNewUpperLimit(
+            upper_bound, mipsolver_.options_mip_->mip_abs_gap,
+            mipsolver_.options_mip_->mip_rel_gap);
+      }
+    }
+    // Can't repair solutions locally, so also buffer infeasible ones
+    solutions_.emplace_back(sol, solobj, solution_source);
+  }
+  return true;
```
fwesselm, line 74: This always returns `true`, correct?

highs/mip/HighsMipSolver.cpp
```diff
@@ -175,54 +209,256 @@ void HighsMipSolver::run() {
...
+    while (mipdata_->domains.size() > 1) {
+      mipdata_->domains.pop_back();
+    }
+    while (mipdata_->lps.size() > 1) {
+      mipdata_->lps.pop_back();
+    }
+    while (mipdata_->pseudocosts.size() > 1) {
+      mipdata_->pseudocosts.pop_back();
+    }
+    while (mipdata_->workers.size() > 1) {
```
fwesselm, line 261: Is the destructor for HighsMipWorker accessing its pools or domain? It may be better to destroy the workers first, and then the other stuff?

Workers, then pools, then domains, then LPs and pseudocosts?

```
auto destroyOldWorkers = [&]() {
      if (mipdata_->workers.size() <= 1) return;
      while (mipdata_->workers.size() > 1) {
        mipdata_->workers.pop_back();
      }
      while (mipdata_->cutpools.size() > 1) {
        mipdata_->cutpools.pop_back();
      }
      while (mipdata_->conflictpools.size() > 1) {
        mipdata_->conflictpools.pop_back();
      }
      while (mipdata_->domains.size() > 1) {
        mipdata_->domains.pop_back();
      }
      while (mipdata_->lps.size() > 1) {
        mipdata_->lps.pop_back();
      }
      while (mipdata_->pseudocosts.size() > 1) {
        mipdata_->pseudocosts.pop_back();
      }
  };
```

highs/mip/HighsCutPool.cpp
```diff
@@ -136,6 +170,31 @@ void HighsCutPool::performAging() {
...
+    } else if (numLps_[i] == 0 && ages_[i] == -1 && rhs_[i] != kHighsInf) {
+      // Cut was removed from the LP, but age changes haven't been made
+      if (matrix_.columnsLinked(i)) {
+        propRows.erase(std::make_pair(ages_[i], i));
+        propRows.emplace(1, i);
+      }
+      ages_[i] = 1;
+      --numLpCuts;
+      ++ageDistribution[1];
+      ageResetWhileLocked_[i].store(0, std::memory_order_relaxed);
```
fwesselm, line 193: Is there a `continue` missing here?

highs/mip/HighsCutPool.cpp
```diff
@@ -136,6 +170,31 @@ void HighsCutPool::performAging() {
...
+    } else if (ageResetWhileLocked_[i].load(std::memory_order_relaxed) == 1) {
+      resetAge(i);
```
fwesselm, line 196: Is there a `continue` missing here? I am not sure.

highs/mip/HighsDomain.cpp
```diff
@@ -372,7 +391,31 @@ HighsDomain::CutpoolPropagation::CutpoolPropagation(
...
+// Warning: When a domain is copy-assigned, e.g., in `resetLocalDomain`,
+// with the line `localdom = getDomain()`, then it is going to notify
+// all cut / conflict pools that the original was propagating.
+// This would be non-deterministic in the order in which the global
+// pool gets notified. Currently, such code is only run in serial.
+
+HighsDomain::CutpoolPropagation& HighsDomain::CutpoolPropagation::operator=(
+    const CutpoolPropagation& other) {
+  if (this == &other) return *this;
+  if (cutpool) cutpool->removePropagationDomain(this);
```
fwesselm, line 409: `assert(!domain->mipsolver->mipdata_->parallelLockActive());` ?

highs/mip/HighsDomain.cpp
```diff
@@ -143,7 +143,26 @@ HighsDomain::ConflictPoolPropagation::ConflictPoolPropagation(
...
+  if (!domain->mipsolver->mipdata_->parallelLockActive() ||
+      conflictpool_ != &domain->mipsolver->mipdata_->getConflictPool())
+    conflictpool_->addPropagationDomain(this);
+}
+
+HighsDomain::ConflictPoolPropagation&
+HighsDomain::ConflictPoolPropagation::operator=(
+    const ConflictPoolPropagation& other) {
+  if (this == &other) return *this;
+  if (conflictpool_) conflictpool_->removePropagationDomain(this);
```
fwesselm, line 155: `assert(!domain->mipsolver->mipdata_->parallelLockActive());` ?

fwesselm (commented):
@Opt-Mucca, some more comments from me!

highs/mip/HighsCutPool.cpp
```diff
@@ -524,3 +635,28 @@ HighsInt HighsCutPool::addCut(const HighsMipSolver& mipsolver, HighsInt* Rindex,
...
+    // Only sync cuts in the LP that are not already synced
+    if ((numLps_[i] > 0 ||
+         ageResetWhileLocked_[i].load(std::memory_order_relaxed) == 1) &&
+        !hasSynced_[i]) {
+      HighsInt Rlen;
+      const HighsInt* Rindex;
+      const double* Rvalue;
+      getCut(i, Rlen, Rindex, Rvalue);
+      // copy cut into something mutable (addCut reorders so can't take const)
+      std::vector<HighsInt> idxs(Rindex, Rindex + Rlen);
```
fwesselm: Would it be worth pre-allocating `idxs` and `vals` outside of the loop, and then assign to them?

```
std::vector<HighsInt> idxs;
std::vector<double> vals;

.....

idxs.assign(Rindex, Rindex + Rlen);
vals.assign(Rvalue, Rvalue + Rlen);
```

highs/mip/HighsCutPool.h
Opt-Mucca: I'm 95% sure this was originally intended. The values are always `-kHighsInf`, so it's not technically needed. If you find this confusing then I can add the line.

highs/mip/HighsPrimalHeuristics.cpp
Opt-Mucca: Nice spot! I apparently just removed the `return`...

highs/mip/HighsCutPool.h
fwesselm: Ok, maybe a comment would be good, but it's up to you.

highs/mip/HighsCutPool.cpp
Opt-Mucca: Very nice spot!!!!!!!!!!!!!!!! Disappointed I didn't spot this with my code highlighting

highs/mip/HighsCutPool.cpp
Opt-Mucca: Agree on both!

highs/mip/HighsMipWorker.cpp
Opt-Mucca: Yup. We're never 100% confident that we couldn't have repaired it until we try. This at some point should be reworked. Maybe at the same time as refactoring the two `addIncumbent` functions?

I wanted this to return `bool` because there's plenty of bits of the code that call functions that will swap between `mipdata_->addIncumbent` and `worker.addIncumbent` dynamically, and I didn't want to handle the return values differently.

fwesselm: Thanks for explaining!

highs/mip/HighsMipSolver.cpp
Opt-Mucca: Isn't that the current order?

I had a lot of problems with `valgrind` when configuring the order of what gets deleted here until finally reaching the above code.

fwesselm: On second thought, I think the original code is fine. Let me resolve this.

fwesselm: > Isn't that the current order?
> 
> I had a lot of problems with `valgrind` when configuring the order of what gets deleted here until finally reaching the above code.

@Opt-Mucca No, I checked one more time, and I think the code is fine. Sorry for the confusion.

highs/mip/HighsCutPool.cpp
Opt-Mucca: I think so.... Very nice spot again!

highs/mip/HighsCutPool.cpp
Opt-Mucca: I don't think this is missing. After resetting the age it still makes sense to age it afterwards (what would have happened in the serial case).

highs/mip/HighsDomain.cpp
Opt-Mucca: That is an incredibly obvious thing to do.....

highs/mip/HighsDomain.cpp
Opt-Mucca: Yup

highs/mip/HighsCutPool.cpp
Opt-Mucca: I'm not against it. I'll make the change.

highs/mip/HighsCutPool.cpp
fwesselm: @Opt-Mucca, thanks for checking.

highs/mip/HighsCutPool.h
Opt-Mucca: Added (edit: as a comment)

Opt-Mucca:
@galabovaa Changed the log line. This branch is good to go for me.

galabovaa:
Sounds great! Please merge it in latest ⭐

PR #2889, opened 2026-03-03
A minor enhancement to `HPresolve::singletonColStuffing`

fwesselm:
- Consider case where all candidates are integer variables with identical weights (coefficients) in `HPresolve::singletonColStuffing`.
- Only 16 models in the test set are affected in terms of node counts and there is only a tiny speedup (0.5 percent), but the extension is simple and cheap and may be worthwhile on some user model.


Opt-Mucca (approved):
Looks good! Is a small and neat change!

Opt-Mucca:
@filikat Are we fine merging with the currently failing HiPO tests?

Edit: Merging after talking to @galabovaa

PR #2891, opened 2026-03-04
Store variable bounds from rows

Opt-Mucca:
This PR is motivated by me strongly believing that we're missing variable bounds compared to commercial solvers, i.e., me desperately searching for reasons why the flow cover cuts aren't showing performance improvements. I noticed that HiGHS is only generating VLB / VUBs when probing, and this potentially misses some obvious ones that can be used for cut strengthening (won't be used for anything else as they're already represented directly in the LP).
I've run a test on 150 instances with a 20min time limit, and solved two more instances and observed a 1.5% positive change on all solved for (time, nodes, LP iterations). This still shouldn't be merged until @fwesselm tests it further.
@fwesselm As I'm still not the most familiar with the presolve code, feel free to change stuff. I'm hoping it's in the right place. The idea could be further generalised to larger rows, but I'm not sure if that's just going to be diminishing returns.


fwesselm:
@Opt-Mucca, very good idea! I have implemented a generalization to rows with more than two elements and I am going to qualify this now.

fwesselm:
@Opt-Mucca, I added my changes to this branch - I hope that's OK.

Opt-Mucca:
@fwesselm Your changes are incredibly welcome. Very neat inclusion of rows with a single `inf`!
I'm happy with the code. My only suggestion would be to check if the varbound is guaranteed to be weaker than the simple bound before trying to add it. 
The failing tests for `unit-tests-external` aren't currently due to this branch.

fwesselm:
> @fwesselm Your changes are incredibly welcome. Very neat inclusion of rows with a single `inf`! I'm happy with the code. My only suggestion would be to check if the varbound is guaranteed to be weaker than the simple bound before trying to add it. The failing tests for `unit-tests-external` aren't currently due to this branch.

Would the following be OK?

 ```
    // add vlb
    if (vlbConstant != -kHighsInf &&
        std::max(0.0, vbCoef) + static_cast<HighsCDouble>(vlbConstant) >
            model->col_lower_[nonzero.index()] + primal_feastol)
      mipsolver->mipdata_->implications.addVLB(nonzero.index(), binCol, vbCoef,
                                               vlbConstant);
    // add vub
    if (vubConstant != kHighsInf &&
        std::min(0.0, vbCoef) + static_cast<HighsCDouble>(vubConstant) <
            model->col_upper_[nonzero.index()] - primal_feastol)
      mipsolver->mipdata_->implications.addVUB(nonzero.index(), binCol, vbCoef,
                                               vubConstant);
```

Opt-Mucca:
@fwesselm very happy with that!

fwesselm:
@Opt-Mucca, I have updated the code. The test results show a tiny increase in the solution times and numbers of branching nodes (1 percent). I think this is just due to randomness. Since extracting the variable bound constraints may be helpful for the flow cover cuts, I would still be in favor of merging.

(Also, there seem to be unrelated test failures in `build-unit-tests-external` that I have not seen before.)

Opt-Mucca:
@fwesselm I also think this is randomness and that having more variable bounds lying around will payoff in the long run. I'm happy to merge it.
The failing test is a non-issue (external tests were changed in preparation of one of the other changes)

Edit: I will check this change on the original set I ran to see if the results are consistent with the basic version (so might merge this on Monday)

Edit: My runs show a 1% improvement for time and LP iterations (same for nodes), and one more instance solved. I'm happy to merge this, but I'm not sure outside of noise why only doing this for rows of size two seemed more effective. (`app1-2` also now failed when I tested this change on master, so that could be related to the parallel branch)

fwesselm:
> @fwesselm I also think this is randomness and that having more variable bounds lying around will payoff in the long run. I'm happy to merge it. The failing test is a non-issue (external tests were changed in preparation of one of the other changes)
> 
> Edit: I will check this change on the original set I ran to see if the results are consistent with the basic version (so might merge this on Monday)
> 
> Edit: My runs show a 1% improvement for time and LP iterations (same for nodes), and one more instance solved. I'm happy to merge this, but I'm not sure outside of noise why only doing this for rows of size two seemed more effective. (`app1-2` also now failed when I tested this change on master, so that could be related to the parallel branch)

@Opt-Mucca, thank you for testing! I was going to investigate whether the variable bound constraints could be tightened before adding them. If you think this should not be merged in its current form, this is also fine by me.

Opt-Mucca:
@fwesselm We can merge first and you can still investigate. Happy with the change (just confused a bit by the results). Will click merge when the tests pass

PR #2897, opened 2026-03-06
`addVariable` no longer creating names with spaces, illegal names for MPS/LP no longer leads to error return, and EMS file facility deleted

jajhall:
- `addVariable` no longer creating names with spaces
- Illegal names for MPS/LP files no longer leads to error return, with spaces in names for MPS being replaced by underscores. If LP file names contain illegal characters or names contain duplicates, then generic names are used
- EMS file facility deleted since it's not been useful


jajhall:
@galabovaa This is the first one to review, since its changes are also in #2895

PR #2898, opened 2026-03-06
Update `meson.build` testing

jajhall:
`meson.build` now creates tests for all cpp files in `check/`, except for `test_cublas_example`, `test_cublas_gpu_start` and `TestLpSolversIterations.cpp`.

Also removed a few unused variables in the test files that were highlighted by meson


jajhall:
@galabovaa This is the last of (#2895, #2897 and #2898) to review, since its changes are also in #2895, and it builds on #2897

PR #2916, opened 2026-03-15
Added unit test to confirm use of advanced basis by all simplex solvers, and updated hot start documentation

### LINKED ISSUE #2913: Document basis warm-start behavior: PAMI ignores readBasis(), presolve invalidates basis
## Summary

When solving a sequence of similar LPs (e.g., rolling-horizon weekly chunks in power system optimization where only RHS changes), basis warm-starting between solves would be valuable. However, several undocumented behaviors make warm-starting ineffective in practice:

### 1. PAMI (`simplex_strategy=2`) silently ignores `readBasis()`

After calling `h.readBasis("previous.bas")`, solving with `simplex_strategy=2` (PAMI parallel dual simplex) cold-starts from scratch — the loaded basis is not used. Only `simplex_strategy=1` (serial dual simplex) respects the loaded basis.

This is surprising because users naturally expect `readBasis()` to work with all simplex strategies. There's no warning or error when PAMI discards the basis.

### 2. `writeBasis()` writes empty basis after IPM without crossover

When solving with IPM (`solver=ipm`) and `run_crossover=off`, calling `writeBasis()` produces a file containing only:
```
HiGHS_basis_file v2
None
```

This is technically correct (IPM produces interior-point solutions, not vertex/basis solutions), but it would be helpful to either:
- Return an error/warning from `writeBasis()` when no basis exists
- Document that crossover must be enabled to get a valid basis from IPM

### 3. Presolve fundamentally conflicts with basis warm-starting

When presolve is ON, it reduces the LP dimensions (e.g., 1.4M → 128K rows in our case, ~89% reduction). A basis from the original LP cannot be used on the presolved LP because the dimensions don't match.

This creates a fundamental trade-off:
- **With presolve**: Cold-start on small LP (~120s in our case)
- **Without presolve**: Warm-start on full LP (~300s in our case)

For large LPs where presolve achieves significant reduction, cold-start with presolve wins. The HiGHS documentation could clarify this trade-off.

## Reproduction

```python
import highspy

h = highspy.Highs()
h.readModel("problem.mps")

# Solve first instance
h.setOptionValue("solver", "ipm")
h.setOptionValue("run_crossover", "on")  # needed for basis!
h.run()
h.writeBasis("warm.bas")  # Valid basis (with crossover)

# Modify RHS for next instance
h.changeRowsBounds(...)

# Attempt warm-start — THIS DOES NOT WORK with PAMI:
h.readBasis("warm.bas")
h.setOptionValue("simplex_strategy", 2)  # PAMI ignores the basis!
h.run()  # Cold-starts from scratch

# THIS WORKS but requires presolve=off (slower on large LPs):
h.readBasis("warm.bas")
h.setOptionValue("simplex_strategy", 1)  # Serial dual simplex
h.setOptionValue("presolve", "off")      # Must disable — presolved LP has different dimensions
h.run()  # Warm-starts from basis (but full LP is much larger)
```

## Context

We're solving 52 weekly DC-OPF chunks (168h each) for power system analysis using PyPSA + Linopy + HiGHS. Each chunk has ~844K columns and ~1.4M rows before presolve (~128K after). Between chunks, only the RHS (load, renewable profiles) changes — the LP structure is identical.

Currently, cold-start with presolve + PAMI (~120s/chunk) is our best approach. If HiGHS could warm-start PAMI from a basis on the presolved LP (mapping the original basis through presolve), this could potentially be much faster.

## Suggested Documentation Improvements

1. Document that `simplex_strategy=2` (PAMI) does not use loaded basis files
2. Document that `writeBasis()` requires crossover when using IPM
3. Add a note about the presolve vs warm-start trade-off for sequential LP solving
4. Consider warning when `readBasis()` is called but the chosen solver strategy will ignore it

jajhall:
This will close #2913

### DIFF
--- a/docs/src/guide/further.md
+++ b/docs/src/guide/further.md
@@ -5,7 +5,7 @@
 HiGHS has a sophisticated presolve procedure for LPs and MIPs that
 aims to reduce the dimension of the model that must be solved. In most
 cases, the time saved by solving the reduced model is very much
-greater than the time taken to perform presolve. Once he presolved
+greater than the time taken to perform presolve. Once the presolved
 model is solved, a postsolve procedure (of minimal computational cost)
 deduces the optimal solution to the original model. Hence presolve is
 performed by default. The only exception occus when there is a valid


docs/src/guide/further.md
```diff
@@ -1,5 +1,29 @@
 # [Further features](@id guide-further)
 
+## [Presolve](@id guide-presolve)
+
+HiGHS has a sophisticated presolve procedure for LPs and MIPs that
+aims to reduce the dimension of the model that must be solved. In most
+cases, the time saved by solving the reduced model is very much
+greater than the time taken to perform presolve. Once he presolved
```
filikat: "Once (T)he presolved model"

jajhall: Thanks!

PR #2920, opened 2026-03-17
Upgrade relaxed to acquire

Opt-Mucca:
On `hmw-mt` the thread sanitizer occasionally fails unless this line is changed. I am unsure if this is a complete fix.

The failure broken down into steps:
- Threads are spawned
- Thread X writes some data (the only thread that is able to write to this place).
- The main thread calls `taskWait()`
- The main thread after waiting (should be the only active thread) reads the data 
Issue: The data being read in is different to what was written

Unless I'm using the wrong `highs::parallel` calls, then one of the two cases must be happening.
1. `taskWait()` is returning while a task is still active (I doubt this because then there'd be odd errors flying all over the place)
2. The main thread is somehow not synchronised correctly, and not recognising the "new" data. (If this is true, then I'm struggling to understand why this change fixes the issue)

There could also be something else I'm missing.......

### DIFF
--- a/highs/parallel/HighsTask.h
+++ b/highs/parallel/HighsTask.h
@@ -122,7 +122,7 @@ class HighsTask {
     uintptr_t xormask = reinterpret_cast<uintptr_t>(owner) ^
                         reinterpret_cast<uintptr_t>(stealer);
     uintptr_t state =
-        metadata.stealer.fetch_xor(xormask, std::memory_order_relaxed);
+        metadata.stealer.fetch_xor(xormask, std::memory_order_acquire);
 
     assert(stealer != nullptr);
 


filikat:
I think it makes sense to have acquire there. My (over-simplified and probably wrong) logic is as follows:
requestNotifyWhenFinished is fetching metadata.stealer and uses it to tell whether the task has finished or not. This is used by waitForTaskToFinish, which returns if the task finished. This is used by sync_stolen_taks, which returns if the task finished. This is then used by sync and taskWait.
So, the data fetched by requestNotifyWhenFinished is eventually used to sync the tasks that taskWait is waiting for. 
If the fetch_xor does not acquire, it does not synchronize with the release in markAsFinished. So, there is nothing preventing the effects of getCallable()(), line 84 (where the task is actually executed), to be seen as happening after the task is mark as finished, line 71, from the point of view of the waiting thread.
In other words, some side effects of the task can be seen as happening after taskWait returns.
This is probably highly timing dependent, because it can only happen if the task completes in between the calls to waitForTaskToFinish and the call to requestNotifyWhenFinished.

PR #2929, opened 2026-03-20
`Highs::postsolve` now uses `lpKktCheck` after postsolve when no basis is supplied

jajhall:
Hence, if the relative tolerances are satisfied, it can declare optimality


filikat (approved):
All good

PR #2934, opened 2026-03-24
Change scaling in HiPO

filikat:
- Currently, HiPO uses Curtis-Reid scaling to scale the problem before solving it.
This PR changes the scaling to infinity-norm equilibration of the rows and columns of the matrix. The scaling procedure takes into account also the magnitude of the diagonal of matrix Q and the width of the bounds interval.
Changing the scaling significantly reduces the number of iterations on some instances, including some of the PyPSA problems, and improves robustness in general.
- Some small pathological tests on infeasible problems occasionally fail with HiPO, due to numerical inaccuracies. These tests are not very useful, since they are not representative of real problems, and their failure is not symptomatic of an error in a commit, but rather of a random numerical perturbation that pushes the test to fail. They have been removed and substituted with tests on more realistic infeasible instances.
- Printing of the ranges of A, b, c and bounds has been slightly changed and made less verbose. Dev printing now shows the ranges of inf and one norms of rows and columns of A.


filikat:
I thought about this, but the results are actually a bit worse that way.
My intuition is that by pushing the values to the nearest power of 2 after each iteration, some scaling coefficients may be prevented from reaching a better value that would take many small steps.

jajhall:
> I thought about this, but the results are actually a bit worse that way.
> My intuition is that by pushing the values to the nearest power of 2 after each iteration, some scaling coefficients may be prevented from reaching a better value that would take many small steps.

Fair point

PR #2935, opened 2026-03-26
Docs: specify architecture for cuda in bazel

galabovaa:


jajhall (approved):
I can well imagine this to be the case!

PR #2937, opened 2026-03-27
bump: update bazel dependencies in MODULE.bazel

TendTo:
Just a small maintenance update to avoid depending on a pinned version of rules_cuda.


galabovaa:
Thank you, @TendTo! This update is very welcome: I only did the override because 0.3.0 was not out yet.

PR #2941, opened 2026-03-31
Allow callbacks to persist when using getIis

haman80:
- Allow original model callbacks to persist throughout elastic and deletion filtering
- Fix a minor bug in the reporting summary for getIIS


haman80:
Thank you for the review, @jajhall
We rely on kCallbackSimplexInterrupt to handle Ctrl+C interrupts from users. I modified the logic to disable all callbacks except for kCallbackSimplexInterrupt and to restore them in getIisInterfaceReturn. Please let me know your thoughts.

Issue #2900, opened 2026-03-08
Presolve incorrectly collapses objective value to exact zero


baharev:
When running the attached model with presolve enabled (default), HiGHS incorrectly reports an optimal objective value of exactly 0. However, with presolve disabled, the solver identifies the expected optimal value of approximately 667,562,736.

While the model exhibits scaling issues in both the objective and constraint RHS, adjusting scaling parameters as suggested by the HiGHS logs does not resolve the behavior. The issue is only bypassed by disabling presolve entirely.

The discrepancy is reproducible across different operating systems and versions:
 
 -  Windows: HiGHS 1.13, NumPy 2.4, Python 3.14
 -  Linux: HiGHS 1.11, NumPy 1.24, Python 3.11
 -  Comparison: SCIP reaches the expected optimal value without issue.

# Input file

 [dbg-presolve-1.lp.zip](https://github.com/user-attachments/files/25828978/dbg-presolve-1.lp.zip)

# Source code

```python
from highspy import Highs

h = Highs()

h.readModel('dbg-presolve-1.lp')

# h.setOptionValue('user_objective_scale', -4)

# h.setOptionValue('user_bound_scale', -5)

# h.setOptionValue('presolve', 'off')

h.run()
```

# Execution Logs

## Presolve Enabled (Incorrect Result)

```plaintext
(dbgpresolve) PS C:\Users\Administrator\dbg_presolve> pip install highspy
Collecting highspy
  Using cached highspy-1.13.1-cp314-cp314-win_amd64.whl.metadata (11 kB)
Collecting numpy (from highspy)
  Downloading numpy-2.4.2-cp314-cp314-win_amd64.whl.metadata (6.6 kB)
Using cached highspy-1.13.1-cp314-cp314-win_amd64.whl (2.3 MB)
Downloading numpy-2.4.2-cp314-cp314-win_amd64.whl (12.4 MB)
   ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 12.4/12.4 MB 75.4 MB/s  0:00:00
Installing collected packages: numpy, highspy
Successfully installed highspy-1.13.1 numpy-2.4.2
(dbgpresolve) PS C:\Users\Administrator\dbg_presolve> python dbg_presolve.py
Running HiGHS 1.13.1 (git hash: 1d267d9): Copyright (c) 2026 under MIT licence terms
[...]

PR #2942, opened 2026-04-01
Check dynamism in Chvatal-Gomory strengthening

fwesselm:
- I noticed that the scaling performed by Chvatal-Gomory strengthening (in `HPresolve::rowPresolve`) creates issues for badly scaled models (e.g., #2900).
- The proposed change is to skip the scaling if the dynamism of the row is large (greater than 1e5).
- Testing on 850+ MIPs showed that the change yields a small speedup (1% for running times and numbers of branching nodes). Moreover, 3 additional models are solved to optimality within the time limit.


fwesselm:
> Looks good! Nice find for this being the reason the other instance was failing. Results sounds super promising too.
> 
> Was `1e5` picked because that's the closest power of 10 that fixes the attached issue?

Thanks, @Opt-Mucca! Yes, but I only did limited testing with the threshold. I could do some more testing in the future.

Opt-Mucca:
> Thanks, @Opt-Mucca! Yes, but I only did limited testing with the threshold. I could do some more testing in the future.

Nonononono. Just out of interest. We use `1e6` for cut generation, but I'm perfectly happy with `1e5` (and you've even proven that it works well). I'm going to merge this now!

Issue #2949, opened 2026-04-03
Make bail-out from dependent equations rule deterministic


jajhall:
Bail-out from the dependent equations rule is much enhanced due to #2950, but is still not deterministic. This should be possible using the pseudo-clock in `HFactor`

PR #2950, opened 2026-04-03
Presolve profile

jajhall:
Adds profiling of presolve and enhancement of dependent equations time-out

Dependent equations time-out is still not deterministic: see #2949


jajhall:
@fwesselm This is now ready for review. It mainly implements profiling for LP presolve and improvements to the dependent equations presolve rule. 

However - motivated by presolving models for a client - I've also introduced the `presolve_light` option which switches off the expensive presolve rules. 

I've also added the method `HighLp::numNz`, and used it wherever there was `a_matrix_.numNz()`, so there are quite a few small changes in a number of files.

There's some more refactoring of presolve that I want to do in order to address #3040, but I'll do that in a separate branch.

highs/util/HFactor.cpp
fwesselm, line 1194: Is 

`num_active_nz_ -= mc_count_a[jColPivot];`

needed before calling `zeroCol`?

jajhall: Yes, I would think so. Good spot!

PR #2951, opened 2026-04-03
Allow kCallbackLogging to persist when using getIis

haman80:
Allow kCallbackLogging to persist when using getIis in order to enable users to redirect output

### DIFF
--- a/highs/lp_data/HighsIis.cpp
+++ b/highs/lp_data/HighsIis.cpp
@@ -624,8 +624,10 @@ HighsStatus HighsIis::compute(const HighsLp& lp, const HighsOptions& options,
     // Setting the logging callbacks currently serves no purpose since
     // output_flag is set to kIisDevReport which is false (unless
     // developing) so that solves with this Highs instance are silent
+    /*
     if (log_options.user_log_callback)
       highs.setLogCallback(log_options.user_log_callback);
+    */
     highs.setCallback(callback.user_callback, callback.user_callback_data);
     if (callback.active[kCallbackLogging])
       highs.startCallback(kCallbackLogging);


haman80:
@jajhall Apologies for opening a PR so close to the release. I missed including kCallbackLogging in PR #2941 which is also required for output redirection.

jajhall:
We should also permit the original logging callback, and have a unit test for handling callbacks in IIS, as code coverage doesn't reach it.

I'll do this

jajhall:
Setting the logging callbacks currently serves no purpose since output_flag is set to `kIisDevReport` which is false (unless developing) so that solves with the Highs instance in `HighsIis::compute` are silent.

However, I have added `log_options.user_log_callback` to the callbacks propagated in `HighsIis::compute` and added `iis-logging-callback` to `TestIis.cpp` for code coverage

jajhall:
These Python failings seem to be due to GitHub @galabovaa: what do you think?

galabovaa:
It is not the CI, I can reproduce it locally and I think it may relate to the use of the deprecated setLogCallback here 

https://github.com/ERGO-Code/HiGHS/pull/2951/changes#diff-89b04c9c7bedd5681c1995b0d01b7caae11ba9530a063c95130837c84657f86aR628

galabovaa:
If I comment out the call and its if statement, the package builds fine. I suppose getIis is exposed in the python bindings but setLogCallback is not, and that somehow causes the failure

jajhall:
Many thanks @galabovaa, I'll just drop the use of the deprecated logging callback - and the logging is muted anyway, so nothing is lost

PR #2962, opened 2026-04-10
Remove continuous singletons from double-sided rows

Opt-Mucca:
General idea: We currently have no explicit reduction for a column `s`, with `0` cost, and that only appears in a single double-sided row `b_0 <= a^Tx + cs <= b_1`. Dual fixing removes all cases where `b_0 == -kHighsInf` or `b_1 == kHighsInf`, but the column is left alone if `b_0` and `b_1` are finite. Removing the column may allow additional reductions to be made.

Motivation: I was playing around with an instance, trying to figure out why HiGHS wasn't making a reduction, and thought this would help (spoiler: it did not), but while doing so I stumbled upon an old TODO in the code with the exact suggestion.

@fwesselm Three major questions:
- Do you think this is worth adding? I haven't performance tested this.
- If so should this be kept behind a MIP-only flag?
- I'm currently skipping the case where `s` has infinite bounds, but in theory we could handle that here. The issue is I'm not sure what to do with infinite values in postsolve.

Edit: Need to look into the failing tests (they're passing locally.....). Windows does not like this change at all.


fwesselm:
This is great, @Opt-Mucca! I would like to experiment with this a llittle more. Also, there is `HPresolve::removeSlacks` which handles a "similar" case. If you are OK with it, I would make some more changes (on top of your change).

Opt-Mucca:
@fwesselm I was worried that I was missing something, so please make any changes you want.

jajhall:
@Opt-Mucca I've merge the fix for `HighsPostsolveStack::ZeroObjSingletonContinuousCol::undo` into this branch, and merged in latest. There's some checking code that prints diagnostic information in the event of errors that assert with debug, and this will disappear when some testing on larger MIPs has been done, and the maths has been looked at properly. I'd like to say that I'll have time to think about the latter in the next few days, but I won't. 

`TEST_CASE("issue-2962", ` is the example that previously yielded iterations after postsolve due to one of the ZeroObjSingletonContinuousCol being made basic in postsolve, leading to an incorrect number of basic variables.

jajhall:
This is now draft until the maths has been checked and the error-checking code has been removed

jajhall:
`HPresolve::redundantSingletonColDoubleSidedSlack` is now renamed `HPresolve::zeroCostSingleton`, and handles zero cost singletons with any bounds. It's also applied to equations, which exposed a serious error in `HPresolve::dualFixing`. This error has been fixed.

`zeroCostSingleton` and `dualFixing` are also logged, and can be switched off.

highs/presolve/HighsPostsolveStack.cpp
```diff
@@ -1392,4 +1392,98 @@ void HighsPostsolveStack::SlackColSubstitution::undo(
...
+          static_cast<double>((origRowLower - act) / coef);
+      double col_value_for_upper =
+          static_cast<double>((origRowUpper - act) / coef);
+      if (col_value_for_lower >= lb - primal_tol &&
+          col_value_for_lower <= ub + primal_tol) {
+        // Can set column basic at this value
+        col_value = col_value_for_lower;
+        row_at_lower = true;
+      } else if (col_value_for_upper >= lb - primal_tol &&
+                 col_value_for_upper <= ub + primal_tol) {
```
Opt-Mucca, line 1408: Should this bit of the code be an `else` with an assert inside of it?

jajhall: No, because this is only determining whether a column value can be chosen so that the row is at a bound - in which case the row is nonbasic at its bound, and the column becomes basic

If both tests fail, then the row remains basic.

jajhall: When I developed this I had a check that `col_value` was assigned to a finite value (having been initialised to `kHighsInf`) but now I don't see necessarily why this is necessarly true.

I'll run a sanity check on my LP test set.

Opt-Mucca: The `col_value` was the bit I should've explicitly mentioned. Maybe we should be initialising the assignment to one of its bounds instead of `kHighsInf`? I guess either way there will be some repair step if this falls through.

jajhall: No, it shouldn't be initialised to one of its bounds - precisely in case the logic falls though and it's set to a bound unjustifiably. 

I believe that, mathematically, one of the tested outcomes must occur - because the row bounds were modified according to the column bounds. So, whatever value the row has, there must be a corresponding column value within its bounds.

The print statement I put in the "else" that would catch `col_value` not being defined isn't triggered for all the LPs I'm testing

jajhall:
@fwesselm, this now looks correct for LPs. 

My `fix-2962` branch fixed an error that caused `pds-20` to have 3 iterations after basis postsolve. I've tested on Netlib, Kennington and relaxations of miplib2017 problems and no unexpected iteraitons after postsolve. Interestingly, Netlib problems `fit1p` and `fit2p`, and the relaxation of Miplib problem `pg` are now reduced in presolve, whereas previously they were unreduced.

This is good to merge into latest, particularly as if fixes a bug in basis postsolve

highs/presolve/HighsPostsolveStack.cpp
```diff
@@ -1345,4 +1345,98 @@ void HighsPostsolveStack::SlackColSubstitution::undo(
   }
 }
 
+void HighsPostsolveStack::ZeroObjSingletonContinuousCol::undo(
+    const HighsOptions& options, const std::vector<Nonzero>& rowValues,
+    HighsSolution& solution, HighsBasis& basis) {
+  // a (removed) cut may have been used in this reduction.
+  bool isModelRow = static_cast<size_t>(row) < solution.row_value.size();
```
Opt-Mucca: I think this check for `isModelRow` is now removed following the recent changes by @fwesselm (only 80% sure)

highs/lp_data/HighsModelUtils.cpp
```diff
@@ -1379,21 +1379,27 @@ std::string utilSolutionStatusToString(const HighsInt solution_status) {
 }
 
 // Return a string representation of HighsBasisStatus
-std::string utilBasisStatusToString(const HighsBasisStatus basis_status) {
+std::string utilBasisStatusToString(const HighsBasisStatus basis_status,
```
fwesselm: Is this change used anywhere?

I was wondering whether it would make sense to return a `std::pair<std::string,std::string>` that could store both the long and short descriptions.

highs/lp_data/HighsSolution.cpp
```diff
@@ -2080,7 +2081,7 @@ bool reportKktFailures(const HighsLp& lp, const HighsOptions& options,
 
   highsLogUser(log_options, log_type, "Solution optimality conditions%s%s\n",
                message == "" ? "" : ": ", message == "" ? "" : message.c_str());
-  if (is_mip && info.max_integrality_violation >= 0)
+  if (solved_as_mip && info.max_integrality_violation >= 0)
```
fwesselm, line 2084: I think I am missing something, but could this not be `if (integrality_error).`..

highs/presolve/HighsPostsolveStack.cpp
fwesselm: Since #3218 the solution (e.g., `solution.row_value`) will be "large" enough to hold data for cuts. Therefore, `isModelRow` will always be `true` and the logic can be removed.

highs/presolve/HPresolve.cpp
```diff
@@ -2376,6 +2376,9 @@ HighsTripletTreeSliceInOrder HPresolve::getSortedRowVector(HighsInt row) const {
 }
 
 void HPresolve::markRowDeleted(HighsInt row) {
+  // Next line was in zero-obj-singleton-double-sided-row, but removed
+  // in latest
+  assert(!analysis_.logging_on_);
```
fwesselm, line 2381: Would you like to keep this?

highs/presolve/HPresolve.cpp
```diff
@@ -2393,6 +2396,9 @@ void HPresolve::markRowDeleted(HighsInt row) {
 }
 
 void HPresolve::markColDeleted(HighsInt col) {
+  // Next line added so markColDeleted has same check as
+  // markRowDeleted
```
fwesselm, line 2400: Would you like to keep this?

highs/presolve/HPresolve.cpp
```diff
@@ -5334,13 +5344,19 @@ HPresolve::Result HPresolve::dualFixing(HighsPostsolveStack& postsolve_stack,
...
   computeLocks(col, true, lockCallback);
 
-  // check if variable can be fixed
+  // Check if variable is fixed: only possible if there are no up
+  // (down) locks, and the cost forces it up (down) to its bound
   if (numDownLocks == 0 || numUpLocks == 0) {
-    // fix variable
-    if (numDownLocks == 0)
+    // fix variable if cost is driving it to its bound
+    if (numDownLocks == 0 &&
```
fwesselm: The number of down-locks and up-locks includes the objective coefficient, so `numDownLocks == 0` implies that `model->col_cost_[col] >= 0`. Similarly, `numUpLocks == 0` implies `model->col_cost_[col] <= 0`.

fwesselm (commented):
Thanks, @jajhall, just a few comments from me.

highs/lp_data/HighsModelUtils.cpp
jajhall: It's not currently used, but I found it very valuable when debugging basis postsolve so I thought it worth keeping.

Since `status_string.first` and `status_string.second` are opaque, I've created the struct `StatusString` with members `full_` and `s2_` to be returned by `utilBasisStatusToString`. Obviously this allows us to extend what's returned - and allows other "toString" methods in `HighsModelUtiuls.h` to return more succinct strings.

`Highs::BasisStatusToString` still only returns the `full_` string, as this would be a breaking change in the `Highs` class and some folk may be using it.

highs/lp_data/HighsSolution.cpp
jajhall: `getKktFailures` determines all the KKT and integrality failures for a `HighsLp` instance, blind to whether the relaxation is being solved. 

If the model is an LP, or the relaxation of a MIP is being solved, `max_integrality_violation` will have its "illegal" value of `inf` to indicate that it's not relevant. Unlike the other violations (such as  `num_primal_infeasibilities`) there's no count of integrality violations (with illegal value of -1) so the corresponding `if (num_integrality_violation >= 0)` can't be used to suppress logging 

> max           inf                                  integrality violations

highs/presolve/HighsPostsolveStack.cpp
jajhall: Thanks: removed `isModelRow`

jajhall: Thanks, observed with full confidence by @fwesselm :-)

highs/presolve/HPresolve.cpp
jajhall: Yes, and that's why I added the corresponding assert in `markColDeleted`. If either of `numDeletedRows` or `numDeletedCols` is updated whilst presolve logging is on, then the accounting is messed up, so this guards against later changes in presolve doing just that.

highs/presolve/HPresolve.cpp
jajhall: If I recall correctly, it was correcting an edge case of your logic that prompted my modification. If the cost is of the right sign but sufficiently small, then it doesn't force the column to its bound. 

That said, I may have a sign wrong

I'll re-investigate and get back to you

jajhall: > In my PR I had something like this:
> 
> ```
>   if (numDownLocks == 0) {
>      // avoid fixing zero-cost column to an infinite bound
>      if (model->col_cost_[col] > 0 || model->col_lower_[col] != -kHighsInf)
>        HPRESOLVE_CHECKED_CALL(fixColToLower(postsolve_stack, col));
>    } else {
>      // avoid fixing zero-cost column to an infinite bound
>      if (model->col_cost_[col] < 0 || model->col_upper_[col] != kHighsInf)
>        HPRESOLVE_CHECKED_CALL(fixColToUpper(postsolve_stack, col));
>    }
> ```

I can't find this PR, but it prompts me to fix what I had, and extend what you did so that we're now always fixing if `numDownLocks == 0 || numUpLocks == 0`.

The code I've just pushed is hopefully clear, but the essence is that if the cost is sufficiently positive (negative) then the column must be fixed to its lower (upper) bound.  Otherwise, the cost is less than `dual_feasibility_tolerance` in magnitude, so the value of the column is not uniquely defined, and can be fixed to any value. Ideally this should be zero - for minimal "disturbance" - but if $$0\not\in(l,u)$$ then the value of $$l$$ and $$u$$ closest to zero is used.

This means that, for 25fv47 - the LP that was originally found to be unbounded or infeasible in error because a variable with bounds $$[0, \infty)$$ and cost 0 was fixed to $$\infty$$ and then not fixed with my previous code - is now fixed to 0.

fwesselm: I also had another look and I think the following would work and handle zero cost correctly (slight modification of the code from my original PR):

 ```
 while (true) {
    if (numDownLocks == 0 &&
        (model->col_cost_[col] > 0 || model->col_lower_[col] != -kHighsInf)) {
      HPRESOLVE_CHECKED_CALL(fixColToLower(postsolve_stack, col));
      break;
    }
    if (numUpLocks == 0 &&
        (model->col_cost_[col] < 0 || model->col_upper_[col] != kHighsInf)) {
      HPRESOLVE_CHECKED_CALL(fixColToUpper(postsolve_stack, col));
    }
    break;
  }
```

jajhall: > I also had another look and I think the following would work and handle zero cost correctly (slight modification of the code from my original PR):
> 
> ```
> while (true) {
>    if (numDownLocks == 0 &&
>        (model->col_cost_[col] > 0 || model->col_lower_[col] != -kHighsInf)) {
>      HPRESOLVE_CHECKED_CALL(fixColToLower(postsolve_stack, col));
>      break;
>    }
>    if (numUpLocks == 0 &&
>        (model->col_cost_[col] < 0 || model->col_upper_[col] != kHighsInf)) {
>      HPRESOLVE_CHECKED_CALL(fixColToUpper(postsolve_stack, col));
>    }
>    break;
>  }
> ```

I appreciate the desire to fix variables, but if `|cost| <= dual_feasibility_tolerance` then the variable is dual feasible at any value, so there's no forcing fix. If a small cost is zero with exact arithmetic, but nonzero due to rounding error then, with your logic, there's a chance of fixing a variable to an infinite bound in error. 

I'd rather be cautious and use 

> while (true) {
   if (numDownLocks == 0 &&
       (model->col_cost_[col] > options->dual_feasibility_tolerance || model->col_lower_[col] != -kHighsInf)) {
     HPRESOLVE_CHECKED_CALL(fixColToLower(postsolve_stack, col));
     break;
   }
   if (numUpLocks == 0 &&
       (model->col_cost_[col] < -options->dual_feasibility_tolerance || model->col_upper_[col] != kHighsInf)) {
     HPRESOLVE_CHECKED_CALL(fixColToUpper(postsolve_stack, col));
   }
   break;
}

I think that very few fixing opportunities will be missed, as original costs are likely to be 0 or greater than `dual_feasibility_tolerance`, and numerical cancelling of a cost is typically exact.

fwesselm: @jajhall, my suggestion was not using a tolerance, thank you for adding this!

fwesselm (approved):

PR #2963, opened 2026-04-10
Round bounds before calling changeColBounds

Opt-Mucca:
Spotted this while going through `HPresolve`: If you pass `model->col_lower` or `model->col_upper` to `changeColBounds` then it will not perform the tightening based on potential integrality upgrades because the `<` `>` condition isn't satisfied. I think we're then not updating some bounds that we could have been tightening. 

Other option: Add something like `bool force` to `changeColBounds` for these cases. (if there's a strong preference to make sure integer rounding is only done in one spot)

### DIFF
--- a/highs/presolve/HPresolve.cpp
+++ b/highs/presolve/HPresolve.cpp
@@ -445,7 +445,8 @@ HPresolve::StatusResult HPresolve::convertImpliedInteger(HighsInt col,
 
   // round and update bounds
   return StatusResult(
-      changeColBounds(col, model->col_lower_[col], model->col_upper_[col]));
+      changeColBounds(col, std::ceil(model->col_lower_[col] - primal_feastol),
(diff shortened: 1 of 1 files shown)


fwesselm:
> Spotted this while going through `HPresolve`: If you pass `model->col_lower` or `model->col_upper` to `changeColBounds` then it will not perform the tightening based on potential integrality upgrades because the `<` `>` condition isn't satisfied. I think we're then not updating some bounds that we could have been tightening.
> 
> Other option: Add something like `bool force` to `changeColBounds` for these cases. (if there's a strong preference to make sure integer rounding is only done in one spot)

Thanks for checking the presolve code, @Opt-Mucca!

I am confused (as usual); could you tell me which condition would prevent rounding of the bounds?

Opt-Mucca:
Ahhhhhhhhhh. You're not confused, I was just reading the function wrong...... My bad.

Issue #2947, opened 2026-04-02
Hessian format square rejected


atraplet:
I am are currently working on a Java wrapper for HiGHS https://github.com/atraplet/highs4j

During development, I found that the Hessian format square seems to be always rejected, in particular Highs::qFormatOk() in highs/lp_data/HighsInterface.cpp:2726 seems to unconditionally reject kSquare. 

Do I miss something, or is this a bug or not yet supported?

jajhall:
How are you passing your Hessian? As part of a model or via passHessian? 

'Highs::qFormatOk' is for use internally, where the only valid format is kTriangular.

jajhall:
Added `HighsHessian::toSquare` and unit test `pass-square-hessian` to test conversion to square Hessian and passing it via `Highs::passHessian`, and via `Highs::passModel` with square Hessian.

No evidence from @atraplet  of square Hessian being rejected, and unit test confirms that it's not happening.

Closed by #2952

atraplet:
@jajhall 

I am sorry for not answering earlier. 

Running latest code (v1.14.0) `call_highs_from_c_minimal.c` with the same Hessian, but in format square gives

`call_highs_from_c_minimal: /home/atrap/github/c/HiGHS/highs/lp_data/HighsInterface.cpp:2881: bool Highs::qFormatOk(HighsInt, HighsInt): Assertion `ok_format' failed.`

Patch for `call_highs_from_c_minimal.c`  to reproduce:

```
Index: examples/call_highs_from_c_minimal.c
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/examples/call_highs_from_c_minimal.c b/examples/call_highs_from_c_minimal.c
--- a/examples/call_highs_from_c_minimal.c	(revision dcc25308d890237531422fc4d0e2e8a4740c8608)
+++ b/examples/call_highs_from_c_minimal.c	(date 1775833454964)
@@ -223,11 +223,17 @@
   const HighsInt a_index[3] = {0, 1, 2};
   const double a_value[3] = {1.0, 1.0, 1.0};
 
-  const HighsInt q_format = kHighsHessianFormatTriangular;
-  const HighsInt q_num_nz = 4;
-  const HighsInt q_start[3] = {0, 2, 3};
-  const HighsInt q_index[4] = {0, 2, 1, 2};
[...]

PR #2965, opened 2026-04-11
`Highs::qFormatOk` now allows square Hessian

jajhall:
Added C API unit tests passing QP with square Hessian, and just square Hessian


jajhall:
This closes #2947

filikat (commented):
Looks good to me

PR #2966, opened 2026-04-11
Added sub-solver timers to `hmw-mt`

jajhall:
The sub-solver timers - sub-MIPs, and LP solves (primal/dual simplex with/without basis and IPM) - are now multi-threaded.

They are now (necessarily) built on common memory for recording sub-solver timing data across all threads that's passed down via a pointer. Hence the calls to `Highs::setGlobalSubSolverCallTime` whenever a new Highs instance is created - either explicitly, or by creating a HighsLpRelaxation instance.

I had to develop the multi-threaded sub-solver timers in a branch `thread-mip-timer` of latest, and then merge them into `hmw-mt-timer` - which is a branch of the most recent `hmw-mt`.

This PR merges `hmw-mt-timer` into `hmw-mt` so that @Opt-Mucca can see what's changed in the MIP solver. Of course, this does mean that a whole bunch of changes in `latest` are merged in, too, but that's not a bad thing!

The changes in `highs/mip` are mainly simplifications of what was there before - when sub-solver data from one level was added in to data at the higher level. All that has gone now that common memory is passed around

In the course of this, I created `Highs::initializeGlobalScheduler();` to call `highs::parallel::initialize_scheduler(this->options_.threads);` (and perform some subsequent error checking and logging)  rather than do it with duplicated code in two places. I also needed to add my own stuff.

Currently the sub-solver timing data are always logged - for ease of development. Normally it is dependent on `this->options_.log_dev_level > 0`

The more detailed timing of MIP computational components will be done the same way, but I wanted to get this working first.


Opt-Mucca:
@jajhall will review this sometime today!

check/TestMipSolver.cpp
```diff
@@ -1162,6 +1162,7 @@ TEST_CASE("mip-lp-solver", "[highs_test_mip_solver]") {
 #endif
 }
 
+/*
```
Opt-Mucca, line 1165: Why is this test now commented out?

highs/lp_data/HighsInterface.cpp
```diff
@@ -4286,59 +4296,267 @@ void HighsSubSolverCallTime::initialise() {
...
       highsLogUser(options_.log_options, HighsLogType::kInfo, "%s\n",
                    ss.str().c_str());
     }
+    hrule();
+    ss.str(std::string());
+    ss << "Total                ";
+    for (HighsInt thread_ix = 0; thread_ix < HighsInt(num_threads_used);
+         thread_ix++) {
+      if (totalPct[thread_ix]) {
+        ss << highsFormatToString(" %5.1f", totalPct[thread_ix]);
```
Opt-Mucca, line 4553: That's a great breakdown! (Did not consider printing results per thread)

highs/mip/HighsMipAnalysis.cpp
```diff
@@ -49,123 +68,63 @@ void HighsMipAnalysis::mipTimerStart(const HighsInt mip_clock
...
   if (!analyse_mip_time) return;
-  HighsInt highs_timer_clock = mip_clocks.clock_[mip_clock];
-  if (highs_timer_clock == check_mip_clock) {
-    std::string clock_name =
-        mip_clocks.timer_pointer_->clock_names[check_mip_clock];
-    printf("MipTimer: stopping clock %d: %s\n", int(check_mip_clock),
-           clock_name.c_str());
-  }
-  mip_clocks.timer_pointer_->stop(highs_timer_clock);
+  HighsInt local_thread_id = 0;  // highs::parallel::thread_num();
```
Opt-Mucca, line 91: Is the comment a TODO?

Opt-Mucca (approved):
Looks good @jajhall ! I've made some minor comments (assuming the `printf` statements will all be removed at some later point). Am a big fan of this abstracting away the `+ / -` changes for run times in the middle of the MIP code.
Do you want to keep this on the separate branch for now, or merge it and finish it directly on `hmw-mt`?

check/TestMipSolver.cpp
jajhall: This is because it's testing consistency between the subsolver times and the corresponding times in the MIP profiling - which isn't working at present. 

I should have a unit test for the sub-solver timing for code coverage

highs/lp_data/HighsInterface.cpp
jajhall: Thanks. I view this as a way to see how many threads are being used, and to what activity.

highs/mip/HighsMipAnalysis.cpp
jajhall: Sorry, there's a lot of incomplete code for the MIP profiling

jajhall:
> Looks good @jajhall ! I've made some minor comments (assuming the `printf` statements will all be removed at some later point). Am a big fan of this abstracting away the `+ / -` changes for run times in the middle of the MIP code. Do you want to keep this on the separate branch for now, or merge it and finish it directly on `hmw-mt`?

Since I'm on the right track, I'll improve let's merge it. I'll do the MIP profiling later on a separate branch

PR #2969, opened 2026-04-13
Mangle symbols of metis, amd, rcm

filikat:
- All symbols exported by metis, amd and rcm are now mangled with ``Highs_`` in front of the symbol name.
- I added a new workflow that checks that there is no conflict with the symbols of Metis from github. The test uses ``nm`` to extract the symbols within the text section of ``libhighs`` and ``libmetis``. It then uses ``comm`` to find the common symbols and checks that the list of common symbols is empty. This test fails using ``latest``, returning the expected 11 duplicate symbols, while it passes on this branch after mangling.

### DIFF
--- a/.github/workflows/hipo-mangle.yml
+++ b/.github/workflows/hipo-mangle.yml
@@ -1,0 +1,80 @@
+name: hipo-mangle
+
+on: [push, pull_request]
+
+jobs:
+  check-metis:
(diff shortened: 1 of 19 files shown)


filikat:
Closes #2823

galabovaa (approved):
Looks good to me!

PR #2972, opened 2026-04-15
Add tiny function for checking row index to simplify postsolve

fwesselm:
Add member function `isModelRow` to struct `HighsSolution` in order to simplify postsolve a little.


Opt-Mucca (approved):
Happy with the change! Everything looks good.

PR #2974, opened 2026-04-16
Bazel sanitizer test action

galabovaa:
@jajhall @Opt-Mucca 

Following our discussion on Tuesday, this is the sanitizer test that is failing atm. 

I have disabled some of it, so only the failing part gets executed quickly. It runs only on workflow dispatch, we will make it run on PR once it is fixed. This way when you merge latest in your branches, we will be able to run it on your branches as well.


Opt-Mucca (approved):
@galabovaa The failing test seems unrelated. Happy with disabling this until `hmw-mt` (and the timers within) are merged.
Also: 3000 tests would have taken a while

PR #2978, opened 2026-04-17
Various changes to HiPO

filikat:
- The structure and values of normal equations or augmented system have been moved to a separate class ``KktMatrix``. Some auxiliary functions used for the factorisation phase have been moved to ``auxiliary``. This allows to use different ``LinearSolver`` objects to solve the linear systems, in preparation to the up-looking factorisation.
- HiPO now computes upper and lower bounds for the number of nonzeros in the normal equations matrix. If the augmented system is much sparser than the lower bound of normal equations, the former is selected automatically. If the augmented system is much dense than the upper bound of normal equations, the latter is selected automatically. This approach avoids applying metis or amd to matrices that are very unlikely to be preferred, saving some time in the analyse phase for some problems. This approach is much more effective than guessing based on the presence of dense columns.
- The HiPO interface to HiGHS logging has been simplified.


jajhall (approved):
Seems to be more elegant and concise. Fine with me

PR #2982, opened 2026-04-20
Bug fixes for getIis

haman80:
- Fixed an issue with near-feasible models where the model status could flip from infeasible to feasible during elasticity filtering: now falls back to the original constraint set and, if requested, proceeds to deletion filtering.
- Improved time-out handling from elasticity filter: If time limit is reached, fall back to the original constraint set and allow deletion filtering to handle the exit
- Improved logging consistency between elasticity filtering and deletion filtering.


jajhall (approved):
Looks fine to me

PR #2984, opened 2026-04-20
Allow slightly asymmetric square Hessians, and revert to using `HighsHashTable` (mainly) In `assessMatrix`

### LINKED ISSUE #2977: Performance regression building LP in v1.14.0
First reported by https://github.com/jump-dev/HiGHS.jl/issues/337

Reproducible MPS:

[medium.zip](https://github.com/user-attachments/files/26803696/medium.zip)

## v1.14.0

```Julia
julia> using JuMP, HiGHS

julia> begin
           model = read_from_file("/Users/odow/Downloads/medium.mps")
           @time MOI.Utilities.attach_optimizer(model)
       end
Running HiGHS 1.14.0 (git hash: 7df0786de3): Copyright (c) 2026 under Apache 2.0 license terms
Using BLAS: blastrampoline 
 19.168187 seconds (18.29 M allocations: 989.328 MiB, 2.97% gc time)
```

## v1.13.1

```Julia
julia> using JuMP, HiGHS

julia> begin
           model = read_from_file("/Users/odow/Downloads/medium.mps")
           @time MOI.Utilities.attach_optimizer(model)
       end
Running HiGHS 1.13.1 (git hash: 1d267d97c): Copyright (c) 2026 under Apache 2.0 license terms
Using BLAS: blastrampoline 
  1.895229 seconds (18.29 M allocations: 989.327 MiB, 33.30% gc time)
```

## Flamegraph

Flamegraph of v1.14 shows time is dominated by calls to `Highs_addRow`:

Image

### LINKED ISSUE #2979: Native heap corruption in v1.14.0 when passHessian receives a square Hessian with epsilon-level asymmetry (via CVXPY/highspy)
## Summary

When a square Hessian with floating-point-level asymmetry (on the order of machine epsilon, ~1e-18) is passed to HiGHS 1.14.0 via `passHessian` from the C API / `highspy`, HiGHS first emits `ERROR: Square Hessian contains N non-symmetries` and then the Python process crashes with native heap corruption (`malloc: Incorrect checksum for freed object`, `free(): invalid size`, `double free or corruption`, `munmap_chunk(): invalid pointer`, SIGABRT / SIGSEGV). The crash is not catchable from Python since the entire interpreter is terminated.

From #2821 / #2857: `passHessian` now returns `kError` on any asymmetry instead of using `(Q+Q^T)/2`. The check at [`highs/model/HighsHessianUtils.cpp#L422`](https://github.com/ERGO-Code/HiGHS/blob/master/highs/model/HighsHessianUtils.cpp) uses strict `!=` equality with no tolerance, so two values that differ by `1e-18` are counted as asymmetric.

The practical impact is that CVXPY's HiGHS backend fails to solve certain QPs whose Hessian has dense off-diagonal entries. In our case, the Hessian originates from `cp.quad_form(w, Sigma)` where Sigma is the result of a PSD projection (`V @ diag(λ) @ V.T`) of a covariance matrix. Although Sigma is mathematically symmetric, the matrix that ultimately reaches HiGHS is not bit-exactly symmetric after floating-point reconstruction and CVXPY's canonicalization, and the new strict equality check flags all such asymmetries as errors.

## Environment

- HiGHS: 1.14.0 (via `highspy 1.14.0`, PyPI wheels)
- CVXPY: 1.7.5
- Python: 3.12
- NumPy 1.26.4, SciPy 1.15.3
- Reproduced on both macOS arm64 (local) and Linux x86_64 (containers). Reproduced sequentially, so not a threading issue.

## Reproduction

```
import numpy as np
import cvxpy as cp
import scipy.sparse as sp

rng = np.random.default_rng(0)

N_PATHS = 150
N_NODES = 50

# ---- Sparse "incidence" matrix mapping paths to nodes ----
rows, cols, vals = [], [], []
for p in range(N_PATHS):
    src, snk = rng.choice(N_NODES, size=2, replace=False)
    rows += [p, p]
    cols += [src, snk]
    vals += [1.0, -1.0]

incidence = sp.csr_matrix((vals, (rows, cols)), shape=(N_PATHS, N_NODES))

# ---- Dense PSD covariance on nodes, built by eigenvalue clamping ----
# (mirrors a typical "project onto PSD cone" step used after estimating a
#  covariance matrix from noisy / partially-missing data)
A = rng.standard_normal((N_NODES, N_NODES))
Sigma_raw = A @ A.T + rng.standard_normal((N_NODES, N_NODES)) * 0.01
Sigma_sym = 0.5 * (Sigma_raw + Sigma_raw.T)
w_eig, V_eig = np.linalg.eigh(Sigma_sym)
w_eig = np.maximum(w_eig, 0.0)
Sigma = V_eig @ np.diag(w_eig) @ V_eig.T
# Sigma is mathematically symmetric and PSD, but not bit-exact symmetric:
print("max |Sigma - Sigma.T|:", np.max(np.abs(Sigma - Sigma.T)))
print("min eigenvalue:", float(np.linalg.eigvalsh(Sigma).min()))

# ---- Build the CVXPY problem ----
w_path = cp.Variable(N_PATHS, nonneg=True)

# Linear transformation to "nodal" space via the sparse incidence matrix.
w_nodal = w_path @ incidence  # shape: (N_NODES,)

# Some realistic constraints so the problem is bounded and not trivial.
path_coef = rng.standard_normal(N_PATHS)
path_upper = 50.0
constraints = [
    w_path <= path_upper,
    cp.sum(w_path) <= 500.0,
]

cov_penalty = 0.1
diag_penalty = 0.01
diag_weights = rng.uniform(0.1, 2.0, size=N_NODES)

obj_terms = [
    cp.sum(cp.multiply(path_coef, w_path)),
    -diag_penalty * cp.sum_squares(cp.multiply(diag_weights, w_nodal)),
    -cov_penalty * cp.quad_form(w_nodal, Sigma),
]
prob = cp.Problem(cp.Maximize(cp.sum(obj_terms)), constraints)

prob.solve(solver=cp.HIGHS, verbose=True)
```

results in

```
Running HiGHS 1.14.0 (git hash: 7df0786): Copyright (c) 2026 under MIT licence terms
ERROR:   Square Hessian contains 743 non-symmetries
QP has 401 rows; 250 cols; 1150 matrix nonzeros; 1325 Hessian nonzeros
Coefficient ranges:
  Matrix  [1e-01, 2e+00]
  Cost    [3e-03, 3e+00]
  Hessian [3e-03, 1e+01]
  Bound   [0e+00, 0e+00]
  RHS     [5e+01, 5e+02]
  Iteration        Objective     NullspaceDim
python(76500,0x1f2066140) malloc: Incorrect checksum for freed object 0x132c4f600: probably modified after being freed.
Corrupt value: 0xc004115f9637ddc9
python(76500,0x1f2066140) malloc: *** set a breakpoint in malloc_error_break to debug
[1]    76500 abort      python
```

jajhall:
Added a test on the difference between corresponding off-diagonal entries in a square Hessian, allowing them to differ by 1e-10. This should close #2979 

Reverted to using `HighsHashTable` (mainly) In `assessMatrix` to identify duplicate entries in a matrix column/row, unless a Hessian is being assessed (so duplicate entries are summed) when an `unordered_map` is used. This closes #2977

### DIFF
--- a/check/TestFilereader.cpp
+++ b/check/TestFilereader.cpp
@@ -566,3 +566,52 @@ TEST_CASE("lp-duplicate-variable", "[highs_filereader]") {
 
   std::remove(lp_file.c_str());
 }
+
+inline double getWallTime() {
+  using namespace std::chrono;
+  return duration_cast<duration<double> >(std::chrono::high_resolution_clock::now().time_since_epoch()).count();
+}
+
+TEST_CASE("efficient-add-row", "[highs_filereader]") {
+  std::string filename;
+  filename =
+    // std::string(HIGHS_DIR) + "/check/instances/adlittle.mps";
+    "/srv/mps_da/cont1.mps.gz";
+  Highs h;
+  //  h.setOptionValue("output_flag", dev_run);
+  REQUIRE(h.readModel(filename) == HighsStatus::kOk);
+  HighsLp lp = h.getLp();
+  h.passModel(lp.num_col_, 0, 0, 0,
+	      0, 0, 0, 0.0,
+	      lp.col_cost_.data(),
+	      lp.col_lower_.data(),
+	      lp.col_upper_.data(),
+	      nullptr, nullptr,
+	      nullptr, nullptr, nullptr,
+	      nullptr, nullptr, nullptr,
+	      nullptr);
+  lp.a_matrix_.ensureRowwise();
+  double tt = -getWallTime();
+  for (HighsInt iRow = 0; iRow < lp.num_row_; iRow++) {
+    HighsInt iEl = lp.a_matrix_.start_[iRow];
+    h.addRow(lp.row_lower_[iRow], lp.row_upper_[iRow],
+	     lp.a_matrix_.start_[iRow+1] - iEl,
+	     &lp.a_matrix_.index_[iEl],
+	     &lp.a_matrix_.value_[iEl]);
+  }
+  tt += getWallTime();
+  printf("Added %d rows individually in %.2gs\n", int(lp.num_row_), tt);
+
+  REQUIRE(h.deleteRows(0, lp.num_row_-1) == HighsStatus::kOk);
+  REQUIRE(h.getLp().num_row_ == 0);
+
+  tt = -getWallTime();
+  h.addRows(lp.num_row_,
+	    lp.row_lower_.data(), lp.row_upper_.data(),
+	    lp.a_matrix_.start_[lp.num_row_],
+	    lp.a_matrix_.start_.data(),
+	    lp.a_matrix_.index_.data(),
+	    lp.a_matrix_.value_.data());
+  tt += getWallTime();
+  printf("Added %d rows together     in %.2gs\n", int(lp.num_row_), tt);
+}
--- a/check/TestFilereader.cpp
+++ b/check/TestFilereader.cpp
@@ -576,7 +576,7 @@ TEST_CASE("efficient-add-row", "[highs_filereader]") {
   std::string filename;
   filename =
     // std::string(HIGHS_DIR) + "/check/instances/adlittle.mps";
-    "/srv/mps_da/cont1.mps.gz";
+    "/srv/mps_da/neos.mps.gz";
   Highs h;
   //  h.setOptionValue("output_flag", dev_run);
   REQUIRE(h.readModel(filename) == HighsStatus::kOk);
--- a/check/TestFilereader.cpp
+++ b/check/TestFilereader.cpp
@@ -569,49 +569,45 @@ TEST_CASE("lp-duplicate-variable", "[highs_filereader]") {
 
 inline double getWallTime() {
   using namespace std::chrono;
-  return duration_cast<duration<double> >(std::chrono::high_resolution_clock::now().time_since_epoch()).count();
+  return duration_cast<duration<double> >(
+             std::chrono::high_resolution_clock::now().time_since_epoch())
+      .count();
 }
 
 TEST_CASE("efficient-add-row", "[highs_filereader]") {
   std::string filename;
-  filename =
-    // std::string(HIGHS_DIR) + "/check/instances/adlittle.mps";
-    "/srv/mps_da/neos.mps.gz";
+  filename = std::string(HIGHS_DIR) + "/check/instances/adlittle.mps";
+  //    "/srv/mps_da/neos.mps.gz";
   Highs h;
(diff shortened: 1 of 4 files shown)


highs/model/HighsHessianUtils.cpp
```diff
@@ -403,10 +403,17 @@ HighsStatus normaliseHessian(const HighsOptions& options,
   upper_off_diagonal.assign(dim, 0);
   // Should be no duplicates
   HighsInt debug_num_duplicate = 0;
-  HighsInt num_non_symmetric = 0;
   HighsInt num_summation = 0;
   HighsInt num_upper_triangle = 0;
   HighsInt num_hessian_el = 0;
+  const double kSquareHessianAsymmetryTolerance = 1e-10;
```
odow, line 409: Open question whether to hard-code this or make it an option.

The argument in favour of this hard-coding is that anyone wanting to change it needs a really good reason, and they should almost certainly just use the triangular input.

check/TestFilereader.cpp
```diff
@@ -566,3 +566,48 @@ TEST_CASE("lp-duplicate-variable", "[highs_filereader]") {
...
+  if (dev_run) tt = -getWallTime();
+  for (HighsInt iRow = 0; iRow < lp.num_row_; iRow++) {
+    HighsInt iEl = lp.a_matrix_.start_[iRow];
+    h.addRow(lp.row_lower_[iRow], lp.row_upper_[iRow],
+             lp.a_matrix_.start_[iRow + 1] - iEl, &lp.a_matrix_.index_[iEl],
+             &lp.a_matrix_.value_[iEl]);
+  }
+  if (dev_run) {
+    tt += getWallTime();
+    printf("Added %d rows individually in %.2gs\n", int(lp.num_row_), tt);
```
odow, line 599: Did you benchmark the before/after or should I do it with the Julia example?

jajhall: When loading the LP neos row-by-row, It's about 3x faster with the two current ways of tracking duplicates than with 1.14. Using HighsHashTable (as before 1.14) is slightly faster than unordered_map (unsurprisingly) so I left that as the default - unless duplicates need to be summed

highs/model/HighsHessianUtils.cpp
jajhall: I'd rather hard code it, as you do in JuMP.

If the difference is larger than 1e-10 due to inexact arithmetic, then users should be aware and handling it themselves

PR #2990, opened 2026-04-23
Add single machine scheduling separator

Opt-Mucca:
This PR implements identification of natural-date single-machine scheduling subproblems and derives valid inequalities. See the description of `highs/mip/HighsMachineSchedSeparator.h` for the exact description.

Motivation: The instance `neos-3046615-murg` from MIPLIB 2017 benchmark now solves well within time limit, and `neos-3046601-motu` from MIPLIB 2017 collection now solves in a second or two (both were previously unsolvable within a few hours). This change should introduce some tiny `epsilon` and very positively affect some instances.

Changes:
- Add `highs/mip/HighsMachineSchedSeparator.h` and `highs/mip/HighsMachineSchedSeparator.cpp`. Currently the separator is only called once, and the cuts are only added to the pool. This is efficient, and I noticed no real change by trying to separate more frequently.
- I have removed some presolve code that shifts integer columns to have a lower bound of zero. Such a shift completely destroys any structure of this subproblem and makes it impossible to efficiently find. I've tested this change in isolation locally, and there was only some (positive) performance noise. I don't believe it is done for any purpose other than cosmetics + potentially making it easier to transform the LP when generating cuts, so it should be fine to remove. @fwesselm I may be missing something obvious there....   

Potential improvements that I don't think need to be made:
- Someone could model jobs from `[-inf, 0]` instead of `[0, inf]`. I'd currently not spot these. I'd also not spot a row where both binary columns share the same coefficient value (different signs), although I think presolve is going to scale these rows most of the time already. 
- Computing these inequalities requires finding a clique. I implemented the most naive greedy approach possible. I don't think we have any max-clique code outside of `HighsCliqueTable`, which I don't think was efficient to use in this case (could be wrong). There's likely going to be some instances that this code could improve but fails to because of this naive approach, however it ensures near-zero overhead is added to all other instances. An example: I suspect `supportcase26` could be improved with some better detection.

This should not be merged until a more thorough performance check has been made! @fwesselm This is a way lower priority than the MIP Worker Refactor to review / test.


Opt-Mucca:
I've added some basic clique extraction (was the largest TODO). It now solves the benchmark instance `neos-5107597-kakapo` within time limit reliably.
That instance had overlapping binary variables, e.g., 
```
x = 1 -> y_0 - y_1 >= b_0
x = 0 -> y_1 - y_0 >= b_1
y = 1 -> y_0 - y_1 >= b_2
y = 0 -> y_1 - y_0 >= b_3
b_i >= 0
```
From this you can conclude that `x=y`. I'm not sure how to efficiently store a multi-graph and find these overlaps though, so I only find one half of it at a time, i.e., `x + ~y <= 1`.

highs/mip/HighsMachineSchedSeparator.cpp
```diff
@@ -0,0 +1,298 @@
...
+  };
+  HighsInt largestDegree = 0;
+  HighsInt largestDegreeCol = -1;
+  std::vector<HighsInt> degrees(mipsolver.numCol());
+  const HighsInt maxRows = std::min(HighsInt{50000}, 2 * mipsolver.numRow());
+  HighsHashTable<std::pair<HighsInt, HighsInt>,
+                 std::tuple<double, HighsInt, ArcType>>
+      adjacency(maxRows + 2);
+  HighsHashTable<std::pair<HighsInt, HighsInt>, std::vector<HighsInt>> arcToBin;
+  // 1 -> (i,j) y = 0, 2 -> (i,j) y = 1, 4 -> (j,i) y = 0, 8 -> (j,i) y = 1
```
BenChampion, line 43: I appreciated this comment! Make it less cryptic? Along the lines of: 

"The keys in this map are triples of column indices representing the arc endpoints (continuous variables) i, j and the binary variable y. The values are bit-fields representing which values of y imply which dependency relationships between i and j, as follows. [...]"?

highs/mip/HighsMachineSchedSeparator.cpp
```diff
@@ -0,0 +1,298 @@
...
+  // Only separate once
+  if (separated) return;
+  const HighsMipSolver& mip = lpRelaxation.getMipSolver();
+  std::vector<std::vector<double>> vals;
+  std::vector<std::vector<HighsInt>> inds;
+  std::vector<double> rhss;
+  has_single_machine_schedule = findSingleMachineScheduleClique(
+      vals, inds, rhss, transLp.getGlobaldom(), mip);
+  if (!has_single_machine_schedule) {
+    separated = true;
```
BenChampion: Since we in fact DON'T separate in this case, we could be more precise and call this variable `already_tried` or something like that.

highs/mip/HighsMachineSchedSeparator.cpp
```diff
@@ -0,0 +1,298 @@
...
+    vals[i].back() = -1;
+  }
+
+  return true;
+}
+
+void HighsMachineSchedSeparator::separateLpSolution(
+    HighsLpRelaxation& lpRelaxation, HighsLpAggregator& lpAggregator,
+    HighsTransformedLp& transLp, HighsCutPool& cutpool) {
+  // Only separate once
```
BenChampion: Similar to my other comment, it's more precise to say we only *try* once. (We may not separate.)

highs/presolve/HPresolve.cpp
```diff
@@ -4472,12 +4472,13 @@ HPresolve::Result HPresolve::colPresolve(HighsPostsolveStack& postsolve_stack,
     // check if variable is implied integer
     HPRESOLVE_CHECKED_CALL(static_cast<Result>(convertImpliedInteger(col)));
 
-    // shift integral variables to have a lower bound of zero
+    // shift "binary" variables to have a lower bound of zero
```
BenChampion, line 4480: I see you have an outstanding comment to @fwesselm about this; I don't think I know enough yet to evaluate the implications of changing this.

highs/mip/HighsMachineSchedSeparator.h
```diff
@@ -0,0 +1,60 @@
...
+/*                                                                       */
+/* * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * */
+/**@file mip/HighsMachineSchedSeparator.h
+ * @brief Class for separating release-date cuts, see Section 4.1 in
+ * Recognition and Exploitation of Single-Machine Scheduling Subproblems in
+ * Mixed Integer Programs, Reinout Lambertus Henricus Wijfjes, Msc Thesis, 2022,
+ * or Cutting Plane Algorithm for the Single Machine Scheduling Problem
+ * with Release Times, G. L. Nemhauser and M. W. P. Savelsbergh, 1992.
+ *
+ * Given a set of jobs N with start times s_j, processing times p_ij,
```
BenChampion, line 11: Explain why there are two different indices for processing times?

highs/mip/HighsMachineSchedSeparator.h
```diff
@@ -0,0 +1,60 @@
...
+ * release times r_j, and binary dependency variables y_ij,
+ * where y = 1 -> s_i - s_j < p_ij.
```
BenChampion:
```suggestion
 * where y = 1 -> s_j >= s_i + p_ij.
```
(Even if I'm missing something, my suggestion seems more natural to me.)

highs/mip/HighsMachineSchedSeparator.h
```diff
@@ -0,0 +1,60 @@
...
+/*                                                                       */
+/*    Available as open-source under the MIT License                     */
+/*                                                                       */
+/* * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * */
+/**@file mip/HighsMachineSchedSeparator.h
+ * @brief Class for separating release-date cuts, see Section 4.1 in
+ * Recognition and Exploitation of Single-Machine Scheduling Subproblems in
+ * Mixed Integer Programs, Reinout Lambertus Henricus Wijfjes, Msc Thesis, 2022,
+ * or Cutting Plane Algorithm for the Single Machine Scheduling Problem
+ * with Release Times, G. L. Nemhauser and M. W. P. Savelsbergh, 1992.
```
BenChampion: Since reading these references isn't necessary to understand what you've written below, put these references at the end?

highs/mip/HighsMachineSchedSeparator.cpp
```diff
@@ -0,0 +1,298 @@
...
+    // y_ji = val -> s_i >= s_j + p_ji - M * val
+    // Make an arc from negCol (j) to posCol (i)
+    if (p < 0) return;
+    const auto it = adjacency.find({negCol, posCol});
+    if (it != nullptr) {
+      if (std::get<0>(*it) < p) {
+        adjacency[{negCol, posCol}] = std::make_tuple(p, binCol, t);
+      }
+    } else {
+      degrees[posCol]++;
```
BenChampion:
```suggestion
      ++degrees[posCol];
```
and similar throughout

highs/mip/HighsMachineSchedSeparator.cpp
```diff
@@ -0,0 +1,298 @@
...
+        if (binCol != -1) {
+          machineSchedRow = false;
+          break;
+        }
+        binCol = col;
+        binCoef = mipsolver.mipdata_->ARvalue_[i];
+      } else if (mipsolver.mipdata_->ARvalue_[i] == -1) {
+        negContCol = col;
+      } else if (mipsolver.mipdata_->ARvalue_[i] == 1) {
+        posContCol = col;
```
BenChampion, line 133: Worth early breaks for `{pos,neg}ContCol != -1` (or a comment explaining why not)?

highs/mip/HighsMachineSchedSeparator.cpp
```diff
@@ -0,0 +1,298 @@
...
+      } else {
+        machineSchedRow = false;
+        break;
+      }
+    }
+    if (!machineSchedRow || binCol == -1 || negContCol == -1 ||
+        posContCol == -1 || posContCol == negContCol)
```
BenChampion: Can the last condition here happen?

highs/mip/HighsMachineSchedSeparator.cpp
```diff
@@ -0,0 +1,298 @@
...
+      continue;
+    // We want to put the row into form:
+    // Mx_ji + s_i - s_j >= p_ji, p_ji >= 0
+    // x_ji = 1 -> s_i >= s_j + p_ij - M
```
BenChampion:
```suggestion
    // x_ji = 1 -> s_i >= s_j + p_ji - M
```
Also
* Explain `x`? 
* make this consistent with the rest of the file? You switch between `x` and `y`. Is one supposed to be the complement of the other?

highs/mip/HighsMachineSchedSeparator.cpp
```diff
@@ -0,0 +1,298 @@
...
+    if (rowUpper != kHighsInf) {
+      // Given My_ij + s_i - s_j <= d
+      // The row becomes (after multiplying by -1):
+      // -My_ij + s_j - s_i >= -d
+      // Add implication s_j >= s_i + p_ij + M, p_ij + M > 0 when binCol = 1
+      // Add implication s_j >= s_i + p_ij, p, p_ij > 0 when binCol = 0
```
BenChampion:
```suggestion
      // Add implication s_j >= s_i + p_ij, p_ij > 0 when binCol = 0
```
Also
* explain how you are trying to find which value of the binary gives the tighter inequality implying the dependency relationship? 
* consider making `ArcType` names clearer? (along the lines of `ImplicationWhenOne`, `ImplicationWhenZero`) 
* explain what you are taking `p_ij` to be?

highs/mip/HighsMachineSchedSeparator.cpp
```diff
@@ -0,0 +1,298 @@
...
+        } else {
+          addEntry(negContCol, posContCol, binCol, rhs_1, ArcType::kIfBinOne);
+        }
+        numRows++;
+      }
+    }
+    if (rowLower != -kHighsInf) {
+      // Given Mx_ij + s_i - s_j >= d
+      // Add implication s_i >= s_j + pji - M, p_ji - M > 0 when binCol = 1
+      // Add implication s_i >= s_j + pji, p_ji > 0 when binCol = 0
```
BenChampion:
```suggestion
      // Add implication s_i >= s_j + p_ji - M, p_ji - M > 0 when binCol = 1
      // Add implication s_i >= s_j + p_ji, p_ji > 0 when binCol = 0
```
Also, similar comments to above.

highs/mip/HighsMachineSchedSeparator.cpp
```diff
@@ -0,0 +1,298 @@
...
+    kIfBinZero,
+  };
+  HighsInt largestDegree = 0;
+  HighsInt largestDegreeCol = -1;
+  std::vector<HighsInt> degrees(mipsolver.numCol());
+  const HighsInt maxRows = std::min(HighsInt{50000}, 2 * mipsolver.numRow());
+  HighsHashTable<std::pair<HighsInt, HighsInt>,
+                 std::tuple<double, HighsInt, ArcType>>
+      adjacency(maxRows + 2);
+  HighsHashTable<std::pair<HighsInt, HighsInt>, std::vector<HighsInt>> arcToBin;
```
BenChampion, line 38: Provide an explanatory comment for these data structures? (Like the one below).

In particular:

For `adjacency`
* make it clearer what the `double` in the `tuple` is?
* explain why it's `maxRows + 2`?

For `arcToBin`
* Explain why we map to a `vector`?

highs/mip/HighsMachineSchedSeparator.cpp
```diff
@@ -0,0 +1,298 @@
...
+    for (const auto& entry : arcToBin) {
+      std::pair<HighsInt, HighsInt> arc = entry.key();
+      HighsInt baseBinCol = -1;
+      bool baseForward = false;
+      for (const HighsInt& binCol : entry.value()) {
+        bool forward = false;
+        bool impliesOrder = false;
+        if ((jobOrder[{arc.first, arc.second, binCol}] & 9) == 9) {
+          impliesOrder = true;
+        } else if ((jobOrder[{arc.first, arc.second, binCol}] & 6) == 6) {
```
BenChampion, line 174: Explain what these conditions mean and why we require opposite implications for the two values of the binary variable?

(The comment in `‎highs/mip/HighsMachineSchedSeparator.h` suggests we only need one direction of the implication. Where do things break down if we don't have both directions? Add any additional conditions?)

highs/mip/HighsMachineSchedSeparator.cpp
```diff
@@ -0,0 +1,298 @@
...
+
+  // Extract binary variables that imply the same job ordering
+  if (!mipsolver.mipdata_->parallelLockActive()) {
+    std::vector<HighsCliqueTable::CliqueVar> clique(2);
+    for (const auto& entry : arcToBin) {
+      std::pair<HighsInt, HighsInt> arc = entry.key();
+      HighsInt baseBinCol = -1;
+      bool baseForward = false;
+      for (const HighsInt& binCol : entry.value()) {
+        bool forward = false;
```
BenChampion: Is there a more intuitive name you could give this? If not, is there a comment you could add to make it clear what it's doing?

highs/mip/HighsMachineSchedSeparator.h
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * Given a set of jobs N with start times s_j, processing times p_ij,
+ * release times r_j, and binary dependency variables y_ij,
+ * where y = 1 -> s_i - s_j < p_ij.
+ * To simplify the problem: We set p_j = min{p_ji : i \in N / j}
+ * A valid inequality is then:
+ * s_j >= min{r_{i} : i \in N} + \sum_{i \in N / j} p_i y_ij
```
BenChampion, line 16: Give a brief explanation in words?

highs/mip/HighsMachineSchedSeparator.cpp
```diff
@@ -0,0 +1,298 @@
...
+        mipsolver.mipdata_->cliquetable.addClique(mipsolver, clique.data(), 2);
+        clique[0] = stayCliqueVar.complement();
+        clique[1] = substCliqueVar;
+        mipsolver.mipdata_->cliquetable.addClique(mipsolver, clique.data(), 2);
+        if (globaldom.infeasible()) return false;
+      }
+    }
+  }
+
+  // A clique of size 3 needs at least 6 arcs
```
BenChampion: Where does 3 come from?

highs/mip/HighsMachineSchedSeparator.cpp
```diff
@@ -0,0 +1,298 @@
...
+        HighsCliqueTable::CliqueVar stayCliqueVar =
+            HighsCliqueTable::CliqueVar(baseBinCol, baseForward);
+        HighsCliqueTable::CliqueVar substCliqueVar =
+            HighsCliqueTable::CliqueVar(binCol, forward);
+        clique[0] = stayCliqueVar;
+        clique[1] = substCliqueVar.complement();
+        mipsolver.mipdata_->cliquetable.addClique(mipsolver, clique.data(), 2);
+        clique[0] = stayCliqueVar.complement();
+        clique[1] = substCliqueVar;
+        mipsolver.mipdata_->cliquetable.addClique(mipsolver, clique.data(), 2);
```
BenChampion, line 195: Say what we're doing here?

(It looks to me like we're enforcing `x1 + ~x2 <= 1` and `~x1 + x2 <= 1` which together imply `x1 == x2`)

highs/mip/HighsMachineSchedSeparator.cpp
```diff
@@ -0,0 +1,298 @@
...
+        if (globaldom.infeasible()) return false;
```
BenChampion, line 196: 1. This seems to be the only way we use the clique table?
1. Why don't we need to back out our changes once we are done?

highs/mip/HighsMachineSchedSeparator.cpp
```diff
@@ -0,0 +1,298 @@
...
+      }
+    }
+  }
+
+  // A clique of size 3 needs at least 6 arcs
+  if (numRows <= 5) return false;
+
+  // Greedily search neighbours of largest degree column for a double-sided
+  // clique (corresponds to a single machine schedule)
```
BenChampion, line 205: Say what you mean by a double-sided clique? (You aren't using any clique data structures here).

highs/mip/HighsMachineSchedSeparator.cpp
```diff
@@ -0,0 +1,298 @@
...
+  std::vector<HighsInt> degrees(mipsolver.numCol());
+  const HighsInt maxRows = std::min(HighsInt{50000}, 2 * mipsolver.numRow());
+  HighsHashTable<std::pair<HighsInt, HighsInt>,
+                 std::tuple<double, HighsInt, ArcType>>
+      adjacency(maxRows + 2);
+  HighsHashTable<std::pair<HighsInt, HighsInt>, std::vector<HighsInt>> arcToBin;
+  // 1 -> (i,j) y = 0, 2 -> (i,j) y = 1, 4 -> (j,i) y = 0, 8 -> (j,i) y = 1
+  HighsHashTable<std::tuple<HighsInt, HighsInt, HighsInt>, uint8_t> jobOrder;
+
+  auto addEntry = [&](HighsInt posCol, HighsInt negCol, HighsInt binCol,
```
BenChampion, line 46:
```suggestion
  auto addEntry = [](HighsInt posCol, HighsInt negCol, HighsInt binCol,
```

For the things you modify, either add explicit captures, or use pass-by-reference.

highs/mip/HighsMachineSchedSeparator.cpp
```diff
@@ -0,0 +1,298 @@
...
+          [&](const HighsInt c1, const HighsInt c2) {
+            return degrees[c1] > degrees[c2];
+          });
+
+  std::vector<HighsInt> neighbours;
+  neighbours.reserve(largestDegree + 1);
+  neighbours.emplace_back(largestDegreeCol);
+  double releaseDate = globaldom.col_lower_[largestDegreeCol];
+  std::vector<double> processingTimes;
+  processingTimes.resize(largestDegree + 1, kHighsInf);
```
BenChampion:
```suggestion
  std::vector<double> processingTimes(largestDegree + 1, kHighsInf);
```

highs/mip/HighsMachineSchedSeparator.cpp
```diff
@@ -0,0 +1,298 @@
...
+  // Iterate over potential neighbours and check validity
+  for (HighsInt col : potentialNeighbours) {
+    bool valid_neighbour = true;
+    for (HighsInt neighbour : neighbours) {
+      const auto fromArc = adjacency.find({col, neighbour});
+      const auto toArc = adjacency.find({neighbour, col});
+      // Need to verify that the to and fromArc exist and are opposites
```
BenChampion, line 232: Why can't we do this earlier?

highs/mip/HighsMachineSchedSeparator.cpp
```diff
@@ -0,0 +1,298 @@
...
+      }
+    }
+    if (!valid_neighbour) continue;
+    const size_t newNeighbourIndex = neighbours.size();
+    // Extract the processing times from the arcs
+    for (size_t i = 0; i != neighbours.size(); ++i) {
+      HighsInt neighbour = neighbours[i];
+      const auto fromArc = adjacency.find({col, neighbour});
+      const auto toArc = adjacency.find({neighbour, col});
+      processingTimes[newNeighbourIndex] =
```
BenChampion: Use a temporary variable and then store the result when you insert the new neighbor, so that it's obvious what we're doing with it? Figuring out the role of `newNeighbourIndex` was a major source of backtracking for me.

highs/mip/HighsMachineSchedSeparator.cpp
```diff
@@ -0,0 +1,298 @@
...
+      if (toArc == nullptr || fromArc == nullptr ||
+          std::get<1>(*fromArc) != std::get<1>(*toArc) ||
+          std::get<2>(*fromArc) == std::get<2>(*toArc)) {
+        valid_neighbour = false;
+        break;
+      }
+    }
+    if (!valid_neighbour) continue;
+    const size_t newNeighbourIndex = neighbours.size();
+    // Extract the processing times from the arcs
```
BenChampion, line 242: Add a comment connecting the computations in the body of this loop with the notation you've defined in `‎highs/mip/HighsMachineSchedSeparator.h`? (`p_j` and `p_ij`)

highs/mip/HighsMachineSchedSeparator.cpp
```diff
@@ -0,0 +1,298 @@
...
+          });
+
+  std::vector<HighsInt> neighbours;
+  neighbours.reserve(largestDegree + 1);
+  neighbours.emplace_back(largestDegreeCol);
+  double releaseDate = globaldom.col_lower_[largestDegreeCol];
+  std::vector<double> processingTimes;
+  processingTimes.resize(largestDegree + 1, kHighsInf);
+  // Iterate over potential neighbours and check validity
+  for (HighsInt col : potentialNeighbours) {
```
BenChampion, line 227: Worth explaining in more detail your strategy here? (In particular, you are enlarging `neighbors` with every iteration.) I had to backtrack multiple times to figure out what this loop was doing.

highs/mip/HighsMachineSchedSeparator.cpp
```diff
@@ -0,0 +1,298 @@
...
+      const auto fromArc = adjacency.find({col, neighbour});
+      const auto toArc = adjacency.find({neighbour, col});
+      processingTimes[newNeighbourIndex] =
+          std::min(processingTimes[newNeighbourIndex], std::get<0>(*fromArc));
+      processingTimes[i] = std::min(processingTimes[i], std::get<0>(*toArc));
+    }
+    releaseDate = std::min(globaldom.col_lower_[col], releaseDate);
+    neighbours.emplace_back(col);
+  }
+  if (neighbours.size() < 3) return false;
```
BenChampion, line 255: Where does 3 come from? (Similar to above).

highs/mip/HighsMachineSchedSeparator.cpp
```diff
@@ -0,0 +1,298 @@
...
+      HighsInt neighbour = neighbours[jj];
+      const auto toArc = adjacency.find({neighbour, col});
+      assert(toArc != nullptr);
+      inds[i][j] = std::get<1>(*toArc);
+      vals[i][j] = processingTimes[jj];
+      if (std::get<2>(*toArc) == ArcType::kIfBinZero) {
+        rhss[i] -= vals[i][j];
+        vals[i][j] *= -1;
+      }
+    }
```
BenChampion, line 277:
```suggestion
    for (size_t k = 0; k < 2; ++k) { // skip i itself
      for (size_t j = k*i; j != i + k*(neighbors.size() - 1 - i); ++j) {
        HighsInt neighbour = neighbours[j + k];
        const auto toArc = adjacency.find({neighbour, col});
        assert(toArc != nullptr);
        inds[i][j] = std::get<1>(*toArc);
        vals[i][j] = processingTimes[j + k];
        if (std::get<2>(*toArc) == ArcType::kIfBinZero) {
          rhss[i] -= vals[i][j];
          vals[i][j] *= -1;
        }
      }
    }
```
I think this also means you have an extra element in some of your vectors?

BenChampion (changes requested):
Thanks for asking me to review this change! It was very educational. (The quantity of comments reflects that: I added the comments as I worked through it.)

To summarize my feedback, some parts of this were much harder to understand than others. It might help to delineate as much as possible
* code that identifies the single machine scheduling structure, ensuring we don't try to build the cuts where they won't be helpful
* code that computes the data we need once we've identified the structure
* the complete preconditions for the valid inequalities
* ideas for each step, intents behind data structures, and your algorithmic strategy

In due course, testing results would also be good to see.

Once you've had a chance to digest my feedback I'd be happy to chat. (It would certainly help me to fill any gaps in my understanding.)

highs/mip/HighsMachineSchedSeparator.cpp
Opt-Mucca: Done.

highs/mip/HighsMachineSchedSeparator.cpp
Opt-Mucca: Done.

highs/mip/HighsMachineSchedSeparator.cpp
Opt-Mucca: No preference here. I've changed it.

highs/presolve/HPresolve.cpp
Opt-Mucca: It is in general nice to have variables have a lower bound of zero. I don't see this making any mathematical difference though, and it completely changes the structure. I don't believe it's done in other solvers. The advantage would be that there's less changes needed to be done in stuff like cut transformation as they already have a lb of 0.
Consider that HiGHS has proven that one job has to start at at-least time 5. It will shift this 0. The job variable no longer represents the start time though, it now represents the start time + 5, which is impossible to recover easily. It can no longer be used to make the inequality.

highs/mip/HighsMachineSchedSeparator.h
Opt-Mucca: Explained.

highs/mip/HighsMachineSchedSeparator.h
Opt-Mucca: Changed. Good suggestion

highs/mip/HighsMachineSchedSeparator.h
Opt-Mucca: Put them at the end

highs/mip/HighsMachineSchedSeparator.cpp
Opt-Mucca: Changed here and in some other places.

highs/mip/HighsMachineSchedSeparator.cpp
Opt-Mucca: I wanted to be careful, but I guess it shouldn't happen. Removed it.

highs/mip/HighsMachineSchedSeparator.cpp
Opt-Mucca: Changed it to `y`

highs/mip/HighsMachineSchedSeparator.cpp
Opt-Mucca: - Is this not obvious from the code?
- Changed the naming to your suggestion
- Isn't this the point of the comment above?

highs/mip/HighsMachineSchedSeparator.cpp
Opt-Mucca: Changed. Similar comments to above.

highs/mip/HighsMachineSchedSeparator.cpp
Opt-Mucca: Added an explanation for the data structures.

highs/mip/HighsMachineSchedSeparator.cpp
Opt-Mucca: Changed it to `arcImplicationWhenOne` following the other variable name changes.

highs/mip/HighsMachineSchedSeparator.h
Opt-Mucca: I've given a non-math explanation below it now

highs/mip/HighsMachineSchedSeparator.cpp
Opt-Mucca: I've changed the comment slightly. I assume any clique of size 2 or smaller can be resolved by branching directly.

highs/mip/HighsMachineSchedSeparator.cpp
Opt-Mucca: Exactly what we're doing. We can't change the matrix and the clique table is the way we store this information. Sadly it will not explicitly add the substitutions if we pass a single equality clique. (needs to be extended)

highs/mip/HighsMachineSchedSeparator.cpp
Opt-Mucca: Just because we've replaced some variables doesn't mean that inequalities can't still be derived. The clique table should also handle all subsitutions going forward.

highs/mip/HighsMachineSchedSeparator.cpp
Opt-Mucca: I imagine this as a directed graph, where a `double-sided clique` is two arcs of different direction inbetween every two members (nodes)

highs/mip/HighsMachineSchedSeparator.cpp
Opt-Mucca: Why? It's easier to just add the `&` here instead of passing three more references.

highs/mip/HighsMachineSchedSeparator.cpp
Opt-Mucca: Added!

highs/mip/HighsMachineSchedSeparator.cpp
Opt-Mucca: Done! Very nice suggestion.

highs/mip/HighsMachineSchedSeparator.cpp
Opt-Mucca: Done!

highs/mip/HighsMachineSchedSeparator.cpp
Opt-Mucca: It doesn't make sense for any clique of less than size 3. This should be clear from the changed comment of the previous `.size()` check

highs/mip/HighsMachineSchedSeparator.cpp
Opt-Mucca: I've added a small comment

highs/mip/HighsMachineSchedSeparator.cpp
Opt-Mucca: I'm finding both my original code and your suggestion horrendously confusing here.....

highs/mip/HighsMachineSchedSeparator.cpp
BenChampion: It's much harder to reason about what a function is doing if it can modify any value in scope rather than just three specific things.

highs/mip/HighsMachineSchedSeparator.cpp
BenChampion: Nice! Thank you!

highs/mip/HighsMachineSchedSeparator.cpp
```diff
@@ -0,0 +1,322 @@
...
+          arcImplicationWhenOne = true;
+        }
+        if (!impliesOrder || binCol == baseBinCol) continue;
+        if (baseBinCol == -1) {
+          baseBinCol = binCol;
+          baseArcImplicationWhenOne = arcImplicationWhenOne;
+          continue;
+        }
+        // Add cliques x1 + ~x2 <= 1 and ~x1 + x2 <= 1 which together imply x1
+        // == x2 (depending on signs may also have x1 == ~x2)
```
BenChampion, line 185: This last addition is a little confusing but I think I know what you mean (that the original variables may already be complemented with respect to each other)

highs/mip/HighsMachineSchedSeparator.cpp
BenChampion: Agree, the improvement is marginal (but it's there, because you can see up front which values it's going to traverse).

There's probably a better solution out there. For now the only improvement I can see would be to make the comment less opaque. (Like: "Skip i itself: process the neighbors before and after as two contiguous blocks")

I spent way too long thinking about this!

highs/mip/HighsMachineSchedSeparator.cpp
```diff
@@ -0,0 +1,322 @@
...
+        machineSchedRow = false;
+        break;
+      }
+    }
+    if (!machineSchedRow || binCol == -1 || negContCol == -1 ||
+        posContCol == -1)
+      continue;
+    // We want to put the row into form:
+    // My_ji + s_i - s_j >= p_ji, p_ji >= 0
+    // y_ji = 1 -> s_i >= s_j + p_jj - M
```
BenChampion:
```suggestion
    // My_ji + s_i - s_j >= p_ji, p_ji >= 0
    // y_ji = 1 -> s_i >= s_j + p_ji - M
```
Also, don't you want `M(1-y_ji)`?

highs/mip/HighsMachineSchedSeparator.cpp
BenChampion: I think I understand this now. The implications go both ways, but one is a big-M "implication".

Follow-up question: what ensures the big M is valid? Or do we just have to assume the user has done this correctly?

BenChampion (approved):
Thanks for addressing my comments. This looks good to me now modulo some new comments/continuing discussions, and proper testing.

That just leaves the clique size at least three thing: why can't we form the inequality for two jobs? Is it just not useful? Or is it something else? (EDIT: this is only for my understanding)

highs/mip/HighsMachineSchedSeparator.cpp
Opt-Mucca: Changed everything to nicer to follow `emplace_back` as discussed!

highs/mip/HighsMachineSchedSeparator.cpp
Opt-Mucca: It's just re-arranging and setting `y_ji = 1` so that it becomes `M`. Edit: Fixed the `p_jj` error

highs/mip/HighsMachineSchedSeparator.cpp
BenChampion: As discussed on our call, and just as a note for future reference, it's possible to do for cliques of size 2, but this case already seems easy to handle with branching.

highs/mip/HighsMachineSchedSeparator.cpp
BenChampion: As discussed on our call, these are in fact true implications both ways. A single machine schedule has the feature that either `i` depends on `j` or `j` depends on `i`, so `y_ij + y_ji == 1` (even if `p_ij != p_ji`)

highs/mip/HighsMachineSchedSeparator.cpp
```diff
@@ -0,0 +1,323 @@
...
+        }
+        // Add cliques x1 + ~x2 <= 1 and ~x1 + x2 <= 1 which together imply x1
+        // == x2 (depending on signs may also have x1 == ~x2)
+        HighsCliqueTable::CliqueVar stayCliqueVar =
+            HighsCliqueTable::CliqueVar(baseBinCol, baseArcImplicationWhenOne);
+        HighsCliqueTable::CliqueVar substCliqueVar =
+            HighsCliqueTable::CliqueVar(binCol, arcImplicationWhenOne);
+        clique[0] = stayCliqueVar;
+        clique[1] = substCliqueVar.complement();
+        mipsolver.mipdata_->cliquetable.addClique(mipsolver, clique.data(), 2);
```
fwesselm, line 192: Maybe use the following instead to pass an equality clique?

`mipsolver.mipdata_->cliquetable.addClique(mipsolver, clique.data(), 2, true);`

highs/mip/HighsMachineSchedSeparator.cpp
```diff
@@ -0,0 +1,323 @@
...
+    }
+    // Store binaries to later check if any two binaries
+    // imply the same ordering of two jobs.
+    HighsInt u = std::min(negCol, posCol);
+    HighsInt v = std::max(negCol, posCol);
+    if (jobOrder.find({u, v, binCol}) == nullptr) {
+      jobOrder[{u, v, binCol}] = 0;
+    }
+    if (negCol < posCol) {
+      arcToBin[{negCol, posCol}].emplace_back(binCol);
```
fwesselm, line 75: Is it OK to add `binCol` twice for ranged rows? Redundant calls to `addClique` are OK, just wanted to ask.

highs/mip/HighsMachineSchedSeparator.cpp
```diff
@@ -0,0 +1,323 @@
...
+        break;
+      }
+      if (globaldom.isBinary(col)) {
+        if (binCol != -1) {
+          machineSchedRow = false;
+          break;
+        }
+        binCol = col;
+        binCoef = mipsolver.mipdata_->ARvalue_[i];
+      } else if (mipsolver.mipdata_->ARvalue_[i] == -1) {
```
fwesselm, line 122: I think this is consistent with the clique code, but this may miss cases where the coefficient is very close to +-1.

highs/mip/HighsMachineSchedSeparator.cpp
```diff
@@ -0,0 +1,323 @@
...
+        break;
+      }
+    }
+    if (!machineSchedRow || binCol == -1 || negContCol == -1 ||
+        posContCol == -1)
+      continue;
+    // We want to put the row into form:
+    // My_ji + s_i - s_j >= p_ji, p_ji >= 0
+    // y_ji = 1 -> s_i >= s_j + p_ji - M
+    if (rowUpper != kHighsInf) {
```
fwesselm, line 145: Please ignore this (nit-picking!) if you do not like it, but this could be a lambda:

```
auto tryAddArc = [&](HighsInt posCol, HighsInt negCol, double rhs_0,
                       double rhs_1) {
    if (rhs_0 > 0 || rhs_1 > 0) {
      if (rhs_0 > rhs_1) {
        addEntry(posCol, negCol, binCol, rhs_0,
                 ArcType::kImplicationWhenZero);
      } else {
        addEntry(posCol, negCol, binCol, rhs_1,
                 ArcType::kImplicationWhenOne);
      }
      ++numRows;
    }
  };

 if (rowUpper != kHighsInf)
    tryAddArc(negContCol, posContCol, -rowUpper, -rowUpper + binCoef);
  if (rowLower != -kHighsInf)
    tryAddArc(posContCol, negContCol, rowLower, rowLower - binCoef);
```

highs/mip/MipTimer.h
```diff
@@ -326,6 +327,8 @@ class MipTimer {
         timer_pointer->clock_def(kPathAggrSepaString.c_str());
     clock[kMipClockModKSepa] =
         timer_pointer->clock_def(kModKSepaString.c_str());
+    clock[kMipClockMachineSchedSepa] =
```
fwesselm, line 330: This is commented out (like the other clocks), is this OK?

fwesselm (approved):
Looks good, @Opt-Mucca!

highs/mip/HighsMachineSchedSeparator.cpp
Opt-Mucca: Passing the equality clique doesn't create the substitutions. This is something on my every-growing TODO list for when I more tightly interface the clique table into presolve.

highs/mip/HighsMachineSchedSeparator.cpp
Opt-Mucca: Yup. I couldn't find a more efficient way to filter them..... It's ugly but should be correct.

highs/mip/HighsMachineSchedSeparator.cpp
Opt-Mucca: I can live with missing those cases. I imagine handling those cases risks introducing tolerance errors too.

highs/mip/HighsMachineSchedSeparator.cpp
Opt-Mucca: Not nitpicky. That's a great suggestion. Will add it.

highs/mip/MipTimer.h
Opt-Mucca: I think so. I can't exactly remember whether or not the clocks are currently working for separators.

Opt-Mucca:
This improves performance by 1% or so on the test set of @fwesselm I'm now going to merge it

PR #2994, opened 2026-04-29
HiPO in Python

galabovaa:
@mathgeekcoder Please let me know if this looks OK to you too!


highs/lp_data/HighsOptions.cpp
```diff
@@ -80,107 +79,121 @@ bool optionOffOnOk(const HighsLogOptions& report_log_options,
...
+          report_log_options, HighsLogType::kError,
+          "The HiPO solver was requested via the \"%s\" option, but this build "
+          "was compiled without HiPO support. Reconfigure with FAST_BUILD=ON "
+          "and -DHIPO=ON to enable HiPO.\n",
+          kSolverString.c_str());
+    } else {
+      highsLogUser(
+          report_log_options, HighsLogType::kError,
+          "The HiPO solver was requested via the \"%s\" option, but the "
+          "HiPO dependencies are missing. Install with highspy[hipo] or "
```
filikat: Can this message be reached only when using Python? 
Otherwise, it shouldn't mention (only) Python instructions.

highspy-extras/README.md
```diff
@@ -0,0 +1,59 @@
+# highspy-extras
+
+Extension package for [highspy](https://pypi.org/project/highspy/) that enables access to external dependencies with licensing terms different from HiGHS, such as Apache 2.0.
+
+The HiPO Interior Point Method (IPM) solver currently uses these external dependencies to provide enhanced performance for linear programming problems. Other algorithms may also rely on `highspy-extras` in the future.
```
filikat: "to provide enhanced performance for linear and quadratic programming problems."

highs/HighsExternalDeps.cpp
```diff
@@ -0,0 +1,190 @@
...
+        inst.status_ = "Extras: Failed to resolve required external functions";
+        inst.unload();
+      } else {
+        std::string highs_version = STRINGFY(HIGHS_VERSION_MAJOR) "." STRINGFY(
+            HIGHS_VERSION_MINOR) "." STRINGFY(HIGHS_VERSION_PATCH);
+        std::string extras_version = get_version();
+        if (extras_version != highs_version) {
+          inst.status_ = "Extras: ABI version mismatch: expected " +
+                         highs_version + ", got " + extras_version +
+                         ". Please reinstall: pip install --force-reinstall "
```
filikat: Same thing, can this be reached only with Python?

filikat:
On Mac, the header reports ``Using BLAS: unknown``, unless I specify manually -DBLAS_LIBRARIES=... when configuring CMake. 
It appears that if it finds Apple BLAS automatically, BLAS_LIBRARIES is not defined in HConfig.h, even though the CMake log correctly prints ``Using BLAS library: ...``

galabovaa:
Thank you @filikat I will address your comments

highs/lp_data/HighsOptions.cpp
mathgeekcoder: Good call.  BTW: you are correct, depending on the build it can be reached by any API (not just highspy).  

If it's not availableAtCompile, users will need the highs_extras library available at runtime.  The highspy-extras package simply includes the highs_extras library.

filikat:
I benchmarked this branch on my set of 330 LPs and QPs and there's no performance difference.

filikat:
If I try to compile with ``ALL_TESTS=ON`` and ``BUILD_SHARED_EXTRAS_LIB=OFF``, I get the error 
```
CMake Error at check/CMakeLists.txt:123 (target_link_libraries):
  The keyword signature for target_link_libraries has already been used with
  the target "unit_tests".  All uses of target_link_libraries with a target
  must be either all-keyword or all-plain.

  The uses of the keyword signature are here:

   * check/CMakeLists.txt:114 (target_link_libraries)
```

filikat:
Everything else looks fine to me.
Great job getting this to work!

mathgeekcoder:
> If I try to compile with `ALL_TESTS=ON` and `BUILD_SHARED_EXTRAS_LIB=OFF`, I get the error

Fixed.  Good catch, thanks!

FYI: Was an issue with the new-style `PRIVATE` syntax for `target_link_libraries`.  The other linker calls were "old-style", and   apparently cmake doesn't like to mix the styles.

Also, `BUILD_SHARED_EXTRAS_LIB=OFF` should be a no-op when `HIPO=OFF`.  This was causing a link error, but I've fixed this too.

mathgeekcoder:
BTW: ignore the "Co-authored-by: Copilot <copilot@github.com>" in my commit message.  It wasn't a big change, but I wrote it all myself.  It's trying to claim co-authorship, even when it just autocompletes a word or two 😆 

Apparently, this is a recent "[feature](https://github.com/orgs/community/discussions/194075)" of vscode.  Thankfully, there's a `"git.addAICoAuthor": "off"` setting.

highs/HighsExternalDeps.cpp
mathgeekcoder: Fixed!

highs/lp_data/HighsOptions.cpp
mathgeekcoder: I've updated the error messages and cleaned up the code by removing the redundancy.  Please @filikat and @galabovaa check if you're happy with the change.  Thanks!

filikat: Looks fine to me

galabovaa: @mathgeekcoder Looks good to me as well!

mathgeekcoder:
BTW: @galabovaa, I noticed and fixed a potential big issue for C++ users of highs.  We weren't including `HIGHS_SHARED_EXTRAS_LIBRARY` define in `HConfig.h`.  We didn't see this in our build system since we always define it appropriately via cmake.

While I fixing this, I also made it much easier to consume/develop.  The appropriate header/linker details are now propagated via libhighs using PUBLIC cmake declarations.

I also made it unnecessary to use #ifdef HIPO everywhere, we can simply rely on `HighsExternalDeps::isAvailable()` instead. Any half-decent compiler will optimize out the unused code if HIPO=OFF.

Most of these changes are here: https://github.com/ERGO-Code/HiGHS/pull/2994/changes/a86d16a9e211e73040d630b7389d7fc242179dd2

I've also updated the bazel and meson builds - but as you already know, they assume HIPO=OFF.  Is this something we need to support in this PR?

highspy-extras/README.md
mathgeekcoder: Fixed!

mathgeekcoder:
> On Mac, the header reports `Using BLAS: unknown`, unless I specify manually -DBLAS_LIBRARIES=... when configuring CMake. It appears that if it finds Apple BLAS automatically, BLAS_LIBRARIES is not defined in HConfig.h, even though the CMake log correctly prints `Using BLAS library: ...`

@filikat: I changed this to return the BLA_VENDOR (if exists).  So, for Apple BLAS, it'll simply say "Apple".  Is this suitable?

galabovaa:
> BTW: @galabovaa, I noticed and fixed a potential big issue for C++ users of highs. We weren't including `HIGHS_SHARED_EXTRAS_LIBRARY` define in `HConfig.h`. We didn't see this in our build system since we always define it appropriately via cmake.
> 
> While I fixing this, I also made it much easier to consume/develop. The appropriate header/linker details are now propagated via libhighs using PUBLIC cmake declarations.
> 
> I also made it unnecessary to use #ifdef HIPO everywhere, we can simply rely on `HighsExternalDeps::isAvailable()` instead. Any half-decent compiler will optimize out the unused code if HIPO=OFF.

Great, thank you for fixing these! 

> I've also updated the bazel and meson builds - but as you already know, they assume HIPO=OFF. Is this something we need to support in this PR?

We can keep HiPO off in bazel and meson for now.

filikat:
> I changed this to return the BLA_VENDOR (if exists). So, for Apple BLAS, it'll simply say "Apple". Is this suitable?

Great thanks.

galabovaa:
I think all the comments are addressed now. Will update documentation separately!

jajhall (commented):
I can't comment on the build changes, but I find the use of `HighsExternalDeps::isAvailable();` to replace `#ifdef HIPO` rather opaque. I agree that it's more elegant than an `#ifdef ... #endif` structure, but it doesn't tell me what external dependency might be available. Further, what do we do as/when we have a second external dependency that we want to handle in the same way?

Also, some of the methods in `HighsExternalDeps`, such as  `openblas_set_num_threads` don't conform to the HiGHS naming convention for method. Perhaps they mimic the name used in external code, and `Highs_genrcm` is as it is because the external method is `genrcm`, so `HighsGenrcm` isinconsistent. If all these deviations can be justified by external naming, then that's OK.

jajhall (commented):
Otherwise, of course, this represents great progress!

mathgeekcoder:
> I can't comment on the build changes, but I find the use of `HighsExternalDeps::isAvailable();` to replace `#ifdef HIPO` rather opaque. I agree that it's more elegant than an `#ifdef ... #endif` structure, but it doesn't tell me what external dependency might be available. Further, what do we do as/when we have a second external dependency that we want to handle in the same way?

Thanks @jajhall, this is a great point to discuss stylistic choices and I'd love your opinion.  I'm currently not 100% happy with it, but I felt like it was a reasonable compromise.  See below for a potentially better idea.

- `HighsExternalDeps` currently has 4 dependencies (`blas`, `metis`, `amd`, and `rcm`), so adding additional dependencies should be straightforward.
- Our assumption is that we split the HiGHS codebase and `HighsExternalDeps` based on licensing (MIT vs Apache). So perhaps it's more accurate to ask: _"what if we add another external dependency with a different license?"_. 
  - The simple approach is to simply update the license for `HighsExternalDeps` to include whatever is necessary.  However, this might be too restrictive in the future.
  - We could separate based on license (e.g., `HighsApacheDeps`, `HighsBsdDeps`, etc.), although this becomes unwieldy.
  - Note: `rcm` is MIT licensed and is always included in the HiGHS build, even though it's accessed via `HighsExternalDeps`.  I don't particularly like this, but it seemed more consistent.
 
 
- I agree that `isAvailable` is opaque, however `#ifdef HIPO` is too specific.  Future (non-HiPO) algorithms may want to use these external dependencies too - and the code could get quite complicated with many build flags.

**Perhaps a better approach is more explict?** e.g., 
```cpp
using hipo = HighsExternalDeps::requires<
    HighsExternalDeps::blas,
    HighsExternalDeps::amd,
    HighsExternalDeps::metis,
    HighsExternalDeps::rcm>;

HighsExternalDeps::isAvailable<hipo>();
HighsExternalDeps::logUnavailable<hipo>(...);

// also supported
HighsExternalDeps::isAvailable<HighsExternalDeps::blas>();
HighsExternalDeps::isAvailable<HighsExternalDeps::blas, HighsExternalDeps::metis>();
``` 
We could also wrap this in a struct for better namespacing:
```cpp
struct HighsExternalRequirement {
  using hipo = HighsExternalDeps::requires<
    HighsExternalDeps::blas,
    HighsExternalDeps::amd,
    HighsExternalDeps::metis,
    HighsExternalDeps::rcm>;

  using new_feature = HighsExternalDeps::requires<...>;
};

HighsExternalDeps::isAvailable<HighsExternalRequirement::hipo>();
HighsExternalDeps::logUnavailable<HighsExternalRequirement::hipo>(...);
```

**Does this look better to you?  Do you have a preference?**  Any opinions @galabovaa, @filikat?

I don't think this would be too difficult to implement.
 
> Also, some of the methods in `HighsExternalDeps`, such as `openblas_set_num_threads` don't conform to the HiGHS naming convention for method. Perhaps they mimic the name used in external code, and `Highs_genrcm` is as it is because the external method is `genrcm`, so `HighsGenrcm` isinconsistent. If all these deviations can be justified by external naming, then that's OK.

Yeah, they're all based on external naming.  It's somewhat distasteful but probably the best option.

jajhall:
> * `HighsExternalDeps` currently has 4 dependencies (`blas`, `metis`, `amd`, and `rcm`), so adding additional dependencies should be straightforward.

My thought was rather more, "What if some external dependencies are available, but not others?" I'd rather have `HighsExternalDeps:hipoIsAvailable()` then, to me, it's a lot clearer what this evaluating to true/false actually means.

> * Our assumption is that we split the HiGHS codebase and `HighsExternalDeps` based on licensing (MIT vs Apache). So perhaps it's more accurate to ask: _"what if we add another external dependency with a different license?"_.

That's a sensible partition. I don't want to think about adding something with a different license!

That said, in [`THIRD_PARTY_NOTICES.md`](https://github.com/ERGO-Code/HiGHS/blob/master/THIRD_PARTY_NOTICES.md) not everything in the `/extern` directory is Apache? Is Highs not always built with `zstr`, and `pdqsort`?

Maybe the contents of the `/extern` directory should only contain the Apache stuff or, more precisely, perhaps we need `extern/MIP` and `extern/Apache` directories?

filikat:
@mathgeekcoder I think what you suggest is good. Having a way of loading and checking availability for each dependency independently would be ideal, also thinking at potential use cases in the future. We could then have some shortcuts to check all dependencies of hipo in one go.

jajhall:
> Perhaps a better approach is more explicit

I don't understand all the C++, but @filikat does, and his interpretation sounds good to me

It also nullifies my comments about whether `HighsExternalDeps::isAvailable();` refers to everything in the `/extern` directory

mathgeekcoder:
> My thought was rather more, "What if some external dependencies are available, but not others?" I'd rather have `HighsExternalDeps:hipoIsAvailable()` then, to me, it's a lot clearer what this evaluating to true/false actually means.

Good to know. `HighsExternalDeps` is currently all-or-nothing.  It starts getting complex if we allow partial support, unless we have some sort of declarative dependency registration (i.e., similar to my above suggestion).

If we go down the `hipoIsAvailable()` path, the code might get a little ugly: `HighsExternalDeps:feature1IsAvailable()`, `HighsExternalDeps:feature2IsAvailable()`, etc.  Plus all the necessary overloads for error messages when things are missing.  

Using C++ templates makes the code more concise - but I'm not sure if we need to expose this directly to the C APIs?

> That's a sensible partition. I don't want to think about adding something with a different license!

That's fair! Though **at some point soon we might need to handle CUDA?**

> That said, in [`THIRD_PARTY_NOTICES.md`](https://github.com/ERGO-Code/HiGHS/blob/master/THIRD_PARTY_NOTICES.md) not everything in the `/extern` directory is Apache? Is Highs not always built with `zstr`, and `pdqsort`?
> 
> Maybe the contents of the `/extern` directory should only contain the Apache stuff or, more precisely, perhaps we need `extern/MIP` and `extern/Apache` directories?

Good point. My personal preference would be to include everything at compile-time and simply document the features that require additional licensing - but I don't think that's an option?

In terms of licensing, I believe: `zlib < MIT < BSD-3 < Apache 2.0 << CUDA`.  Where `highs_extras` uses the `Apache 2.0` license.  If we separate files based on license, we'd probably want `BSD` and `zlib` to be consistent - alternatively, we can just simply attribute the license in each class. e.g., 

```cpp
enum class HighsLicense {
    kMit,
    kBsd3Clause,
    kApache2_0,
    kZlib,
    ...
  };

template <>
struct HighsExternalDeps::feature_traits<HighsExternalDeps::metis> {
  static const char* name() { return "metis"; }
  static HighsLicense license() { return HighsLicense::kApache2_0; }
  static const char* notice() { return "metis/LICENSE.txt"; }
};


HighsExternalDeps::getLicenseSummary<hipo>()

HiPO dependencies:
- amd: BSD-3-Clause
- metis: Apache-2.0
- rcm: MIT
- blas: OpenBLAS (BSD-3-Clause)
See THIRD_PARTY_NOTICES.md for full notices.
```

galabovaa:
> Good point. My personal preference would be to include everything at compile-time and simply document the features that require additional licensing - but I don't think that's an option?

I don't think it is an option, at least I read that if you distribute compiled code it changes the license even if it is unused at runtime. 

I like the look of the more explicit approach:
```
using hipo = HighsExternalDeps::requires<
    HighsExternalDeps::blas,
    HighsExternalDeps::amd,
    HighsExternalDeps::metis,
    HighsExternalDeps::rcm>;

HighsExternalDeps::isAvailable<hipo>();
HighsExternalDeps::logUnavailable<hipo>(...);

// also supported
HighsExternalDeps::isAvailable<HighsExternalDeps::blas>();
HighsExternalDeps::isAvailable<HighsExternalDeps::blas, HighsExternalDeps::metis>();
```

Adding an attribute with the license in each class would makes it clear and keep it simpler.

> Using C++ templates makes the code more concise - but I'm not sure if we need to expose this directly to the C APIs?

I am not sure how much of this we will need in the HiGHS C API. If we need to add a method that checks whether hipo is available from the C Api I guess we can add a method to the C++ class which "hides" the templates from C? Is this what you meant?

mathgeekcoder:
> I don't think it is an option, at least I read that if you distribute compiled code it changes the license even if it is unused at runtime.

Thanks! Yeah, I thought as much but wanted to confirm again for the PR.

> I am not sure how much of this we will need in the HiGHS C API. If we need to add a method that checks whether hipo is available from the C Api I guess we can add a method to the C++ class which "hides" the templates from C? Is this what you meant?

Agreed, I don't think the C api really needs to know about it.  And we can always wrap something if needed.

I'm halfway through implementing something.  I hope to push something tomorrow (or day after) and see what everyone thinks.  I think this approach can also be used for optional CUDA (with versioning) support too.

galabovaa:
Great, thank you @mathgeekcoder ! Since this is already a large update, and functionally OK, I am tempted to merge this now and have your next updates in another PR when you are happy with them!

PR #3004, opened 2026-05-06
Minor fixes for HighsCliqueTable

fwesselm:
@Opt-Mucca, I think there is an index error in `HighsCliqueTable::extractCliques`. Maybe I am just confused again...

### DIFF
--- a/highs/mip/HighsCliqueTable.cpp
+++ b/highs/mip/HighsCliqueTable.cpp
@@ -909,7 +909,7 @@ void HighsCliqueTable::extractCliques(
 
   // check if this is a set packing constraint (or easily transformable
   // into one)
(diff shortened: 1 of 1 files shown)


Opt-Mucca (approved):
That seems to be a pretty clear index error. Very nice spot! It's even an outright error potentially?
Double nice spot on the `minEntries`.

I've been looking at these functions for the last few days and missed it completely....

fwesselm:
> That seems to be a pretty clear index error. Very nice spot! It's even an outright error potentially? Double nice spot on the `minEntries`.
> 
> I've been looking at these functions for the last few days and missed it completely....

I agree, `vals[0]` will be some element (not the first one), thus the conclusion (set packing constraint or not) may be incorrect.

PR #3005, opened 2026-05-06
Fix computation of fractional value for variable with finite lower and upper bounds

fwesselm:
@Opt-Mucca, it looks like an `else` is missing here. What do you think?


Opt-Mucca (approved):
Agree! Nice find.

PR #3009, opened 2026-05-07
Use std::move in OpenNode constructor

fwesselm:
Vectors `domchgstack` and `branchings` are rvalue references. I think the intention was to move them, but they are currently copied (which may be less efficient).

### DIFF
--- a/highs/mip/HighsNodeQueue.cpp
+++ b/highs/mip/HighsNodeQueue.cpp
@@ -57,9 +57,10 @@ class HighsNodeQueue::NodeLowerRbTree : public CacheMinRbTree<NodeLowerRbTree> {
     return nodeQueue->nodes[node].lowerLinks;
   }
   std::tuple<double, HighsInt, double, int64_t> getKey(int64_t node) const {
-    return std::make_tuple(nodeQueue->nodes[node].lower_bound,
-                           HighsInt(nodeQueue->nodes[node].domchgstack.size()),
-                           nodeQueue->nodes[node].estimate, node);
+    return std::make_tuple(
+        nodeQueue->nodes[node].lower_bound,
+        static_cast<HighsInt>(nodeQueue->nodes[node].domchgstack.size()),
+        nodeQueue->nodes[node].estimate, node);
   }
 };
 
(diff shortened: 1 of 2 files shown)


highs/mip/HighsNodeQueue.cpp
```diff
@@ -56,7 +56,7 @@ class HighsNodeQueue::NodeLowerRbTree : public CacheMinRbTree<NodeLowerRbTree> {
   const RbTreeLinks<int64_t>& getRbTreeLinks(int64_t node) const {
     return nodeQueue->nodes[node].lowerLinks;
   }
-  std::tuple<double, HighsInt, double, int64_t> getKey(HighsInt node) const {
+  std::tuple<double, HighsInt, double, int64_t> getKey(int64_t node) const {
```
Opt-Mucca, line 59: Why did this need to change?

highs/mip/HighsNodeQueue.h
```diff
@@ -162,8 +162,8 @@ class HighsNodeQueue {
     OpenNode(std::vector<HighsDomainChange>&& domchgstack,
              std::vector<HighsInt>&& branchings, double lower_bound,
              double estimate, HighsInt depth)
-        : domchgstack(domchgstack),
-          branchings(branchings),
+        : domchgstack(std::move(domchgstack)),
```
Opt-Mucca, line 165: I thought this was covered by always using `std::move` when passing the arguments. My IDE is even highlighting that this is copied though. I've still got a fair bit of C++ to learn.....
This should have a small but universal performance bump (Potentially even quite large on some problems)!

Opt-Mucca (approved):

highs/mip/HighsNodeQueue.h
filikat: Have a look here @Opt-Mucca 

https://stackoverflow.com/questions/58761148/reason-to-use-stdmove-on-rvalue-reference-parameter

highs/mip/HighsNodeQueue.cpp
fwesselm: @Opt-Mucca This is a type mismatch; the method returns an `int64_t`. Also, the caller passes a `LinkType` (which is `int64_t` as far as I can see).

Opt-Mucca: I am apparently blind.

highs/mip/HighsNodeQueue.h
Opt-Mucca: @filikat Thank you! I think I've actually got it.

PR #3010, opened 2026-05-08
Keep strengthened variable bounds

fwesselm:
@Opt-Mucca, I realized that, if an existing VLB / VUB is updated, the code does not use the strengthened version. This may not improve much, but anyway...


Opt-Mucca (approved):
I don't see how this can do anything but help! Quite funny that we missed this completely when adding the strengthening.

fwesselm:
> I don't see how this can do anything but help! Quite funny that we missed this completely when adding the strengthening.

My fault, I am blind.

PR #3012, opened 2026-05-08
Fixed errors exposed by AI agent that do not lead to incorrect execution

jajhall:
The first is a missing "%" in a logging statement that's never executed. I guess that the use of `HIGHSINT_FORMAT` leads to the compiler failing to spot it.

The second is, an error, but the two parameter values are associated with a development check that's not executed. In development, the check was used in the context of the serial dual simplex solver, so this error for the parallel dual simplex solver was never exposed.


fwesselm (approved):
Looks good @jajhall! Thanks.

jajhall:
> Looks good @jajhall! Thanks.

Thanks: I'll look at the other one you flagged up before merging

jajhall:
Error in parallel dual simplex Devex weight update fixed

PR #3013, opened 2026-05-08
Improvements to free variables in HiPO

filikat:
Free variables now receive an artificial upper and lower bound. If the variable gets too close to the bound during the IPM iterations, the bound is moved. This improve robustness on some problems. I added the instance ``perold`` from Netlib, which has a large number of free variables, and a unit test.

### DIFF
--- a/check/TestHipo.cpp
+++ b/check/TestHipo.cpp
@@ -112,11 +112,19 @@ TEST_CASE("test-hipo-qp", "[highs_hipo]") {
   highs.resetGlobalScheduler(true);
 }
 
-TEST_CASE("test-hipo-infeas", "[highs)hipo]") {
+TEST_CASE("test-hipo-infeas", "[highs_hipo]") {
   const HighsModelStatus expected_status = HighsModelStatus::kInfeasible;
   Highs highs;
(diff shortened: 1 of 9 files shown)


jajhall (approved):
Fine by me. Since free variables are unusual, there's an argument to have a list of free variables, allowing treatment of free variables to be skipped if there are none, and performed more efficiently when there are some. However, this is more a matter of principle, ans the overhead of handling free variables as implemented is minimal.

filikat:
Yes, you're right, the same could be achieved using a list. I will make the change Monday.

odow:
> Since free variables are unusual

I'm not sure this is true. They may seem unusual from a theoretical perspective, but they appear quite often in models when people just forget to put bounds, or enforce bounds via other linear constraints. I guess you hope that, after presolve, free variables are unusual?

Somewhat related: last week we found a bug in Gurobi where the problem was incorrectly returned as unbounded (instead of either infeasible or optimal, depending on tolerances) because when JuMP adds a binary variable it didn't also set `[0, 1]` bounds.

filikat:
It's okay. Using a list of free variables should be better in any case.

PR #3014, opened 2026-05-08
Fast triangular solves in HiPO

filikat:
Triangular solves (forward and backward) with small supernodes are now performed without calling BLAS functions, since the cost of extra BLAS function calls is not worth it. This brings a reduction in the time taken to perform solves for most problems, and a very large reduction for sparse problems (like PyPSA).


jajhall (approved):
I can't vouch for this being correct, but I know you have tested it well - and I'm sure errors would immediately show

PR #3016, opened 2026-05-11
Added logging and failure return when Highs::presolve() is called and…

jajhall:
Spotted - thanks to @gghisa that if a QP has been passed to HiGHS, and then `Highs::presolve()` is called, the corresponding LP is presolved.

This prints a logging line and returns `HighsStatus::kError`. Tested in `test-qjh`


jajhall:
Maybe my `clang-format` is out of date?

filikat (approved):
Looks good to me.

> Maybe my `clang-format` is out of date?

The formatting looks correct to me. I think some files in the hipo-in-python PR were not formatted correctly.

PR #3017, opened 2026-05-11
Collection of several minor changes

fwesselm:
- Made 4 vectors members of `HighsCutGeneration` to avoid allocating memory each time a cut is separated. 
- Fixed some typos.
- Several other minor fixes.
- Ran 850+ MIPs; behavior is not affected.

### DIFF
--- a/highs/mip/HighsCutGeneration.cpp
+++ b/highs/mip/HighsCutGeneration.cpp
@@ -128,8 +128,6 @@ bool HighsCutGeneration::determineCover(bool lpSol) {
 }
 
 void HighsCutGeneration::separateLiftedKnapsackCover() {
-  const double feastol = lpRelaxation.getMipSolver().mipdata_->feastol;
-
   const HighsInt coversize = cover.size();
 
   std::vector<double> S;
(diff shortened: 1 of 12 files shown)


highs/mip/HighsCutGeneration.cpp
```diff
@@ -958,15 +956,6 @@ bool HighsCutGeneration::preprocessBaseInequality(bool& hasUnboundedInts,
...
       hasContinuous = true;
-      // if (lpRelaxation.isColIntegral(inds[i]))
-      //   printf("vals[i] = %g  upper[i] = %g\n", vals[i], upper[i]);
-
-      if (vals[i] > 0) {
-        if (upper[i] == kHighsInf)
-          maxact = kHighsInf;
-        else
-          maxact += vals[i] * upper[i];
-      }
```
Opt-Mucca, line 969: Nice spot on this being redundant!

Opt-Mucca (approved):
Happy with the changes!

highs/mip/HighsCutGeneration.cpp
jajhall: ... in that maxact is then never used?

jajhall (approved):
All OK to my eyes

highs/mip/HighsCutGeneration.cpp
Opt-Mucca: `maxact` is still updated correctly below. It was just that the `if` statment of `val > 0` here could never actually trigger given the other `if` statements above it.

Issue #3015, opened 2026-05-09
MIP claims optimality with row infeasibility = semi-continuous lower bound (1.12.0)


molnarg:
## Summary

When a MIP contains semi-continuous variables (`integrality=2`) and the LP
relaxation places one or more of them at their lower bound `lo` exactly,
HiGHS returns a "feasible / Optimal" solution that its own internal
feasibility check passes (`row viol. 0`), while a separate post-validation
rejects it with `primal(1/X/X) infeasibilities` where `X` equals the SC
variable's `lo` bound, exactly.

## Smoking gun

Two adjacent lines in the same `disp=True` report disagree:

```
Solution status   feasible
                  -1407973.68089 (objective)
                  0 (bound viol.)
                  1.50000000758e-07 (int. viol.)
                  0 (row viol.)              ← solver: feasible
ERROR:   MIP solver claims optimality, but with num/max/sum
         primal(1/63.125/63.125) infeasibilities      ← post-check: 1 row off by 63.125
ERROR:   Setting model status to Solve error
```

`63.125` is the `lo` of one specific SC variable in the model, to all
significant digits. Across runs of the same model with slightly different
inputs, the reported infeasibility is always one of `{59.412, 63.125, 74.001}`,
each matching one specific SC `lo` exactly.

## Reproducer

Self-contained single-file repro (LP data embedded as base64 NPZ inline):

https://gist.github.com/molnarg/c5cc72c2f951791737cd5834eac959e4

```
python highs_sc_bug_repro.py
```

The script:

1. Solves the original SC formulation → `(HiGHS Status 4: Solve error)`.
2. Same model with each SC var (`integrality=2`, `min=lo, max=hi`) rewritten as
   a continuous var `q ∈ [0, hi]` + binary indicator `b ∈ {0,1}` + two coupling
   rows `q − hi·b ≤ 0` and `lo·b − q ≤ 0` → returns optimal cleanly
   (`objective = -1407973.680891`).

Both formulations encode the same `{0} ∪ [lo, hi]` feasible region. Only the
SC variant triggers the bug. The reduced LP has 866 rows × 696 cols, 78 SC
variables.

## Environment

- HiGHS 1.12.0 (git hash: 4f96ee8) bundled with scipy 1.17.1
- Python 3.12, Linux
- Default options + `mip_rel_gap=1e-7`. Reproduces with `presolve=off` too.

## Related but distinct existing issues

[...]

PR #3018, opened 2026-05-12
Correct handling of semi-variable infeasibilities

jajhall:
When reformulating semi-variables `[l, u]` with a binary, if its optimal value is not exactly 0 or 1, then infeasibilities in the semi-continuous variable may exceed the primal feasibility tolerance for `[l, u]`, or the MIP feasibility tolerance at 0. This is now correctly identified and logged specifically, suggesting the use of a smaller MIP feasibility tolerance

The instance `3015.mps` also highlighted the issue of postsolving parallel free columns, so is added as a unit test for both issues.


jajhall:
This closes #3015

fwesselm (approved):
Thank you for fixing this.

PR #3019, opened 2026-05-12
Ml

yunusemrecagliyan:


Opt-Mucca:
@yunusemrecagliyan Have you noticed any improvements from computing parallel cuts? I don't see any changes here on suppressing global information that could be derived during separation, e.g., cliques.

yunusemrecagliyan:
> @yunusemrecagliyan Have you noticed any improvements from computing parallel cuts? I don't see any changes here on suppressing global information that could be derived during separation, e.g., cliques.

This was an accidental PR; I was just doing some experiments. Honestly, it triggered a bunch of RAM access issues, and the performance gains were either negligible or even worse in most cases. Making this work properly would require significant time and effort.

Also, the parallel improvements in this version were different from the version I actually intended to merge. While parallel processing can speed up certain problems almost proportionally to the thread count, it is far from being a universal solution that works well across all problem types.

The main gain actually came from using insertion sort instead of pdqsort, especially for speeding up the early stages. However, even that only improved early-stage performance for some specific problems, nothing more.

Opt-Mucca:
@yunusemrecagliyan is the parallel performance comment specifically related to cut generation, or something more general?

Thanks for the interesting note on `pdqsort`!

yunusemrecagliyan:
@Opt-Mucca It was more of a general parallelization attempt. I didn't specifically benchmark or track the performance gains of each individual task implementation across different workloads.

Opt-Mucca:
There's a general parallel attempt in #2886 
Any feedback is welcome!

PR #3024, opened 2026-05-14
Hipo python docs

galabovaa:


docs/src/interfaces/python/hipo-in-python.md
```diff
@@ -0,0 +1,69 @@
+# [HiPO in Python](@id hipo-in-python)
+
+The HiPO Interior Point Method (IPM) solver currently uses external dependencies to provide enhanced performance for linear and quadratic programming problems. The required dependencies are packaged in the `highspy-extras`. The packaged dependencies have licensing terms different from HiGHS, such as Apache 2.0. Other algorithms may also rely on `highspy_extras` in the future.
```
filikat: I suppose `highspy_extras` should be `highspy-extras`

docs/src/interfaces/python/hipo-in-python.md
```diff
@@ -0,0 +1,69 @@
...
+
+```bash
+pip install highspy[extras]
+```
+
+At present, the optional dependency installs support needed for HiPO. Both `highspy` and `highspy-extras` are available on PyPI and conda-forge.
+
+## Usage
+
+When `highspy_extras` is installed, HiGHS can use algorithms that depend on these external libraries. At present this primarily means the HiPO solver. You can explicitly select HiPO:
```
filikat: I suppose `highspy_extras` should be `highspy-extras`

README.md
```diff
@@ -228,12 +228,20 @@ pip install .
...
+The python package `highspy` depends on the `numpy` package and `numpy` will be installed as well, if it is not already present.
 
 The installation can be tested using the small example `HiGHS/examples/call_highs_from_python_highspy.py`.
 
 The [Google Colab Example Notebook](https://colab.research.google.com/drive/1JmHF53OYfU-0Sp9bzLw-D2TQyRABSjHb?usp=sharing) also demonstrates how to call `highspy`.
 
+To use HiPO in Python, the additional `highspy-extras` is required. It contains the HiPO dependencies and is Apache 2.0 licensed. It can be installed with
+
+```shell
+$ pip install highspy[hipo]
```
filikat: I suppose `highspy[hipo]` should be `highspy[extras]`

jajhall (approved):
Now that @filikat has identified the typos, this looks good to me!

docs/src/interfaces/python/hipo-in-python.md
mathgeekcoder: Yeah, this is a good point - and is actually a slightly larger issue that's been on my mind.

If we're going to standardize, we probably should choose `highspy_extras` instead of `highspy-extras`, since `import highspy-extras` will give a syntax error.

Python actually treats `-` and `_` in package names as equivalent, that is, it's valid to use `pip install highspy-extras` and then `import highspy_extras`.  However, users should never have to manually `import` this package themselves.

Choosing `highspy_extras` is also consistent with the associated `highs_extras` library.  That said, `highspy-extras` looks better to me as a package name (I don't know why).  

**Options:**
1. keep `highspy-extras` and refer to it everywhere as `highspy-extras` (except in advanced examples involving `import`)
2. rename to `highspy_extras` for consistency
3. rename to `highspyextras` and avoid the whole `-` and `_` weirdness.

I don't really have a strong preference, though, I'm leaning towards option 1. Thoughts @galabovaa?

docs/src/interfaces/python/hipo-in-python.md
```diff
@@ -0,0 +1,69 @@
...
+print(highspy_extras.__version__)
+print(highspy_extras.get_library_version())
+```
+
+## Local installation requirements
+
+To install locally, you will need
+
+- Python >= 3.8
+- BLAS library (bundled or system)
```
mathgeekcoder, line 57: Just to clarify, for BLAS, the `system` requirement is only for Apple right?  Since we're bundling OpenBLAS.

It is technically correct as written, but users might think they need to install additional libraries.

mathgeekcoder (commented):
Looks good!  Just want to clarify our `highspy-extras` vs `highspy_extras` decision.

docs/src/interfaces/python/hipo-in-python.md
filikat: > However, users should never have to manually `import` this package themselves.

Is this worth mentioning explicitly in the documentation? I'm not a Python person, so I don't know how likely users are to try importing it, receiving a syntax error.

mathgeekcoder: It certainly wouldn't hurt to explicitly mention.  Possibly under `## Usage`??  e.g., 

> ## Usage
> 
> When `highspy-extras` is installed, HiGHS can use algorithms that depend on these external libraries. At present this primarily means the HiPO solver. **Note that `highspy-extras` is automatically consumed by `highspy` and does not need to be imported manually.** You can explicitly select HiPO:

or

> When `highspy-extras` is installed, HiGHS can enable algorithms that rely on these external libraries, such as the HiPO solver. `highspy-extras` is automatically detected and used by `highspy`, so no explicit import is required.  You can select HiPO:

or something to that effect

galabovaa:
Thank you for your comments! 

@mathgeekcoder, I agree that

> 1. keep highspy-extras and refer to it everywhere as highspy-extras (except in advanced examples involving import)

for some reason, seems most visually pleasing. I will make the corresponding updates. 

I did not realize that for python it does not matter. I guess I assumed that there was a reason we were using both, like the highspy_extras being a small wrapper over the C++ lib and the highspy-extras being the python package, but that seems to not be the case :sweat_smile:

docs/src/interfaces/python/hipo-in-python.md
galabovaa: Done!

docs/src/interfaces/python/hipo-in-python.md
galabovaa: Done!

docs/src/interfaces/python/hipo-in-python.md
galabovaa: I added a few clarifications there.

README.md
galabovaa: Yes, I updated this to extras. I guess highspy[hipo] sounded nice, but now extras may contain other deps like cuda for pdlp.

PR #3025, opened 2026-05-15
sanitizer tests

galabovaa:
std::vector<bool> is the one STL container with no element-independence guarantee


highs/lp_data/HStruct.h
```diff
@@ -178,7 +178,7 @@ struct HighsSubSolverCallTime {
   HighsInt mip_clock_running;
   std::vector<double> submip_start_time;
   std::vector<HighsInt> submip_clock_running;
-  std::vector<bool> submip;
+  std::vector<char> submip;
```
Opt-Mucca: I'd suggest changing this to `uint8_t`

Opt-Mucca (approved):
@galabovaa only one point (I do think that should be changed though). 

Is enabling the sanitizers going to cause test times in general to now increase a lot?

galabovaa:
> @galabovaa only one point (I do think that should be changed though).
> 
> Is enabling the sanitizers going to cause test times in general to now increase a lot?

We should just run it on PR, but until all is in latest I left it on push 😅

highs/lp_data/HStruct.h
galabovaa: Sure, can do

jajhall (approved):
Looks neater than before, so what's not to like!

highs/lp_data/HStruct.h
jajhall: It'll change anyway when `hmw-mt-mip-profile` is merged

galabovaa:
@Opt-Mucca Is there a reason we have not merged this yet?

Opt-Mucca:
> @Opt-Mucca Is there a reason we have not merged this yet?

@galabovaa I think I misinterpreted a comment and thought this would be merged with `hmw-mt-mip-profile`. This should be merged.

galabovaa:
tbh that's what I thought at the time. can you please merge it? I see a conflict now with HStruct that you will know better about!

PR #3031, opened 2026-05-19
Updated third party dependency support

mathgeekcoder:
This PR makes several changes to improve external dependency support in HiGHS.

1. Optional dependencies are now more explicit, with `isAvailable<Feature1, Feature2>()` support
2. Features can be bundled into feature sets, e.g., `hipo = require<amd, blas, ...>`
3. Uses recursive templates so that nested feature sets also supported, e.g., `isAvailable<hipo, require<require<...>, ...>`
4. Uses template extensions so that new features can be added to the `app`, `libhighs`, or `highs_extras`, but still use the same API
5. Tidied up `highs_extras` API, so that adding new features is easier and more consistent
6. Added `getThirdPartyNotice()` and `--notice` to get current list of enabled dependencies (with version and license info)
7. Added `HighsDynamicLibrary`, which can be used in the near future for dynamic CUDA support
8. Tidied up `blas` cmake routines and `highs_extras` cmake for increased maintainability
9. Added `README.md` with details on how to add new dependencies and how the system works

Example of output:
```
> highs --notice
HiGHS version 1.14.0 Githash a372b82b08. Copyright (c) 2026 under MIT licence terms
Includes third-party software components, see THIRD_PARTY_NOTICES.md for full details

Third-party components:

key      name             version       license     
-------  ---------------  ------------  ------------
amd      SuiteSparse AMD  7.12.1+       BSD-3-Clause
blas     OpenBLAS         0.3.29        BSD-3-Clause
cli11    CLI11            2.5.0         BSD-3-Clause
metis    METIS-GKlib      5.2.1+        Apache-2.0  
pdqsort  pdqsort          git:b1ef26a   Zlib        
rcm      SPARSEPAK        unversioned+  MIT         
zlib     ZLIB             1.3.1         Zlib        
zstr     zstr             1.0.6         MIT  
```

PR #3033, opened 2026-05-19
Significantly improve macos build times (i.e., github actions)

mathgeekcoder:
* I noticed that a few github actions were taking >1 hour for macos builds.  
* Most of this time was spent linking libhighs. 
* This change enables `UNITY_BUILD` on `APPLE` and build time reduces to ~1 minute.

I haven't seen any issues with using it, but I noticed that it had been previously disabled.  I'm not an Apple user, so please reject this PR if it doesn't make sense.

Note: This is PR is set to merge to my `third-party-dependencies` branch, i.e., PR #3031.  I can rewrite it to work directly with latest if preferred.

### DIFF
--- a/highs/CMakeLists.txt
+++ b/highs/CMakeLists.txt
@@ -132,8 +132,18 @@ else()
   # Define library in modern CMake using target_*()
   # No interfaces (apart from c); No ipx; New (short) ctest instances.
   add_library(highs)
-
   add_library(${PROJECT_NAMESPACE}::highs ALIAS highs)
+
+  if(APPLE)
+    # linker was taking ~1 hour, unity build reduces it to ~1 minute
+    set_target_properties(highs PROPERTIES
+      UNITY_BUILD ON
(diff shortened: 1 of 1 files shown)


mathgeekcoder:
In particular look at:

[build-macos](https://github.com/ERGO-Code/HiGHS/actions/workflows/build-macos.yml?query=is%3Asuccess): `1+ hour wall clock, 6+ hours compute effort` vs `10 min wall clock, 16 minute compute effort`

Helps other workflows too, but harder to see due to other bottlenecks.

mathgeekcoder:
FYI: I also tried several alternative 'fixes':
- Reducing exported symbols
- Disabling IPO
- Static lib vs dynamic 
- Using the legacy linker
- Using no-deduplication

None of them helped.

galabovaa (approved):
Great!

galabovaa:
Thank you, @mathgeekcoder! Merging this. No need for a separate PR here!

Issue #2926, opened 2026-03-19
c# nuget packing missing setBasis


EEParker:
Hello, 

Thank you for all your work on this library, it is a genuine help!

Question: it seems as though `setBasis` is missing from the nuget package and c# api.

c#
getBasis https://github.com/ERGO-Code/HiGHS/blob/master/highs/interfaces/highs_csharp_api.cs#L347
setBasis missing?

c++
getBasis https://github.com/ERGO-Code/HiGHS/blob/master/highs/interfaces/highs_c_api.h#L998
setBasis https://github.com/ERGO-Code/HiGHS/blob/master/highs/interfaces/highs_c_api.h#L1250

For context, I'm using HiGHS for real-time power system dispatch (~2,000 variables) where the LP structure stays fixed between solves and only bound values change (simplex warm start). The current work around is P/Invoking Highs_setBasis directly, but it would be nice to have it in the nuget API.

I wasn't sure if you were taking PRs, so I've included my workaround here. This is just replicating what is already in the c++ api.

Thank you for your time.

```csharp
  [DllImport(highslibname)]
  private static extern int Highs_setBasis(IntPtr highs, int[] col_status, int[] row_status);

  [DllImport(highslibname)]
  private static extern int Highs_setLogicalBasis(IntPtr highs);

  Public wrapper methods:

  public HighsStatus setBasis(HighsBasis basis)
  {
      int[] colbasstat = basis.colbasisstatus.Select(x => (int)x).ToArray();
      int[] rowbasstat = basis.rowbasisstatus.Select(x => (int)x).ToArray();
      return (HighsStatus)HighsLpSolver.Highs_setBasis(this.highs, colbasstat, rowbasstat);
  }

  public HighsStatus setLogicalBasis()
  {
      return (HighsStatus)HighsLpSolver.Highs_setLogicalBasis(this.highs);
  }
```

Gbtank:
> For context, I'm using HiGHS for real-time power system dispatch (~2,000 variables) where the LP structure stays fixed between solves and only bound values change (simplex warm start). The current work around is P/Invoking Highs_setBasis directly, but it would be nice to have it in the nuget API.

Hi not a developer but I am curious about this usecase also. I wanted to ask whether you see runtime improvements with a warm start versus starting from scratch? I have been looking into hot starting optimisations but I don't necessarily want the solution to look the same every time. In fact, the more variable the solution the better.

EEParker:
[...]

Issue #2970, opened 2026-04-14
C# NuGet package doesn't expose Highs_changeObjectiveOffset


ethylon:
Hello team,

I'm currently using .NET package v1.14.0 and noticed that Highs_changeObjectiveOffset is missing from the exposed functions.

I'm using my own [DllImport] as a workaround for now.

Thanks!

jajhall:
I'll address this in due course in tandem with #2926

ethylon:
Here's some other functions that I need that I found missing, but I didn't want to open an issue for each of them:

Highs_version
Highs_setCallback
Highs_startCallback
Highs_stopCallback
Highs_getRanging

I appreciate your work!

jajhall:
Closed by #3041

Issue #2667, opened 2025-11-26
C# API does not support adding multiple objectives


Vab123:
The HighsLpSolver-class in the C# API does not support adding and clearing multiple linear objectives, despite this being exported by the DLLs.

https://github.com/ERGO-Code/HiGHS/blob/869cbd4df4eba475f5df0774036dce1fce490c14/highs/interfaces/highs_csharp_api.cs#L180

This could be added by the following code:
```csharp
[DllImport(Highslibname)]
private static extern int Highs_clearLinearObjectives(IntPtr highs);

[DllImport(Highslibname)]
private static extern int Highs_addLinearObjective(IntPtr highs, double weight, int offset, double[] coefficients, double absTolerance, double relTolerance, int priority);

public HighsStatus clearLinearObjectives() 
{
	return (HighsStatus)Highs_clearLinearObjectives(_highs);
}

public HighsStatus addLinearObjective(double weight, int offset, double[] coefficients, double absTolerance, double relTolerance, int priority) 
{
	return (HighsStatus)Highs_addLinearObjective(_highs, weight, offset, coefficients, absTolerance, relTolerance, priority);
}
```

jajhall:
Thanks, we will add this to the C# API

jajhall:
Closed by #3041

PR #3039, opened 2026-05-21
Implement setBasis and setLogicalBasis methods

Carl-Hossler:
Added methods to set basis and logical basis in the C# API.


jajhall:
Closes #2926

jajhall:
Thanks for this: it would be great if you could add interfaces corresponding to #2667 and #2970

PR #3041, opened 2026-05-21
Implemented additional methods to the csharp wrapper

Carl-Hossler:


jajhall (commented):
Thanks for these. Have you tested these - and the other modifications you made?

Carl-Hossler:
@jajhall - I have not tested the C++ method.

jajhall (approved):
I'm confident enough that this is correct

PR #3042, opened 2026-05-26
Prevent parallel simplex when solving MIPs

jajhall:
If  the `parallel` option is set "on" by a user in an attempt to run the parallel MIP solver, don't allow this to cause a switch from `kSimplexStrategyDual` to `kSimplexStrategyDualMulti` (in `HEkk::chooseSimplexStrategyThreads`) otherwise the MIP solver may use parallel dual simplex - which is less robust.

A couple of un-used options have been deleted

### DIFF
--- a/highs/lp_data/HighsOptions.h
+++ b/highs/lp_data/HighsOptions.h
@@ -438,9 +438,7 @@ struct HighsOptionsStruct {
   HighsInt presolve_rule_off;
   bool presolve_rule_logging;
   bool presolve_remove_slacks;
-  bool simplex_initial_condition_check;
   bool no_unnecessary_rebuild_refactor;
-  double simplex_initial_condition_tolerance;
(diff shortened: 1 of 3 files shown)


Opt-Mucca (approved):
Very happy with removing the unused options!

I'm assuming this call-path never affects any LP solve that's unrelated to a MIP solve? If so then I'm happy with the change!

jajhall:
> I'm assuming this call-path never affects any LP solve that's unrelated to a MIP solve? If so then I'm happy with the change!

It will if someone wants to invoke the parallel simplex solver by setting `parallel="on"` and `simplex_strategy=kSimplexStrategyDual`, but I'll document this in the release notes. However, 

- there are two parallel dual simplex solvers, so it's better to force users to select the one they want, rather than only using PAMI
- the use of parallel dual simplex is niche
- it's preferable to having the MIP solver fail due to parallel simplex being used

PR #3048, opened 2026-06-03
Fix HiPDLP GPU initialization consistency

Yanyu000:
This PR fixes HiPDLP GPU startup consistency.

Changes:
- keep `params_.omega`, `primal_weight_`, and `stepsize_.beta` consistent as cupdlp-C style
- use `powerMethodGpu()` for GPU step-size initialization
- project the initial primal vector before copying it to GPU state
- initialize Halpern anchors from the projected device current vector

This avoids inconsistent startup behavior where step sizes used one primal-dual weight while restart/PID bookkeeping used another.

Benchmarks at `1e-8` show mixed performance, so this PR should be treated as a correctness/stability fix rather than a performance optimization.

### DIFF
--- a/highs/pdlp/hipdlp/pdhg.cc
+++ b/highs/pdlp/hipdlp/pdhg.cc
@@ -1663,6 +1663,119 @@ double PDLPSolver::powerMethod() {
   return op_norm_sq;
 }
 
+#ifdef CUPDLP_GPU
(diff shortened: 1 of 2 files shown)


jajhall (approved):
If this improves correctness, then we have to take what performance consequences there are

PR #3052, opened 2026-06-08
Fix numpy version parsing

### LINKED ISSUE #3051: Bug: highspy incompatible with numpy pre-releases
If I install highspy using the `--pre` flag of pip, a pre-release version of numpy is installed with it:
```
pip install --pre highspy
```
gets me the environment:
```
Package Version
------- --------
highspy 1.14.0
numpy   2.5.0rc1
pip     26.1.2
```

The problem is that when simply running:
```python
import highspy
```
I then get into the following error:
```
Traceback (most recent call last):
  File "<python-input-0>", line 1, in <module>
    import highspy
  File "/home/valerian/.pyenv/versions/test-highspy/lib/python3.14/site-packages/highspy/__init__.py", line 47, in <module>
    from .highs import Highs
  File "/home/valerian/.pyenv/versions/test-highspy/lib/python3.14/site-packages/highspy/highs.py", line 22, in <module>
    np_version = tuple(map(int, np.__version__.split('.')))
ValueError: invalid literal for int() with base 10: '0rc1'
```

I think this line of code should be changed so that highspy is compatible with pre-release versions of numpy.

ValerianRey:
- Fixes #3051

### DIFF
--- a/highs/highspy/highs.py
+++ b/highs/highspy/highs.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 import sys
+import re
 import numpy as np
 from numbers import Integral
 from itertools import product
@@ -19,7 +20,9 @@ from ._core import (
 )
 
 # backwards typing support information for HighspyArray
-np_version = tuple(map(int, np.__version__.split('.')))
+# Use re.match to strip non-numeric suffixes (e.g. "0rc1") before converting
+# to int, so that pre-release numpy versions don't raise a ValueError here.
+np_version = tuple(int(re.match(r'\d+', part).group()) for part in np.__version__.split('.'))
 if sys.version_info >= (3, 9) and np_version >= (1,22,0):
     ndarray_object_type = np.ndarray[Any, np.dtype[np.object_]]
 else:


mathgeekcoder (approved):
Looks good!

@galabovaa I changed the base branch from master to latest, and it also flagged a change to CONTIBUTING.md, but seems reasonable.

jajhall:
That was my modification to CONTRIBUTING.md - about users working from latest rather than master when creating PRs! 😅

jajhall:
We need to merge this into a local branch to even view the merge conflicts

jajhall:
@mathgeekcoder and @galabovaa 

Having merged `ValerianRey:fix-numpy-version-parsing` into `fix-3052` (which is a branch off `master`), I've tried merging `latest` into `fix-3052`. The merge conflict is small, but not one that I should try to fix. It relates to `TYPE_CHECKING` being replaced by 

> np_version = tuple(int(re.match(r'\d+', part).group()) for part in np.__version__.split('.'))
if sys.version_info >= (3, 9) and np_version >= (1,22,0)

Please would one of you take this form here

mathgeekcoder:
> The merge conflict is small, but not one that I should try to fix. It relates to `TYPE_CHECKING` being replaced by 
> 
> 
> Please would one of you take this form here

If it's not urgent, I can do this next week after I get back from Korea (don't have laptop here).

PR #3055, opened 2026-06-09
Model pic

jajhall:
The facility to generate images of constraint matrices has been extended to Hessians, and is now available by setting the `write_matrix_image` or `write_hessian_image` options to be true, or by calling `Highs::matrixImage`


galabovaa (approved):
Looks good to me!

Issue #3049, opened 2026-06-05
Non-empty blank line in options file causes read error


jajhall:
A line with just a space character in an options file causes an error

jajhall:
Closed by #3057

PR #3057, opened 2026-06-10
Lines in options files with only spaces are now ignored, and text of first line with error is printed

jajhall:
I found it irritating that an options file with a line containing only spaces caused an error

Such lines are now skipped, and if an error occurs (elsewhere) the text of the line is now logged, as well as the line number

### DIFF
--- a/check/TestOptions.cpp
+++ b/check/TestOptions.cpp
@@ -1,4 +1,5 @@
 #include <cstdio>
+#include <fstream>
 
 #include "HCheckConfig.h"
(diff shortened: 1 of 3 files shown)


jajhall:
Changes to `pdhg.cpp` are due to me running clang-format on it. No reason why clang-format CI shouldn't include `HiPDLP`

jajhall:
This closes #3049

galabovaa (approved):
Looks good to me!

PR #3059, opened 2026-06-11
Implement initial RKO knapsack heuristic

123yongzheng:


jajhall:
Thanks, but please keep your work in your fork, as both of you will be working from this HiGHS branch.

In passing I have a few comments

- This implements everything in rkoHeuristic, and the code will have to be modular. Hence the set-up for knapsack problems should be in a separate method, and the search you've implemented should be in a separate method.
- The constraint matrix will always be column-wise. I'll add an assert to emphasise this. Also (more trivially) you can assume that the start vector for the constraint matrix is of size `lp.num_col_+1` so you don't have to consider using the `p_end_` vector. Note that `p_end_` is only used with row-wise storage for a special case in the simplex solver, and will always be empty, even if you generate a row-wise copy of the constraint matrix.
- Your loop 
> for (HighsInt pass = 0; pass < 64; ++pass)

contains a lot of hard-coded values. Defining these as constants in `HighsRko.h` would make the code much clearer.

PR #3062, opened 2026-06-12
Nuget api key to trusted publishing

galabovaa:
https://learn.microsoft.com/en-us/nuget/nuget-org/trusted-publishing

Closes https://github.com/ERGO-Code/HiGHS/issues/3065


jajhall (approved):
I trust you!

PR #3066, opened 2026-06-12
Add C interface to HiPO linear solver

filikat:
Changes requested in order to use the HiPO linear solver within UNO.

New features:

- pivoting can now be switched off at runtime.
- the expected sign of a pivot, used by the factorisation to decide how to perform dynamic regularisation, can now be unknown.
- the interface to the linear solver now uses pointers `T*` instead of `std::vector<T>`.
- logging for errors or statistics of the linear solver can now use `printf`, if HiGHS logging is not available.
- a function to compute the inertia of the factorisation has been added.
- the linear solver can now be used outside of HiGHS, through the C API in `highs/ipm/hipo/factorhighs/FactorHighs_c_api.h`. An example is provided in `highs/ipm/hipo/factorhighs/FactorHighs_c_api_example.c`.


filikat:
@amontoison @cvanaret could you have a look and check if these changes are ok?

filikat:
@amontoison the C API is more or less ready. Please have a look at `highs/ipm/hipo/factorhighs/FactorHighs_c_api.h` and `highs/ipm/hipo/factorhighs/FactorHighs_c_api_example.h`.

@cvanaret it's better if you use the C API as well, rather then CPP. I set up the C API to be usable outside of HiGHS, so that it automatically sets up everything by itself. If you want to use the CPP interface, you would need to perform the actions in `FactorHighs_initialise` and `FactorHighs_terminate` explicitly.

Issue #2968, opened 2026-04-13
highs.dll in NuGet package is dynamically linked to Visual C++ library


ap5d:
My program using HiGHS was initially unable to complete optimizations due to crashing on a test system. I then realized that the unamanaged highs.dll included in the NuGet package has dependencies on Visual C++ standard library dlls, such as MSVCP140.dll. The test system had an old version of Visual C++ redistributable installed, which caused the crashes.

I can think of a couple options to prevent this issue:
1. Document which version of Visual C++ redistributable needs to be installed
2. Link the highs library statically with MSVC STL

I would prefer option 2 so that the NuGet package would contain all dependencies.

jajhall:
Another one for you @galabovaa, thanks

galabovaa:
Thank you for this, I think I have sorted the static link. Would it be possible for you to test this, if I publish a pre-release version on NuGet?

galabovaa:
the pre-release NuGet can be tested with 

dotnet add package Highs.Native --version 1.14.0-preview.1

Please let us know if you run into any more issues!

Issue #2967, opened 2026-04-13
highs.dll in NuGet package not copied to output directory automatically


ap5d:
I would expect the unmanaged highs.dll included in the package to be copied to build output directory next to Highs.Native.dll.

I made a manual copy step in the build process to workaround this.

I'm using the 32-bit version because of other restrictions. NuGet package directory for the 32-bit native dll is runtimes/native/win-x32. This would typically be runtimes/native/win-x86. Does this atypical folder naming cause the dll to be missing from output?

jajhall:
One for you @galabovaa, thanks

galabovaa:
@ap5d apologies for the delay, you are right, we have an issue there which I think have resolved on a development branch. I will open a PR shortly

galabovaa:
the pre-release NuGet can be tested with

dotnet add package Highs.Native --version 1.14.0-preview.1

Please let us know if you run into any more issues!

PR #3070, opened 2026-06-15
Csharp msvc stl static link and win32 installation fix

galabovaa:
Closes https://github.com/ERGO-Code/HiGHS/issues/2968 and https://github.com/ERGO-Code/HiGHS/issues/2967

Also used to test the new NuGet Trusted Publishing: seems to work with a published pre-release version Highs.Native --version 1.14.0-preview.1


jajhall (approved):
Can't comment on its correctness, so I have to trust you

Issue #2906, opened 2026-03-09
CUDA build for python


slowjazz:
Hello,
I was having some trouble building `highspy` with CUDA pdlp support off the v1.13.1 release. 
```
$ cd HiGHS
$ uv pip install . --config-settings=cmake.define.CUPDLP_GPU=ON
```

After running `import highspy`:
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
    import highspy, pathlib; print(next(pathlib.Path(highspy.__file__).parent.rglob('*.so')))
    ^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/kli/HiGHS/.venv/lib/python3.13/site-packages/highspy/__init__.py", line 1, in <module>
    from ._core import (
    ...<43 lines>...
    )
ImportError: /data/kli/HiGHS/.venv/lib/python3.13/site-packages/highspy/_core.cpython-313-x86_64-linux-gnu.so: undefined symbol: cupdlp_projSameub_cuda
```
It doesn't appear that CUDA is linked properly.

The following diff fixes it for me:
```diff
diff --git a/cmake/python-highs.cmake b/cmake/python-highs.cmake
index 8c3bd1535..d39dd7796 100644
--- a/cmake/python-highs.cmake
+++ b/cmake/python-highs.cmake
@@ -16,6 +16,11 @@ set(headers_python ${highs_headers_python}
                    ${ipx_headers_python}
                    ${basiclu_headers_python})

+if(CUPDLP_GPU)
+  list(APPEND sources_python ${cuda_sources_python})
+  list(APPEND headers_python ${cuda_headers_python})
+endif()
+
 # Find Python 3
 find_package(Python COMPONENTS Interpreter Development.Module REQUIRED)
 find_package(pybind11 CONFIG)
@@ -45,6 +50,10 @@ target_link_libraries(_core PRIVATE pybind11::headers)
 # sources for python
 target_sources(_core PUBLIC ${sources_python} ${headers_python})

+if(CUPDLP_GPU)
+  target_link_libraries(_core PRIVATE ${CUDA_LIBRARY})
+endif()
+
 # include directories for python
 target_include_directories(_core PUBLIC ${include_dirs_python})

diff --git a/cmake/sources-python.cmake b/cmake/sources-python.cmake
index c60b37613..8cb27b9cc 100644
--- a/cmake/sources-python.cmake
+++ b/cmake/sources-python.cmake
[...]

Issue #3008, opened 2026-05-07
Issues when linking cuda files on windows for python


MateuszKrol13:
Following the steps listed in #2906 seemed to fix the issue with linking files, they cannot be processed as nvcc sees multiple inputs for singular output. I have little to no knowledge on nvcc at this time, would appreciate any help.

run command: python -m pip install . --config-settings=cmake.define.CUPDLP_GPU=ON

` Compiling CUDA source file F:\dev\python\HiGHS\highs\pdlp\cupdlp\cuda\cupdlp_cuda_kernels.cu...
[...]

PR #3071, opened 2026-06-16
GPU support for local Python pip install

galabovaa:
NVidia GPUs now supported locally from Python via 

pip install . --config-settings=cmake.define.CUPDLP_GPU=ON

Tested locally on Windows and Linux with NVIDIA GeForce RTX 4050

Closes https://github.com/ERGO-Code/HiGHS/issues/2906 and https://github.com/ERGO-Code/HiGHS/issues/3008

Will likely be updated later for dynamic loading, like in highspy_extras


jajhall (approved):
Good to have, and I trust you to have done it correctly

PR #3072, opened 2026-06-17
Add explicit -fexceptions copts to Bazel config

mkruskal-google:
This allows users to build with -fno-exceptions, and smoothly turns exceptions into crashes at the boundary.


galabovaa:
Changed base from master, to latest. Will close and re-open to trigger CI

Did not work. CI running in https://github.com/ERGO-Code/HiGHS/pull/3075

PR #3077, opened 2026-06-19
Add blas compile def to highs library

filikat:
`HIGHS_BLAS_COMPILE_DEFINITION` is defined only for the `highs_extras` library and not for the `highs` library, but it is needed in both.


filikat:
@galabovaa I need to define `HIGHS_BLAS_COMPILE_DEFINITION` for both libraries, but there's something going wrong for the python wheels. Could you have a look?

PR #3080, opened 2026-06-22
Check BLAS used in HiPO at runtime

filikat:
HiPO was using `HIPO_USES_APPLE_BLAS` for making algorithmic choices.
However, this is not defined in `highs` anymore, but only in `highs_extras`.
The use of Apple Accelerate can be checked at runtime using the HighsExtras interface.


galabovaa:
Looks great!

galabovaa (approved):

Issue #2910, opened 2026-03-12
Feature request: Expose additional HiPO timing fields via HighsInfo


MaykThewessen:
## Summary

HiPO's internal `hipo::Info` struct contains detailed timing data that is valuable for profiling and optimization, but only 4 of 9+ timing fields are currently exposed via `HighsInfo`.

## Currently Exposed (via IpxWrapper.cpp)

These 4 fields are forwarded in `solveHipo()`:
- `ipm_analyse_time` — symbolic analysis (ordering + elimination tree)
- `ipm_factor_time` — numerical factorisation
- `ipm_solve_time` — triangular solves
- `ipm_matrix_time` — matrix assembly

## Proposed Additions

The following fields exist in `hipo::Info` but are not forwarded to `HighsInfo`:

| hipo::Info field | Proposed HighsInfo field | Description |
|-----------------|------------------------|-------------|
| `residual_time` | `ipm_residual_time` | Residual computation (primal/dual infeasibility) |
| `omega_time` | `ipm_omega_time` | Omega (scaling weight) computation |
| `AS_structure_time + NE_structure_time` | `ipm_structure_time` | Sparsity structure detection |
| `factor_number` | `ipm_factor_count` | Number of factorisations (integer) |
| `solve_number` | `ipm_solve_count` | Number of triangular solves (integer) |

## Use Case

We're benchmarking HiPO on PyPSA DC-OPF problems (168-672h chunks, ~600k constraints). Having detailed timing breakdowns helps us:
1. Identify which IPM phases dominate solve time
2. Guide optimisation efforts (e.g., symbolic reuse, parallel factorisation)
3. Compare HiPO vs IPX phase-by-phase rather than just total time

## Implementation

The changes are straightforward:
1. Add 5 fields to `HighsInfoStruct` in `HighsInfo.h`
2. Register them as `InfoRecord` entries in `initRecords()`
3. Forward from `hipo_info` in `IpxWrapper.cpp::solveHipo()`

We have a working implementation on our fork — happy to submit a PR.

jajhall:
Interesting request. `HighsInfo` is not the place for such "niche" developer information relating to timing. In the MIP solver we currently compute the time required for execution of sub-solvers - which are of comparable interest to developers and expert users. Perhaps a `HighsTime` struct can be created to communicate back all such times. 

That does nothing for your desire to have `factor_number` and `solve_number`. We can consider adding these to `HighsInfo`. I'll talk to @filikat

[...]

PR #3081, opened 2026-06-22
Fix column stuffing

jajhall:
This identifies and fixes the source of the postsolve errors with column stuffing when applied to the relaxation of 'neos-787933'. It's a draft PR because I've left in the debug printing that was useful in pinpointing the error.

It also introduces two other development utilities

**Presolve rule testing** so that presolve can be run to test specific reductions on minimal instances. This greatly expands the scope for testing presolve rules

**The `HighsRunData` class** to collect run-time data that had not been available via the API. In the case of presolve rule testing, it provides sanity checks that
 
- minimal instances are reduced to empty;
- after dual postsolve the solution is still optimal;
- after basis postsolve no simplex iterations are required

This class is also a route towards closing #2910

### DIFF
--- a/highs/presolve/HPresolve.cpp
+++ b/highs/presolve/HPresolve.cpp
@@ -4990,10 +4990,6 @@ HPresolve::Result HPresolve::singletonColStuffing(
   HighsInt numFixedCols = 0;
   // Temporary for fix-col-stuffing
   bool report_stuffing = true;
-  if (report_stuffing && col == 0) {
-    printf("HPresolve::singletonColStuffing for col = 0\n");
-  }
-
   struct candidate {
     HighsInt col;
     double val;
@@ -5113,15 +5109,16 @@ HPresolve::Result HPresolve::singletonColStuffing(
       // abandoned.
       //
       // Question: The greedy algorithm is only guaranteed to assign
-      // values optimally if the variables are continuous. What
-      // presolve consequences are there if integer variables are
-      // assigned according to a sub-optimal assignment of values?
-      //
-      // Perhaps the criterion minWeight != maxWeight (ie not all |aj|
-      // values being equal) leading to abandonment of stuffing if
-      // there are only integer columns overcomes this. However, is it
+      // values optimally if the variables are continuous. The
+      // criterion minWeight != maxWeight (ie not all |aj| values
+      // being equal) leading to abandonment of stuffing if there are
+      // only integer columns overcomes this. However, is it
       // guaranteed that with those continuous decisions the knapsack
       // is solved optimally?
+      //
+      // Question: Is there value in solving the knapsack problem by
+      // DP if all variables are integer, and not all |aj| values are
+      // equal?
       if (isSingleton(j)) {
         // check singleton
         if (aj > 0) {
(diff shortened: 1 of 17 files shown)


highs/presolve/HPresolve.cpp
```diff
@@ -5170,12 +5228,38 @@ HPresolve::Result HPresolve::singletonColStuffing(
...
+              double(delta), double(delta),
+              double(direction * rhs - sumUpper + primal_feastol));
         numFixedCols++;
         HPRESOLVE_CHECKED_CALL(fixCol(t.col, t.multiplier));
       } else if (sumLowerFinite &&
                  direction * rhs <= sumLower + primal_feastol) {
-        numFixedCols++;
-        HPRESOLVE_CHECKED_CALL(fixCol(t.col, -t.multiplier));
+        // Only allow fixing if there is no degeneracy
+        const bool allow_fixing =
```
jajhall: Only functional change is setting and using `allow_fixing`

highs/presolve/HPresolve.cpp
```diff
@@ -5170,12 +5228,38 @@ HPresolve::Result HPresolve::singletonColStuffing(
...
+              "%11.4g | "
+              "Logic: delta = %11.4g <= %11.4g = direction * rhs - sumUpper + "
+              "primal_feastol\n",
+              int(t.multiplier), int(t.col),
+              t.multiplier < 0 ? "lower" : "upper", model->col_cost_[t.col],
+              double(delta), double(delta),
+              double(direction * rhs - sumUpper + primal_feastol));
         numFixedCols++;
         HPRESOLVE_CHECKED_CALL(fixCol(t.col, t.multiplier));
       } else if (sumLowerFinite &&
```
fwesselm, line 5177: I would prefer just having a single check here:
```
        else if (sumLowerFinite &&
                 delta <= sumLower - direction * rhs + primal_feastol) {
        numFixedCols++;
        HPRESOLVE_CHECKED_CALL(fixCol(t.col, -t.multiplier));
      }
```

highs/presolve/HPresolve.cpp
```diff
@@ -5068,7 +5071,54 @@ HPresolve::Result HPresolve::singletonColStuffing(
       double sumUpperBound = model->col_upper_[j];
       bool isCandidate = allowIntegerCandidates ||
                          model->integrality_[j] != HighsVarType::kInteger;
-
+      // Considering row as a knapsack
```
fwesselm: Could these comments be moved to an issue?

fwesselm (commented):
Thank you, @jajhall. I like the new testing code.

Are you planning to keep the `Temporary for fix-col-stuffing` and `report_stuffing` lines?

fwesselm:
The `report_stuffing` code could also live in a lambda (see [here](https://github.com/ERGO-Code/HiGHS/commit/6a6efc3e0c0bf3db2b1dc3bdeb02d4c4d380df66)).

jajhall:
> The `report_stuffing` code could also live in a lambda (see [here](https://github.com/ERGO-Code/HiGHS/commit/6a6efc3e0c0bf3db2b1dc3bdeb02d4c4d380df66)).

I'm happy to just delete it. I left it there in case it was useful to you in checking my fix and studying further testing examples

> Could these comments be moved to an issue?

Yes, they are now largely unnecessary since I understand what column stuffing is doing, and the questions are more issue-related

highs/presolve/HPresolve.cpp
jajhall: > I would prefer just having a single check here:
> 
> ```
>         else if (sumLowerFinite &&
>                  delta <= sumLower - direction * rhs + primal_feastol) {
>         numFixedCols++;
>         HPRESOLVE_CHECKED_CALL(fixCol(t.col, -t.multiplier));
>       }
> ```

Sure, this is natural if you agree that my fix is correct. Some comment about the logic in this (and the initial "if") would be in order. Also, I guess that there is a degenerate situation with the "if" that needs to be avoided

I'll make the changes and flag up the PR as not being draft

fwesselm (approved):

PR #3082, opened 2026-06-22
Minor changes to FindHipoDeps to support Intel MLK BLAS library.

mathgeekcoder:
**Note:** we currently define `blasint` as 32bit, so need to use `64lp` rather than `64ilp` (or update blasint to `int64_t`).

**Build notes:**

I tested this by installing Intel MKL via vcpkg: `intel-mkl`. I modified the `extern/vcpkg.json` to replace `OpenBLAS` dependency. Since `x64-windows-static` is defined vcpkg installed `Intel10_64ilp_seq` on my machine.

There were some issues with `find_package(MKL ...)`, so I needed to specify `MLK_LINK` and `MKL_THREADING` for it to find the correct files. Since vcpkg installs `64ilp` by default, I needed to modify `mycblas.h` to change `blasint` definition.  Lastly, without specifying `Release`, the `find_package(...)` found the debug lib files from vcpkg. 

It should be possible to configure MKL differently (e.g., not using vcpkg), just make sure the build flags match whatever you're trying to do.

**Full parameters:**

```cmake
HIGHS_VCPKG=ON 
HIPO=ON 
CMAKE_BUILD_TYPE=Release 
BLA_STATIC=ON
BLA_VENDOR=Intel10_64ilp_seq 
MKL_LINK=static 
MKL_THREADING=sequential
```


galabovaa:
LGTM!

galabovaa (approved):
Looks great!

PR #3084, opened 2026-06-23
Update parallel documentation

Opt-Mucca:
@galabovaa Are documentation changes supposed to target latest?

### DIFF
--- a/docs/src/parallel.md
+++ b/docs/src/parallel.md
@@ -59,17 +59,16 @@ order to select the best one.
 
 ## MIP
 
-The only parallel computation currently implemented in the MIP solver
-occurs when performing symmetry detection on the model, when querying
+If the [parallel](@ref option-parallel) option is set to "on", the MIP solver
+will explore the branch-and-bound tree using multiple threads.
+This exploration includes cuts and heuristics that are not run from the root node.
+In addition, the MIP solver utilises parallelism when performing 
+symmetry detection on the model, when querying
 clique tables, and when the interior point solver is used to compute
 the analytic centre. This parallelism is always advantageous, so is
 performed regardless of the value of the [parallel](@ref option-parallel) option.
 
 ## Future plans
-
-The MIP solver has been written with parallel tree search in mind. A
-prototype multithreaded solver was completed in February 2026, and is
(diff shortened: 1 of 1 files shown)


jajhall (changes requested):
I'd clarify the role of the `threads` option more explicitly at the top of the file. Rather than 

> By default, when running in parallel, HiGHS will use half the available threads on a machine. This number can be modified by setting the value of the [threads](https://ergo-code.github.io/HiGHS/stable/options/definitions/#option-threads) option.

I'd write

The value of the [threads](https://ergo-code.github.io/HiGHS/stable/options/definitions/#option-threads) option determines the number of threads used by HiGHS. By default it is zero, in which case HiGHS will use half the available threads on a machine. If it is set to one, then HiGHS will never use more than one thread. The maximum value that is advantageous is machine-dependent, but it is unlikely to be more than eight due to most computation in HiGHS being memory-bound.

I'd put "In addition, the MIP solver utilises... " in a separate paragraph so that the comment at the end about "This parallelism ... is performed regardless of the value of the parallel option." is totally unambiguous. I'd close the MIP section by observing that 

The extent of the parallelism depends on the value of the [threads](https://ergo-code.github.io/HiGHS/stable/options/definitions/#option-threads) option (see above).

I feel that the same statement should also be added below the IPM section @filikat

Opt-Mucca:
@jajhall I've made the suggested changes.

I haven't added the sentence to the IPM section (will wait for thoughts of @filikat)

filikat:
For HiPO, there is basically nothing run in parallel if `parallel=off`.
Maybe running multiple orderings in parallel should happen regardless of the value of `parallel`, but I'm not going to make this change for this release.
We can add the same statement about threads in the IPM section.
The sentence "Multi-threading within HiPO will be extended to other phases of the solver, including the solve phase of the factorisation and the process of assembling the matrices" in Future plans is outdated and should be removed.
I can make changes tomorrow, or feel free to make them yourself @Opt-Mucca.

Opt-Mucca:
@filikat I'll add the changes now

galabovaa (approved):
LGTM!

Issue #2996, opened 2026-04-30
Subprocesses terminate while testing clang builds


orlitzky:
I should know better, but in https://github.com/ERGO-Code/HiGHS/issues/2690 I reported two different issues, only one of which is fixed. After building HiGHS with **clang++**, several tests fail:

```
The following tests FAILED:
	  1 - capi_unit_tests (Subprocess aborted)
	  3 - unit_tests_all (Subprocess aborted)
	104 - small_mip--presolve=off (Subprocess aborted)
	...
	153 - dcmulti--random_seed=3 (Subprocess aborted)
```

These all look the similar in the log:

```
Solving MIP model with:
   5 rows
   8 cols (1 binary, 0 integer, 0 implied int., 6 continuous, 1 domain fixed)
   14 nonzeros

Src: B => Branching; C => Central rounding; F => Feasibility pump; H => Heuristic;
     I => Shifting; J => Feasibility jump; L => Sub-MIP; P => Empty MIP; R => Randomized rounding;
     S => Solve LP; T => Evaluate node; U => Unbounded; X => User solution; Y => HiGHS solution;
     Z => ZI Round; l => Trivial lower; p => Trivial point; u => Trivial upper; z => Trivial zero

        Nodes      |    B&B Tree     |            Objective Bounds              |  Dynamic Constraints |       Work
Src  Proc. InQueue |  Leaves   Expl. | BestBound       BestSol              Gap |   Cuts   InLp Confl. | LpIters     Time

 J       0       0         0   0.00%   -inf            3.5                Large        0      0      0         0     0.1s
 T       0       0         0   0.00%   3.236842105     3.236842105        0.00%        0      0      0         0     0.1s
terminate called after throwing an instance of 'HighsTask::Interrupt'
<end of output>
```

filikat:
We didn't really address this issue yet, so it's not surprising that you still notice it.
This is due to #2574.

orlitzky:
Oh, I didn't realize (or forgot) that the cause was known. I only opened this to make sure that the issue didn't get lost.

filikat:
That's fine. We will eventually get to the bottom of this. The problem doesn't appear often (or at all) for most users, and it involves parts of the code that are very complicated, so we're taking some time to fix this.

PR #3087, opened 2026-06-24
Fix two bugs in parallel scheduler

### LINKED ISSUE #2574: Destructor of TaskGroup can throw exception
The destructor of highs::parallel::TaskGroup calls:

~TaskGroup()
taskWait()
sync()
sync_stolen_task()
leapfrogStolenTask()
runStolenTask()
checkInterrupt()

which may throw HighsTask::Interrupt(). 
There is no try-catch in this chain of calls, and the destructor is noexcept by default.
So, if an exception is thrown, std::terminate is automatically called.
This sometimes happens in HiPO, when running test tsp-p01 on Mac in Debug.

filikat:
Multiple crashes with message `terminating due to uncaught exception of type HighsTask::Interrupt` have been noticed. This PR fixes two bugs in the parallel scheduler that were responsible for this behaviour.

In detail:
- The function `HighsSplitDeque::checkInterrupt` now sets `rootTask` to be null before throwing an exception. This ensures that more exceptions cannot be thrown during the process of stack unwinding.
- The destructor of `TaskGroup` now sets `rootTask` to be null before waiting for the cancelled tasks to finish. This guarantees that all tasks are completed when the object is destroyed. It then reinstates the previous `rootTask` and calls `HighsSplitDeque::checkInterrupt`, potentially throwing an exception.
- The destructor of `TaskGroup` is now declared `noexcept(false)` to allow the behaviour of the previous point to work.

Additional changes:
- A new unit test `CancelNestedTasks` that uses nested `TaskGroup` objects to spawn tasks and immediately cancel them has been added. This was failing with `latest` but it passes after the changes.
- The object `TaskGroupSpecial` that was used inside of HiPO in order for the tests to pass has now been removed.

Closes #2574 (and #2690 and #2996)


galabovaa (approved):
Looks great! Thank you @filikat!

Issue #2521, opened 2025-09-03
Error hot-starting the QP solver


jajhall:
The QP solver fails to hot start if HiGHS is given a basis and then a solution, since the solution invalidates the incumbent basis. It should be done in the reverse order. Create Highs::passSolutionAndBasis to take both and apply them in the right order. 

However the Atwood repeated model solve exposes an error when the lower bound of a constraint is reduced (so the previous solution and basis are primal feasible), but are considered optimal incorrectly.

PR #3089, opened 2026-06-25
QP hot start

jajhall:
As exposed by #2521 the existing QP hot start can fail. Some study was done in [fix-2521](https://github.com/ERGO-Code/HiGHS/tree/fix-2521) but until proper work is done with QP hot start, disable it using the option `allow_qp_hot_start`. This is false by default. However, since QP hot start works for the `test-qp-hot-start` unit test and robust OCS (see [IAA](https://github.com/jajhall/IAA)) still allow it to be switched on.

Also deletes `highs/qpsolver/crashsolution.hpp`, transferring its one method to one instance of a lambda


galabovaa (approved):
Looks good to me!

PR #3091, opened 2026-06-25
Fix 3090: Re-check implied bound

### LINKED ISSUE #3090: Incorrectly detected infeasibility due to presolve
Dear all,

I have encountered the following issue.

[model.mps.txt](https://github.com/user-attachments/files/29341269/model.mps.txt)
[model.sol.txt](https://github.com/user-attachments/files/29341270/model.sol.txt)

Highs 1.14.0 incorrectly declares the above problem to be infeasible.

Executing
```shell
highs model.mps
``` 
results in Status=Infeasible, while both
```shell
highs --read_solution_file model.sol model.mps
```
and
```shell
highs --presolve=off model.mps
```
result in Status=Optimal.

The model is small and well-conditioned, in my opinion.
```shell
Running HiGHS 1.14.0 (git hash: n/a): Copyright (c) 2026 under MIT licence terms
Command line parsed using CLI11 2.5.0: Copyright (c) 2017-2025 University of Cincinnati
Number of BV entries in BOUNDS section is 96
MIP model has 73 rows; 112 cols; 177 nonzeros; 96 integer variables (96 binary)
Coefficient ranges:
  Matrix  [1e+00, 1e+02]
  Cost    [0e+00, 0e+00]
  Bound   [1e+00, 2e+01]
  RHS     [1e+00, 6e+02]
```
Props to my coworker Ricardo Euler for first reproducing this phenomenon.

peterlietz:
fixes #3090

Full disclosure:
I used Claude to analyze and fix the problem. Since the change consists of only two lines, I hope this is OK.
I added a test case, too.

### DIFF
--- a/check/TestPresolve.cpp
+++ b/check/TestPresolve.cpp
@@ -910,3 +910,13 @@ TEST_CASE("presolve-issue-2874", "[highs_test_presolve]") {
   REQUIRE(highs.presolve() == HighsStatus::kOk);
   REQUIRE(highs.getModelPresolveStatus() == HighsPresolveStatus::kInfeasible);
 }
+
+TEST_CASE("bound_implied", "[highs_test_presolve]") {
+  std::string model_file =
+      std::string(HIGHS_DIR) + "/check/instances/bound_implied.mps";
+  Highs highs;
+  highs.setOptionValue("output_flag", dev_run);
+  highs.readModel(model_file);
+  highs.run();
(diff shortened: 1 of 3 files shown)


Opt-Mucca (approved):
@peterlietz Very nice spot! I believe Claude is correct here.

The instance is worth adding to our tests too (we were outputting an incorrect result).

Issue #3044, opened 2026-06-01
Populating dict with highs_var object throws delayed exception w/o description


RicardoEuler:
Hi everybody,

I've encountered the following issue: 

### Issue
The following code throws an exception without description at the last line:

  ```
import highspy

model = highspy.Highs()

# Add variables
var = model.addBinary(name=f"x{0}")

values = {var: 0}

print(model.getVariables()[0] is var)  # prints False: different object, same hash

# calls __eq__ due to hash conflict, which returns highs_linear_expression,
# triggers call to __bool__ in highs_linear_expression, setting the thread-local chain variable
values[model.getVariables()[0]]

model.addConstr(
    model.getVariables()[0] <= 1
)  # exception after false positive in highs_linear_expression.__is_active_chain()
```

By accessing the `values` dict, the `__bool__` method of `highs_linear_expression` is called, which sets the thread-local `chain` variable.
This is most likely triggered (I didn't step through the Cpython implementation, so I'm not 100% certain here) by a call to `highs_var.__eq__` which returns a `highs_linear_expression` object instead of a Boolean.

###  Expected behavior: 

I would expect one of the following:
- `highs_var` is not hashable
- the above code produces no exception and correctly adds the constraint
- The raised exception describes the root cause

### Workaround
One can just store integers instead of the `highs_var`'s themselves.

### Version
Tested on highspy 1.13.1 and 1.14.0 with Python 3.13.5 on  Ubuntu 24.04.2 LTS

**Comment by jajhall** [comment]
Can you see a fix for this, @mathgeekcoder?

**Comment by mathgeekcoder** [comment]
Thanks for the mention @jajhall.  

Good spotting @RicardoEuler, this looks like a real edge case bug.  I'm curious to know when it's helpful to use the `highs_var` as a dictionary key, but it certainly should be possible.

**FYI:** expression chains allow us to write constraints like `1 <= x + y <= 5`. Python parses this as `bool(1 <= x + y) <= 5`, and we need to do some tricks to interpret it correctly. The `dict[x]` above is basically calling `bool(x == x)`. Expression chains don't make sense for equality constraints, so we can simply ignore them.

I'm too tired to trust myself with a PR right now (2am), but it's a one-line fix:

```python
# ignore equality constraints or expressions without bounds
highs_linear_expression.__chain.check = (
    self if self.bounds is not None and self.bounds[0] != self.bounds[1] else None
)
```

RicardoEuler:
[...]

PR #3092, opened 2026-06-25
Fixed issue #3044

mathgeekcoder:
Fixed issue #3044, an edge case with lb <= expr <= ub chains.  Also added exception if addConstr fails


galabovaa (approved):
Looks good to me! Thank you!

PR #3093, opened 2026-06-25
Add context manager support to highspy

mathgeekcoder:
For example: 

```python
with Highs() as h:
   ...
   h.solve()
```

Feature requested by Adrián during HiGHS 2026 workshop.

Note that this still might not actually release all memory immediately; it is still somewhat dependent on the python garbage collection implementation.


Opt-Mucca (approved):
I never would have assumed that someone wanted this functionality. I'm amazed and find it pretty cool that it takes so few lines to implement!

GarnierAdrian:
Thanks for implementing this! I had a similar one in my code to make sure the memory was cleared out. I think the plotting library was holding a reference to some variables, so it was not freeing the memory each iteration, so having this context manager made sure it would be cleared after it was used.

PR #3097, opened 2026-06-26
Implement strongly connected component analysis on the clique table

Opt-Mucca:
This PR implements an iterative form of tarjans algorithm that traverses the clique table and tries to find strongly connected component, e.g. a proof that `x = y = ~z`, as well as  infeasible assignments, e.g., `x -> y -> ~x => x can't be true`.

It is first mentioned in [Mixed integer programming: Analyzing 12 years of progress](https://www.researchgate.net/profile/Tobias-Achterberg/publication/285227411_Mixed_Integer_Programming_Analyzing_12_Years_of_Progress/links/569d69b608ae00e5c98ed97d/Mixed-Integer-Programming-Analyzing-12-Years-of-Progress.pdf), albeit some improvements of the algorithm are intentionally not described in depth. The implementation is largely copied from [SCIP 5.0](https://opus4.kobv.de/opus4-zib/files/6629/scipopt-50.pdf), where it gave a ~2% performance improvement (only getting ~0.2% on my smaller test set)

Some notes:
- I did not implement the topological sorting that SCIP does for a second DFS sweep. I don't think it's that important, but it still may be worth adding.
- The algorithm is kind of hacked in because the clique table is not always up to date with the presolved problem. Therefore I call it at the start of probing and stash all the reductions in `HighsDomain` and `HighsCliqueTable`, which will then be applied in `finaliseProbing`

@fwesselm No rush to review this. I'd rather attention is on #2990 first.


Opt-Mucca:
This was -0.5% on the larger test set and solved the same amount of instances (one or two changes between the two solved sets). I'd say this is complete noise, as was my 0.2% improvement observation. I'd therefore conclude that either this technique is not useful and its relevance is entirely SCIP specific, or we're not testing on instances where this technique shines. I'd put this PR on ice and re-evaluate it in another 6 months or a year.

jajhall:
Shame, but was sensible to try it.

PR #3108, opened 2026-06-28
All values passed for string option values (except for file names) are converted to lower case

### LINKED ISSUE #3107: Case when setting string options should not matter
In #3106 the call

> `model.setOptionValue("parallel", "On")`

fails to set the parallel option to "on".

This shouldn't happen

jajhall:
Closes #3107

### DIFF
--- a/check/TestOptions.cpp
+++ b/check/TestOptions.cpp
@@ -560,3 +560,22 @@ TEST_CASE("incomplete-options-file-line", "[highs_options]") {
 
   std::remove(incomplete_options_file.c_str());
 }
+
+TEST_CASE("string-option-case-insensitivity", "[highs_options]") {
+  Highs highs;
+  highs.setOptionValue("output_flag", dev_run);
+  REQUIRE(highs.setOptionValue("output_flag", "True") == HighsStatus::kOk);
+  REQUIRE(highs.setOptionValue("output_flag", "TRUE ") == HighsStatus::kOk);
+  REQUIRE(highs.setOptionValue("output_flag", "FaLsE") == HighsStatus::kOk);
+  REQUIRE(highs.setOptionValue("output_flag", " false") == HighsStatus::kOk);
+  REQUIRE(highs.setOptionValue(kParallelString, "On") == HighsStatus::kOk);
+  REQUIRE(highs.setOptionValue(kParallelString, "OFF ") == HighsStatus::kOk);
+  REQUIRE(highs.setOptionValue(kSolverString, " Choose") == HighsStatus::kOk);
+  REQUIRE(highs.setOptionValue(kSolverString, "HiPO") == HighsStatus::kOk);
+  REQUIRE(highs.setOptionValue(kHipoOrderingString, "Metis") ==
+          HighsStatus::kOk);
+  REQUIRE(highs.setOptionValue(kHipoParallelString, "Tree") ==
+          HighsStatus::kOk);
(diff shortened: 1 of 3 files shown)


odow:
The counter point to this is that you could accept only the case sensitive strings and error when something doesn't match? (Shouldn't that already be the case?)

odow:
```Julia
julia> highs = Highs_create()
Ptr{Nothing}(0x0000000818ef0000)

julia> Highs_setStringOptionValue(highs, "parallel", "On")
ERROR:   Value "On" for parallel option is not one of "off", "choose" or "on"
-1
```
Seems good to me... I wonder why the OP didn't realise in https://github.com/ERGO-Code/HiGHS/discussions/3106?

jajhall:
That's what was done before - other than strings for bool options - but today someone used "On", not "on" for the parallel option. He and I both lost time trying to work out what happened. 

Of course he didn't check the HighsStatus::kError return 🙄

jajhall:
> ```julia
> julia> highs = Highs_create()
> Ptr{Nothing}(0x0000000818ef0000)
> 
> julia> Highs_setStringOptionValue(highs, "parallel", "On")
> ERROR:   Value "On" for parallel option is not one of "off", "choose" or "on"
> -1
> ```
> 
> Seems good to me... I wonder why the OP didn't realise in #3106?

He's just acknowledged that the error logging is there for him to see!

galabovaa:
LGTM, should we do it for all options? 

I guess you are missing the "solver" option in the list 

```
-------------------------------------------------------------------------------
string-option-case-insensitivity
-------------------------------------------------------------------------------
/home/ivet/code/HiGHS/check/TestOptions.cpp:564
...............................................................................

/home/ivet/code/HiGHS/check/TestOptions.cpp:574: FAILED:
  REQUIRE( highs.setOptionValue(kSolverString, "HiPO") == HighsStatus::kOk )
with expansion:
  -1 == 0
```

galabovaa (approved):

jajhall:
> LGTM, should we do it for all options?
> 
> I guess you are missing the "solver" option in the list
> 
> ```
> -------------------------------------------------------------------------------
> string-option-case-insensitivity
> -------------------------------------------------------------------------------
> /home/ivet/code/HiGHS/check/TestOptions.cpp:564
> ...............................................................................
> 
> /home/ivet/code/HiGHS/check/TestOptions.cpp:574: FAILED:
>   REQUIRE( highs.setOptionValue(kSolverString, "HiPO") == HighsStatus::kOk )
> with expansion:
>   -1 == 0
> ```

Do you get this failure with fix-3107 built locally?

It runs fine on oronsay and fletcher

I saw all the CI test failures, but couldn't see from the logs what was failing

jajhall:
> LGTM, should we do it for all options?

What I've implemented trims all prospective string option values and converts them to lower case for all but the options that are file names

> I guess you are missing the "solver" option in the list

Which list? I'm confused

galabovaa:
No, I got confused! 

I thought the list of options in possibleLowerCaseOptionValue(.. is a set of which options we will convert to lower case, but it is just an exclusion of the file options, as you say.

Yes, I got the failure on my linux laptop! Tested the changes quickly with 

```
cmake -S. -B build -DALL_TESTS=ON
cmake --build build -j30
./build/bin/unit_tests [highs_options]
```

Possibly deleting the build folder may help you to reproduce

jajhall:
I was building with -DHIPO=on !

PR #3110, opened 2026-06-29
Implement HighsDynamicLibrary::call in a C++11 & C++20 compatible way

hedtke:
The usage of `std::result_of` within the header makes it impossible for this code to compile when the user uses C++20 but has HiGHS as dependency.

Added a new guard to change the function signature to be compatible with C++20 and later.

I tried to create the conan package for this release, but with my local setting of cppstd=20, compilation fails


galabovaa:
The CI looked fine, apart from the julia tests which I modified in the linked PR

galabovaa:
I did reproduce the issue locally and verified that the fix works for the conan build. Will add this to the release tests, thank you @hedtke

mathgeekcoder:
Thanks @hedtke! The fix is now incorporated into #3114, so we can close this PR _without merge_.  I also changed the code to avoid the preprocessor check, which works for C++11 through to C++20 (and beyond):

```cpp
  template <typename FuncType, typename... Args>
  auto call(const char* name, Args... args) const
      -> decltype(std::declval<FuncType>()(std::declval<Args>()...)) {
    FuncType func;
    if (!resolve(func, name)) {
      throw std::runtime_error(std::string("Failed to resolve symbol: ") +
                               name);
    }
    return func(args...);
  }
```

Issue #3105, opened 2026-06-28
highs-1.15.0-arm-apple-static-apache hipo features unavailable


ollie-bell:
Downloading and unzipping from the releases page, then running:

```bash
highs-1.15.0-arm-apple-static-apache/bin/highs --solver=hipo HiGHS/check/instances/afiro.mps
```

results in an error:

```
ERROR:   The HiPO solver was requested via the "solver" option. The following features are unavailable: amd, blas, metis, rcm
```

1.13.1 and 1.14.0 both run hipo as expected with the same command.

jajhall:
Please would you look at how this has slipped through, @galabovaa

ollie-bell:
Adding `-DBUILD_SHARED_EXTRAS_LIB=OFF` to the cmake command appears to fix the issue.

galabovaa:
Resolved in https://github.com/ERGO-Code/HiGHS/pull/3112, will be in the patch release.

PR #3112, opened 2026-06-29
Static packages generation bug fix 

galabovaa:
Fix to the build of static release binaries. 

Closes https://github.com/ERGO-Code/HiGHS/issues/3105 and https://github.com/ERGO-Code/HiGHS/issues/3117

added tests too :sweat_smile:


jajhall (approved):
I can't comment on this, so have to trust you

mathgeekcoder (approved):
Looks good.  I'm assuming `BUILD_STATIC_EXE` is released as apache license (when `HIPO=ON`).

At some point we'll need to see why `NO_AVX512` is necessary.

galabovaa:
BUILD_STATIC_EXE is used for both the apache and the MIT licenses. Should I set some other flags in CMake? Is this not OK for the extras? 

For the MIT build, I get locally 

```
[100%] Linking CXX executable ../bin/highs
/usr/bin/ld: ../lib/libhighs.a(HighsDynamicLibrary.cpp.o): in function `HighsDynamicLibrary::load(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)':
HighsDynamicLibrary.cpp:(.text+0x1a2): warning: Using 'dlopen' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
[100%] Built target highs-bin
```

This is the workflow: 

eg this line was only necessary for the apache binaries https://github.com/ERGO-Code/HiGHS/blob/37b80048f2a5dd65347ee35308bc03a0c6c76691/.github/workflows/release-cpack.yml#L24

perpaps I should add it now to the MIT ones as well? Even with that I still get the warnings above

mathgeekcoder:
> BUILD_STATIC_EXE is used for both the apache and the MIT licenses. Should I set some other flags in CMake? Is this not OK for the extras?

Interesting.  For `BUILD_STATIC_EXE` we probably want both `BUILD_SHARED_LIBS=OFF` and `BUILD_SHARED_EXTRAS_LIB=OFF`. This would be completely static and will never call `dlopen`, which is in line with the assumptions of the `-static` flag.

I've tested the `BUILD_SHARED_LIBS=OFF` and `BUILD_SHARED_EXTRAS_LIB=ON`, which I assume is the combination that `-static` doesn't like.  If we want this combination, we can simply drop the `-static` flag and it'll be partially static.

mathgeekcoder:
> BUILD_STATIC_EXE is used for both the apache and the MIT licenses. Should I set some other flags in CMake? Is this not OK for the extras?

Ignore my previous comment. For `BUILD_STATIC_EXE` we already have both `BUILD_SHARED_LIBS=OFF` and `BUILD_SHARED_EXTRAS_LIB=OFF`. 

The warning is for `HighsDynamicLibrary::load`, which won't be called, but is still compiled. I believe we can safely ignore this warning for now. 

It might become relevant depending on how we support cuda etc. in the future.

galabovaa:
Yes, I think that is right! Optionally adding HighsDynamicLibrary to cmake seems to clear the warning: https://github.com/ERGO-Code/HiGHS/commit/63be5094f437bdf37c3d019b18644c5849d82239

If you are happy with this I can merge it to this branch, but I don't think it's essential

PR #3113, opened 2026-06-29
ipo off for debug builds

galabovaa:
Switch off IPO for debug builds

Was causing trouble to the debugger, as spotted by @filikat 

Possibly occuring now because of highs_extras? 

Not super sure if we should merge this, one possible drawback is that I think it may be slowing down the valgrind tests 

https://github.com/ERGO-Code/HiGHS/actions/runs/28382771311
https://github.com/ERGO-Code/HiGHS/actions/runs/28380402678

1:13 vs 1:51 in this case, but I see one 2:07 valgrind without the changes, so I am really not sure if it would affect it or just a runner feature. 

What do you all think?


jajhall (approved):
Seems sensible

filikat (approved):
It makes sense that it would slow down a bit the tests, as there is less optimization. But the times for the valgrind workflows seem to have a large variability anyway.
It's not a problem to leave this out. I just need to remember to set it manually when using the debugger.

mathgeekcoder (approved):
Makes sense to me too.  Debug builds aren't meant to be that fast anyhow.

PR #3114, opened 2026-06-29
Pr 3110 with julia updates

galabovaa:
Updates to the julia workflow prompted by failing tests at https://github.com/ERGO-Code/HiGHS/pull/3110


highs/util/HighsDynamicLibrary.h
```diff
@@ -42,8 +42,14 @@ class HighsDynamicLibrary {
 
   // resolve and call a function directly (throws on failure)
   template <typename FuncType, typename... Args>
+#if (defined(_MSVC_LANG) && _MSVC_LANG >= 201703L) || \
```
mathgeekcoder: This is reasonable, but please let me check something first.  I believe I can rewrite this so that it doesn't need a preprocessor check.

galabovaa: Of course! I can check the conan build afterwards, having just reproduced the problem.

I just added a workflow to attempt to catch the cpp20 issues :sweat_smile:

galabovaa: just updated the workflow, it was passing on latest because of the CXX standard override from cmake, I made it more conan-like

mathgeekcoder: Looks like my change works for C++11 & C++20, so marking this as resolved.  Still need to fix some other build issues though.

mathgeekcoder:
@galabovaa I believe these latest build errors are resolved by #3112?

galabovaa:
> @galabovaa I believe these latest build errors are resolved by #3112?

Yes, apologies for merging the wrong one!

odow (commented):
There's a few different things happening here, but if the Julia tests pass I'm happy.

PR #3115, opened 2026-06-29
MinGW vs MSVC 

galabovaa:
Trying to get started on the MinGW incompatibilities brought up by @amontoison

I had some occasional failures in the julia windows test that I thought may be related. 

Since we are doing a patch release I thought I would get started on this, but I am really not sure what I am doing here or how to test it

@odow have you encountered issues in the mingw case?

### DIFF
--- a/cmake/highs-config.cmake.in
+++ b/cmake/highs-config.cmake.in
@@ -20,3 +20,29 @@ set(HIGHS_HAVE_BLAS ${_HIGHS_HAVE_BLAS})
 @CONF_ZLIB@
 
 include("${CMAKE_CURRENT_LIST_DIR}/highs-targets.cmake")
+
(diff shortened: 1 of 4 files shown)


odow:
> Trying to get started on the MinGW incompatibilities brought up by

Where were they brought up?

> @odow have you encountered issues in the mingw case?

Nope. The Julia builds run fine on Windows?

amontoison:
We got an issue in the past when we wanted to link HiGHS in Uno.
The reason is that the `.cmake` files generated by HiGHS expected MSVC as compiler on Windows.

Oscar probably never encountered this problem because he never built a JLL with `HiGHS_jll.jl` as a dependency.
The dependent package also needs to rely on a CMake build system (which is the case for Uno).

galabovaa:
Is there any chance you could give us an example to reproduce the issue?

galabovaa:
closing for now.

PR #3119, opened 2026-06-30
fix: wrong comment in postsolve basis

Wronskode:
The comments in the `case HighsBasisStatus::kUpper` branch of
`HighsPostsolveStack::SingletonRow::undo` incorrectly referred to a
"tightened lower bound" and swapped the row bound origin, even though
this branch only handles the case where the column's *upper* bound was
tightened.


jajhall (approved):
OK

PR #3122, opened 2026-06-30
Fixed issue with OpenBLAS not recognized when using BLAS_LIBRARIES 

mathgeekcoder:
Changed how `HIPO_USES_OPENBLAS` and `HIPO_USES_APPLEBLAS` are configured, so that it works correctly when using `BLAS_LIBRARIES`.

Note that when using `BLAS_LIBRARIES`, the version number is not known unless you manually set `OpenBLAS_VERSON="x.x.x"`.


galabovaa (approved):
Looks great!

PR #3124, opened 2026-07-01
Parallel early termination

Opt-Mucca:
This PR adds a deterministic stopping criteria during parallel MIP, so that some workers may terminate early in the case where another worker has already proven global optimality / infeasibility.

This should only affect cases where (1) a worker finds a primal solution where the existing global dual bound is within tolerances (2) a worker concludes that the problem is infeasible. In my test set it saves 5% of time on 3% of instances, and has no overhead + no effect on the solution path of everything else. I don't see a reason not to add this.

The biggest "change" here would be adding a `std::atomic` to `HighsMipSolverData`, and when checking limits doing a `std::memory_order_relaxed` load. 

Edit: The "improvement" on LP iterations and nodes is much higher and might make this PR look good, but it's simply cheating by discarding the information from all other workers (necessary for determinism)

### DIFF
--- a/highs/mip/HighsMipSolver.cpp
+++ b/highs/mip/HighsMipSolver.cpp
@@ -949,19 +949,27 @@ restart:
         }
       }
       assert(early_terminated_worker != -1);
-      // Clear all buffered solve information from workers that are not
-      // the earliest to terminate
-      for (const HighsInt i : search_indices) {
-        if (i == early_terminated_worker) continue;
-        HighsMipWorker& worker = mipdata_->workers[i];
-        worker.nodequeue.clear();
-        worker.search_ptr_->resetStatistics();
-        worker.resetHeurStats();
-        worker.resetSepaStats();
-        worker.solutions_.clear();
+      if (early_terminated_worker == -1) {
+        mipdata_->worker_lp_iterations_stop.store(
+            std::numeric_limits<int64_t>::max(), std::memory_order_relaxed);
+        for (const HighsInt i : search_indices) {
+          mipdata_->workers[i].early_termination = false;
+        }
+      } else {
+        // Clear all buffered solve information from workers that are not
+        // the earliest to terminate
+        for (const HighsInt i : search_indices) {
+          if (i == early_terminated_worker) continue;
+          HighsMipWorker& worker = mipdata_->workers[i];
+          worker.nodequeue.clear();
+          worker.search_ptr_->resetStatistics();
+          worker.resetHeurStats();
+          worker.resetSepaStats();
+          worker.solutions_.clear();
+        }
+        search_indices.clear();
+        search_indices.emplace_back(early_terminated_worker);
       }
-      search_indices.clear();
-      search_indices.emplace_back(early_terminated_worker);
     }
 
     // Sync statistics, check infeasibility, and flush nodes from worker queues
--- a/highs/mip/HighsMipSolver.cpp
+++ b/highs/mip/HighsMipSolver.cpp
@@ -681,6 +681,13 @@ restart:
     return mipdata_->workers[i].getGlobalDomain().infeasible() || pruned;
   };
 
+  auto assignEarlyTermination = [&](HighsMipWorker& worker) {
+    if (worker.getGlobalDomain().infeasible() ||
+        mipdata_->lower_bound > worker.optimality_limit) {
+      mipdata_->updateWorkerEarlyTermination(worker);
+    }
+  };
+
   auto separateAndStoreBasis = [&](HighsInt i) -> bool {
     HighsMipWorker& worker = mipdata_->workers[i];
     if (options_mip_->mip_allow_cut_separation_at_nodes) {
@@ -699,6 +706,7 @@ restart:
                                         ? worker.nodequeue
                                         : mipdata_->nodequeue;
       worker.search_ptr_->openNodesToQueue(globalqueue);
+      assignEarlyTermination(worker);
       return true;
     }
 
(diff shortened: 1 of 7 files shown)


highs/mip/HighsMipSolver.cpp
```diff
@@ -681,6 +681,13 @@ void HighsMipSolver::run() {
     return mipdata_->workers[i].getGlobalDomain().infeasible() || pruned;
   };
 
+  auto assignEarlyTermination = [&](HighsMipWorker& worker) {
+    if (worker.getGlobalDomain().infeasible() ||
+        mipdata_->lower_bound > worker.optimality_limit) {
```
fwesselm: Could you use `getOptimalityLimit()` here?

highs/mip/HighsMipSolver.cpp
```diff
@@ -928,9 +934,47 @@ void HighsMipSolver::run() {
 
     root_node = false;
 
+    // Check if one worker sent an early global termination signal
+    const int64_t early_termination_lp_iterations =
+        mipdata_->worker_lp_iterations_stop.load(std::memory_order_relaxed);
```
fwesselm, line 939: Noob question: If I understand correctly, you are using the number of LP iterations to identify the worker that terminated. Is there a different way to do this? Some kind of ID?

highs/mip/HighsMipSolver.cpp
```diff
@@ -957,13 +1001,15 @@ void HighsMipSolver::run() {
       break;
     }
 
-    mipdata_->updateLowerBound(std::min(
-        mipdata_->upper_bound, mipdata_->nodequeue.getBestLowerBound()));
-
     syncSolutions();
 
+    if (early_terminated_worker == -1) {
```
fwesselm, line 1006: Is it intended that this now done after `syncSolutions()`?

highs/mip/HighsMipSolverData.h
```diff
@@ -262,6 +264,17 @@ struct HighsMipSolverData {
     return (parallel_lock && hasMultipleWorkers());
   }
 
+  void updateWorkerEarlyTermination(HighsMipWorker& worker) {
+    int64_t current = worker_lp_iterations_stop.load(std::memory_order_relaxed);
+    const int64_t candidate = worker.search_ptr_->lpiterations;
+    while (candidate < current &&
```
fwesselm, line 270: What is this loop doing?

fwesselm (approved):
Looks good, @Opt-Mucca. Just some minor comments/questions from me.

highs/mip/HighsMipSolver.cpp
Opt-Mucca: I can. I'll move `getOptimalityLimit` from `HighsSearch` to `HighsMipWorker`! That makes it nicer to call in other places (like in this example)

highs/mip/HighsMipSolver.cpp
Opt-Mucca: After some of the early prototypes of the parallel MIP solver I've tried to avoid any form of ID tracking. I don't think it's needed here, and I always ended up constrained by the design at some future point. (A counter example is the cutpools.....) 
Currently the ordering would be a pareto ordering of `(lp_iters, i)`, where `i` is the index in `mipdata_.workers`. That should always be deterministic.

highs/mip/HighsMipSolver.cpp
Opt-Mucca: Yes, but now I've completely forgotten why I did this..... Do you see any consequences?

highs/mip/HighsMipSolverData.h
Opt-Mucca: This was a beautiful piece of wizardry that I recently learnt! (It's called compare-and-swap or something similar) Let's imagine we've got multiple workers all terminating, e.g., all have identifies infeasibility. We want to make sure the minimum `lp_iters` is stored from all three, and we want to avoid having to self-program any locks. In the loop, we stop if either `candidate >= current` or that `compare_exchange_weak` return `true`.

`compare_exchange_weak` will return `true` if the value is exchanged, i.e., `worker_lp_iterations_stop` is updated with the new value. It will return `false` if the value of `worker_lp_iterations_stop` has changed compared to what it loaded in `current`.

So the loop terminates if:
1. `candidate >= current`, i.e., we can't decrease the current best `worker_lp_itertations_stop`
2. `worker_lp_iterations_stop` gets updated successfully (in which case it had to be smaller)

The design essentially allows us to not add any fancy mutex stuff, and there's no risk of deadlock because one of the workers will always successfully update `worker_lp_iterations_stop` if multiple workers are trying. Then the next ones will just re-try, where one of them is then guaranteed to be successful (provided the value continues to decrease otherwise they fail and exit the loop)

Opt-Mucca:
@fwesselm ready for round 2 of review

highs/mip/HighsMipSolver.cpp
fwesselm: No, I just wanted to make sure that this was not changed accidentally.

highs/mip/HighsMipSolverData.h
fwesselm: Nice!

highs/mip/HighsMipSolver.cpp
```diff
@@ -928,9 +934,47 @@ void HighsMipSolver::run() {
...
+        for (const HighsInt i : search_indices) {
+          mipdata_->workers[i].early_termination = false;
+        }
+      } else {
+        // Clear all buffered solve information from workers that are not
+        // the earliest to terminate
+        for (const HighsInt i : search_indices) {
+          if (i == early_terminated_worker) continue;
+          HighsMipWorker& worker = mipdata_->workers[i];
+          worker.nodequeue.clear();
```
fwesselm, line 964: This looks familiar. If it is repeated somewhere, maybe it's worth adding a method to clear this.

fwesselm:
> @fwesselm ready for round 2 of review

Thanks, @Opt-Mucca. Just one more comment, and I think it's good to go.

highs/mip/HighsMipSolver.cpp
Opt-Mucca: I'd avoid that for now. Similar code is sprinkled everywhere, but it's always calling different `sync` functions and is missing some components depending on when it's called. Any central function would have so many arguments that it'd be confusing currently.

PR #3131, opened 2026-07-03
Misc

jschueller:


jajhall (commented):
In the first case, I expect that `multi_nFinish` is necessarily positive, otherwise `iRows[multi_nFinish - 1] ` would always have been an error. Do you have evidence of `multi_nFinish` being zero?

In the second case, I'd have thought that ` h_pivot_value[i];` is necessarily nonzero, so your safeguard is unnecessary. Do you have evidence of ` h_pivot_value[i];` being zero?

jschueller:
hello, I have no evidence of actual errors, but it may be safer that way

jajhall:
> hello, I have no evidence of actual errors, but it may be safer that way

The test for zero is unnecessary, since the value is a simplex pivot that is logically nonzero when chosen. Hence the conditional branch in the loop is a performance hit (in principle).

jschueller:
ok, nevermind

jajhall:
> ok, nevermind

The first case does look odd - the sort of thing that AI might well pick up - so I'll look into it.

PR #3134, opened 2026-07-06
Remove assert for empty workers

Opt-Mucca:


odow (commented):
I don’t have the ability to review this. As a high level comment: tests, or it didn’t happen. I would love to better understand how we can more robustly test highs, instead of running a few large scale instances and declaring success.

jajhall:
> I don’t have the ability to review this. As a high level comment: tests, or it didn’t happen. I would love to better understand how we can more robustly test highs, instead of running a few large scale instances and declaring success.

Although the need to test HiGHS more thoroughly on LPs, and to test presolve rules on curated special cases has been exposed in recent weeks - and is being addressed, the testing of HiGHS has always been rather more than "running a few large scale instances and declaring success." @fwesselm is performing regression testing on about 850 MIPs.

Issue #3118, opened 2026-06-30
Setting a "solution" for an infeasible problem can make solver return an incorrect status


BenChampion:
If the user sets a "solution" for a problem that presolve determines infeasible, the solver can return the incorrect status `Optimal`.

jajhall:
Please create a `fix-3118` branch with a unit test that exposes this, and I can look into it if you want

BenChampion:
> Please create a `fix-3118` branch with a unit test that exposes this, and I can look into it if you want

@jajhall Done. I know this example has bad scaling, but it exhibits the issue and it was the simplest one I could find that did.

[Branch fix-3118](https://github.com/ERGO-Code/HiGHS/tree/fix-3118)

odow:
I actually use this behavior in my tolerance tutorial: https://jump.dev/JuMP.jl/stable/tutorials/getting_started/tolerances/#Contradictory-results

jajhall:
Closed by #3135

PR #3135, opened 2026-07-06
Fix 3118

Opt-Mucca:
This fixes issue #3118 

General concept: Let's assume a model is infeasible due to something like integrality arguments, but that we have access to a solution that is feasible within tolerances. We can decide that everything but the exact solution that we have is infeasible (a completely valid deduction given we wanted to claim the instance was infeasible), and therefore the solution is optimal (clearly not true mathematically, but is us playing around with tolerances).

### DIFF
--- a/check/TestMipSolver.cpp
+++ b/check/TestMipSolver.cpp
@@ -1483,3 +1483,39 @@ TEST_CASE("issue-2975", "[highs_test_mip_solver]") {
 
   highs.resetGlobalScheduler(true);
 }
+
+TEST_CASE("issue-3118", "[highs_test_mip_solver]") {
+  //   min  a + b
+  //   s.t. a + 1e7*b = 1
+  //        1e7*a + b = 1
+  //        a, b binary
+  //   Initial "solution" a = 1e-7, b = 1e-7
+  Highs highs;
+  highs.setOptionValue("output_flag", dev_run);
+
+  HighsInt a = 0;
+  HighsInt b = 1;
+  HighsLp lp;
+  lp.num_col_ = 2;
+  lp.num_row_ = 2;
+  lp.col_lower_ = {0., 0.};
+  lp.col_upper_ = {1., 1.};
+  lp.col_cost_ = {1., 1.};
+  lp.integrality_ = {HighsVarType::kInteger, HighsVarType::kInteger};
+  lp.row_lower_ = {1., 1.};
+  lp.row_upper_ = {1., 1.};
+  lp.a_matrix_.format_ = MatrixFormat::kRowwise;
+  lp.a_matrix_.start_ = {0, 2, 4};
+  lp.a_matrix_.index_ = {a, b, a, b};
+  lp.a_matrix_.value_ = {1., 1e7, 1e7, 1};
+  highs.passModel(lp);
+
+  std::vector<HighsInt> solution_index = {0, 1};
+  std::vector<double> solution_values = {1e-7, 1e-7};
+  highs.setSolution(2, solution_index.data(), solution_values.data());
+
+  highs.run();
+  REQUIRE(highs.getModelStatus() == HighsModelStatus::kInfeasible);
+
+  highs.resetGlobalScheduler(true);
+}
--- a/check/TestMipSolver.cpp
+++ b/check/TestMipSolver.cpp
@@ -1515,7 +1515,8 @@ TEST_CASE("issue-3118", "[highs_test_mip_solver]") {
   highs.setSolution(2, solution_index.data(), solution_values.data());
 
   highs.run();
-  REQUIRE(highs.getModelStatus() == HighsModelStatus::kInfeasible);
+  REQUIRE(highs.getModelStatus() == HighsModelStatus::kOptimal);
+  REQUIRE(std::abs(2e-7 - highs.getInfo().objective_function_value) < 1e-9);
 
   highs.resetGlobalScheduler(true);
 }
--- a/check/TestMipSolver.cpp
+++ b/check/TestMipSolver.cpp
@@ -1486,16 +1486,20 @@ TEST_CASE("issue-2975", "[highs_test_mip_solver]") {
 
 TEST_CASE("issue-3118", "[highs_test_mip_solver]") {
   const double M = 1e10;
-  //   min    a +   b
-  //   s.t.   a + M*b = 1
-  //        M*a +   b = 1
-  //          a, b binary
-  //   Initial "solution" a = b = 1/M
+  //   min    x +   y
+  //   s.t.   x + M*y = 1
+  //        M*x +   y = 1
+  //          x, y binary
+  //   Initial "solution" x = y = 1/M
+  //
+  // This problem is found infeasible in presolve, but the point x = y
+  // = 1/M is integer feasible to within the tolerances, so is
+  // acccepted as an optimal solution to the problem
   Highs highs;
-  // highs.setOptionValue("output_flag", dev_run);
+  highs.setOptionValue("output_flag", dev_run);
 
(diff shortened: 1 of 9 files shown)


check/TestMipSolver.cpp
```diff
@@ -1483,3 +1483,122 @@ TEST_CASE("issue-2975", "[highs_test_mip_solver]") {
...
+
+  highs.resetGlobalScheduler(true);
+}
+
+TEST_CASE("issue-3118a", "[highs_test_mip_solver]") {
+  const double M = 1e7;
+  //   max f = x
+  //   s.t.    x - M*y  = -1
+  //           x       <=  b
+  //           x, y integer in {0, 2}x{0, 1}
```
BenChampion:
```suggestion
  //           x, y integer in [0, 2]x[0, 1]
```

check/TestMipSolver.cpp
```diff
@@ -1483,3 +1483,122 @@ TEST_CASE("issue-2975", "[highs_test_mip_solver]") {
...
   highs.resetGlobalScheduler(true);
 }
+
+TEST_CASE("issue-3118", "[highs_test_mip_solver]") {
+  const double M = 1e10;
+  //   min    a +   b
+  //   s.t.   a + M*b = 1
+  //        M*a +   b = 1
+  //          a, b binary
+  //   Initial "solution" a = b = 1/M
```
BenChampion: Can we make the point of this test case clearer (perhaps expand the comment as in the test `issue-3118a` below)?

check/TestMipSolver.cpp
```diff
@@ -1483,3 +1483,122 @@ TEST_CASE("issue-2975", "[highs_test_mip_solver]") {
...
+  Highs highs;
+  // highs.setOptionValue("output_flag", dev_run);
```
BenChampion: Uncomment or remove

check/TestMipSolver.cpp
```diff
@@ -1483,3 +1483,122 @@ TEST_CASE("issue-2975", "[highs_test_mip_solver]") {
...
+  lp.col_cost_ = {1, 0};
+  lp.integrality_ = {HighsVarType::kInteger, HighsVarType::kInteger};
+  lp.row_lower_ = {-1, -kHighsInf};
+  lp.row_upper_ = {-1, 0};
+  lp.a_matrix_.format_ = MatrixFormat::kRowwise;
+  lp.a_matrix_.start_ = {0, 2, 3};
+  lp.a_matrix_.index_ = {x, y, x};
+  lp.a_matrix_.value_ = {1., -M, 1};
+
+  for (HighsInt k = 0; k < 3; k++) {
```
BenChampion: The `k == 2` case just seems to be a dupe of `k == 1`. Perhaps I'm missing something?

highs/Highs.h
```diff
@@ -266,8 +266,11 @@ class Highs {
 
   /**
    * @brief Assess the validity, integrality and feasibility of the
-   * current primal solution. Of value after calling
-   * Highs::readSolution
+   * current primal solution. Row values are computed and checked
+   * against what's in Highs::solution_.row_value and, if the
+   * differences exceed a tolernace, valid returns false.  If any of
```
BenChampion:
```suggestion
   * differences exceed a tolerance, valid returns false.  If any of

BenChampion (commented):
Just getting back up to speed on the various communications around this. It seems the conclusion is that the only thing to fix from #3118 is the incorrect bounds. When we have a feasible solution but identify infeasibility in presolve, the idea is that we identify `kOptimal` because we won't be able to find a better feasible solution. This is just one of those solver grey areas.

My only other comment is that since this may be counter-intuitive for some users, perhaps we want to add some specific logging for this case?

check/TestMipSolver.cpp
jajhall: Done!

check/TestMipSolver.cpp
jajhall: Done!

check/TestMipSolver.cpp
jajhall: Needed to be uncommented so that the unit test runs silently

check/TestMipSolver.cpp
jajhall: Yes, there was once a third case, so upper limit of 3 needs to be 2

highs/Highs.h
jajhall: Corrected!

Opt-Mucca:
@BenChampion The other change would be that the `presolve: infeasible` line now no longer happens if there's a feasible solution in storage. That removes the counter-intuitive logging (from my perspective)

BenChampion:
> @BenChampion The other change would be that the `presolve: infeasible` line now no longer happens if there's a feasible solution in storage. That removes the counter-intuitive logging (from my perspective)

It seems like we still don't have anything that would warn the user of potential strange results with such a model. (We don't, for example, warn about the large matrix coefficients.)

BenChampion:
> It seems like we still don't have anything that would warn the user of potential strange results with such a model. (We don't, for example, warn about the large matrix coefficients.)

Whether or nor we print the presolve infeasibility message, the fact that presolve infeasibility conflicts with the existence of a feasible solution is a hint that something numerically odd is going on.

jajhall:
Dilemma: although Gurobi gives no such insight, I think that we could be more helpful in indicating that infeasible in presolve means that there is probably no exact integer solution to the problem, but that a possibly optimal integer solution is known to within the tolerances and return the model status "unknown"

The last point comes from Gurobi finding the better near-integer solution for 3118a with b=1

BenChampion:
> Whether or nor we print the presolve infeasibility message, the fact that presolve infeasibility conflicts with the existence of a feasible solution is a hint that something numerically odd is going on.

(And presolve infeasibility being inconsistent with solution feasibility may identify different cases to say, matrix coefficient range.)

BenChampion:
> Dilemma: although Gurobi gives no such insight, I think that we could be more helpful in indicating that infeasible in presolve means that there is probably no exact integer solution to the problem, but that a possibly optimal integer solution is known to within the tolerances and return the model status "unknown"
> 
> The last point comes from Gurobi finding the better near-integer solution for 3118a with b=1

Yes, something along these lines.

jajhall:
> > @BenChampion The other change would be that the `presolve: infeasible` line now no longer happens if there's a feasible solution in storage. That removes the counter-intuitive logging (from my perspective)
> 
> It seems like we still don't have anything that would warn the user of potential strange results with such a model. (We don't, for example, warn about the large matrix coefficients.)

Good point, unlike Gurobi, we only warn about small and large values in the costs and bounds - the values that can be modified by setting the user scale options

Opt-Mucca:
> Whether or nor we print the presolve infeasibility message, the fact that presolve infeasibility conflicts with the existence of a feasible solution is a hint that something numerically odd is going on.

I don't think this is always the case (even if it's probably most of the cases). I think we'd further open this can of worms by adding any extra logging.

> Dilemma: although Gurobi gives no such insight, I think that we could be more helpful in indicating that infeasible in presolve means that there is probably no exact integer solution to the problem, but that a possibly optimal integer solution is known to within the tolerances and return the model status "unknown"

I would not be in favour of this. It is further opening the can of worms, and creating a more confusing problem from my perspective. We constantly risk missing optimal solutions within tolerances, e.g., when we branch and remove all wiggle room for that column in the subtree (binary column is fixed in all LPs that follow), including the heuristics that run there. I don't see why we need to change the model status if presolve does the same thing.

jajhall (approved):
Although interesting, I think we've all spent at least as long as can be justified on this!

PR #2493, opened 2025-08-13
Reset global scheduler in C and Cpp examples

galabovaa:
Closes https://github.com/ERGO-Code/HiGHS/issues/2403

Added valgrind tests to CI

PR #3136, opened 2026-07-07
Update Julia CI job to use libblastrampoline and remove --parallel=false

odow:
An attempt to reproduce the segfaults in https://github.com/jump-dev/MultiObjectiveAlgorithms.jl/pull/211, where we can additionally make changes to HiGHS.

### DIFF
--- a/.github/julia/build_tarballs.jl
+++ b/.github/julia/build_tarballs.jl
@@ -12,9 +12,6 @@ version = VersionNumber(ENV["HIGHS_RELEASE"])
 sources = [GitSource(ENV["HIGHS_URL"], ENV["HIGHS_COMMIT"])]
 
 script = raw"""
-export BUILD_SHARED="ON"
-export BUILD_STATIC="OFF"
-
 cd $WORKSPACE/srcdir/HiGHS
 
 # Remove system CMake to use the jll version
@@ -23,19 +20,21 @@ apk del cmake
 rm -rf build
 mkdir build
 
-# Do fully static build only on Windows
-if [[ "${BUILD_SHARED}" == "OFF" ]] && [[ "${target}" == *-mingw* ]]; then
-    export CXXFLAGS="-static"
+if [[ "${target}" == *-mingw* ]]; then
+    LBT=blastrampoline-5
+else
+    LBT=blastrampoline
 fi
 
 cmake -S . -B build \
     -DCMAKE_INSTALL_PREFIX=${prefix} \
     -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} \
     -DCMAKE_BUILD_TYPE=Release \
-    -DBUILD_SHARED_LIBS=${BUILD_SHARED} \
-    -DZLIB_USE_STATIC_LIBS=${BUILD_STATIC} \
-    -DHIPO=ON -DBUILD_SHARED_EXTRAS_LIB=OFF \
-    -DBLAS_LIBRARIES="${libdir}/libopenblas.${dlext}"
+    -DBUILD_SHARED_LIBS=ON \
+    -DHIPO=ON \
+    -DBUILD_SHARED_EXTRAS_LIB=OFF \
+    -DBLA_VENDOR=blastrampoline \
+    -DBLAS_LIBRARIES=\"${LBT}\"
 
 if [[ "${target}" == *-linux-* ]]; then
     make -C build -j ${nproc}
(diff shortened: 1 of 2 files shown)


.github/julia/build_tarballs.jl
```diff
@@ -75,6 +67,6 @@ build_tarballs(
     platforms,
     products,
     dependencies;
-    preferred_gcc_version = v"11",
-    julia_compat = "1.6",
+    preferred_gcc_version = v"6",
```
odow:
```suggestion
    preferred_gcc_version = v"11",
```

odow:
@filikat this branch now uses HiGHS#latest to run the Julia tests, as well as the tests from https://github.com/jump-dev/MultiObjectiveAlgorithms.jl/pull/211 (not that I can reproduce the segfault there though).

image

I wonder if it's something like this:

1. We start one HiGHS instance
2. We run, do a bunch of stuff
3. We start a new HiGHS instance
4. We run, do a bunch of stuff
5. Sometime during (4), the GC runs and collects the first HiGHS instance, including calling `Highs_destroy`.

filikat:
This is totally possible. In the C api, `Highs_destroy` not only deletes the Highs instance, but also kills the parallel scheduler. The scheduler relies on static variables, so there is only one of it, regardless of how many Highs instances are created. If you create two Highs instances, destroy the first and still use the second, this may cause issues because the parallel scheduler is terminated while the second instance is still using it.

I don't think that this should happen. The destructor of the static instance of `ExecutorHandle` already kills the scheduler at process termination, so technically there is no need to shut it down explicitly. I wonder if calling shutdown explicitly is done to solve some of the issues that we were having on Windows. Maybe @galabovaa @jajhall have some insights into this.
Maybe we should have a `Highs_finalise` function in the C api that does not depend on any instance and simply kills the scheduler, so that it is not done by `Highs_destroy`.

Tomorrow I will check if any of the unit tests creates this exact situation of two separate instances where one is destroyed before the other. If not, I will write such a test and try to verify if the issue is there. 
One easy way to check is also to comment the line `Highs::resetGlobalScheduler(true);` in `Highs_destroy`.

odow:
> One easy way to check is also to comment the line Highs::resetGlobalScheduler(true); in Highs_destroy.

Ooooooo.

https://github.com/ERGO-Code/HiGHS/blob/04024d701f79feb8e2f18bc3df0dffc04ef05088/highs/interfaces/highs_c_api.cpp#L176-L178

Based on this scary warning, it seems highly likely that's the cause:

https://github.com/ERGO-Code/HiGHS/blob/04024d701f79feb8e2f18bc3df0dffc04ef05088/highs/interfaces/highs_c_api.h#L2494-L2495

I may be able to fix this in JuMP, because we explicitly mark that the GC can run during a call to `Highs_run`:

https://github.com/jump-dev/HiGHS.jl/blob/5f3e3b7a6af3b988186f64934416f0241d5eefb2/src/MOI_wrapper.jl#L2295-L2297

odow:
Added by https://github.com/ERGO-Code/HiGHS/pull/2493

There is already `oid Highs_resetGlobalScheduler(const HighsInt blocking);` in the C API.

The C API should never call `resetGlobalScheduler` because you can never do so in a thread-safe way. This must be called only by the user.

odow:
So I think this was probably the cause of the failures in https://github.com/ERGO-Code/HiGHS/pull/2998. I should have taken a deeper look, instead of just working around the cause.

odow:
I'm fairly happy that `resetGlobalScheduler` was the culprit. Once https://github.com/ERGO-Code/HiGHS/pull/3141 is merged I'll update this PR to be just the changes for the Julia CI tests.

filikat:
> The C API should never call `resetGlobalScheduler` because you can never do so in a thread-safe way. This must be called only by the user.

I agree. Indeed, the cpp functions that run Highs never call `resetGlobalScheduler` explicitly. It is always called by the user after Highs has finished running.

jajhall:
Am I right in thinking that nothing is lost by removing 'resetGlobalScheduler' from the C API method that destroys HiGHS.

All we need is a C API method that calls 'resetGlobalScheduler' so that we, and users, can achieve this from C. Something we'll need for CI tests?

filikat:
There already is such a function, `Highs_resetGlobalScheduler`.
We need to make sure that users are aware that this should be called by them when they are finished using all Highs instances.

filikat:
> Am I right in thinking that nothing is lost by removing 'resetGlobalScheduler' from the C API method that destroys HiGHS.

I think so. It was added to avoid some valgrind errors when running the minimal C example, but the same can be achieved by calling `Highs_resetGlobalScheduler` explicitly in `call_highs_from_c_minimal.c`, like Oscar did.

> Something we'll need for CI tests?

The tests in #3141 are passing, so I don't think there's much else to fix.

odow:
Let's merge #3141 into `latest`, and then I'll re-target this PR to merge into `latest` after that.

PR #3137, opened 2026-07-07
Processor affinity

fwesselm:
This is a first attempt to handle [processor affinity](https://en.wikipedia.org/wiki/Processor_affinity). I am not an expert here, so any feedback is appreciated.

### DIFF
--- a/check/TestHighsParallel.cpp
+++ b/check/TestHighsParallel.cpp
@@ -294,6 +294,12 @@ static void nestedTask() {
   tg.spawn([]() { loseTime(); });
 }
 
+TEST_CASE("AvailableConcurrency", "[parallel]") {
+  unsigned int cores = highs::parallel::available_concurrency();
+  REQUIRE(cores > 0);
+  REQUIRE(cores <= std::thread::hardware_concurrency());
+}
+
 TEST_CASE("CancelNestedTasks", "[parallel]") {
   highs::parallel::initialize_scheduler();
 
--- a/check/TestHighsParallel.cpp
+++ b/check/TestHighsParallel.cpp
@@ -305,8 +305,8 @@ static void nestedTask() {
 }
 
 TEST_CASE("AvailableCoreCount", "[parallel]") {
-  unsigned int cpus = highs::parallel::available_core_count();
-  REQUIRE(cpus > 0);
+  unsigned int cores = highs::parallel::available_core_count();
+  REQUIRE(cores > 0);
 }
 
 #if defined(_WIN32) || defined(_WIN64)
@@ -321,9 +321,9 @@ TEST_CASE("AffinityReducedCoreCount", "[parallel]") {
   SetProcessAffinityMask(GetCurrentProcess(), 1);
 
   // available_core_count should now report 1
-  unsigned int cpus = highs::parallel::available_core_count();
-  REQUIRE(cpus == 1);
-  REQUIRE(initial_threads >= static_cast<int>(cpus));
+  unsigned int cores = highs::parallel::available_core_count();
+  REQUIRE(cores == 1);
+  REQUIRE(initial_threads >= static_cast<int>(cores));
 
   // Restore original affinity
   SetProcessAffinityMask(GetCurrentProcess(), original_mask);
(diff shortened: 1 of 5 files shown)


Opt-Mucca (approved):
I like the change! I'd defer to @mathgeekcoder for a more informed opinion though

mathgeekcoder:
I think this is great! 

My only feedback would be to be careful about interchanging the terminology vthreads with cores. Specifically, `unsigned int available_cores = highs::parallel::available_concurrency();`

The `CPU_COUNT`, `hardware_concurrency` return vthreads, not cores. However, we'd like to create a thread pool with the number of cores (not vthreads)!

On that note, I believe the current (and previous) code assumes 2 vthreads per core, however this is not always true; and gets more complicated when running in the cloud.

Ideally, we'd like to get the count of physical cores that we are mapped and masked by affinity, and use this as our number of threads in the thread pool.

fwesselm:
> I think this is great!
> 
> My only feedback would be to be careful about interchanging the terminology vthreads with cores. Specifically, `unsigned int available_cores = highs::parallel::available_concurrency();`
> 
> The `CPU_COUNT`, `hardware_concurrency` return vthreads, not cores. However, we'd like to create a thread pool with the number of cores (not vthreads)!
> 
> On that note, I believe the current (and previous) code assumes 2 vthreads per core, however this is not always true; and gets more complicated when running in the cloud.
> 
> Ideally, we'd like to get the count of physical cores that we are mapped and masked by affinity, and use this as our number of threads in the thread pool.

Thank you, @mathgeekcoder!

I tried to change the code according to your comments. I hope I did not introduce new bugs, it would be great if you could check.

mathgeekcoder:
Oh wow!  I said it, but didn't expect you to implement it :) 

I think this is a great start.  If I'm being _really_ pedantic:
* I believe the `cpu` terminology is the same as `vthread` (rather than `core`), i.e., `cpu` can mean physical or logical processor, whereas `core` is only physical processor
* We might want to use fallback if *any* of the topology files don't exist (rather than if `physical_cores.empty()`)
* Technically `CPU_SETSIZE = 1024` and larger systems exist. Can use `sysconf(_SC_NPROCESSORS_CONF)` and `CPU_ALLOC` to get and set correct size
* I think you can use `GetLogicalProcessorInformationEx` when >64 CPUs on Windows

But as I said... pedantic.  Honestly, it's fine as is!

Ultimately, it would be fantastic to also exploit numa nodes and shared cache (e.g., L3).  Once @galabovaa gets back we should check if it's okay to use something like [hwloc](https://www.open-mpi.org/projects/hwloc/) to get this information. hwloc also has cross-platform code to count the cores.

For example, it'd be great to configure work-stealing to prefer cores within same numa node, and perhaps use logical processors (i.e., vthread, if available) to sync/exchange information.

fwesselm:
> Oh wow! I said it, but didn't expect you to implement it :)
> 
> I think this is a great start. If I'm being _really_ pedantic:
> 
> * I believe the `cpu` terminology is the same as `vthread` (rather than `core`), i.e., `cpu` can mean physical or logical processor, whereas `core` is only physical processor
> * We might want to use fallback if _any_ of the topology files don't exist (rather than if `physical_cores.empty()`)
> * Technically `CPU_SETSIZE = 1024` and larger systems exist. Can use `sysconf(_SC_NPROCESSORS_CONF)` and `CPU_ALLOC` to get and set correct size
> * I think you can use `GetLogicalProcessorInformationEx` when >64 CPUs on Windows
> 
> But as I said... pedantic. Honestly, it's fine as is!
> 
> Ultimately, it would be fantastic to also exploit numa nodes and shared cache (e.g., L3). Once @galabovaa gets back we should check if it's okay to use something like [hwloc](https://www.open-mpi.org/projects/hwloc/) to get this information. hwloc also has cross-platform code to count the cores.
> 
> For example, it'd be great to configure work-stealing to prefer cores within same numa node, and perhaps use logical processors (i.e., vthread, if available) to sync/exchange information.

Thanks again, @mathgeekcoder! I tried to address your comments. For Windows, I am trying to follow `hwloc`'s approach.

Issue #2866, opened 2026-02-21
Large column bounds lead to MIP presolve and solve errors


jajhall:
Problems [2866a.mps](https://github.com/user-attachments/files/25459179/2866a.mps.txt) and [2866b.mps](https://github.com/user-attachments/files/25459180/2866b.mps.txt) have column bounds of 1e7 - with the maximum solution value being 3. The correct optimal objective value is 417. 

HiGHS gives 420 for 2866a, and 451 for 2866b

Using [2866.cpp](https://github.com/user-attachments/files/25459236/2866.cpp), if the bounds are pushed up to infinity, HiGHS get the right objective. As they are reduced by powers of 10 from 1e7, at 1e3 (1e4) HiGHS gets the right objective with (without) presolve. 

They are small problems with "release" solve times below 1s.

Hence @fwesselm and @Opt-Mucca, there are issues in both presolve and solve.

Opt-Mucca:
I can look into this later in the week!

zeddybot:
Chiming in here because this looks like a **hard-crash manifestation of the same "large column bounds" problem**: HiGHS 1.15.1 **segfaults in MIP presolve** on a small, feasible model whose only unusual feature is large (`1e12`) column bounds.

**Versions:** HiGHS 1.15.1 (git 04024d7), highspy 1.15.1, Linux x86_64. Also reproduces with the HiGHS bundled in Google OR-Tools 9.15 (HiGHS 1.12.0).

## Summary

`Highs::run()` with **default options segfaults during MIP presolve**. The model is a small, well-scaled MILP (matrix coefficients in `[7e-1, 2e+1]`), valid and feasible — SCIP solves it to `Optimal`, objective `0.0`. It has some large continuous column bounds (`1e12`); HiGHS warns "excessively large column bounds", then crashes. `presolve=off` avoids the crash but then returns `kInfeasible` (also wrong), so neither path is usable.

## Reproduction

Repro MPS (62 cols, 104 rows, 12 integer / 4 binary): https://gist.github.com/zeddybot/54ba51c0fec844f5eab369f7a49eb6e5

```python
import highspy
h = highspy.Highs()
h.readModel("highs-mip-presolve-segfault.mps")
h.run()   # SIGSEGV, pure defaults
```

[...]

PR #3139, opened 2026-07-07
Fix 2866

Opt-Mucca:
This fixes the crash from the instance in the comments of #2866 

Explanation: While iterating over `singletonRows`, presolving one of the singletons reduces multiple (probably identical ones) to empty rows. Despite them being reduced `singletonRows` is not updated (is waiting to be cleared). When the size 0 row is then processed, they are not yet deleted and `rowPresolve` will be called on them during `removeRowSingletons`. This will then result in a nested call of `removeRowSingletons`. This is all fine, until we're back at the original loop, where the counter is at something non-zero, e.g., 2, and the loop is continuing because `2 != 0` (the new size of `singletonRows` from the nested call.

Fix: I've just changed the `!=` to a `<`.

@zeddybot I'm assuming you're a bot, and a lot of the information wasn't necessary, but the instance was helpful and the issue was clearly explained.

### DIFF
--- a/highs/presolve/HPresolve.cpp
+++ b/highs/presolve/HPresolve.cpp
@@ -6944,7 +6944,7 @@ void HPresolve::removeFixedCol(HighsInt col, double fixval) {
 
 HPresolve::Result HPresolve::removeRowSingletons(
     HighsPostsolveStack& postsolve_stack) {
-  for (size_t i = 0; i != singletonRows.size(); ++i) {
+  for (size_t i = 0; i < singletonRows.size(); ++i) {
     HighsInt row = singletonRows[i];
     if (rowDeleted[row] || rowsize[row] > 1) continue;
     // row presolve will delegate to rowSingleton() if the row size is 1
--- a/highs/presolve/HPresolve.cpp
+++ b/highs/presolve/HPresolve.cpp
@@ -6959,7 +6959,7 @@ HPresolve::Result HPresolve::removeRowSingletons(
 
 HPresolve::Result HPresolve::presolveColSingletons(
     HighsPostsolveStack& postsolve_stack) {
-  for (size_t i = 0; i != singletonColumns.size(); ++i) {
+  for (size_t i = 0; i < singletonColumns.size(); ++i) {
     HighsInt col = singletonColumns[i];
     if (colDeleted[col]) continue;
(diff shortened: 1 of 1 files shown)


highs/presolve/HPresolve.cpp
```diff
@@ -6944,7 +6944,7 @@ void HPresolve::removeFixedCol(HighsInt col, double fixval) {
 
 HPresolve::Result HPresolve::removeRowSingletons(
     HighsPostsolveStack& postsolve_stack) {
-  for (size_t i = 0; i != singletonRows.size(); ++i) {
+  for (size_t i = 0; i < singletonRows.size(); ++i) {
```
fwesselm: What happens if `singletonRows` grows? Maybe use `while` loop? Should we implement the same for `singletonColumns`?

```
size_t i = 0;
while (i < singletonRows.size()) {
...
i++;
}
```

fwesselm (approved):
@Opt-Mucca, thanks for fixing this!

highs/presolve/HPresolve.cpp
Opt-Mucca: Doesn't the for loop already cover the case where `singletonRows` grows?

Yes, we should add the same to ` singletonColumns`. I can't see why this wouldn't be more safe

fwesselm: I am not sure since a compiler may try to be clever and pre-compute `size()`. But, I did not double-check in this case.

Opt-Mucca: Puhhhhhhhh. I'd never assume a compiler could do that. Regardless, I can change the code. Is quick.

fwesselm: Sorry, Mark, on second thought, I think your initial fix was better. I forgot that this is a vector (variable size), so I think it is unlikely that the compiler does optimize away `size()`. Apologies!

Opt-Mucca: I've removed the commit. Should now just be the `!=` that has been changed

fwesselm: @Opt-Mucca, looks good, sorry for the confusion!

Opt-Mucca: All good! Will merge this now

PR #3141, opened 2026-07-08
Remove resetGlobalScheduler calls from C API

odow:
It is not thread-safe for the C API to call resetGlobalScheduler because another thread may be in the middle of a Highs_ call.

Extracted from https://github.com/ERGO-Code/HiGHS/pull/3136

This should fix the random segfaults we are seeing with the Julia tests.

The Julia tests are a prime candidate for encountering this bug because they spawn many instances of HiGHS, `Highs_destroy` is called when the instances are garbage collected, and Julia's garbage collector runs in a separate thread.

### DIFF
--- a/.github/julia/build_tarballs.jl
+++ b/.github/julia/build_tarballs.jl
@@ -12,9 +12,6 @@ version = VersionNumber(ENV["HIGHS_RELEASE"])
 sources = [GitSource(ENV["HIGHS_URL"], ENV["HIGHS_COMMIT"])]
 
 script = raw"""
-export BUILD_SHARED="ON"
-export BUILD_STATIC="OFF"
-
 cd $WORKSPACE/srcdir/HiGHS
 
 # Remove system CMake to use the jll version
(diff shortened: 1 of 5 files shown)


filikat:
I managed to reproduce segfaults and other weird behaviour with a unit test that runs two separate Highs instances in separate threads. The bad behaviour disappears using this branch.

tobiass-sdl:
Sorry for highjacking this but if resetGlobalScheduler is not thread safe how do I change the number of threads used by a HiGHS instance?
Use case: running HiGHS inside a server that depending on user input either wants to put all cpu cores on one MIP or solve a bunch of small MIP in parallel with 1 thread each - the later is useful if you have a problem that is decomposable by domain knowledge...
If you set the option twice within the same thread, you get 
"ERROR:   Option 'threads' is set to 3 but global scheduler has already been initialized to use 2 threads. The previous scheduler instance can be destroyed by calling Highs::resetGlobalScheduler()."
Given this PR I guess the recommendation is not safe?

filikat:
It is not thread safe to call `resetGlobalScheduler` while a Highs instance is running.
If you generate multiple instances and run them in parallel, you should not call `resetGlobalScheduler` while another thread is calling `Highs_run`, or any other function that runs Highs.
It is completely safe to call `resetGlobalScheduler` while no Highs instance is running (e.g. before starting running any Highs instance, or after all instances have finished).

It is not possible to have multiple Highs instances running with different values of the `thread` option within the same process, as the parallel scheduler is shared by all instances.

tobiass-sdl:
> It is not thread safe to call `resetGlobalScheduler` while a Highs instance is running. If you generate multiple instances and run them in parallel, you should not call `resetGlobalScheduler` while another thread is calling `Highs_run`, or any other function that runs Highs. It is completely safe to call `resetGlobalScheduler` before starting running any Highs instance, or after all instances have finished.

If you're using HiGHS within a web server that serves more then one user you always have to assume that another HiGHS instance is running in parallel - or you keep track yourself how many highs instances are running and only set the thread call if this count is 0 and if not you can't set the thread count for the instance you are about to start... Which means that in my use case it's probably best disable the parallel mip solver and go for threads=1 all the time.

filikat:
You can set `parallel=off` for a given instance and most parallelism will be switched off. There is only a limited number of things that can run in parallel when `parallel=off` (see https://ergo-code.github.io/HiGHS/stable/parallel/).
If you're looking to fine tune how many threads each Highs instance is allowed to use within the same process, I don't think this is possible using the current Highs parallel scheduler.

odow:
Ooops. Failing Julia tests were because I jumped the gun to make a change in HiGHS. I've reverted it for now: https://github.com/jump-dev/HiGHS.jl/commit/c8a94c633a7be47c1c91601749c9b76d9fdd61bd and restarted the tests.

(I thought it would just silently ignore the option, not error.)

odow:
Okay, all tests passing. This is ready to be merged.

filikat (approved):

filikat:
Shall we merge this now, or is it better to wait for @galabovaa to have a look?

odow:
The original commits of this PR could be merged without @galabovaa as straightforward bug fix. She might want some visibility on the CI changes to JuMP, which is why I originally kept the PRs separate. Let me know if you want to split them back out.

Otherwise, I'd support just merging this. It fixes the segfaulting of tests.

filikat:
The CI changes seem all related to JuMP, so as long as you're happy with them and she gets notified about these changes, I think it's fine to merge now.

PR #3143, opened 2026-07-08
fix compiler warnings from version 1.15.1

svigerske:
Remove unused local variable `record_int64`.

Fix warning that `numSetPackingRows` may not be initialized in a symmetry class.


Opt-Mucca (approved):
Don't see any problems with the the changes! The Julia test fails seem unrelated, and the other two failures seem to be HiPO build related.

jajhall (approved):
Thanks. I'm fine with these. If we need an `Int64` record in the future, it's soon restored

PR #3145, opened 2026-07-09
Test an LP with HiPO in the Julia CI job

odow:
v1.15 broke HiPO support in HiGHS, even though I thought we were testing it: https://github.com/jump-dev/HiGHS.jl/pull/347

It might be fixed in `latest`, so testing this. I need to figure out a more robust way to check if HiPO is actually compiled in the binary.


odow:
I've added proper test to HiGHS.jl: https://github.com/jump-dev/HiGHS.jl/pull/347

Issue #3074, opened 2026-06-17
Presolve worsens solution and hangs


igofunke:
Hi team!

I bring you an instance of a model that seems to give a bit of trouble to HiGHS with presolver (Gap worsens and the solver hangs). When I switch off the presolver it seems to work fine.

I would appreciate it if someone could have a look and advise us further, as turning off the presolver is a workaround but not the ideal solution. Thanks in advance!

To reproduce:
1. Download model  [model.lp.zip](https://github.com/user-attachments/files/29049403/model.lp.zip).
2. Execute with presolve on. Note how Gap increases instead of decreasing. HiGHS seems to hang after 189.7s (execution was in reality stopped after 30+ min but logging stopped earlier)

```
% highs ./model.lp
Running HiGHS 1.14.0 (git hash: n/a): Copyright (c) 2026 under MIT licence terms
Command line parsed using CLI11 2.5.0: Copyright (c) 2017-2025 University of Cincinnati
MIP model has 17658 rows; 17228 cols; 50676 nonzeros; 10988 integer variables (10988 binary)
Coefficient ranges:
  Matrix  [8e-02, 3e+04]
  Cost    [9e+03, 6e+10]
  Bound   [1e+00, 1e+00]
  RHS     [1e+00, 1e+00]
WARNING: Problem has some excessively large costs
WARNING:    Consider scaling the objective by 1e-5, or setting the user_objective_scale option to -16
Presolving model
16890 rows, 16448 cols, 48340 nonzeros 0s
14381 rows, 14909 cols, 48639 nonzeros 0s
Presolve reductions: rows 14381(-3277); columns 14909(-2319); nonzeros 48639(-2037) 

Solving MIP model with:
   14381 rows
   14909 cols (10393 binary, 0 integer, 103 implied int., 4413 continuous, 0 domain fixed)
   48639 nonzeros

Src: B => Branching; C => Central rounding; F => Feasibility pump; H => Heuristic;
     I => Shifting; J => Feasibility jump; L => Sub-MIP; P => Empty MIP; R => Randomized rounding;
     S => Solve LP; T => Evaluate node; U => Unbounded; X => User solution; Y => HiGHS solution;
     Z => ZI Round; l => Trivial lower; p => Trivial point; u => Trivial upper; z => Trivial zero

        Nodes      |    B&B Tree     |            Objective Bounds              |  Dynamic Constraints |       Work      
[...]

PR #3146, opened 2026-07-09
Fix assertion failure

fwesselm:
Fix assertion failure encountered when running the reproduction example from #3074 (by using `HighsCDouble`):

> Assertion failed: (std::abs(model->col_cost_[col]) <= std::max(options->dual_feasibility_tolerance, kHighsTiny * std::abs(static_cast<double>(objscale)))), function substitute, file HPresolve.cpp, line 2970.

Note that this does not resolve the slow convergence with presolve.


fwesselm:
Need to investigate another assertion failure. Closing this one.

PR #3147, opened 2026-07-10
Bazel fixup

Mizux:
in `BUILD.bazel` explicitly link to `ld` and `pthread` on linux to avoid error on debian11 runner from BCR (Bazel Central Registry).

related to:
* https://github.com/bazelbuild/bazel-central-registry/pull/9635


galabovaa (approved):
LGTM, thank you @Mizux

PR #3149, opened 2026-07-15
Fix edge-case assertion failures in presolve and domain propagation code

fwesselm:
This is motivated by #3074:
1. Use `HighsCDouble` precision for objective accumulations (avoid summing up doubles).
2. `HighsDomain::resolveLinearGeq` and `HighsDomain::resolveLinearLeq`: Added a guard condition before dropping a bound change from the explanation to verify that the explanation remains valid.


Opt-Mucca (approved):
Would the guard alone have been sufficient to fix the bug for the issue?

I'm fine with upgrading a lot of the accumulations to `HighsCDouble`, but it would be nice to know what price we're paying for this extra level of safety (assuming it's well south of `0.1%` and everything's fine)

Edit: Maybe wasn't clear: Happy with the PR! Just want a sanity check we haven't added meaningful overhead (I can try and do this next week)

PR #3152, opened 2026-07-17
Claude/hipo parallel solve sketch

IainDM:


jajhall:
Thanks for this. We can't accept AI-generated code, or tune any of our solvers for specific problem classes, but please would you see if anything useful can be distilled from this investigation, @filikat, focusing initially on [this](https://github.com/ERGO-Code/HiGHS/pull/3152/changes/12d0756c31756d299211cb2af9b182c755a099a8).

filikat:
Thanks, I will have a look.

filikat:
There seems to be three contributions in this PR:
- Exposing HiPO internal parameters as HiGHS options. I don't have any issue with that, apart from polluting the list of HiGHS options with lots of very technical stuff. If users want to play with that it's fine for me. However, the current approach is not acceptable as it relies on a static object to communicate the options to the factorisation. If there are multiple concurrent HiGHS instances trying to set different options, they would overwrite each other and it would not be thread safe. In `latest` I already added a struct containing factorisation options, for other reasons. I can use this struct to pass other options too and this would be thread safe.
- Speed-up the serial solve phase by removing memory allocations and pivoting swaps. This is a great contribution. I noticed that there was some time that was unaccounted for in my profiling and I was already working on removing unnecessary swaps. Memory allocation is another low hanging fruit that will definitely provide a benefit. However, the next release will contain a second factorisation that is much simpler than the current multifrontal one and that works very well for the PyPSA problems (and any other problem that has similar properties). My benchmark on medium sized problems show large speed-ups on both factorisation and solve times. The changes that are proposed here are still useful in general, but they will not have an impact on the running time of energy models in the future, as another factorisation will be used.
- Parallel solve. I didn't read these changes yet, as they are long and complicated, but getting parallel solve to work has been on my list of things to do for a long time. It didn't have high priority for me, so I didn't put too much time into it, but I am definitely curious to have a look at these changes. As I mentioned before though, this will not have an impact on running times of energy models, as a separate factorisation is already much faster (which may explain why you didn't observe a benefit, the PyPSA problems are simply not suited for this kind of parallelised factorisation or solve, and a simpler approach works much better).

As @jajhall mentioned, we do not accept AI generated PRs, especially to the core code of the solvers. I will implement the additional options and the improvements to the serial solve phase, as these should be quick to do. I will have a look at the parallel solve implementation and check if it's worth having it, and in that case I will include these ideas in what I already have (but this will take some time).

jajhall:
Thanks @IainDM

IainDM:
@jajhall @filikat apologies! I meant to just make this change to my own fork, not to the main ERGO-Code repository. I spend my time running PyPSA, so these changes were really for my own use at least for now.

IainDM:
obviously if anything here is any use feel free to use it, but it wasn't intended for the main repo

jajhall:
> @jajhall @filikat apologies! I meant to just make this change to my own fork, not to the main ERGO-Code repository. I spend my time running PyPSA, so these changes were really for my own use at least for now.
> 

No problem. As you see, although much of what you suggested is in the pipeline for @filikat, it was still working looking at.

PR #3153, opened 2026-07-17
Hessian oracle

jajhall:
This allows the QP active set solve (QpAsm) and HiPO to operate using an oracle for the Hessian $$Q$$. 

- For HiPO, the oracle is used to generate the explicit Hessian
- For QpAsm, the oracle will provide all data required during the course of the algorithm.

This oracle must be able to compute $$Qx$$, but there is greater efficiency if it can supply a column of the Hessian. 

Also incorporates a method for users to check that their Hessian oracle is correct

C API methods have been written, and tested in `check/testCAPI.c`

Review of QpAsm identified the dense Cholesky code in `highs/qpsolver/feasibility_bounded.hpp` as being wholly incorrect, requiring a complete re-write. 

Documentation is in C API header file, and `check/TestQpOracle.cpp`. Formal documentation will follow


highs/lp_data/HighsOptions.h
```diff
@@ -586,6 +587,7 @@ struct HighsOptionsStruct {
         pdlp_cupdlpc_restart_method(0),
         pdlp_step_size_strategy(0),
         pdlp_optimality_tolerance(0.0),
+        test_qp_oracle(false),
```
filikat, line 590: Is this options supposed to be here? Or is it just for development?

jajhall: In the unit tests I wanted to have it on and off, but I'll change it to a const bool internally. Thanks

highs/interfaces/highs_c_api.cpp
```diff
@@ -287,6 +287,19 @@ HighsInt Highs_passHessian(void* highs, const HighsInt dim,
       ->passHessian(dim, num_nz, format, start, index, value);
 }
 
+HighsInt Highs_passHessianOracle(void* highs, const HighsInt dim,
+                                 HighsCHessianFunctionType oracleCall,
+                                 void* oracle_data) {
+  auto status =
+      static_cast<Highs*>(highs)->passHessian(dim, oracleCall, oracle_data);
```
filikat, line 294: Here you are passing the `HighsCHessianFunctionType` as second argument to `passHessian` (hence casting to `HighsHessianFunctionType`), while there is an unused fourth argument which takes exactly `HighsCHessianFunctionType`.

filikat: BTW, I think std::function can be constructed from a C-style function pointer if the signatures are exactly the same, so this may be totally fine and the fourth argument of passHessian may actually not be needed at all.

highs/lp_data/HStruct.h
```diff
@@ -246,4 +247,38 @@ struct HighsUserScaleData {
...
+  // or shift_
+  HighsInt dim_ = 0;
+  double multiplier_ = 1.0;  // Minimize
+  double shift_ = 0.0;       // No regularization
+  HighsHessianFunctionType call_ = nullptr;
+  void* data_ = nullptr;
+  void clear();
+  bool isValid() const { return hasProductCall(); }
+  bool hasProductCall() const;
+  void formFromOracle();
```
filikat, line 268: `formFromOracle` is unused, as you are doing it as part of HighsHessian and not HessianOracle

filikat:
Looks great, apart from minor things and documentation to be added later.
I didn't check the changes in qpsolver though.

highs/interfaces/highs_c_api.cpp
jajhall: Thanks, I just followed what was done for the call-backs, but then the C-style function pointer had to be retrofitted.

filikat: Oh I see, then it's fine if it's consistent with the rest of the callbacks.

jajhall:
> Looks great, apart from minor things and documentation to be added later. I didn't check the changes in qpsolver though.

Thanks. I've been checking that I get the same iteration counts and objective with/without the oracle using the 80 or so easiest Maros-Meszaros problems, so I'm fairly confident!

filikat (approved):

PR #3155, opened 2026-07-18
Feature/grpc server impl

awakeningofthetrailblazer:
This fork adds a gRPC server subproject (server/) that exposes HiGHS LP/MIP solving as a remote service with CPU/GPU adaptive builds (PDLP on GPU when CUDA available, IPM on CPU otherwise).


jajhall:
Thanks for this, we will consider accepting this PR. In the meantime, please remove all Chinese text and replace it with English.

jajhall:
Thanks, but please make this available as an external interface

jajhall:
We're happy to add your interface to those that we document

PR #3159, opened 2026-07-21
Simplified factorisation for HiPO

filikat:
- Add up-looking factorisation. This is a simplified sparse factorisation that works well when the matrix is very sparse. 
- Add criterion to select whether to use the up-looking factorisation or the multi-frontal one.
- Add option `hipo_factor` to force the choice of a specific factorisation.
- Add possibility of switching between the two factorisations. This is used mainly if the up-looking factorisation (which is slightly less robust for now) fails, to continue iterating using the multi-frontal one
- If HiPO fails, it resets to the best iteration that it found before switching to IPX or switching factorisation.
- Changed the way HiPO manages the status of the solve process, keeping track of errors separately from the status.
- Removed unnecessary memory allocations and swaps from the triangular solves, as suggested by #3152.
- Changed the internal time profiling.
- Removed a lot of useless comments, unused functions, and improved general readability of the code.

### DIFF
--- a/cmake/sources.cmake
+++ b/cmake/sources.cmake
@@ -267,7 +267,8 @@ set(hipo_sources
     ipm/hipo/ipm/Model.cpp
     ipm/hipo/ipm/PreProcess.cpp
     ipm/hipo/ipm/Refine.cpp
-    ipm/hipo/ipm/Solver.cpp)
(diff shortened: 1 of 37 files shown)


jajhall (approved):
Fine by me

PR #3162, opened 2026-07-22
fix message when solver option value is not ok

svigerske:
Quotes were misplaces in the message
```
ERROR:   Value "cupdlp" for MIP LP solver option ("mip_lp_solver") is not one of hipo", ""choose", "simplex", "ipm" or "ipx"
```
This PR adds the missing quote in front of `hipo` and removes the duplicate quote in front of `choose`.


jajhall:
Thanks @svigerske !

Issue #3040, opened 2026-05-21
Perform initial sweep of simple presolve rules without forming presolve data structures


jajhall:
`HPresolve::initialRowAndColPresolve` can account for a significant proportion of the time when presolving models with a lot of fixed columns. This is due to the overhead of creating and then updating the dynamic storage of the model during presolve. 

As an example, a model from a client has 35631 / 108047 fixed columns, and removing them accounts for 1/3 of the time in presolve. They also constitute a non-trivial overhead when creating the initial dynamic storage of the model

Some non-trivial code changes are required, since the initial fixes will have to be accommodated in the postsolve stack before creating the copy of the model used in presolve

jajhall:
Closed by #3167

PR #3167, opened 2026-07-25
Introduces presolve "light"

jajhall:
Presolve "light" performs an initial sweep removing empty and fixed columns, and empty, singleton and redundant rows. For some practical problems, this removes significant numbers of rows and columns, and avoids the overhead of forming the general presolve data structure for the whole original model. 

After the initial sweep, presolve light avoids applying the following expensive presolve rules
- DependentEquations
- DependentFreeCols
- Aggregator
- ParallelRowsAndCols
- Sparsify
- Probing
- Enumeration
- DualFixing
- ColStuffing


jajhall:
This closes #3040

highs/presolve/HighsPostsolveStack.h
```diff
@@ -593,9 +612,13 @@ class HighsPostsolveStack {
     }
     std::copy(valuesNew.cbegin(), valuesNew.cend(), values.begin());
 #else
+    for (size_t i = index.size(); i < static_cast<size_t>(origSize); i++)
+      values[i] = zero;
```
fwesselm, line 616: values[i] = T{};

But, this would initialise the basis with `HighsBasisStatus::kLower`. I am not sure if this is problematic.

highs/presolve/HPresolve.cpp
```diff
@@ -545,11 +652,20 @@ void HPresolve::markChangedCol(HighsInt col) {
   }
 }
 
-double HPresolve::getMaxAbsColVal(HighsInt col) const {
+double HPresolve::getMaxAbsColVal(HighsInt col,
+                                  const bool initial_sweep) const {
```
fwesselm: Maybe add a member variable bool `bool in_initial_sweep_`?

Thereby, the signatures of `HPresolve::getMaxAbsColVal` etc. would not need to be changed.

highs/presolve/HighsPostsolveStack.cpp
```diff
@@ -34,27 +34,34 @@ void HighsPostsolveStack::initializeIndexMaps(HighsInt numRow,
...
+}
 
-  // now compress the column array
-  HighsInt numCol = origColIndex.size();
-  for (size_t i = 0; i != newColIndex.size(); ++i) {
-    if (newColIndex[i] == -1)
-      --numCol;
+void HighsPostsolveStack::compressIndexMap(
+    const std::vector<HighsInt>& newIndex, std::vector<HighsInt>& origIndex) {
+  // loop over entries, decrease entry counter for deleted entriess
```
fwesselm: entries?

highs/presolve/HighsPostsolveStack.h
```diff
@@ -634,32 +657,86 @@ class HighsPostsolveStack {
...
     }
 
+    // Initialise to illegal values so that initial values are logged
+    double report_col_value = kHighsInf;
+    double report_col_dual = kHighsInf;
+    HighsBasisStatus report_col_status = HighsBasisStatus::kNonbasic;
+    size_t check_reduction = -35044;
+
+    auto solutionLogging = [&](const std::string& message) {
+      return;
```
fwesselm: The lambda directly returns?

fwesselm (approved):
I can't really comment on the changes to HFactor etc.

The presolve changes look reasonable, although I am wondering whether it would be better to have a member variable (like `bool in_initial_sweep_`), thereby avoiding changes to function signatures.

highs/presolve/HPresolve.cpp
jajhall: Yes, this is more elegant. The explicit change was helpful in development

highs/presolve/HighsPostsolveStack.cpp
jajhall: Corrected

highs/presolve/HighsPostsolveStack.h
jajhall: It's not problematic.

The method scatters "values" for the presolved problem into their position in the original problem. The values not assigned in this process are for the rows and columns removed. Their values are assigned during postsolve, without reference to what was stored in the particular location - so it doesn't matter what was stored in that location. Indeed, previously, what was in the particular location was the value for a row or column in the presolved problem. I found this very confusing when developing. 

The "it doesn't matter" argument is also reflected in the use of  `std::numeric_limits<T>::signaling_NaN()` to fill these locations when `DEBUG_EXTRA` is assigned. This NaN assignment ensures that a check can be made (with `bool containsNanOrInf`) that a value is assigned in postsolve for all the (double) locations corresponding to rows and columns removed.

In the case of `T` = `HighsBasisStatus`, ideally `zero` would yield `HighsBasisStatus::kNonbasic` as that's never assigned by the simplex solver or crossover, so would be clearly "not assigned". However, this would require the definition of `HighsBasisStatus` to be changed so `kNonbasic = 1` and, even if the values of `kHighsBasisStatus*` were changed in `interfaces/highs_c_api.h` there are sure to be users who are assuming that `HighsBasisStatus::kLower` casts to 0, who would then be messed up, so I'll live with initialising to `HighsBasisStatus::kLower`.

highs/presolve/HighsPostsolveStack.h
jajhall: Yes, I see that it's unnecessary. That said, I've commented out the debugging lambdas and the calls to them to improve code coverage! I think they are worth keeping for future debugging

jajhall:
One consequence of this PR (@fwesselm and @Opt-Mucca) is that the initial sweep is always performed, so the behaviour of any solve involving presolve (for a problem with empty columns, fixed columns, empty rows, singleton rows or redundant rows) will change. Clearly this will change MIP performance (in particular) but the initial sweep is of very low cost if it doesn't yield reductions, and it's more efficient than achieving these reductions in "standard" presolve. Hence I think that
- We have to accept the volatility
- We don't need the ability to switch off the initial sweep

Issue #3150, opened 2026-07-15
HiGHS 1.15.1: C++ API crashes with "IOT instruction (core dumped)", CLI executable succeeds


sducomman:
Hello,

For some MIP model I have some IOT instruction (core dumped) when parallel is activated.
But when I run this MIP model in command line, Highs solved it.

Example with the file last_model.lp : 
With this code : 
``` 
#include <Highs.h>

int main(int argc, char **argv) {

    std::string filename = "last_model.lp";
    Highs highs;
    highs.readModel(filename);
    highs.setOptionValue("parallel", "on");
    highs.run();
}
``` 
The output is 
```
Running HiGHS 1.15.1 (git hash: 04024d7): Copyright (c) 2026 under MIT licence terms
Includes third-party software components, see THIRD_PARTY_NOTICES.md for full details
MIP last_model has 21180 rows; 22063 cols; 170585 nonzeros; 22063 integer variables (8839 binary)
Coefficient ranges:
  Matrix  [1e+00, 1e+06]
  Cost    [1e+04, 1e+04]
  Bound   [1e+00, 6e+03]
  RHS     [1e+00, 3e+03]
Presolving model
20760 rows, 20614 cols, 166957 nonzeros 0s
19099 rows, 19230 cols, 154917 nonzeros 0s
Presolve reductions: rows 19099(-2081); columns 19230(-2833); nonzeros 154917(-15668)
Objective function is integral with scale 0.0001

Solving MIP model with:
   19099 rows
   19230 cols (7937 binary, 11293 integer, 0 implied int., 0 continuous, 0 domain fixed)
   154917 nonzeros
   Thread count 11 (of 22 threads). Using 19 max workers. Parallel search on

Src: B => Branching; C => Central rounding; F => Feasibility pump; H => Heuristic;
     I => Shifting; J => Feasibility jump; L => Sub-MIP; P => Empty MIP; R => Randomized rounding;
     S => Solve LP; T => Evaluate node; U => Unbounded; X => User solution; Y => HiGHS solution;
     Z => ZI Round; l => Trivial lower; p => Trivial point; u => Trivial upper; z => Trivial zero

        Nodes      |    B&B Tree     |            Objective Bounds              |  Dynamic Constraints |       Work
Src  Proc. InQueue |  Leaves   Expl. | BestBound       BestSol              Gap |   Cuts   InLp Confl. | LpIters     Time

[...]

PR #3168, opened 2026-07-26
Wrapped the calls to `Highs::calledOptimizeModel()`, simplex, `callSolveQp` and MIP solver in try-catch blocks

### LINKED ISSUE #2924: Wrap active set QP and simplex in try-catch blocks
The calls to IPX, HiPO and pdlp are wrapped in try-catch blocks (in HighsSolve.cpp), so that HiGHS can catch any exception thrown. But the QP solver and simplex do not have a try-catch block, so an exception can escape, leading to the abnormal termination of HiGHS (and potential untold damage in downstream interfaces).

jajhall:
Closes #2924 and #3150

Although the `Highs::calledOptimizeModel()` try-catch block is sufficient, the lower level try-catch blocks enable the particular solver causing the exception to be identified.


filikat:
Looks good! However, if there is an exception thrown by the parallel scheduler (there shouldn't be, I hope I fixed all the bugs, but who knows), this will not be caught here, because it does not inherit from `std::exception`. To catch it, we would need a specific catch block with `catch (const HighsTask::Interrupt&)`, or just `catch(...)`, but then we cannot know what exception it is.

jajhall:
> Looks good! However, if there is an exception thrown by the parallel scheduler (there shouldn't be, I hope I fixed all the bugs, but who knows), this will not be caught here, because it does not inherit from `std::exception`. To catch it, we would need a specific catch block with `catch (const HighsTask::Interrupt&)`, or just `catch(...)`, but then we cannot know what exception it is.

Thanks. I've done quite a bit of refactoring of the interface to the solves. Now 

- In the Highs class methods, there's minimal code relating to solvers - just the MIP solver for `Highs::presolve()`. The Highs class methods just identify which optimization problems are to be solved, using the incumbent model or its presolved instance
- The QP solvers are fed a `HighsQpSolverObject` instance in the same way as the LP solvers, rather than drawing directly on data members of the `Highs` class. This will make it simple to solve presolved QP problems.
- The active set QP solver has a wrapper
- All solvers are chosen in `HighsSolve`, and called in a `try... catch block`, catching `const std::exception& exception` and `const HighsTask::Interrupt&`, with the LP and QP solvers using the same generic #define method. 
- All solver-specific code is in the corresponding wrapper - except for the MIP solver, where it's still in `HighsSolve`

`Highs.cpp` is now significantly smaller

All this means that there are a lot of diffs - mainly due to moving code. The most important changes are in `HighsSolve.cpp`

filikat:
Looks good to me

filikat (approved):

PR #3172, opened 2026-07-26
Fix #3170: discard a redundant variable bound instead of substituting it as if tight

### LINKED ISSUE #3170: MIP: invalid cut from a redundant variable bound gives a suboptimal solution reported as Optimal at gap 0
### Summary

`HighsTransformedLp::transform()` ignores the `redundant` flag returned by
`HighsImplications::cleanupVub()` / `cleanupVlb()`. A redundant variable bound is deliberately
returned *untightened* for the caller to discard; `transform()` substitutes it anyway. Cut
generation is then given a false upper bound on the transformed variable, and the cut that comes
out is not globally valid. It can remove the optimum, after which the search closes and reports the
incumbent as `Optimal` at a zero gap, with no warning and no violated row.

Two small models that current `master/latest` solves incorrectly are attached as `3170-1.mps.txt` (39
rows, 30 columns, 137 nonzeros, 10 binaries) and `3170-2.mps.txt` (40 rows, 24 columns, 125
nonzeros, 8 binaries) — rename them to `.mps`.

### Reproducing

[3170-1.mps.txt](https://github.com/user-attachments/files/30398388/3170-1.mps.txt) fails with a single option and nothing else:

```
$ highs --presolve off 3170-1.mps

Solving report
  Model             3170-1
  Status            Optimal
  Primal bound      -21262719.4302
  Dual bound        -21262719.4302
  Solution status   feasible
                    -21262719.4302 (objective)
                    0 (bound viol.)
                    0 (int. viol.)
                    0 (row viol.)
  Nodes             1
```

The optimum is `-21446579.364727`, so this is suboptimal by **183,860 — a relative error of
0.86%, about 86x the default relative gap tolerance**. It is not a tolerance artefact.

[3170-2.mps.txt](https://github.com/user-attachments/files/30398387/3170-2.mps.txt) needs the option set below (with default options it is solved correctly), and then
reports `-516096644.644` against an optimum of `-516307714.782682` — suboptimal by **211,070**, a
relative error of 0.041%, again at `Gap 0% (tolerance: 0%)`, 1 node, zero violations:

```
threads=1, random_seed=0, parallel=off, run_crossover=on,
mip_rel_gap=0, mip_abs_gap=1e-6,
mip_feasibility_tolerance=1e-7, primal_feasibility_tolerance=1e-7, dual_feasibility_tolerance=1e-7,
all six mip_heuristic_run_* = false,
presolve=off
```

Both solve in a few hundredths of a second at 1 node. The failure is silent: the returned point is
feasible, integral and self-consistent, and the gap is certified 0.

Both optima were verified **independently of the MIP solver**, by enumerating all 2^n binary
assignments and solving the resulting LP for each — so the claim above does not rest on trusting
the branch-and-bound search:

```python
import itertools, highspy

def brute_force(path):
    h = highspy.Highs(); h.setOptionValue("output_flag", False)
    h.readModel(path)
    lp = h.getLp()
    ints = [j for j in range(lp.num_col_)
            if lp.integrality_[j] == highspy.HighsVarType.kInteger]
    best = float("inf")
    for bits in itertools.product((0.0, 1.0), repeat=len(ints)):
        g = highspy.Highs(); g.setOptionValue("output_flag", False)
        g.readModel(path)
        for j, v in zip(ints, bits):
            g.changeColBounds(j, v, v)
            g.changeColIntegrality(j, highspy.HighsVarType.kContinuous)
        g.run()
        if g.getModelStatus() == highspy.HighsModelStatus.kOptimal:
            best = min(best, g.getObjectiveValue())
    return best

print(brute_force("3170-1.mps"))   # -21446579.364727
print(brute_force("3170-2.mps"))   # -516307714.782682
```

> **On the `presolve` option.** Disabling presolve is only what makes these two models expose the
> defect — presolve itself is not doing anything wrong, and the faulty code is in cut generation,
> which runs either way. Disabling it changes the relaxation that reaches the separators, which is
> what determines whether the unsound substitution below is reached at all.

### Cause

`highs/mip/HighsTransformedLp.cpp`, `transform()`.

Cut generation substitutes bounds to move each variable into `[0, ub-lb]`. When it substitutes a
*variable* upper bound `x <= a*z + b` (`z` binary), the transformed variable is the slack
`y = a*z + b - x`, and the code passes `upper[j] = ub - lb` to the cut generator as `y`'s upper
bound. That is sound only if the variable bound is at least as tight as the simple bound, i.e.
`maxValue() <= ub`. The code checks exactly that and tries to repair it:

```cpp
if (bestVub[col].first != -1 &&
    bestVub[col].second.maxValue() > ub + mip.mipdata_->feastol) {
  bool redundant = false;
  bool infeasible = false;
  mip.mipdata_->implications.cleanupVub(col, bestVub[col].first,
                                        bestVub[col].second, ub, redundant,
                                        infeasible, false);
}
```

but the `redundant` output is discarded. `cleanupVub()` tightens the bound only when it is *not*
redundant:

```cpp
if (minub >= ub - mipsolver.mipdata_->feastol) {
  redundant = true;          // left unchanged, for the caller to discard
} else if (maxub > ub + mipsolver.mipdata_->epsilon) {
  ...                        // tightened so that maxValue() == ub
}
```

so on the `redundant` path the variable bound comes back with `maxValue()` still far above `ub`,
and the invariant `transform()` just tried to establish still fails. `cleanupVarbounds()` handles
this correctly — it collects redundant bounds and erases them — but `transform()` simply carries
on and substitutes it. `cleanupVlb()` has the identical contract and the identical caller bug.

On a larger model where I first hit this, a continuous column with global bounds
`[0.0298, 0.0683]` (so `ub-lb = 0.0385`) was substituted with

```
x <= 0.29538846115832368 * z + 0.4420490388416824       minValue 0.4421   maxValue 0.7374
```

`cleanupVub()` returned `redundant = 1` and changed nothing. The real range of the slack is
`[0, 0.7077]`, but the cut generator was told `y <= 0.0385` — 18x too small. The cut that came out
had a minimum activity over the global box of `+1.28` against a right-hand side of `-0.55`: it is
violated by *every* point in the box. Once such a cut is in the global pool every node LP is
infeasible, the tree closes at once, and whatever incumbent existed is declared optimal.

That last detail may also be worth checking against reports where a feasible MIP is returned
`Infeasible` — a globally infeasible cut added before any incumbent exists produces exactly that.

### How the state arises

`addVUB()` refuses to store an already-redundant bound, so this needs one that was valid when
stored and became redundant later:

1. probing stores a variable upper bound for `x` with `minValue = c` and `maxValue = ub(x)`
   (probing-derived bounds always have `maxValue` equal to the column's bound at derivation time);
2. `x`'s upper bound is later tightened to `<= c` — for instance because the binary gets fixed;
3. `getBestVub()` had already selected that bound for the current separation round, and
   `transform()` then takes the `redundant` path above.

Step 3 is why the binary being fixed does not save you: `getBestVub()` skips fixed binaries, but
the selection happened earlier in the round.

### Fix

A candidate fix, a regression test using the two attached models that fails without it, and the
supporting measurements are on PR #3172, so as to keep this report to the diagnosis.

### Version

master @ 04024d701f and latest @ bc30ccf1a1 (1.15.1). Also reproduces on 1.13.1; I have not tried other versions systematically.

### Related

There is a second, independent unsoundness in the same area — cut coefficients built with one bound
substitution and untransformed with another: #3171.

Both come from cut construction reading bound-substitution state back out of a shared, still-
mutating `HighsTransformedLp`, rather than the cut carrying the transformation it was built with.
They are independent in practice though: neither fix subsumes the other, and each can be reviewed
on its own.

#3173 is a third defect found on the same class of model, but in presolve rather than in cut
generation - it is decided before branch-and-bound runs and is unaffected by the fix here.

EamonHetherton:
Fixes #3170.

The first commit adds the two reproducers as a failing regression test, on its own, so the defect
is visible without any fix applied. The second commit contains the fix.

### The models

* `check/instances/3170-1.mps` — 39x30, 10 binaries. Solved incorrectly with `presolve=off` alone:
  reports `-21262719.4302` against an optimum of `-21446579.364727`, a relative error of 0.86%,
  about 86x the default relative gap tolerance, so not a tolerance artefact.
* `check/instances/3170-2.mps` — 40x24, 8 binaries. Needs the tighter option set used in the test;
  reports `-516096644.644` against an optimum of `-516307714.782682`.

Both optima were verified independently of the MIP search, by enumerating every binary assignment
and solving the resulting LP for each.

### The defect

`HighsTransformedLp::transform()` substitutes a variable upper bound `x <= a*z + b` by the slack
`y = a*z + b - x`, and passes `upper[j] = ub - lb` to the cut generator as `y`'s upper bound. That
is sound only if the variable bound is at least as tight as the simple bound, i.e.
`maxValue() <= ub`, so the code calls `cleanupVub()` to re-establish that when it does not hold.

`cleanupVub()` only tightens a variable bound that is *not* redundant. When
`minValue() >= ub - feastol` it reports `redundant = true` and deliberately leaves the bound
unchanged, for the caller to discard — which is what `cleanupVarbounds()` does. The caller here
ignored that flag and substituted the bound anyway, so the slack could range far above the
`ub - lb` it was declared to have, and the cut derived from it was not globally valid. Such a cut
can remove the optimum, after which the search closes and reports the incumbent as `Optimal` at a
zero gap.

`cleanupVlb()` has the identical contract and the identical caller bug.

### The fix

Discard the variable bound when cleanup reports it redundant, and defensively whenever the
assumption still fails afterwards, falling back to the simple bound. Nothing is lost, since a
redundant variable bound is by definition no stronger than the simple bound.

```cpp
      if (redundant ||
          bestVub[col].second.maxValue() > ub + mip.mipdata_->feastol) {
        bestVub[col].first = -1;
        ubDist[col] = simpleUbDist[col];
        boundDist[col] = std::min(lbDist[col], ubDist[col]);
      }
```

and symmetrically for the variable lower bound.

### Two things to be aware of when reviewing

**Only the variable upper bound path is exercised by the tests.** The variable lower bound half is
fixed by symmetry, since `cleanupVlb()` has the same contract and the same caller bug, but neither
attached model reaches it.

**The second disjunct is belt-and-braces.** After a successful `cleanupVub()` the invariant holds,
so `|| maxValue() > ub + feastol` should not fire; the `redundant` flag alone is what these models
hit. Happy to drop it if you would rather the change were minimal.

### Adverse effects

The concern with this change is that it discards a variable bound that would otherwise have been
used for cut generation, which could weaken cuts. Measured on `latest`:

| | result |
|---|---|
| 35 MIP models in `check/instances` | identical status, objective, node count **and LP iteration count** — every one |
| 400 generated MIPs | identical objectives, none differing |
| 9 models attached to open correctness issues | identical |
| full unit test suite | 1259332 assertions in 335 test cases, pass |

So there is no collateral change: no weakened cuts, no extra nodes, no extra iterations anywhere
the defect is not present. The fix appears to be inert unless the defect actually fires. That is
not a proof that it never activates — those 400 models come from the same generator that produces
triggers for this defect at roughly 1 in 500, so most likely none of them hit it.

One thing that might look like a regression but is not: on the larger model this was first found
on, the node count rises with the fix (7 to 18). That is the search doing its job — without the fix
an invalid cut was closing the tree early.

### Testing

* the new `issue-3170` test fails on `latest` at the first commit and passes at the second;
* both attached models return the brute-force optimum under every option set tried;

### DIFF
--- a/highs/mip/HighsTransformedLp.cpp
+++ b/highs/mip/HighsTransformedLp.cpp
@@ -195,6 +195,11 @@ bool HighsTransformedLp::transform(std::vector<double>& vals,
       mip.mipdata_->implications.cleanupVub(col, bestVub[col].first,
                                             bestVub[col].second, ub, redundant,
                                             infeasible, false);
+      if (redundant) {
+        bestVub[col].first = -1;
+        ubDist[col] = simpleUbDist[col];
+        boundDist[col] = std::min(lbDist[col], ubDist[col]);
+      }
     }
 
     // the code below uses the difference between the column upper and lower
@@ -210,6 +215,11 @@ bool HighsTransformedLp::transform(std::vector<double>& vals,
       mip.mipdata_->implications.cleanupVlb(col, bestVlb[col].first,
                                             bestVlb[col].second, lb, redundant,
                                             infeasible, false);
+      if (redundant) {
+        bestVlb[col].first = -1;
+        lbDist[col] = simpleLbDist[col];
+        boundDist[col] = std::min(lbDist[col], ubDist[col]);
+      }
     }
 
     // store the old bound type so that we can restore it if the continuous


jajhall:
Thanks for this. We don't accept AI-generated code, but I expect that @Opt-Mucca and @fwesselm  will identify what needs to be changed and fix it themselves.

EamonHetherton:
@jajhall  ok, I have two other issues I'm working through for a minimal reproduction, one is already logged as an issue with more detail to come.  Essentially I am verifying a larger model across hundreds of thousands of solves with different inputs and occasionally HiGHS was returning a noticeably sub-optimal solution and found three discrete issues causing them.  Would you rather I just report the issue as best I can or open a pull request with a demonstrated fix (albeit with AI assistance) for you to cherry pick from?

jajhall:
If you can generate a fix, then it it certainly helps. Our developers may well implement it as it is, but any adverse effects need to be considered

EamonHetherton:
Thanks, I'll continue as is then; log the issue with as much detail as I can gather and open a PR with a failing test followed by a proposed fix.  Cherry pick as you like for what goes upstream. (and i'll rebase to latest too)

Opt-Mucca (changes requested):
Nice fix! The changes I'd request:
- Remove the extra test. The instances have pretty large objective coefficient ranges, and I'd rather not introduce a test that may cause problems later down the road.
- Please remove the comments between the `if (redundant` and the tightening of the variable bounds. I don't think this needs any explanation.
- Please remove the additional `||` check for whether the max values make the variable bound redundant. This case will only trigger when `epsilon < feastol`. That is, make the `if` statement only have `if (redundant)`

Edit: I've tested locally and removing the additional `||` check still fixes the attached instance.

EamonHetherton:
Thanks.

- Removed the regression test and both instances.
- Removed the comments in that block.
- The `if` is now `if (redundant)` only, for both the variable upper and lower bound.

Some AI commentary on the changes though that are worth hearing (I will create a new issue and PR probably tomorrow for the new bug mentioned in there around HPresolve::scaleMIP...bed time shortly)

**On the dropped `||`.** Agreed it is unreachable in any sane configuration, but the reason is the other way round from the one given: the two conditions can only hold together when `feastol < epsilon`, not `epsilon < feastol`. After a successful `cleanupVub()` the bound satisfies `maxValue() == ub`, and otherwise `maxValue() <= ub + epsilon`, so `maxValue() > ub + feastol` needs `feastol < epsilon`. That is reachable in principle, since `epsilon` is `small_matrix_value` (default `1e-9`) while `mip_feasibility_tolerance` can legally be set down to `kMinimumMipTolerance` = `1e-10`. Not worth guarding against on the evidence I have, and the `redundant` flag is the actual defect either way — just flagging it so the reasoning on record is right.

**On removing the test.** Understood on the objective coefficient ranges, and I think the instinct is right: I have since traced a separate wrong-answer defect on this same class of model to `HPresolve::scaleMIP` scaling rows *down*, where the feasibility tolerance enforced on the presolved row corresponds to a violation of `feastol / |scale|` on the original row. On one model that turned a `1e-7` tolerance into `1e-5` of real violation, so solutions were discarded as having untransformed violations and the optimum was lost. Wide coefficient ranges genuinely do make these models tolerance-sensitive, so a test pinned to an exact objective on one is a fair flake risk.

The trade-off is that nothing now guards this fix against regression. If it would help, I am happy to supply a numerically tamer instance for the test suite, or leave it out entirely — your call. Both models remain attached to #3170 for anyone who wants to reproduce it.

Opt-Mucca:
@EamonHetherton Thanks for the fast response and fix! I'm opening an equivalent PR in #3177 Sorry if this does not credit you as the author of the change. In general we try not to merge AI generated code (@jajhall would have the full take if this is a problem for you). I'll see if I can somehow get GitHub to still credit you.

A proper thank you for distilling the issue into something nice.

EamonHetherton:
all good, don't need the credit, just the bugs out of my way :)

jajhall:
> @EamonHetherton Thanks for the fast response and fix! I'm opening an equivalent PR in #3177 Sorry if this does not credit you as the author of the change. In general we try not to merge AI generated code (@jajhall would have the full take if this is a problem for you). I'll see if I can somehow get GitHub to still credit you.
> 
> A proper thank you for distilling the issue into something nice.

Indeed, thanks from me. The issue with AI-generated code is that it may be valuable to say that we don't accept it, and if someone objects to their AI-generated PR being rejected, we don't want a precedent that they can point to.

Issue #3171, opened 2026-07-26
MIP: path-mixing cut is untransformed with different bound substitutions than its coefficients were built with, producing an invalid cut


EamonHetherton:
### Summary

In `HighsPathSeparator::separateLpSolution()`, a path-mixing cut is assembled from several base
rows, each transformed by its own `HighsTransformedLp::transform()` call, and mapped back by a
single `untransform()` at the end. The chosen substitution per column lives in the shared
`boundTypes` array, and `untransform()` reads whatever is in there when it runs.

A base row that is transformed and then **discarded** leaves its changes behind. If it flipped the
complementation chosen for a binary, `untransform()` resubstitutes a different variable than the
coefficients were computed against, and the resulting cut is not globally valid. Such a cut can
remove the optimum, after which the search closes and reports the incumbent as `Optimal` at a zero
gap.

### Reproducing

[3171-1.mps.txt](https://github.com/user-attachments/files/30408319/3171-1.mps.txt) (85 rows, 99 columns, 49 binaries, 300 nonzeros) is solved incorrectly under **default
options**, with no option pinning:

```
$ highs 3171-1.mps

  Status            Optimal
  Primal bound      42332.2356068
  Dual bound        42332.2356068
  Gap               0% (tolerance: 0.01%)
  Nodes             0
```

The optimum is `42215.5250005`, so this is suboptimal by **116.71 — a relative error of 0.276%,
about 28x the default relative gap tolerance**. Not a tolerance artefact, and decided at the root
in zero nodes.

The optimum is confirmed by HiGHS itself, without trusting a patched build: fixing all 49 integer
columns to the values of the better point and solving the remaining LP returns
`Optimal 42215.5250005`. So a feasible integral point 116.71 better than the reported optimum
demonstrably exists. `--presolve off` returns `42215.5250005` too, so the contradiction is visible
within one unmodified binary.

The failure is seed-dependent — it occurs on seeds 0 and 5 of the first 8 — so a regression test
needs `random_seed` pinned to 0 (which is the default).

### Cause

The bound type for a binary depends on the sign of its coefficient in the row being transformed
(`highs/mip/HighsTransformedLp.cpp`):

```cpp
} else if (vals[i] > 0) {
  boundTypes[col] = BoundType::kSimpleLb;   // y = x - lb
} else {
  boundTypes[col] = BoundType::kSimpleUb;   // y = ub - x   (complemented)
}
```

[...]

PR #3174, opened 2026-07-27
Fix #3173: bound the objective perturbation when treating a near-fixed column as fixed

### LINKED ISSUE #3173: MIP: presolve fixes a column whose bounds are equal only within tolerance without bounding the objective error, returning a suboptimal solution as Optimal
### Summary
Note: Authored in conjunction with Claude AI

`HPresolve::checkColBounds()` treats a column as fixed when its bounds are equal only *within the
feasibility tolerance*. The guard bounds the resulting perturbation of the **row activities**, but
not of the **objective**. A column can be nearly fixed in the constraints while still being worth a
great deal in the objective, and `colPresolve()` then pins it at its **lower** bound, losing up to
`|cost| * (ub - lb)`.

On the attached 40x33 model that loses 2496.5 — presolve reduces the model to empty, reports
`Presolve: Optimal`, and returns an objective **6.3% away from the true optimum**, with default
options and 0 branch-and-bound nodes.

**Provenance, in the interests of not overselling this.** I did not hit this in my own models. It
turned up in a randomly generated model, from a generator I wrote to produce minimal reproducers
for #3170 and #3171. On the two real models that motivated those two issues, this condition is
reached zero times, so this is not a defect I have observed in practice.

That said, I do not think it is only an artefact of odd generated data. The column being fixed has
a cost of `-7.29e10`, but the model's own cost range is `[1e-06, 1e+06]` — the coefficient is about
73,000x larger than anything in the input, built up by presolve's own substitutions before the
column reaches `checkColBounds()`. So reaching this state does not require an extreme cost in the
MPS file, only an awkward one plus a few substitutions.

### Reproducing

[3173-1.mps.txt](https://github.com/user-attachments/files/30396626/3173-1.mps.txt) is attached — rename it to `3173-1.mps`. 40 rows, 33 columns, 138 nonzeros,
8 binaries.

```
$ highs 3173-1.mps

Presolve reductions: rows 0(-40); columns 0(-33); nonzeros 0(-138) - Reduced to empty
Presolve: Optimal

  Status            Optimal
  Primal bound      -37333.0762329
  Dual bound        -37333.0762329
  Solution status   feasible
                    0 (bound viol.)
                    0 (int. viol.)
                    0 (row viol.)
  Nodes             0
```

The true optimum is **-39829.5816585168**. `presolve=off` returns it.

The optimum was verified independently of the MIP search, by enumerating all 2^8 binary assignments
and solving the resulting LP for each — so the claim does not rest on trusting branch-and-bound:

```python
import itertools, highspy

h = highspy.Highs(); h.setOptionValue("output_flag", False)
h.readModel("3173-1.mps")
lp = h.getLp()
ints = [j for j in range(lp.num_col_)
        if lp.integrality_[j] == highspy.HighsVarType.kInteger]

best = float("inf")
for bits in itertools.product((0.0, 1.0), repeat=len(ints)):
    g = highspy.Highs(); g.setOptionValue("output_flag", False)
    g.readModel("3173-1.mps")
    for j, v in zip(ints, bits):
        g.changeColBounds(j, v, v)
        g.changeColIntegrality(j, highspy.HighsVarType.kContinuous)
    g.run()
    if g.getModelStatus() == highspy.HighsModelStatus.kOptimal:
        best = min(best, g.getObjectiveValue())

h.run()
print("brute-force optimum:", best)                  # -39829.5816585168
print("HiGHS reports      :", h.getObjectiveValue())  # -37333.0762503876
```

Also reproduces on 1.13.1 (`-37333.076250`).

### Cause

`highs/presolve/HPresolve.cpp`, `checkColBounds()`:

```cpp
double boundDiff = model->col_upper_[col] - model->col_lower_[col];
if (boundDiff <= primal_feastol &&
    (boundDiff <= options->small_matrix_value ||
     getMaxAbsColVal(col) * boundDiff <= primal_feastol)) {
  ...
  if (isFixed != nullptr) *isFixed = true;
}
```

and `colPresolve()`, which acts on it:

```cpp
  HPRESOLVE_CHECKED_CALL(checkColBounds(col, &isFixed));
  if (isFixed) {
    postsolve_stack.removedFixedCol(col, model->col_lower_[col], ...);
    removeFixedCol(col);
    return checkLimits(postsolve_stack);
  }

  switch (colsize[col]) {
    case 0:
      return emptyCol(postsolve_stack, col);   // never reached for this column
```

The state on the attached model, at the point the column is removed:

```
lb           1.3502174835205079
ub           1.3502175177853328
boundDiff    3.426482e-08        <= mip_feasibility_tolerance (1e-7)  -> "fixed"
maxAbsColVal 0                   <- the column is empty by this stage
cost         -72859132125.478302
```

`|cost| * boundDiff = 2496.5050`, against an observed objective error of **2496.5054**.

Two things compound here:

1. the column is **empty** by this stage, so `getMaxAbsColVal(col) * boundDiff` is `0 <= 1e-7`,  
   the row-activity guard is vacuously satisfied no matter how large `boundDiff` is;
2. `checkColBounds()` runs *before* the `switch (colsize[col])` that would dispatch to
   `emptyCol()`, which exists precisely to fix an empty column at its **cost-optimal** bound. The
   tolerance check pre-empts the routine that would have got it right, and pins at the lower bound.

Dropping `mip_feasibility_tolerance` below `boundDiff`, so the column is no longer treated as
fixed, removes the error entirely:

| `mip_feasibility_tolerance` | objective |
|---|---|
| `1e-6` | `-37333.0762329` (wrong) |
| `1e-7` | `-37333.0762329` (wrong) |
| `1e-8` | `-39829.5816345` (correct) |
| `1e-9` | `-39829.5816345` (correct) |

### Version

`master` @ `04024d701f` and `latest` @ `bc30ccf1a1` (1.15.1); also reproduces on 1.13.1.

### Related

#3170 and #3171 are two invalid-cut defects in MIP cut generation, found on the same class of
model. This one is separate: it is decided entirely in presolve, before branch-and-bound runs, and
the fixes for those two leave it unchanged.

I have a branch carrying this reproducer as a failing regression test, and a candidate change, but
the appropriate threshold for the missing condition is a judgement call I would rather leave to
you, so I have kept it out of this report. It is written up on the PR instead.

EamonHetherton:
Addresses #3173.

The first commit adds the reproducer as a failing regression test, on its own, so the defect is
visible without any fix applied. The second commit contains a candidate change. They are separate
deliberately: the test is the part I am confident about, and the threshold in the fix is a
judgement call I would rather you made — see below. Taking only the first commit is a perfectly
reasonable outcome.

### Provenance

This was not encountered in a real model. It turned up in a randomly generated one, from a
generator written to produce minimal reproducers for #3170 and #3171, and the condition is reached
zero times on the two real models that motivated those. So this is a defect found by construction
rather than one observed in practice, and that is worth weighing when deciding how much churn it
justifies.

The counterweight: the column being fixed has a cost of `-7.29e10`, but the model's own cost range
is `[1e-06, 1e+06]`. The coefficient is about 73,000x larger than anything in the input, built up
by presolve's own substitutions before the column reaches `checkColBounds()`. Reaching this state
does not require an extreme cost in the MPS file, only an awkward one plus a few substitutions.

### What the defect is

`HPresolve::checkColBounds()` treats a column as fixed when its bounds are equal only within the
feasibility tolerance. The condition bounds the resulting perturbation of the row activities:

```cpp
getMaxAbsColVal(col) * boundDiff <= primal_feastol
```

but not the perturbation of the objective, and `colPresolve()` then pins the column at its lower
bound. On the model added here:

```
lb           1.3502174835205079
ub           1.3502175177853328
boundDiff    3.426482e-08     <= mip_feasibility_tolerance, so "fixed"
maxAbsColVal 0                <- empty column, the row-activity guard is vacuous
cost         -72859132125.478302
```

`|cost| * boundDiff = 2496.5050` against an observed objective error of `2496.5054`. Presolve
reduces the model to empty, reports `Presolve: Optimal`, and returns a value 6.3% away from the
optimum with default options and no branch-and-bound nodes.

The optimum was verified independently of the MIP search, by enumerating all 2^8 binary assignments
and solving the resulting LP for each.

### The candidate change

Require the objective perturbation to be negligible as well:

```cpp
  if (boundDiff <= primal_feastol &&
      (boundDiff <= options->small_matrix_value ||
       (getMaxAbsColVal(col) * boundDiff <= primal_feastol &&
        std::fabs(model->col_cost_[col]) * boundDiff <= primal_feastol))) {
```

The column then falls through to `emptyCol()`, which fixes it at the cost-optimal bound. The error
drops from 2496.5 to 2.4e-5 (6e-10 relative) and the regression test passes.

### Why the threshold is an open question

`primal_feastol` is a *feasibility* tolerance being used to bound an *objective* quantity. It is
convenient and symmetric with the row term, not principled. Alternatives, none obviously right:

* `mip_abs_gap` is dimensionally correct for a MIP, but meaningless for an LP, and users
  legitimately set it to 0, which would disable the reduction unconditionally;
* something relative to the objective magnitude is most defensible in principle, but there is no
  reliable objective scale available at that point in presolve;
* dispatching to `emptyCol()` whenever `colsize[col] == 0`, before the tolerance check, fixes the
  empty-column case specifically but not a non-empty column with a large cost and a small
  `boundDiff`, where the row-activity term is satisfied and `|cost| * boundDiff` is still large.

Detection is insensitive to the choice: the failing case exceeds any candidate threshold by about
nine orders of magnitude. The risk is in the other direction, suppressing reductions that were
doing useful work.

### How often the condition is reached, and what suppressing it costs

I instrumented the condition and counted every column reaching it with `boundDiff > 0` and the
row-activity term already satisfied.

**Small models** — 35 MIP models from `check/instances`, 400 generated MIPs, and this reproducer,
435 in total. Only **11 columns** reach it at all. Their `|cost| * boundDiff`:

```
min     2.502295e-14
median  5.889381e-06
max     2.496505e+03      <- this bug
```

`primal_feastol` blocks 10 of those 11. Across all 435 models, status, objective, node count and LP
iteration count are identical with and without the extra condition.

**Larger models** — two MIPs of 13724 x 24810 and 10404 x 18899. The condition is reached **zero**
times on both, and presolve reduction counts, node counts and LP iteration counts are identical
with and without the change:

```
                   presolve reductions                              nodes  LP iters
model 1 before     rows 10134(-3590); cols 19268(-5542); nz 64905   1      7087
model 1 after      rows 10134(-3590); cols 19268(-5542); nz 64905   1      7087
model 2 before     rows  7350(-3054); cols 14210(-4689); nz 48859   135    27980
model 2 after      rows  7350(-3054); cols 14210(-4689); nz 48859   135    27980
```

**What that does and does not establish.** On the larger models the change is *inert* — the
condition never fires, so those runs say nothing about the cost of suppressing the reduction, only
that the change does not perturb models where the situation does not arise. The evidence that
suppression is free comes only from the small models, where it is free but the models are small.

So I cannot rule out a performance cost on a large model where this reduction does fire and is
load-bearing. I have not benchmarked against MIPLIB or anything at production scale. What the
counts do suggest is that the situation is rare: 11 occurrences in 435 small models and none in two
models of ~10^4 rows. That bounds how often any cost could be incurred, not how large it might be
when it is.

If that matters for the decision, a run over your own benchmark set would settle it, and I am happy
to adjust the threshold to whatever you would prefer.

### A caveat on the change itself

Bounding `|cost| * boundDiff` bounds only the *direct* objective error. Fixing a column also
perturbs row activities, and downstream that can flip an integer decision worth far more than the
bound suggests. The condition is necessary; I would not claim it is sufficient.

### An approach I would advise against

Pinning at the cost-optimal bound instead of always the lower one. It fixes this model, but only
converts a pessimistic error into an optimistic one of the same size, i.e. a dual bound better than
anything actually achievable, which is worse.

### Testing

* the new `issue-3173` test fails on `latest` at the first commit and passes at the second;
* the full unit test suite passes (1259304 assertions in 335 test cases);
* the 435 small models and 2 larger models above are unchanged in status, objective, node count and
  LP iteration count.

### DIFF
--- a/check/TestMipSolver.cpp
+++ b/check/TestMipSolver.cpp
@@ -1605,3 +1605,39 @@ TEST_CASE("issue-3118a", "[highs_test_mip_solver]") {
 
   highs.resetGlobalScheduler(true);
 }
+
+TEST_CASE("issue-3173", "[highs_test_mip_solver]") {
+  // HPresolve::checkColBounds() treats a column as fixed when its bounds are
+  // equal only within the feasibility tolerance. The condition bounds the
+  // resulting perturbation of the row activities, but not of the objective, so
+  // a column that is nearly fixed in the constraints but has a large cost can
+  // be pinned at the wrong end of its range, losing |cost| * (ub - lb).
+  //
+  // On this model the column has bounds differing by 3.4e-8 and a cost of
+  // -7.3e10, so pinning it at the lower bound costs 2496.5. Presolve reduces
+  // the model to empty and reports that value as optimal, with default options
+  // and no branch-and-bound nodes.
+  //
+  // The optimal objective was verified independently of the MIP solver by
+  // enumerating all 2^8 binary assignments and solving the resulting LP for
+  // each.
+  const std::string filename =
+      std::string(HIGHS_DIR) + "/check/instances/3173-1.mps";
+  const double optimal_objective = -39829.5816585168;
+  Highs highs;
+  highs.setOptionValue("output_flag", dev_run);
+  REQUIRE(highs.readModel(filename) == HighsStatus::kOk);
+  REQUIRE(highs.run() == HighsStatus::kOk);
+  REQUIRE(highs.getModelStatus() == HighsModelStatus::kOptimal);
+  // The column is still pinned within the feasibility tolerance, so allow a
+  // small relative error - but nothing like the 6.3% the defect produces.
+  const double relative_error =
+      std::fabs(highs.getInfo().objective_function_value - optimal_objective) /
+      std::max(1.0, std::fabs(optimal_objective));
+  if (dev_run)
+    printf("issue-3173: objective %g, require %g, relative error %g\n",
+           highs.getInfo().objective_function_value, optimal_objective,
+           relative_error);
+  REQUIRE(relative_error < 1e-6);
+  highs.resetGlobalScheduler(true);
+}
--- a/check/instances/3173-1.mps
+++ b/check/instances/3173-1.mps
@@ -1,0 +1,288 @@
+NAME          HIGHSBUG
+ROWS
+ N  COST
+ L  r0
+ L  r1
+ L  r2
+ L  r3
+ L  r4
+ L  r5
+ L  r6
+ L  r7
+ L  r8
+ L  r9
+ L  r10
+ L  r11
+ L  r12
+ L  r13
+ L  r14
+ L  r15
+ L  r16
+ L  r17
+ L  r18
+ L  r19
+ L  r20
+ L  r21
+ L  r22
+ L  r23
+ L  r24
+ L  r25
+ L  r26
+ L  r27
+ L  r28
+ L  r29
+ L  r30
+ L  r31
+ L  r32
+ L  r33
+ L  r34
+ L  r35
+ L  r36
+ L  r37
+ L  r38
+ L  r39
+COLUMNS
+    MARKER                 'MARKER'                 'INTORG'
+    c0         COST      -1000000
+    c0         r0         -128
+    c0         r7         128
+    c0         r13        64
+    c0         r15        10
+    c0         r17        2.5
+    c0         r19        -2.5
+    c0         r22        10
+    c0         r25        -12
+    c0         r33        -3
+    c0         r39        17
+    c1         COST      -1
+    c1         r0         -20
+    c1         r2         2
+    c1         r3         -20
+    c1         r5         -2
+    c1         r11        128
+    c1         r18        -2.5
+    c1         r21        -1
+    c1         r23        -16
+    c1         r30        -40
+    c1         r34        -1.5
+    c1         r35        -1
+    c1         r36        -2.5
+    c2         COST      -10000
+    c2         r2         -32
+    c2         r10        3
+    c2         r22        -8
+    c2         r29        1
+    c2         r37        -10
+    c3         COST      -0.25
+    c3         r4         -1.5
+    c3         r9         -100
+    c3         r12        -3
+    c3         r13        50
+    c3         r20        -0.5
+    c3         r22        -10
+    c3         r29        1
+    c3         r32        -300
+    c3         r34        -10
+    c4         COST      -0.5
+    c4         r1         4
+    c4         r8         -3
+    c4         r9         -20
+    c4         r11        -10
+    c4         r16        50
+    c4         r20        1
+    c4         r21        -16
+    c4         r34        50
+    c4         r38        -100
+    c5         COST      -5
+    c5         r3         100
+    c5         r5         16
+    c5         r6         128
+    c5         r7         -100
+    c5         r13        -10
+    c5         r15        5
+    c5         r18        5
+    c5         r20        -64
+    c5         r23        5
+    c5         r28        1
+    c5         r36        -62.5
+    c6         COST      -5
+    c6         r0         -10
+    c6         r1         -8
+    c6         r3         -32
+    c6         r21        64
+    c6         r23        -20
+    c6         r26        -4
+    c6         r28        1
+    c6         r31        -10
+    c6         r35        -2.5
+    c6         r37        20
+    c6         r39        8
+    c7         COST      -5
+    c7         r1         1.5
+    c7         r4         19
+    c7         r14        64
+    c7         r24        -200
+    c7         r27        -2
+    c7         r29        1
+    MARKER                 'MARKER'                 'INTEND'
+    c8         COST      -10000
+    c8         r0         64
+    c8         r31        1
+    c9         COST      -2.5
+    c9         r0         -20
+    c9         r1         3
+    c10        COST      -2.5
+    c10        r1         -100
+    c10        r2         2
+    c10        r33        0.5
+    c10        r35        20
+    c11        COST      -0.5
+    c11        r2         -16
+    c11        r3         100
+    c12        COST      -1000000
+    c12        r3         -32
+    c12        r4         64
+    c12        r34        -2
+    c13        COST      -0.5
+    c13        r4         -5
+    c13        r5         100
+    c14        COST      -100
+    c14        r5         -16
+    c14        r6         64
+    c14        r24        1
+    c14        r25        1
+    c14        r26        1
+    c14        r27        1
+    c15        COST      -10
+    c15        r6         -10
+    c15        r7         5
+    c16        COST      -100
+    c16        r7         -0.5
+    c16        r8         20
+    c17        COST      -1
+    c17        r8         -4
+    c17        r9         32
+    c17        r32        1
+    c17        r37        -8
+    c18        COST      -10000
+    c18        r9         -3
+    c18        r10        50
+    c18        r37        128
+    c19        COST      -100
+    c19        r10        -8
+    c19        r11        16
+    c20        COST      -0.25
+    c20        r11        -5
+    c20        r12        50
+    c20        r35        1
+    c21        COST      -100
+    c21        r12        -128
+    c21        r13        16
+    c22        COST      -2.5
+    c22        r13        -64
+    c22        r14        64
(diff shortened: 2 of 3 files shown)


odow:
It's worth noting that HiGHS differs from Gurobi here.

```Julia
julia> optimize!(model)
Using BLAS: Apple Accelerate 
MIP has 40 rows; 33 cols; 138 nonzeros; 8 integer variables (8 binary)
Coefficient ranges:
  Matrix  [5e-01, 3e+02]
  Cost    [1e-06, 1e+06]
  Bound   [1e+00, 1e+03]
  RHS     [1e+00, 4e+01]
WARNING: Problem has some excessively small costs
Presolving model
38 rows, 30 cols, 114 nonzeros 0s
28 rows, 25 cols, 98 nonzeros 0s
23 rows, 20 cols, 82 nonzeros 0s
17 rows, 14 cols, 50 nonzeros 0s
13 rows, 11 cols, 41 nonzeros 0s
11 rows, 8 cols, 24 nonzeros 0s
0 rows, 0 cols, 0 nonzeros 0s
Presolve reductions: rows 0(-40); columns 0(-33); nonzeros 0(-138) - Reduced to empty
Presolve: Optimal

         0       0         0   0.00%   -37333.076244   -37333.076244      0.00%        0      0      0         0     0.0s

Solving report
  Status            Optimal
  Primal bound      -37333.0762443
  Dual bound        -37333.0762443
  P-D integral      0
  Solution status   feasible
                    -37333.0762504 (objective)
                    0 (bound viol.)
                    0 (int. viol.)
                    0 (row viol.)
  Timing            0.00
                    0.00 (Presolve)
                    0.00 (Solve)
                    0.00 (Postsolve)
  Max sub-MIP depth 0
  Nodes             0
  Repair LPs        0
  LP iterations     0

julia> exit()
(base) odow@Mac SDDP % gurobi_cl ~/Downloads/3173-1.mps
Set parameter WLSAccessID
Set parameter WLSSecret
Set parameter LicenseID to value 722777
Set parameter LogFile to value "gurobi.log"
Using license file /Users/odow/gurobi.lic
WLS license 722777 - registered to JuMP Development

Gurobi Optimizer version 12.0.3 build v12.0.3rc0 (mac64[arm] - Darwin 25.5.0 25F80)
Copyright (c) 2025, Gurobi Optimization, LLC

Read MPS format model from file /Users/odow/Downloads/3173-1.mps
Reading time = 0.00 seconds
HIGHSBUG: 40 rows, 33 columns, 138 nonzeros

Using Gurobi shared library /Library/gurobi1203/macos_universal2/lib/libgurobi120.dylib

CPU model: Apple M4
Thread count: 10 physical cores, 10 logical processors, using up to 10 threads

WLS license 722777 - registered to JuMP Development
Optimize a model with 40 rows, 33 columns and 138 nonzeros
Model fingerprint: 0x5b1968c2
Variable types: 25 continuous, 8 integer (0 binary)
Coefficient statistics:
  Matrix range     [5e-01, 3e+02]
  Objective range  [1e-06, 1e+06]
  Bounds range     [1e+00, 1e+03]
  RHS range        [1e+00, 4e+01]
Found heuristic solution: objective -13807.35189
Found heuristic solution: objective -14074.01856
Presolve removed 40 rows and 33 columns
Presolve time: 0.00s
Presolve: All rows and columns removed

Explored 0 nodes (0 simplex iterations) in 0.01 seconds (0.00 work units)
Thread count was 1 (of 10 available processors)

Solution count 4: -39829.6 -39829.6 -14074 -13807.4 
No other solutions better than -39829.6

Optimal solution found (tolerance 1.00e-04)
Best objective -3.982958165852e+04, best bound -3.982958165852e+04, gap 0.0000%
```

Opt-Mucca:
This will always be playing a bit of whack-a-mole, but it makes sense to add some safeguard. I'd say this is too strict if it's appearing that many times already for `check/instances`. I'd wait for @fwesselm for the actual decision (I think he's already toyed with the current conditions a bit)

PR #3175, opened 2026-07-27
fix gcc for hipo sanitizer workflow

galabovaa:
flaky sanitizer hipo test

### DIFF
--- a/extern/HighsExtrasApi.cpp
+++ b/extern/HighsExtrasApi.cpp
@@ -11,6 +11,7 @@
 
 #include "HighsExtrasApi.h"
 
+#include "HConfig.h"
 #include "HighsExtrasExternalDeps.h"
 
 using namespace HighsExtras;
--- a/extern/HighsExtrasApi.cpp
+++ b/extern/HighsExtrasApi.cpp
@@ -11,7 +11,6 @@
 
 #include "HighsExtrasApi.h"
 
-#include "HConfig.h"
 #include "HighsExtrasExternalDeps.h"
 
 using namespace HighsExtras;
--- a/extern/HighsExtrasExternalDeps.cpp
+++ b/extern/HighsExtrasExternalDeps.cpp
@@ -9,7 +9,6 @@
  * @brief Defines the set of external features available
  */
 
-#include "HConfig.h"
 #include "HighsExtrasExternalDeps.h"
 
 namespace HighsExtras {
--- a/extern/HighsExtrasExternalDeps.cpp
+++ b/extern/HighsExtrasExternalDeps.cpp
@@ -9,6 +9,7 @@
  * @brief Defines the set of external features available
  */
 
+#include "HConfig.h"
 #include "HighsExtrasExternalDeps.h"
 
 namespace HighsExtras {
--- a/.github/workflows/sanitizers-cmake.yml
+++ b/.github/workflows/sanitizers-cmake.yml
@@ -152,3 +152,41 @@ jobs:
         working-directory: ${{github.workspace}}/build
         run: |
           ./bin/unit_tests [highs_hipo]
+
+  hipo_macos:
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [macos-latest]
+        config: [Debug]
+        sanitizer: [Address, Thread]
+
+    steps:
+      - uses: actions/checkout@v6
+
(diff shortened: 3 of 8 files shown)


filikat:
Thanks for looking into this!
In the CI, there are a few build failures, which is what I was referring to.
Some of the sanitizers also fail with the message that HiPO is unavailable, which is exactly what I observed on my mac, but using Clang.

extern/HighsExtrasApi.cpp
```diff
@@ -11,6 +11,7 @@
 
 #include "HighsExtrasApi.h"
 
+#include "HConfig.h"
```
galabovaa: @mathgeekcoder should this be here?

extern/HighsExtrasExternalDeps.cpp
```diff
@@ -9,6 +9,7 @@
  * @brief Defines the set of external features available
  */
 
+#include "HConfig.h"
```
galabovaa: @mathgeekcoder same, should this be here?

galabovaa:
@filikat The flaky ones seem to be resolved here, and the macos sanitizers now should work. 

However, I get new data race(s) on ubuntu:

https://github.com/ERGO-Code/HiGHS/actions/runs/30262939950/job/89966915486?pr=3175
https://github.com/ERGO-Code/HiGHS/actions/runs/30262936828/job/89966906925?pr=3175

Would you mind having a look?

filikat:
I cannot find any faulty logic with that piece of code. 
Looking at tsan stack trace, it seems like it's complaining about one of its own functions.

I found this:
https://github.com/google/sanitizers/issues/1106
https://github.com/google/sanitizers/issues/1218
https://stackoverflow.com/questions/73136532/where-is-the-data-race-in-this-simple-c-code

It seems that ubsan and tsan do not work well together. 
To fix, in `CMakeLists.txt` change `-fsanitize=thread,undefined` to `-fsanitize=thread`.

extern/HighsExtrasExternalDeps.cpp
mathgeekcoder: Not strictly necessary, but `#define HIPO` is in `HConfig.h`.  If we remove this include, I think we'll need to add a `target_compile_definitions(highs_extras PRIVATE HIPO)` to the `extern\CMakeLists.txt`.

mathgeekcoder: Oh wait, I just noticed you're including this in the PR.  Were you getting build errors beforehand?

extern/HighsExtrasApi.cpp
mathgeekcoder: `HighsExtrasApi.h` includes `HConfig.h`, so it's not strictly necessary.  Are you wanting to move it from the header file?

jajhall (approved):
Impressive, and thankful for @mathgeekcoder, because I can't comment!

extern/HighsExtrasApi.cpp
galabovaa: No,  I will just remove this. Thank you!

extern/HighsExtrasExternalDeps.cpp
galabovaa: I was getting some issues but they were unrelated! Thank you!

galabovaa:
> I cannot find any faulty logic with that piece of code. Looking at tsan stack trace, it seems like it's complaining about one of its own functions.
> 
> I found this: [google/sanitizers#1106](https://github.com/google/sanitizers/issues/1106) [google/sanitizers#1218](https://github.com/google/sanitizers/issues/1218) https://stackoverflow.com/questions/73136532/where-is-the-data-race-in-this-simple-c-code
> 
> It seems that ubsan and tsan do not work well together. To fix, in `CMakeLists.txt` change `-fsanitize=thread,undefined` to `-fsanitize=thread`.

Thank you for tracking this @filikat!

filikat:
I checked the sanitizer that fails on mac and I think it's a false positive. 
The reads and writes that are flagged are synchronised, so this is not a data race.
If it keeps happening, we can add TSAN_ANNOTATE_HAPPENS_BEFORE/AFTER

galabovaa:
@filikat I think it is worth adding TSAN_ANNOTATE_HAPPENS_BEFORE/AFTER as it may pop up for users elsewhere too

galabovaa:
Thank you @filikat!

PR #3181, opened 2026-07-28
Do not discard a MIP solution when its repair LP fails to solve (#3180)

### LINKED ISSUE #3180: MIP: a solution whose repair LP fails to solve is discarded, losing the optimum and reporting a suboptimal solution as Optimal at gap 0
### Summary

When a MIP solution is feasible for the presolved model but violates the *original* model by more
than `mip_feasibility_tolerance`, `HighsMipSolverData` repairs it: the integer columns are fixed and
the LP is re-solved in the original space. The repaired solution is accepted only if that LP returns
a feasible primal solution:

```cpp
tmpSolver.optimizeLp();
...
if (tmpSolver.getInfo().primal_solution_status == kSolutionStatusFeasible) {
  this->total_repair_lp_feasible++;
  solution = tmpSolver.getSolution();
  allow_try_again = false;
  goto try_again;
}
```

Anything else — including the LP solver *failing* — falls through, and the solution is discarded
with `kHighsInf` returned so that it is not used for bounding.

So a failure of the solver is treated identically to a solution that is genuinely bad. When the
discarded solution is the optimum, the search goes on to report a worse incumbent as `Optimal` at a
zero gap, with nothing in the result indicating that anything went wrong. The repair LP is solved
with `output_flag = false`, so even the error it raised is never shown.

### Reproducing

[repair-discard-1.mps.txt](https://github.com/user-attachments/files/30441316/repair-discard-1.mps.txt), 134 rows x 356 columns, 72 integers, 2690 nonzeros, with
`mip_rel_gap = 0`, `mip_abs_gap = 1e-6`, `mip_feasibility_tolerance = 1e-7`, `random_seed = 0`,
`threads = 1`, `parallel = off`:

| run | objective | |
|---|---|---|
| `--presolve on` | `-7.60954221608e+14` | reported `Optimal`, gap 0 |
| `--presolve off` | `-7.60954325469e+14` | better by `1.04e+08` |

One repair LP is attempted during the `presolve=on` solve and it fails, and that is the solution
that is lost. The better point is feasible, verified independently of the solver by evaluating `Ax`
and `c'x` directly from the MPS: row violation `6.4e-10`, bound violation `5.0e-14`, integrality
violation `0`, objective `-7.60954325469e+14`.

On a 12710 x 23091 production MIP the same mechanism is much more visible: **34 repair LPs are
attempted, 17 fail, and 17 solutions are discarded**, the optimum among them. HiGHS reports
`Optimal` at a zero gap with `-3755449.094926` against a true optimum of `-3755469.767048`,
suboptimal by 20.67. That model is available on request.

### The failure is not tied to one cause

Two distinct LP failures reach the same discard, driven by different model properties:

| failure raised by the repair LP | driven by | demonstrated on |
|---|---|---|
| `Dual simplex ratio test failed due to excessive dual values` | large objective coefficients | `chuzc-1.mps`, 14 x 28 — reported separately |
| `Dual simplex detected excessive primal values` | large variable bounds | [primal-values-1.mps.txt](https://github.com/user-attachments/files/30441315/primal-values-1.mps.txt), 102 x 376, attached |

`primal-values-1.mps` returns `Solve error` under the dual simplex and `Optimal` under the primal
simplex, deterministically over five runs. It is attached as evidence that the discard is not
specific to one numerical fault; it is not being reported as a defect in its own right, since
bailing out on a model with bounds of `1e13` may well be intended behaviour.

A third route needs no numerical difficulty at all. The repair LP is given

```cpp
double time_available = std::max(
    mipsolver.options_mip_->time_limit - mipsolver.timer_.read(), 0.1);
```

so with a user-set `time_limit` it can be given as little as 0.1 seconds. A timeout also leaves
`primal_solution_status != kSolutionStatusFeasible` and so also discards the solution. I have not
constructed that case, and the surrounding MIP would probably terminate on the time limit rather
than report `Optimal`, so I mention it only as part of the argument that the acceptance test is too
broad.

### Where the violation that triggers the repair comes from

On the models above it comes from `HPresolve::scaleMIP`, which scales each row by a power of two
derived from the largest absolute coefficient among its continuous columns. The scaling is exact,
but the feasibility tolerance is not scale invariant: a row scaled by `s` and satisfied to `feastol`
in the presolved space is satisfied only to `feastol / |s|` in the original space. On the production
model every one of the 5341 scaled rows is scaled *down*, 2453 of them by 256x or more, so a `1e-7`
tolerance is being applied as up to `2e-4`.

I would advise against addressing this by suppressing the down-scaling. Only scaling rows up does
fix the production model, and leaves the 26 MIP models in `check/instances` unchanged in objective,
node count and LP iteration count — but on a second production model of 14412 x 26520 it *causes* a
wrong answer, `-19104998.2316` against a correct `-19105203.4816`, even though that model never
reaches the untransformed-violation path at all. Any change to `scaleMIP` perturbs models that have
no defect. I mention it because it is the obvious first thing to try, and it looked clean on every
measurement I had until the second model turned up.

The repair mechanism is the right place for this: it already exists, it already recovers 17 of the
34 solutions on the production model, and it is inert on models where no solution is ever discarded.

### Why this is worth separating from the LP failure

Fixing the ratio test would remove today's trigger without changing the behaviour reported here: an
unsolved repair LP would still discard a solution silently, for any of the reasons above. The two
live in different subsystems — `highs/simplex/HEkkDualRow.cpp` and `highs/mip/HighsMipSolverData.cpp`
— and neither fix touches the other file.

The difference in severity also matters. A failed LP solve is a loud failure: the status is
`Solve error` and the caller knows. This one converts that into a silently suboptimal answer
reported as `Optimal` at a zero gap, which is much harder to detect downstream.

### A second, related inconsistency in the same code

Retrying the failed repair LP is the obvious remedy, and on its own it is not enough. Doing only
that, one generated model produced a repaired solution that passes the MIP's own check and then
fails the check applied to the returned solution, at `1.5e-7` against a `1e-7` tolerance, turning a
reported `Optimal` into `Solve error`.

The reason is that the two checks measure different things. `solutionFeasible()` is called as

```cpp
mipsolver.solutionFeasible(mipsolver.orig_model_, solution.col_value,
                           &solution.row_value, ...)
```

and with a row-value vector supplied it trusts it rather than recomputing:

```cpp
if (pass_row_value) {
  ...                                    // used as given
} else {
  calculateRowValuesQuad(*lp, col_value, row_value);
}
```

After a repair, `solution.row_value` comes from the repair LP, and those activities need only
satisfy that LP to its own tolerance. The check later applied to the returned solution recomputes
them in quad precision, so the two can disagree by more than the tolerance. Recomputing the row
activities from the repaired column values before the feasibility test removes the discrepancy.

### Relation to existing issues

The LP failure itself is already known and was judged to be expected behaviour. #1484 reports the
dual simplex failing on a model with costs of `1e11`-`1e12`, with the diagnosis that dual
feasibility to `1e-7` against values of order `1e14` is beyond machine precision; it was resolved by
improving the message rather than changing the ratio test, and HiGHS now warns
`Problem has some excessively large costs` on such models. #2866 is the same thing driven by large
column *bounds* rather than costs — its two attached models now return the correct objective 417 on
`latest`.

This report is not asking for either of those to change. The point is what the MIP solver does when
that expected failure occurs inside it: a solution is discarded, silently, and the optimum can be
lost. A user solving the LP directly sees `Solve error` and knows; a user solving the MIP sees
`Optimal` at a zero gap and does not.

#1543 asks what the `has untransformed violations` warning means, and jajhall's answer there — that
such messages indicate an error in presolve and/or postsolve — is consistent with the origin
described above. It does not cover the discard behaviour.

### Version

`latest` @ `bc30ccf1a1` (1.15.1), with the fixes for #3170, #3171 and #3173 applied; the behaviour
is unchanged without them.

EamonHetherton:
Fixes #3180.

### The defect

When a MIP solution is feasible for the presolved model but violates the original model by more
than `mip_feasibility_tolerance`, `HighsMipSolverData` repairs it by fixing the integer columns and
re-solving the LP in the original space. The repaired solution is accepted only if that LP returns a
feasible primal solution, so a *failure of the solver* is treated exactly like a solution that is
genuinely bad: it is discarded, and `kHighsInf` is returned so it is not used for bounding.

The repair LP is solved in the original space with the dual simplex, which on a badly scaled model
can fail the ratio test — "excessive dual values", or its primal counterpart — and return
`kSolveError`. When the discarded solution is the optimum, the search reports a worse incumbent as
`Optimal` at a zero gap. The repair LP runs with `output_flag = false`, so the error it raised is
never shown either.

### The change

Two parts, both in `HighsMipSolverData::transformNewIntegerFeasibleSolution`.

**Retry a failed repair LP without presolve.**

```cpp
    tmpSolver.optimizeLp();
    if (tmpSolver.getModelStatus() == HighsModelStatus::kSolveError) {
      tmpSolver.setOptionValue("presolve", kHighsOffString);
      tmpSolver.optimizeLp();
    }
```

The failing LPs solve to optimality that way; primal simplex and IPM also solve them, so if you
would rather fall back differently, any of the three recovers the solution.

**Recompute the row activities from the repaired column values.**

```cpp
      solution = tmpSolver.getSolution();
      calculateRowValuesQuad(*mipsolver.orig_model_, solution.col_value,
                             solution.row_value);
```

This part is not optional, and it is the subtler of the two. `solutionFeasible()` is passed
`&solution.row_value` and, when a row-value vector is supplied, trusts it instead of recomputing:

```cpp
if (pass_row_value) {
  ...                                    // used as given
} else {
  calculateRowValuesQuad(*lp, col_value, row_value);
}
```

After a repair those activities come from the repair LP and need only satisfy that LP to its own
tolerance, whereas the check later applied to the returned solution recomputes them in quad
precision. Without recomputing, a repaired solution can pass here at `1e-7` and then fail that check
at `1.5e-7`, turning a reported `Optimal` into `Solve error`. With the retry alone I reproduced
exactly that on a generated model; with both parts the same model correctly rejects the marginal
point and keeps a genuinely feasible one.

### Effect

| | result |
|---|---|
| 12710 x 23091 production MIP | `-3755449.094926` -> `-3755469.767048` (correct); 17 previously discarded solutions recovered |
| 14412 x 26520 production MIP | unchanged and correct at `-19105203.4816` |
| generated MIP, repair LP fails | `-7.60954221608e+14` -> `-7.60954325469e+14` (correct) |
| generated MIP, marginal repair | now `Optimal` with no error, where the retry alone gave a false optimum |
| the 26 MIP models in `check/instances` | identical objectives to stock `latest` |
| full unit test suite | 1259300 assertions in 334 test cases, pass |

Measured on this branch alone — `latest` plus this commit, with no other changes applied — so the
result does not depend on anything else I have reported.

The retry is reached only when a repair LP has already failed, and the recomputation only when a
repair has already succeeded, so neither can perturb a model where no solution is ever repaired.

### What I would advise against

Addressing this in `HPresolve::scaleMIP`, which is where the violation that triggers the repair
comes from — it scales rows down, and the feasibility tolerance is not scale invariant, so a row
scaled by `s` and satisfied to `feastol` in the presolved space is satisfied only to `feastol / |s|`
in the original one.

Suppressing the down-scaling does fix the first model and leaves `check/instances` unchanged in
objective, node count and LP iteration count. But on the second production model above it *causes* a
wrong answer — `-19104998.2316` against a correct `-19105203.4816` — even though that model never
reaches the untransformed-violation path. Any change to `scaleMIP` perturbs models that have no
defect. The repair path is the right place: it already exists, it already recovers 17 of the 34
repairs on the first model, and it is inert otherwise.

### On a regression test

I have a generated MIP that reproduces the discard — 134 rows x 356 columns, 72 integers — but it
needs the pinned option set and its objective coefficients span `1e-6` to `1e12`, which is the kind
of instance that was objected to on #3172. I have not added it as a test for that reason, and am
happy to supply it if you would like it, or to look for a numerically tamer one.

### Testing

* both production models correct under `presolve` on and off;
* the reproducers for #3170, #3171 and #3173 unchanged and correct;

### DIFF
--- a/highs/mip/HighsMipSolverData.cpp
+++ b/highs/mip/HighsMipSolverData.cpp
@@ -11,6 +11,7 @@
 #include <sstream>
 
 #include "../extern/pdqsort/pdqsort.h"
+#include "lp_data/HighsLpUtils.h"
 #include "lp_data/HighsModelUtils.h"
 #include "mip/HighsPseudocost.h"
 #include "mip/HighsRedcostFixing.h"
@@ -1239,11 +1240,31 @@ try_again:
     // HiPO or IPX to solve an LP without a basis, use simplex
     tmpSolver.setOptionValue("solver", kSimplexString);
     tmpSolver.optimizeLp();
+    // The repair LP is solved in the original space, which for a badly scaled
+    // model can defeat the dual simplex ratio test ("excessive dual values")
+    // or its primal counterpart. That is a failure of the solver, not a
+    // statement about the solution, so retry once without presolve rather than
+    // discarding a solution that may be the incumbent - or the optimum. Only
+    // worth doing if presolve was used for the first attempt: otherwise the
+    // retry would solve an identical LP and fail in the same way.
+    if (use_presolve &&
+        tmpSolver.getModelStatus() == HighsModelStatus::kSolveError) {
+      tmpSolver.setOptionValue("presolve", kHighsOffString);
+      tmpSolver.optimizeLp();
+    }
     this->total_repair_lp_iterations +=
         tmpSolver.getInfo().simplex_iteration_count;
     if (tmpSolver.getInfo().primal_solution_status == kSolutionStatusFeasible) {
       this->total_repair_lp_feasible++;
       solution = tmpSolver.getSolution();
+      // Recompute the row activities from the repaired column values. The
+      // feasibility test below is given solution.row_value and trusts it, but
+      // the values returned by the LP solver need only satisfy the LP to its
+      // own tolerance, so they can differ from a quad-precision recomputation -
+      // which is what the check applied to the returned solution uses. Without
+      // this, a repaired solution can pass here and fail that later check.
+      calculateRowValuesQuad(*mipsolver.orig_model_, solution.col_value,
+                             solution.row_value);
       allow_try_again = false;
       goto try_again;
     }
--- a/highs/mip/HighsMipSolverData.cpp
+++ b/highs/mip/HighsMipSolverData.cpp
@@ -1240,13 +1240,6 @@ try_again:
     // HiPO or IPX to solve an LP without a basis, use simplex
     tmpSolver.setOptionValue("solver", kSimplexString);
     tmpSolver.optimizeLp();
-    // The repair LP is solved in the original space, which for a badly scaled
-    // model can defeat the dual simplex ratio test ("excessive dual values")
-    // or its primal counterpart. That is a failure of the solver, not a
-    // statement about the solution, so retry once without presolve rather than
-    // discarding a solution that may be the incumbent - or the optimum. Only
-    // worth doing if presolve was used for the first attempt: otherwise the
-    // retry would solve an identical LP and fail in the same way.
     if (use_presolve &&
         tmpSolver.getModelStatus() == HighsModelStatus::kSolveError) {
       tmpSolver.setOptionValue("presolve", kHighsOffString);
@@ -1257,12 +1250,6 @@ try_again:
     if (tmpSolver.getInfo().primal_solution_status == kSolutionStatusFeasible) {
       this->total_repair_lp_feasible++;
       solution = tmpSolver.getSolution();
-      // Recompute the row activities from the repaired column values. The
-      // feasibility test below is given solution.row_value and trusts it, but
-      // the values returned by the LP solver need only satisfy the LP to its
-      // own tolerance, so they can differ from a quad-precision recomputation -
-      // which is what the check applied to the returned solution uses. Without
-      // this, a repaired solution can pass here and fail that later check.
       calculateRowValuesQuad(*mipsolver.orig_model_, solution.col_value,
                              solution.row_value);
       allow_try_again = false;


jajhall (approved):
Fair enough, this does no harm

highs/mip/HighsMipSolverData.cpp
```diff
@@ -1239,11 +1240,28 @@ double HighsMipSolverData::transformNewIntegerFeasibleSolution(
     // HiPO or IPX to solve an LP without a basis, use simplex
     tmpSolver.setOptionValue("solver", kSimplexString);
     tmpSolver.optimizeLp();
+    // The repair LP is solved in the original space, which for a badly scaled
+    // model can defeat the dual simplex ratio test ("excessive dual values")
+    // or its primal counterpart. That is a failure of the solver, not a
+    // statement about the solution, so retry once without presolve rather than
+    // discarding a solution that may be the incumbent - or the optimum.
+    if (tmpSolver.getModelStatus() == HighsModelStatus::kSolveError) {
```
Opt-Mucca: Presolve may have already been disabled in the lines directly above. In this case there's no need to re-run the LP.

Opt-Mucca:
@EamonHetherton you keep mentioning "production models" and "available upon request". Any chance you're willing to share them? If so we'd store them, use them for testing, and make them public with appropriate credit in a couple of years.

highs/mip/HighsMipSolverData.cpp
jajhall: True, in which case the LP could be solved with presolve "on" as another attempt to get an optimal solution. 

If it matters enough to a user, IPM could also be tried.

highs/mip/HighsMipSolverData.cpp
```diff
@@ -1239,11 +1240,28 @@ double HighsMipSolverData::transformNewIntegerFeasibleSolution(
...
+      tmpSolver.setOptionValue("presolve", kHighsOffString);
```
jajhall, line 1245: Change this so that the attempt with presolve off only happens if presolve was on in the first attempt

jajhall (changes requested):
See comment

EamonHetherton:
@Opt-Mucca yes, I can share more, see https://github.com/ERGO-Code/HiGHS/issues/3180#issuecomment-5109878723 for some detail.  Would you like an MPS file where I find a bug? they are usually of the order of 2-3MB.

EamonHetherton:
Good catch, thanks — applied.

`use_presolve` is `!mip_root_presolve_only`, so with that option set presolve is already off for
the first attempt and the retry would solve an identical LP and fail in the same way. The retry is
now guarded:

```cpp
    if (use_presolve &&
        tmpSolver.getModelStatus() == HighsModelStatus::kSolveError) {
      tmpSolver.setOptionValue("presolve", kHighsOffString);
      tmpSolver.optimizeLp();
    }
```

Re-verified after the change: the two production models and the generated reproducer still return
the correct optima, including with `mip_root_presolve_only = true`, and the full unit test suite
passes (1259300 assertions in 334 test cases).

I have not added the further fallbacks you both suggested — retrying with presolve on when it was
off first, or trying IPM — since that widens the change beyond what the reported defect needs. Both
would fit in the same place if you want them; on the LPs I have, presolve off, primal simplex and
IPM all solve what the dual simplex fails on, so any of the three works as a fallback.

Opt-Mucca (approved):
Happy with the changes! I'd consider removing all comments though as they're pretty verbose and it's always clear what's happening.

Opt-Mucca:
@EamonHetherton I'd like one or two of the instances that you think are representative of the application. It doesn't have to be the ones that exposed some bugs.

jajhall:
Closes #3180

EamonHetherton:
@Opt-Mucca how would you like me to get the mps files to you? I would rather not make them public at this stage.

Opt-Mucca:
@EamonHetherton please email me using the information here https://www.zib.de/members/turner (if the files are more than a couple of megabytes compressed then please send a good drive link or something similar)

PR #3182, opened 2026-07-28
Fix 3166

### LINKED ISSUE #3166: Out-of-bounds access in HighsSymmetryDetection::partitionRefinement
The following model crashes on Linux x86-64 and also MacOS arm64:

[highs_crashmodel.mps.gz](https://github.com/user-attachments/files/30368711/highs_crashmodel.mps.gz)

It acrashes with default options but also deterministically (`highs --parallel off --threads 1 highs_crashmodel.mps`). I was unable to meaningfully reduce the size of the crashing instance, looks like a fairly specific path is needed.

ASAN nails it down to an out-of-bounds access: at HighsSymmetry.cpp around line 912 we have `cellEnd < cellStart` which makes `std::partition` fault.

It seems that the pruning around line 979:

```      
      currentPartitionLinks[firstCellStart] = cellEnd;

      // undo possibly incomplete changes done to the cells
      for (--i; i >= refineStart; --i)
        updateCellMembership(i, firstCellStart, false);
```

causes this to become corrupt. I tried to remove that and rely on them be removed by backtracking of the caller together with all other cells created for this node, and that fixes the issue.

Opt-Mucca:
This fixes #3166 

I believe the error was that some data structures were not being updated when pruning in `partitionRefinement`. This had absolutely no effect on my run of MIPLIB 2017 with a 30min time limit. Given I'm not that comfortable with the symmetry code still, it should probably wait for a review.


fwesselm (approved):
@Opt-Mucca, I am still learning about the symmetry detection code, but this looks correct!

I added a test, I hope that's OK, your code changes are now covered.

Opt-Mucca:
@fwesselm I'm still learning it too. Will merge this when the tests pass!

PR #3184, opened 2026-07-28
Upgrade activity contributions to highscdouble

Opt-Mucca:
This fixes the infeasible instance attached by @zeddybot in #2866 (completely missed it in the wall of AI text)

Similar to the change @fwesselm did recently in #3149 (upgrades the precision of some terms). Given all calls to this were already summing their results into a `HighsCDouble`, I assume the performance impact is truly negligible, but I'll still run a test to check.

### DIFF
--- a/highs/mip/HighsDomain.cpp
+++ b/highs/mip/HighsDomain.cpp
@@ -17,29 +17,29 @@
 #include "mip/HighsCutPool.h"
 #include "mip/HighsMipSolverData.h"
 
-static double activityContributionMin(double coef, const double& lb,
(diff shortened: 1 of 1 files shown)


fwesselm (approved):
Looks good, @Opt-Mucca!

Opt-Mucca:
@galabovaa I believe the two failing tests are completely unrelated. Are we fine to merge this?

fwesselm:
@Opt-Mucca, I restarted the failing tests, and they all passed on re-run.

Issue #3120, opened 2026-06-30
Check for time limit when constructing the starting basis in IPX


jajhall:
Constructing the starting basis in IPX can take a _long_ time, and there appears to be no time_limit check

jajhall:
Closed by #3185

PR #3185, opened 2026-07-28
Added timeout to forming initial basis and factorization in IPX

jajhall:
Timeout when forming the initial basis is achievable with `control_.InterruptCheck()`, and `bool interrupt` used to communicate interruption down to levels where `IPX_ERROR_time_interrupt` can be returned or assigned to `info->errflag`. 

Spotted that `Int Load(const int* basic_status);` is redundant.

Timeout within `basiclu_factorize` needs an independent clock, time start and time limit. 

When IPX takes "forever" there was no logging, so I've added that at the end of each pass through the loop in `BasicLu::_Factorize`, with an interval of at least 5s. It's interesting to see how `BasicLu::_Factorize` keeps returning to pick up more memory. There can still be long gaps, but it's not feasible to put HiGHS logging into `BasicLu::_Factorize`.

As with all time-out code, unit tests are troublesome, but I've developed it with the examples that made me realise that there's no time-out when forming the initial basis and factorization in IPX


jajhall:
Closes #3120

highs/ipm/ipx/lp_solver.cc
```diff
@@ -405,7 +405,9 @@ void LpSolver::RunIPM() {
         if (info_.status_ipm != IPX_STATUS_not_run)
             return;
     }
+    printf("LpSolver::RunIPM before BuildStartingBasis(); timer.Elapsed() = %s\n", time(this->control_.Elapsed()).c_str());
```
filikat: These 2 printf should be removed

filikat:
Apart from that, it looks good

highs/ipm/ipx/lp_solver.cc
jajhall: Thanks: done!

PR #3187, opened 2026-07-29
Normal equations constructed in parallel

filikat:
- The structure (once per IPM solve) and values (once per IPM iteration) of the normal equations are built in parallel. This is only done if matrix A is sufficiently large (1e4 rows or columns) and dense (10 nz per col or 30 nz per row). Building values is parallelised easily, as the structure is already computed and different threads write to different entries. Building the structure is more complicated, as the total number of nonzeros is not known. Each thread uses a local buffer and the final structure is assemble serially.
- Running of the multiple orderings in parallel is now performed regardless of the `parallel` option.
- The Highs random seed is now passed to Metis (with an offset, so that current behaviour is not perturbed, and it's easier to compare results with previous versions). This is the only place where the random seed has an effect on HiPO.


jajhall (approved):
I'm a bit shocked by the comment "vector<bool> is not thread-safe" - why? - as I'm sure it's used elsewhere! 

Otherwise, apart from the `random_seed` offset of 42 (why?) this looks good to me

filikat:
> I'm a bit shocked by the comment "vector is not thread-safe" - why? - as I'm sure it's used elsewhere!

`std::vector<bool>` is _not_ thread-safe (from https://en.cppreference.com/cpp/container/vector_bool: "Does not guarantee that different elements in the same container can be modified concurrently by different threads"). 
`std::vector<T>` for any other T _is_ thread safe.


> Otherwise, apart from the `random_seed` offset of 42 (why?) this looks good to me

It's just the random seed that I was passing to Metis up to now (https://en.wikipedia.org/wiki/Phrases_from_The_Hitchhiker's_Guide_to_the_Galaxy#The_Answer_to_the_Ultimate_Question_of_Life,_the_Universe,_and_Everything_is_42).
I can remove the offset, but it makes it a bit harder to compare future results with past runs.

jajhall:
> > I'm a bit shocked by the comment "vector is not thread-safe" - why? - as I'm sure it's used elsewhere!
> 
> `std::vector<bool>` is _not_ thread-safe (from https://en.cppreference.com/cpp/container/vector_bool: "Does not guarantee that different elements in the same container can be modified concurrently by different threads"). `std::vector<T>` for any other T _is_ thread safe.

So, do we get rid of other occurrences of  `std::vector<bool>` in HiGHS? Or only the ones where modification by different threads is possible - and I'm unaware of any such cases.

> > Otherwise, apart from the `random_seed` offset of 42 (why?) this looks good to me
> 
> It's just the random seed that I was passing to Metis up to now (https://en.wikipedia.org/wiki/Phrases_from_The_Hitchhiker's_Guide_to_the_Galaxy#The_Answer_to_the_Ultimate_Question_of_Life,_the_Universe,_and_Everything_is_42). I can remove the offset, but it makes it a bit harder to compare future results with past runs.

No problem, I know the reference, and it adds a human touch to the code

filikat:
> So, do we get rid of other occurrences of `std::vector<bool>` in HiGHS? Or only the ones where modification by different threads is possible - and I'm unaware of any such cases.

This has been an issue for me (thread sanitizer was failing and it took some time to figure out that `std::vector<bool>` was the problem) and for Ivet (#3025, same issue). I think it would be good to eventually get rid of all instances and replace them with `std::vector<char>` or `std::vector<uint8_t>`, but there is no rush to do this. As you mentioned, I don't think that `std::vector<bool>` is used concurrently anywhere at the moment.

jajhall:
> > So, do we get rid of other occurrences of `std::vector<bool>` in HiGHS? Or only the ones where modification by different threads is possible - and I'm unaware of any such cases.
> 
> This has been an issue for me (thread sanitizer was failing and it took some time to figure out that `std::vector<bool>` was the problem) and for Ivet (#3025, same issue). I think it would be good to eventually get rid of all instances and replace them with `std::vector<char>` or `std::vector<uint8_t>`, but there is no rush to do this. As you mentioned, I don't think that `std::vector<bool>` is used concurrently anywhere at the moment.

I've just done a search, and there are about 20 std::vector<bool> in HiGHS, including the following

> /home/jajhall/HiGHS/highs/ipm/hipo/factorhighs
Analyse.cpp:213:  std::vector<bool> is_sn(n_, false);
/home/jajhall/HiGHS/highs/ipm/hipo/ipm
KktMatrix.cpp:137:  std::vector<bool> is_nz(m, false);
Model.cpp:47:  std::vector<bool> mark(m_, false);
UpLookingSolver.cpp:95:  std::vector<bool> mark(n_, false);

I'll create an issue to remind ourselves to clear them all!

Opt-Mucca:
@filikat I've been using `uint8_t` for parallel MIP (I find it nicer than `char`)

PR #3191, opened 2026-08-03
Fourier-Motzkin presolve

fwesselm:
This PR adds the Fourier-Motzkin presolve reduction as described by [Zhang, Y., Ploskas, N. & Sahinidis, N.V. A novel linear optimization presolve technique based on Fourier-Motzkin elimination. Math. Prog. Comp. 18, 345–378 (2026)](https://doi.org/10.1007/s12532-026-00316-3):
- Currently, impact on MIP solver performance is very small (around 1% slowdown on average). The reduction is enabled in the default settings, which may need to be changed.
- I made changes to postsolve that allow for adding rows/columns (during presolve).


highs/presolve/HPresolve.cpp
```diff
@@ -1329,12 +1328,8 @@ HPresolve::Result HPresolve::dominatedColumns(
         if (!tryToFix) numDomChecksPredBndAnalysis++;
         // check for domination
         if (checkDomination(direction, col, direction_k, k)) {
-          // Re-check the implied bound condition since earlier fixings in
-          // this dominatedColumns call may have changed the model state
-          bool currentBoundImplied =
-              direction > 0 ? isUpperImplied(col) : isLowerImplied(col);
           if (tryToFix &&
-              (currentBoundImplied ||
+              (boundImplied ||
```
Opt-Mucca, line 1465: Is this an old artifact that a merge with `latest` missed?

check/TestPresolveRules.cpp
```diff
@@ -79,6 +79,67 @@ TEST_CASE("test-col-stuffing", "[highs_test_presolve_rules]") {
   h.resetGlobalScheduler(true);
 }
 
+TEST_CASE("test-fourier-motzkin", "[highs_test_presolve_rules]") {
+  Highs h;
+  //  h.setOptionValue("output_flag", dev_run);
+  h.setOptionValue("presolve_rule_test", kPresolveRuleFourierMotzkin);
+  h.setOptionValue("presolve_rule_logging", true);
```
Opt-Mucca, line 86: Should this always be `true`? I'm assuming the `dev_run` changes is also just a forgotten local change.

highs/presolve/HighsPostsolveStack.h
```diff
@@ -237,69 +340,110 @@ class HighsPostsolveStack {
...
     reductions.emplace_back(type, position);
   }
 
+  bool isModelRow(HighsInt row) const { return row < nextRowIndex; }
+
  public:
-  const HighsInt* getOrigRowsIndex() const { return origRowIndex.data(); }
+  const std::vector<HighsInt>& getOrigColIndex() const { return origColIndex; }
+
+  const std::vector<HighsInt>& getOrigRowIndex() const { return origRowIndex; }
```
Opt-Mucca, line 381: Should there be a second function `const std::vector<HighsInt>& getOrigRowIndex(HighsInt row) const { return origRowIndex[row]; }`? 

Is nitpicky, but the constant `getOrigRowIndex()[i]` looks a bit funny.

highs/presolve/HPresolve.cpp
```diff
@@ -5912,10 +6006,19 @@ HPresolve::Result HPresolve::presolve(HighsPostsolveStack& postsolve_stack) {
...
+        HPRESOLVE_CHECKED_CALL(
+            fourierMotzkin(postsolve_stack, numColsEliminatedFourierMotzkin));
+
       if (analysis_.allow_rule_[kPresolveRuleAggregator])
         HPRESOLVE_CHECKED_CALL(aggregator(postsolve_stack));
 
-      if (problemSizeReduction() > 0.05) continue;
+      // check if there were reductions
+      bool haveReductions = problemSizeReduction() > 0.05;
+      tryFourierMotzkin = haveReductions || numColsEliminatedFourierMotzkin > 0;
```
Opt-Mucca, line 6678: Were there any cases where it found further reductions in a second round?

highs/presolve/HPresolve.cpp
```diff
@@ -6806,6 +6934,848 @@ HPresolve::Result HPresolve::aggregator(HighsPostsolveStack& postsolve_stack) {
...
+        postsolve_stack.fourierMotzkinBlockFinalise(blockSteps, rowAncestry);
+        printLog(numColsEliminatedBlock, numRowsEliminated, numRowsAdded);
+        blockSteps.clear();
+        rowAncestry.clear();
+        numColsEliminatedBlock = 0;
+        numRowsEliminated = 0;
+        numRowsAdded = 0;
+      }
+      // reformulate objective
+      reformulateObjective();
```
Opt-Mucca, line 8323: Aren't we leaving the model with an erased objective vector? Doesn't this have tonnes of run-off effects?

highs/presolve/HPresolve.cpp
```diff
@@ -6806,6 +6934,848 @@ HPresolve::Result HPresolve::aggregator(HighsPostsolveStack& postsolve_stack) {
...
+  HighsInt numRowsAdded = 0;
+
+  // FM block data for postsolve
+  std::vector<FmeBlockStep> blockSteps;
+
+  // surviving row to its ancestry (which parent rows it descends from)
+  std::unordered_map<HighsInt, std::vector<FmeAncestryEntry>> rowAncestry;
+
+  // distinct original parent rows for each derived row (Cernikov check)
+  std::unordered_map<HighsInt, std::set<HighsInt>> rowOriginals;
```
Opt-Mucca, line 8299: Can this become outdated when there's a call `reformulateObjective -> shrinkProblem`?

highs/presolve/HighsPostsolveStack.cpp
```diff
@@ -1392,4 +1439,366 @@ void HighsPostsolveStack::SlackColSubstitution::undo(
...
+      stack.pop(steps[s].plusEntries[r]);
+  }
+
+  return steps;
+}
+
+void HighsPostsolveStack::undoFourierMotzkinBlock(
+    const std::vector<FmeStepData>& steps, const HighsOptions& options,
+    HighsSolution& solution, HighsBasis& basis) {
+  const double tol = options.mip_feasibility_tolerance;
```
Opt-Mucca: `primal_feasibility_tolerance`?

highs/presolve/HPresolve.cpp
```diff
@@ -6806,6 +6934,848 @@ HPresolve::Result HPresolve::aggregator(HighsPostsolveStack& postsolve_stack) {
...
+        collectRowEntries(pRow, col, pDirection * pScale, newRowEntries,
+                          newRowMark);
+        collectRowEntries(mRow, col, mDirection * mScale, newRowEntries,
+                          newRowMark);
+
+        // reset marker before removing near-zeros
+        for (const auto& e : newRowEntries) newRowMark[e.col] = -1;
+
+        // remove near-zero entries
+        newRowEntries.erase(
```
Opt-Mucca, line 8374: Does any relaxation need to be done here as opposed to directly removing the coefficient?

Opt-Mucca (commented):
I've finished my first round of review. I've got the general gist, but I failed completely at some point to keep track of the `lambdas` and the basis postsolve is brutal...... I'm hoping all my comments are still valid. I'm just going to put down some comments and questions here:
- The functionality to add rows and columns during presolve is a must. This opens the door for other techniques. Is super cool that you've added it!
- In general I'm against adding "dead-code" to the repository, and I'd prefer not adding such a large feature if it's not enabled by default.
- Do you think there's a window of MIP instances where this is super helpful? 1% performance degradation isn't too bad, and I'd still be fine merging it if (1) we're confident that the chance of worst-case performance is near zero (2) it is really helpful for some family of instances we care about
- Is it helpful for LPs over a general test set? If so then I'd just enable it there by default.

check/TestPresolveRules.cpp
jajhall: `dev_run` is set true or false at the top of a file of unit tests so that logging is on/off due to `h.setOptionValue("output_flag", dev_run);` according to whether `dev_run` is true/false

Sometimes I'm lazy and comment out `h.setOptionValue("output_flag", dev_run);` rather than change `dev_run`  to true

jajhall:
> * Is it helpful for LPs over a general test set? If so then I'd just enable it there by default.

Depending on what LP testing @fwesselm has done, I'll test over a good set of LPs - for both performance and number of iterations after postsolve (which can be a test of correctness for the dual/basis postsolve).

highs/presolve/HPresolve.cpp
fwesselm: No, I just cleaned up the code a little.

highs/presolve/HighsPostsolveStack.h
fwesselm: Good point. I added methods for this.

highs/presolve/HPresolve.cpp
fwesselm: In order to be able to eliminate a column with non-zero cost, FM presolve has to add the objective as a constraint. I agree, this may have all kinds of side effects.

highs/presolve/HighsPostsolveStack.cpp
fwesselm: I double-checked, and I think you are right, thank you. The `primal_feasibility_tolerance` is used now.

highs/presolve/HPresolve.cpp
fwesselm: Good catch! The `rowAncestry` was already cleared, but `rowOriginals` was not. I fixed this.

check/TestPresolveRules.cpp
fwesselm: Thanks, I uncommented `h.setOptionValue("output_flag", dev_run);`.

fwesselm:
I have merged `latest` (including presolve light) into this branch.

highs/presolve/HPresolve.cpp
fwesselm: Good point. I checked the remaining presolve code, and currently coefficients with absolute value less than or equal to `options->small_matrix_value` are just treated as zero. Therefore, I would like to keep the code as it is right now for the time being. I am open to revisit this in the future.

PR #3192, opened 2026-08-03
Parallel redesign

Opt-Mucca:
This PR is a first redesign of the parallel MIP code. I'm just going to dot point the major changes:
- Adds nodes to the `preparedNodes` of each worker. A worker now only processes nodes assigned to it.
- Adds a max node processing limit of 100 to each worker. This tries to balance thread utilisation.
- Change `backtrackPlunge` to a simple DFS style algorithm without the ability to skip a level of the tree while backtracking based on pseudocosts.
- Change the `nodequeue` attached to each worker to a simple `std::vector`. 

For my tests on 8 threads this gives a 6.5% time improvement and 6% LP iteration improvement. It gives a 1% performance degradation on single-threaded performance. Edit: This change should also use 40% less memory.

Reason for single-threaded degradation: There is one outlier instance `tbfp-network` where the sub-mip no longer finds the optimal solution instantly. This means the problems solves 5x slower. It stems from the `backtrackPlunge` changes, but I'd mostly chalk it up to randomness. Without this instance the change is performance neutral on single-threaded and more than 6.5% improving on 8 threads.


highs/mip/HighsMipSolver.cpp
```diff
@@ -462,6 +457,7 @@ void HighsMipSolver::run() {
       master_worker.search_ptr_->getLocalDomain());
 
   profiling_->start(kMipClockSearch);
+  HighsInt maxNodesPerWorkerLim = max_num_workers > 1 ? 100 : kHighsIInf;
```
fwesselm: Did you experiment with different values? It is not worth making this an option?

Or maybe add `constexpr HighsInt someGoodVarName = 100;` and a description.

highs/mip/HighsMipSolver.cpp
```diff
@@ -837,52 +824,110 @@ void HighsMipSolver::run() {
...
+          stallNodes[i]++;
         }
-        considerHeuristics = false;
-        if (worker.getGlobalDomain().infeasible()) break;
-        if (dive(i, nodeLim)) break;
-        if (worker.search_ptr_->checkLimits(
-                worker.search_ptr_->getLocalNodes())) {
-          break;
+        if (!skip_separation) {
+          evaluateNode(i);
```
fwesselm, line 850: The return value of `evaluateNode` is ignored here? It could return `true`, right? Does this overcount the `total_nodes_explored`?

highs/mip/HighsMipSolver.cpp
```diff
@@ -837,52 +824,110 @@ void HighsMipSolver::run() {
...
+        worker.getLpRelaxation().setIterationLimit(iterlimit);
+        while (true) {
+          if (considerHeuristics &&
+              (skip_separation || nodeLim == maxNodesPerWorkerLim) &&
+              worker.getAllowHeuristics() &&
+              mipdata_->moreHeuristicsAllowed()) {
+            if (runHeuristics(i, stop)) break;
+            if (stop) break;
+          }
+          considerHeuristics = false;
```
fwesselm, line 880: `considerHeuristics` is never reset to `true` again?

highs/mip/HighsMipSolver.cpp
```diff
@@ -674,7 +662,14 @@ void HighsMipSolver::run() {
...
 
-  auto pruneNode = [&](HighsInt i) -> bool {
+  auto assignEarlyTermination = [&](HighsMipWorker& worker) {
+    if (worker.getGlobalDomain().infeasible() ||
+        mipdata_->lower_bound > worker.getOptimalityLimit()) {
+      mipdata_->updateWorkerEarlyTermination(worker);
+    }
+  };
+
+  auto pruneNode = [&](const HighsInt i, bool& infeasible) -> bool {
```
fwesselm, line 675: Just to double check, it is correct that `infeasible` is not reset to `false` by this lambda?

highs/mip/HighsMipSolver.cpp
```diff
@@ -609,62 +606,53 @@ void HighsMipSolver::run() {
...
-            mipdata_->nodequeue.popBestBoundNode());
-        lastLbLeave = numQueueLeaves;
-      } else {
-        HighsInt bestBoundNodeStackSize =
-            mipdata_->nodequeue.getBestBoundDomchgStackSize();
-        double bestBoundNodeLb = mipdata_->nodequeue.getBestLowerBound();
-        HighsNodeQueue::OpenNode nextNode(mipdata_->nodequeue.popBestNode());
-        if (nextNode.lower_bound == bestBoundNodeLb &&
-            (HighsInt)nextNode.domchgstack.size() == bestBoundNodeStackSize)
+  auto prepareNodes = [&](std::vector<HighsInt>& indices) {
```
fwesselm: `const std::vector<HighsInt>& indices`?

highs/mip/HighsMipSolver.cpp
```diff
@@ -837,52 +824,110 @@ void HighsMipSolver::run() {
...
         }
-
-        if (worker.search_ptr_->getLocalNodes() + nodes_explored >= plungeLimit)
-          break;
-        if (!mipdata_->parallelLockActive()) {
-          nodes_explored += worker.search_ptr_->getLocalNodes();
+        worker.getConflictPool().performAging();
+        HighsInt iterlimit = 10 * std::max(avgiter, mipdata_->avgrootlpiters);
+        iterlimit = std::max({HighsInt{10000}, iterlimit,
+                              HighsInt((3 * mipdata_->firstrootlpiters) / 2)});
```
fwesselm: `static_cast`? :)

highs/mip/HighsMipSolver.cpp
```diff
@@ -837,52 +824,110 @@ void HighsMipSolver::run() {
...
-      worker.getLpRelaxation().setIterationLimit(iterlimit);
+      worker.prepNodeIdx = 0;
+      int64_t total_nodes_explored = 0;
       bool considerHeuristics = true;
-      while (true) {
-        if (considerHeuristics && (skip_separation || nodeLim == kHighsIInf) &&
-            worker.getAllowHeuristics() && mipdata_->moreHeuristicsAllowed()) {
-          if (runHeuristics(i)) break;
+      bool stop = false;
+      while (worker.prepNodeIdx != worker.preparedNodes.size()) {
```
fwesselm, line 840: Is it worth reserving memory for `preparedNodes` in this lambda?

`worker.processedNodes.reserve(nodeLim + worker.preparedNodes.size());`

highs/mip/HighsSearch.cpp
```diff
@@ -730,20 +731,22 @@ void HighsSearch::currentNodeToQueue(HighsNodeQueue& nodequeue) {
   if (!prune) {
     std::vector<HighsInt> branchPositions;
```
fwesselm: I think this block is repeated. Would it be worth making a method?

```
 void HighsSearch::stashNodeToProcessed(double lb, double estimate, HighsInt depth) {
      std::vector<HighsInt> branchPositions;
      auto domchgStack = localdom.getReducedDomainChangeStack(branchPositions);
      mipworker.processedNodes.emplace_back(
          std::piecewise_construct,
          std::forward_as_tuple(std::move(domchgStack), std::move(branchPositions),
                                lb, estimate, depth),
          std::forward_as_tuple(countTreeWeight));
  }
```

And here we could have:

```
stashNodeToProcessed(std::max(nodestack.back().lower_bound,
                                localdom.getObjectiveLowerBound()),
                       nodestack.back().estimate, getCurrentDepth());
```

highs/mip/HighsSearch.cpp
```diff
@@ -771,12 +774,14 @@ void HighsSearch::openNodesToQueue(HighsNodeQueue& nodequeue) {
     if (!prune) {
       std::vector<HighsInt> branchPositions;
```
fwesselm:
```
stashNodeToProcessed(std::max(nodestack.back().lower_bound,
                                localdom.getObjectiveLowerBound()),
                       nodestack.back().estimate, getCurrentDepth());
```

highs/mip/HighsSearch.cpp
```diff
@@ -1738,49 +1736,17 @@ bool HighsSearch::backtrackPlunge(HighsNodeQueue& nodequeue) {
...
-        //   ancestorScore);
-        nodeToQueue = ancestorScoreInactive - ancestorScoreActive >
-                      nodeScore + getFeasTol();
-        break;
-      }
-    }
 
     if (nodeToQueue) {
       // if (!mipsolver.submip) printf("node goes to queue\n");
       std::vector<HighsInt> branchPositions;
```
fwesselm: `stashNodeToProcessed(nodelb, nodestack.back().estimate, getCurrentDepth() + 1);`

fwesselm (commented):
@Opt-Mucca, this looks great! Just a couple of uninformed comments from me.

highs/mip/HighsMipSolver.cpp
Opt-Mucca: I've added `constexpr HighsInt kMaxNodesPerWorker = 100` to the line before. 

I am yet to experiment with any other values, and would be confident that there's more performance to be had by tuning this (similar to the strong branching PR). However, I don't want to make it an open parameter because I'm still not 100% sure that the design won't be re-designed again in the somewhat near-future. (Example: We've now removed the original parameter idea of `num_workers_per_thread`)

highs/mip/HighsMipSolver.cpp
Opt-Mucca: It's set to `true` again when a new batch is created. I kept the logic similar to what we had originally.

Currently: Load two nodes onto a worker. Before the dive on the first node, run heuristics. After running heuristics, disable heuristics for all future dives based on this node (including backtracked ones) and never run heuristics based on the second node. 

The naming of some lambdas are now a bit confusing.... I'll probably now rename `processNode` as it's misleading. Edit: Renamed to `processWorkerNodes`

highs/mip/HighsMipSolver.cpp
Opt-Mucca: I don't think it's incorrect or correct. Any case where `infeasible = true` is being passed would mean that something's gone wrong and we haven't exited when we could've. I'll add an assert.

highs/mip/HighsMipSolver.cpp
Opt-Mucca: Yup! Nice spot.

highs/mip/HighsMipSolver.cpp
Opt-Mucca: Cast!

highs/mip/HighsMipSolver.cpp
Opt-Mucca: I'm assuming this was meant as `processedNodes`? I'm a bit scared about reserving infinite space.

Would it make sense to reserve this space when the worker is created? We should know `maxNodesPerWorkerLim` already?

highs/mip/HighsSearch.cpp
Opt-Mucca: It would be worth making a method. I've added it. Great suggestion!

highs/mip/HighsMipSolver.cpp
Opt-Mucca: I don't think this overcounts `total_nodes_explored`. The node should be processed by `pruneNode` and then increment the counter. Edit: The node is already sent of to `processedNodes`, but `pruneNode` should recognise this.

The `bool` return seems to be something left-over from early drafts (isn't even used on `master` currently)

highs/mip/HighsMipSolver.cpp
fwesselm: I don't think this is critical. I was just wondering if vectors could be pre-allocated here to avoid re-allocations.

highs/mip/HighsMipSolver.cpp
fwesselm: @Opt-Mucca, I am fine with keeping it, if ignoring the return value is OK.

PR #3195, opened 2026-08-03
Certify simplex optimality only from a refactorized basis

### LINKED ISSUE #3194: LP warm-started via setBasis returns status Optimal with a solution that is not the basic solution of its own returned basis
# LP warm-started via setBasis returns status Optimal with a solution that is not the basic solution of its own returned basis

**Version:** official v1.15.1 release binary and current `latest` (6dd8b97ccc) behave bit-identically. Deterministic, 10 of 10 runs, ~0.25 s per run.

## Summary

A pure LP (11877 rows, 21553 cols). Solved cold it returns Optimal at **818203670.005**, which two independent solvers and a from-text KKT recomputation confirm as the optimum. Warm-started with a basis that `Highs_setBasis` accepts (returns kOk), the same library on the same file returns Optimal at **818203761.820**, which is 91.81 dearer (relative 1.1e-7, but the objective coefficients reach 2.4e7, so this is far above tolerance dust). My understanding is that an LP's optimum does not depend on the starting basis.

## Reproducer

[warmstart-reproducer.zip](https://github.com/user-attachments/files/30658107/warmstart-reproducer.zip), [warmstart-basis.txt](https://github.com/user-attachments/files/30658114/warmstart-basis.txt) (column and row status arrays, one line each, numeric HighsBasisStatus codes), [hbasis-replay.c.txt](https://github.com/user-attachments/files/30658148/hbasis-replay.c.txt) (readModel, setBasis, run; ~40 lines of logic). All options at defaults.

```
hbasis-replay warmstart-wrong-optimal.mps /dev/null warmstart-basis.txt   ->  Optimal 818203761.81996739 (2654 iters)  WRONG
hbasis-replay warmstart-wrong-optimal.mps /dev/null -                     ->  Optimal 818203670.00500774 (3201 iters)  correct
```

The basis is one our application constructs from a related prior solve: 2595 basic columns plus 3656 basic rows, so 6251 basic in an 11877-row LP. It is alien in exactly the sense `HighsBasis` documentation permits, and `Highs_setBasis` accepts it with kOk.

## Why the Optimal result is false

HiGHS's own info reports max_primal_infeasibility 1.09e-11 and dual infeasibility and complementarity exactly 0. Recomputing from the MPS text and the returned solution instead: one row is violated by 1.17e-05, and 100 basic columns have reduced costs up to 0.025 when z = c - A'y is formed from HiGHS's own returned row duals (a basic column must have z = 0). The same checker run on the cold solution is clean, which validates its conventions.

## The defect is in the returned values, not the pivoting

Feeding the warm solve's FINAL basis (from `Highs_getBasis`) to a fresh Highs instance certifies Optimal in **0 iterations at the correct objective 818203670.005**. So the warm-started solve does reach an optimal basis. What it returns alongside that basis is stale: the entire 91.81 discrepancy is a single basic column with cost 7.8e6 reported at 1.17493e-05 where the true basic solution of that basis has it at 0 (plus knock-on drift through 80 related columns; the products c_j times the value error sum to exactly 91.81496). The same 1.17e-05 appears as the row residual above. Every returned nonbasic value does sit on its status bound, so the drift is confined to basic values, and the internal feasibility and KKT checks evidently measure against the same drifted values, which is why nothing trips.

## Discriminators (all measured on this input, default options unless stated)

- dual simplex (default) from the supplied basis: wrong, 818203761.820
- `dual_simplex_cost_perturbation_multiplier=0`, same basis: correct
- `simplex_strategy=4` (primal), same basis: correct
- logical basis via setBasis (all rows basic): correct
- supplied column statuses with logical rows: correct
- logical columns with supplied row statuses: correct
- cold: correct

So the failure needs the joint structure of this under-complete alien basis together with the dual simplex path with cost perturbation active.

As stated previously, I know the model is badly scaled (costs 1e-6 to 2.4e7; HiGHS warns on load). The range is inherent to the market rules being modelled and is what I have to work with. Hostile numerics can excuse slow progress or a degraded path, but not an internally inconsistent solution and basis pair certified Optimal at default tolerances, with the library's own KKT report showing zeros while the true residuals are 6 orders larger.

## Suggested direction

After termination from an alien start, recompute the primal values from the final factorization before reporting (the 0-iteration re-solve proves the correct values are available at that basis), or invalidate the value state carried out of the alien-basis completion.  I will try to create a PR with a fix.

EamonHetherton:
Fixes #3194.

When dual or primal simplex concludes optimality at a rebuild whose factor still carries updates, force a refactorization and reassess instead of certifying. On the LP of #3194 the values computed through the update chain are wrong by 91.81 in the objective while showing no infeasibility; the same basis refactorized yields the correct solution.

### DIFF
--- a/highs/simplex/HEkk.cpp
+++ b/highs/simplex/HEkk.cpp
@@ -1009,6 +1009,7 @@ HighsStatus HEkk::solve(const bool force_phase2) {
     analysis_.simplexTimerStart(SimplexTotalClock);
   dual_simplex_cleanup_level_ = 0;
   dual_simplex_phase1_cleanup_level_ = 0;
+  force_refactor_on_rebuild_ = false;
 
   previous_iteration_cycling_detected = -kHighsIInf;
 
@@ -1968,7 +1969,14 @@ void HEkk::computeDualObjectiveValue(const HighsInt phase) {
 
 bool HEkk::rebuildRefactor(HighsInt rebuild_reason) {
   // If no updates have been performed, then don't refactor!
-  if (info_.update_count == 0) return false;
+  if (info_.update_count == 0) {
+    force_refactor_on_rebuild_ = false;
+    return false;
+  }
+  if (force_refactor_on_rebuild_) {
+    force_refactor_on_rebuild_ = false;
+    return true;
+  }
   // Otherwise, refactor by default
   bool refactor = true;
   double solution_error = 0;
--- a/highs/simplex/HEkk.h
+++ b/highs/simplex/HEkk.h
@@ -35,6 +35,7 @@ class HEkk {
         iteration_count_(0),
         dual_simplex_cleanup_level_(0),
         dual_simplex_phase1_cleanup_level_(0),
+        force_refactor_on_rebuild_(false),
         previous_iteration_cycling_detected(-kHighsIInf),
         solve_bailout_(false),
         called_return_from_solve_(false),
@@ -205,6 +206,7 @@ class HEkk {
   HighsInt iteration_count_;
   HighsInt dual_simplex_cleanup_level_;
   HighsInt dual_simplex_phase1_cleanup_level_;
+  bool force_refactor_on_rebuild_;
 
   HighsInt previous_iteration_cycling_detected;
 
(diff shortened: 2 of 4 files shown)


jajhall:
I'll look at #3194 in due course, but the proposed certification of simplex optimality only from a refactorized basis hasn't been necessary previously, and would be a performance hit

PR #3196, opened 2026-08-03
Add extra safety net for relaxing out bounds

Opt-Mucca:
This fixes the bug presented in #3190 

What I think the problem is: A bug was fixed in #3149 to stop conflict analysis dropping an explanation that would've invalidated the conflict. This somehow has lead to the chain of conflicts exhausting itself, and having no guards against it stuff gets a bit funny when a `-1` index appears. My change just returns `false` when this happens.

### DIFF
--- a/highs/mip/HighsDomain.cpp
+++ b/highs/mip/HighsDomain.cpp
@@ -3043,10 +3043,8 @@ bool HighsDomain::ConflictSet::resolveLinearGeq(HighsCDouble M, double Mupper,
             M -= reasonDomchg.delta;
             // ++numDropped;
           } else {
-            while (locdomchg.pos != -1 &&
-                   relaxLb <= localdom.prevboundval_[locdomchg.pos].first)
-              locdomchg.pos = localdom.prevboundval_[locdomchg.pos].second;
-            if (locdomchg.pos == -1) return false;
+            if (!localdom.walkBoundChain(locdomchg.pos, relaxLb, HighsInt{1}))
+              return false;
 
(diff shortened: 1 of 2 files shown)


highs/mip/HighsDomain.cpp
fwesselm: Does this loop need to be updated as well?

fwesselm: Does this loop need to be updated as well?

fwesselm: Does this loop need to be updated as well?

fwesselm: Does this loop need to be updated as well?

fwesselm (approved):
Thank you, @Opt-Mucca, the changes are fine. Would it make sense to have a method for this? See [here](https://github.com/ERGO-Code/HiGHS/compare/fix-3190...fwesselm:HiGHS:fix-3190-fw).

Opt-Mucca:
@fwesselm I'm not sure whether the others can trigger, but I very much like having a function for this, and it's way safer than the existing assert! I'm super happy if you swap this PR for your changes.

fwesselm:
> @fwesselm I'm not sure whether the others can trigger, but I very much like having a function for this, and it's way safer than the existing assert! I'm super happy if you swap this PR for your changes.

@Opt-Mucca, I could just pull my branch into this one, if that is OK for you.

Opt-Mucca:
@fwesselm sounds good!

PR #3198, opened 2026-08-04
Bazel: do not try to get cuda if option is not specified

galabovaa:


galabovaa:
Closing this for now, will re-open after more testing

PR #3200, opened 2026-08-05
Parallel solves in HiPO

filikat:
- Forward, backward and diagonal solves with the factorisation now exploit multi-threading. This is achieved by dividing the supernodes into tasks, with each tasks processing "enough work" to make multi-threading worth it. From the supernodal elimination tree, a task elimination tree is derived giving the dependencies among tasks. 
- Diagonal solves are easily parallelised, as different supernodes require access to different entries, an there is no need to satisfy tree dependencies. Backward solves are easy too, because the tree is processed from roots to leaves, so no data race is possible. Forward solves are more tricky, because the tree is processed from leaves to roots, so local buffers are needed to avoid data races. A heuristic chooses whether to use serial or parallel solve, for each of the three types of solves, depending on the properties of the elimination tree.
- Given the many phases that can run in parallel within HiPO, two new options are added `hipo_parallel_force` and `hipo_parallel_forbid`, to force or forbid the use of specific parallel techniques. They are both bit maps and their details are explained in the documentation.
- General improvements and simplifications of the factorisation code


jajhall (changes requested):
I trust you on the technical changes, but in the documentation there are typos in `parallel.md` since it compiles to give 

> Setting hipo\\_parallel\_force (resp. hipo\\_parallel\\_forbid) to one of these values, or a sum of values, forces (resp. forbids) the use of parallelism in the corresponding phases. These options override any other behaviour enforced by other options. If a given phase is both forced and forbidden, the default behaviour is used instead. For instance, setting hipo\\_parallel\\_force to 81 = 1+16+64 and hipo\\_parallel\\_forbid to 68 = 4+64 forces the use of parallelism in the analyse phase and for building the normal equations values, and forbids it for the reordering of augmented system. 

Also, could you comment on the default value of  `hipo_parallel_type`, `hipo_parallel_force` and `hipo_parallel_forbid`, as well as adding a health warning about changing `hipo_parallel_force` or `hipo_parallel_forbid`: something along the lines of "the heuristics are very much likely to be correct, but force/forbid are ways of overriding them"

PR #3202, opened 2026-08-05
`HighsRunData` now correct when presolve is not used, or no reductions are achieved

jajhall:
Minor fix


galabovaa (approved):
Looks good to me!

PR #3203, opened 2026-08-06
fix(presolve): prevent floating-point underflow in dynamism-scaled tolerances

HuAliAsu:
### Problem
When solving MILP models containing large matrix coefficients (such as micrometer dimensions `10^12` or large Big-M indicators), HiGHS Presolve can incorrectly declare feasible models as `Infeasible` or fail to reduce valid forcing rows and columns.

### Technical Root Cause
In `HPresolve.cpp` (lines 4300 and 4587), `primal_feasibility_tolerance` and `dual_feasibility_tolerance` are scaled by dividing by the matrix `dynamism` ratio.

When dynamism is high (`>= 10^8`), `tolerance / dynamism` underflows below the IEEE 754 double precision machine epsilon (`2.22e-16`). For example, `1e-7 / 10^12 = 1e-19`.

As a result, comparisons like `std::abs(bound) <= 1e-19` fail due to standard floating-point noise (`~10^-15`), causing valid presolve reductions to fail and leading to false `Infeasible` status returns.

### The Fix
Clamp the dynamism-scaled tolerances to a lower bound of `1e-15` (IEEE 754 machine epsilon safeguard):
```cpp
std::max(double(primal_feastol / dynamism), 1e-15)
std::max(double(options->dual_feasibility_tolerance / dynamism), 1e-15)
```

### Reproducible LP Benchmark
```lp
Minimize
  obj: 2976800000000 x1 + 1000 y1
Subject To
  c1: 480000000000 z1 + 600000000000 z2 <= 2976800000000 x1
  c2: pos_x1 + 600000 - pos_x2 - 3000000 delta1 <= 0
Bounds
  0 <= pos_x1 <= 2440000
Binary
  x1 z1 z2 delta1
End
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,7 +22,7 @@ You can raise an [issue](https://github.com/ERGO-Code/HiGHS/issues) with HiGHS i
 
 HiGHS is open source for distribution rather than contribution. This applies particularly to the core C++ code of the solvers. However, there is definitely scope for external contribution to interfaces and documentation. If you want to contribute in this way, please open an issue before making a pull request, since pull requests to the HiGHS solvers will not normally be accepted. 
 
-Please note that any pull requests to HiGHS must be made to the `latest` branch, so that full testing can take place befor updating `master`. This ensures that `master` is always good to be downloaded by users. As a consequence, any changes should be developed as a branch of `latest` to reduce the chance of merge conflicts.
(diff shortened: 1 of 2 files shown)


Opt-Mucca (changes requested):
I don't see how this change can hurt. It needs to be formatted correctly though. Nice spot!

fwesselm:
I don't really like hard-coding `1e-15`, and I would prefer something like `std::numeric_limits<double>::epsilon()`.

jajhall:
> I don't really like hard-coding `1e-15`, and I would prefer something like `std::numeric_limits<double>::epsilon()`.

Agreed

PR #3206, opened 2026-08-07
Added `static_cast<Int>` to remove compiler warning in line 226 of `ipm/hipo/factorhighs/HybridSolveHandler.cpp`

jajhall:


filikat (approved):
Thanks

PR #3209, opened 2026-08-10
Keep implications between restarts

Opt-Mucca:
This PR changes the data structure holding implication information to a `HighsHashTree` (similar to vbs). It now transfers implication information between restarts, shrink-problem calls, and sub-mips. This has been a long-standing TODO in the code, but with low priority because it's not going to be too performance changing.

Pros:
- We now save implications from a cut or conflict that were not copied over between restarts. We also save implications from columns that may not have been re-probed on.
- Implications are now valid throughout the entire presolve code and can be used in more complicated reductions down the line.
- Can now add implications from multiple places without corrupting the data structure

Cons:
- There's some overhead to all the new data structures (probably minor, but still).

Testing locally I'm getting a 0.5% slowdown but solving two more instances. I'd say this is mostly noise and I'd wait for @fwesselm to try it on a larger test set.


highs/mip/HighsImplications.cpp
```diff
@@ -380,14 +359,37 @@ bool HighsImplications::runProbing(HighsInt col, HighsInt& numReductions) {
...
+    if (insertresult.first->ub == kHighsInf && implic.ub != kHighsInf)
+      ++numImplications;
+    insertresult.first->lb = std::max(implic.lb, insertresult.first->lb);
+    insertresult.first->ub = std::min(implic.ub, insertresult.first->ub);
+  } else {
+    if (implic.lb != -kHighsInf) ++numImplications;
+    if (implic.ub != kHighsInf) ++numImplications;
+  }
+
+  reverseImplications[implCol].insert_or_get(std::floor(loc / 2));
```
fwesselm: Is the `std::floor` needed? I think `loc / 2` would work.

highs/mip/HighsImplications.cpp
fwesselm, line 351: Is it worth checking for infeasibility here after `changeBound`?

fwesselm (commented):
Looks good @Opt-Mucca!

highs/mip/HighsImplications.cpp
Opt-Mucca: Yup. This would change if we merge your branch already?

highs/mip/HighsImplications.cpp
Opt-Mucca: It definitely is....... I can make the `for_each` stop early too. I'll add this in a few mins.

highs/mip/HighsImplications.cpp
fwesselm: @Opt-Mucca, thanks, I just merged this!

highs/mip/HighsImplications.cpp
Opt-Mucca: Added!

Opt-Mucca:
This change has come out as performance neutral (0.01% performance improving), but is mostly there to enable future presolve rules to use any implication information. Therefore it should be merged.

PR #3210, opened 2026-08-10
Changes to presolve initial sweep

fwesselm:
- Remove `std::vector<bool>`
- Use `HighsCDouble` for activity bounds (similar to `HighsLinearSumBounds`)

### DIFF
--- a/highs/presolve/HPresolve.cpp
+++ b/highs/presolve/HPresolve.cpp
@@ -274,11 +274,15 @@ bool HPresolve::isRanged(HighsInt row) const {
           model->row_upper_[row] != kHighsInf);
 }
 
+bool HPresolve::isRedundant(HighsInt row, double sumLower,
+                            double sumUpper) const {
+  return sumLower >= model->row_lower_[row] - primal_feastol &&
+         sumUpper <= model->row_upper_[row] + primal_feastol;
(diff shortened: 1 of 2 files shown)


jajhall (approved):
Thanks. I thought it was OK to compute infinite implied bounds explicitly, rather than tracking them with the `HighsBool` vectors `implied_row_*_finite`, but fair enough

fwesselm:
> Thanks. I thought it was OK to compute infinite implied bounds explicitly, rather than tracking them with the `HighsBool` vectors `implied_row_*_finite`, but fair enough

I think using `HighsCDouble` is beneficial if the activity bounds are large.

`kHighsInf` cannot be converted to `HighsCDouble`, at least I ran into issues last time I tried, hence the additional indicator vectors.

jajhall:
> kHighsInf cannot be converted to HighsCDouble, at least I ran into issues last time I tried, hence the additional indicator vectors.

Very true

fwesselm:
> > kHighsInf cannot be converted to HighsCDouble, at least I ran into issues last time I tried, hence the additional indicator vectors.
> 
> Very true

However, I can still try to get rid of the `HighsBool` vectors. I will update the PR.

PR #3211, opened 2026-08-10
Update CONTRIBUTING.md

galabovaa:


jajhall (commented):
This is good as far as it goes, but when we have templated issue and PR submission, some comment on this will be required in `CONTRIBUTING.md`

PR #3213, opened 2026-08-11
Reduce strong branching budget before two restarts

Opt-Mucca:
From the mouth of some commercial solver developers, it is purportedly a standard technique to "reduce the strong branching budget until the first or second restart". I therefore made the budget a quarter of the default until two restarts have happened. From testing locally, this seems to have extremely good results, which now makes me doubt my setup completely. I'm getting a 7% time reduction, 9% LP iteration reduction, and an 8% increase in number of nodes over all-optimal. I'm also solving a couple more instances. This should be tested using @fwesselm setup.


fwesselm:
@Opt-Mucca, I'll start testing this shortly and will report back.

jajhall:
> From testing locally, this seems to have extremely good results, which now makes me doubt my setup completely. I'm getting a 7% time reduction, 9% LP iteration reduction, and an 8% increase in number of nodes over all-optimal. I'm also solving a couple more instances.

Great to hear the seasoned MIP developer's instinct that there's no such thing as an easy 7% improvement!

fwesselm (approved):
I like this change, @Opt-Mucca!

Would it make sense to increase the threshold with increasing number of restarts:

`if (mipsolver.mipdata_->numRestarts <= 2) sbmaxiters = ((mipsolver.mipdata_->numRestarts + 1) * sbmaxiters) >> 2;`

Opt-Mucca:
@fwesselm My gut is telling me that it wouldn't be better, but I'd say we should probably be testing many variations of such a rule.

Opt-Mucca:
For future reference: This was a 4.1% performance improvement over a larger test set with a larger time limit. My explanation would be that the change disproportionately improves mid-size instances by a marginal amount (can see in the test set of @fwesselm where there's not that many actual major wins or losses, but there's tonnes of minor improvements). The change was also more performant with `<= 2` instead of `<= 1`.

PR #3215, opened 2026-08-12
Reduce amount of duplicate code in `HighsSymmetry`

fwesselm:
Added methods to reduce amount of similar/duplicate code in `HighsSymmetry` class


Opt-Mucca (approved):
Looks good!

fwesselm:
> Looks good!

@Opt-Mucca, I added a test for this. I hope you agree that this is worth having.

Opt-Mucca:
@fwesselm tests are rarely worth not having. I didn't expect you manage to more code simplifications too! Happy the new changes as well.

PR #3218, opened 2026-08-17
Presolve infrastructure for adding rows/columns

fwesselm:
- This is just the infrastructure part from #3191.
- `ImpliedEquation` reduction type was added to properly handle conversion of inequality constraints into equality constraints.
- Benchmarking (on 100 MIPs) did not reveal any unintended side effects.
- Added a test to exercise new code.


highs/presolve/HighsPostsolveStack.cpp
```diff
@@ -41,29 +45,14 @@ void HighsPostsolveStack::compressIndexMaps(
 void HighsPostsolveStack::compressRowIndexMap(
     const std::vector<HighsInt>& newRowIndex) {
   compressIndexMap(newRowIndex, this->origRowIndex);
+  compressIndexMap(newRowIndex, this->origRowType);
```
Opt-Mucca: Placeholder: This looks incorrect.

highs/presolve/HighsPostsolveStack.h
```diff
@@ -347,38 +397,63 @@ class HighsPostsolveStack {
...
+          origRowIndex[newSize] = origRowIndex[i];
+          origRowType[newSize] = origRowType[i];
+        }
+        ++newSize;
+      }
     }
-
-    origRowIndex.resize(origRowIndexSize);
+    origRowIndex.resize(newSize);
+    origRowType.resize(newSize);
```
Opt-Mucca, line 406: Does this also need to change `nextRowIndex`?

highs/presolve/HPresolve.cpp
```diff
@@ -2267,6 +2261,102 @@ void HPresolve::addToMatrix(const HighsInt row, const HighsInt col,
...
+  for (HighsInt i = 0; i < num_rows; i++) {
+    // new row index
+    HighsInt row = oldNumRows + i;
+
+    // add non-zeros
+    for (size_t j = 0; j < row_indices[i].size(); j++)
+      addToMatrix(row, row_indices[i][j], row_values[i][j]);
+
+    // add row singleton
+    if (rowsize[row] == 1) singletonRows.push_back(row);
```
Opt-Mucca, line 2343: Just to check because I've been fighting this on another branch: Would `singletonRows` (potentially `equations` too but less sure) get screwed up if we were passing values smaller than `small_matrix_value`? The "non-zero" then wouldn't actually exist in the matrix because `addToMatrix`* didn't add it.

Opt-Mucca (commented):
I am a hugeeeeeeeeeeeeeeeeee fan of this PR!!!!! It somehow doesn't complicate the code at all, and even does a latent TODO for equation upgrades!

I've only found two potential issues. Both could be me missing the larger picture (not unlikely), but I think they're potential bugs.

highs/presolve/HighsPostsolveStack.h
fwesselm: No, `origRowIndex` and `origRowType` just the original row index and type for each row that is currently in the (presolved) model, so these just have `newSize` elements here. `nextRowIndex` is the next available index for a new / appended row.

highs/presolve/HPresolve.cpp
fwesselm: I think the current `addToMatrix` method does not check the size of the new coefficient (`pos == -1` branch). Should this be added there?

highs/presolve/HighsPostsolveStack.h
Opt-Mucca: Then I've misunderstood something...... When does `isModelRow` ever return false? Shouldn't this be checking if the row is in the model?

highs/presolve/HPresolve.cpp
Opt-Mucca: Ahhhhhhhhhh. It was only checking the sum of the two values...... I am absolutely blind. I guess this is fine then?

fwesselm: I think you have a point because `HPresolve::addToMatrix(const HighsInt row, const HighsInt col, const double val)` does not check `val` against `small_matrix_value`. I tried to change this in another PR some time ago and observed some behavior changes. I am fine with investigating this.

highs/presolve/HighsPostsolveStack.h
fwesselm: Thanks for asking these questions. Yes, `isModelRow` will return `false` for cuts that have been added (after presolve but before a restart). After a restart cuts are converted into model rows (from the postsolve point of view) by calling `appendCutsToModel`.

fwesselm: @Opt-Mucca, would you prefer a different name for `isModelRow`. Maybe `hasPostsolveIndex`?

fwesselm: @Opt-Mucca, I deleted my previous comments. You were right, since the new code resizes the postsolved solution vectors to size `nextRowIndex`, `isModelRow` was never returning false. I removed `isModelRow`, thereby simplifying postsolve a little. I hope you agree with these updates.

highs/presolve/HPresolve.cpp
Opt-Mucca: Happy if this gets put down somewhere for future investigation! I don't see a reason for it to slow down this PR.

highs/presolve/HighsPostsolveStack.h
Opt-Mucca: Glad to see the misunderstanding led to something, but I'm still slightly lost (although I'd just simply trust the change). How does postsolve now make sure the problem is the same dimension as the original space? Isn't it now potentially acting on cuts?

fwesselm: @Opt-Mucca, good question again:

1. `latest` branch: Solution vectors are resized to `origNumRow`. Each `undo` method uses `isModelRow` to check that `row < row_value.size()`. After a restart, cut row indices are `>= origNumRow`, so `isModelRow` returns `false` and those rows are skipped in postsolve.
2. This PR: Solution vectors are resized to `nextRowIndex` (which includes slots for cuts and appended rows). `isModelRow` is removed, all undo operations execute unconditionally. After postsolve, `resize(origNumRow)` discards the extra slots.

So, yes, postsolve may operate on cut indices but the corresponding entries in the solution are discarded at the end of postsolve.

The design rationale is that postsolve can operate in a single index space without needing to distinguish row types.

The downside is that more memory is used by postsolve temporarily (for slots that are later discarded).

Opt-Mucca: @fwesselm beautiful answer! I'm going to accept and merge this now.

Opt-Mucca (approved):

PR #3219, opened 2026-08-17
Presolve/postsolve debug utilities

jajhall:
Introduces the struct `MipViolation` to gather data from `HighsMipSolver::solutionFeasible` rather than just `bound_violation_`, `row_violation_`, `integrality_violation_`. This allows more informative logging of "untransformed violations" when `log_dev_level` is positive. 

Introduced `reportOriginalPresolvedCol(report_col, sol);` to log model data (in both the original and presolved problem) about a column `report_col` (when non-negative) before calling `postSolveStack.undoPrimal` in `HighsMipSolverData::transformNewIntegerFeasibleSolution` since model data is unknown in `postSolveStack.undo`.

Introduced `getPresolvedColumnIndex` and `getPresolvedRowIndex` in `HighsPostSolveStack.h` so that the index of a column/row in the presolved model can be determined, allowing the model data for the column/row in the presolved problem to be logged.

Introduced `presolveTypeToString` in `HighsPostSolveStack.h` so that presolve reductions can be logged more clearly. 

Introduced `last_reduction_ ` as a data member of `HighsPresolve` so that calls to `HighsPresolve::checkLimits` following a reduction can be identified, simplifying the task of stopping VScode before the last (error-causing) reduction has been performed.

### DIFF
--- a/highs/lp_data/Highs.cpp
+++ b/highs/lp_data/Highs.cpp
@@ -3999,7 +3999,7 @@ HighsPostsolveStatus Highs::runPostsolve() {
     return HighsPostsolveStatus::kNoPrimalSolutionError;
   const bool have_dual_solution =
       presolve_.data_.recovered_solution_.dual_valid;
-  const HighsInt report_3040_col = -21792;
+  const HighsInt report_3040_col = -578;
   presolve_.data_.postSolveStack.undo(
       options_, presolve_.data_.recovered_solution_,
       presolve_.data_.recovered_basis_, report_3040_col);
--- a/highs/mip/HighsMipSolver.cpp
+++ b/highs/mip/HighsMipSolver.cpp
@@ -1381,15 +1381,22 @@ bool HighsMipSolver::solutionFeasible(const HighsLp* lp,
     assert(col_value.size() == static_cast<size_t>(lp->num_col_));
 
   // Lambda to update violation record
-  auto updateViolation = [&](double v, HighsInt index, HighsInt& num, double& max_v, HighsInt& max_index) {
-      if (v <= 0) return;
-      if (v > mip_feasibility_tolerance) num++;
-      if (v > max_v) { max_v = v; max_index = index; }
+  auto updateViolation = [&](double v, HighsInt index, HighsInt& num,
+                             double& max_v, HighsInt& max_index) {
+    if (v <= 0) return;
+    if (v > mip_feasibility_tolerance) num++;
+    if (v > max_v) {
+      max_v = v;
+      max_index = index;
+    }
   };
 
-  auto updatePrimalViolation = [&](double value, HighsInt index, bool is_column) {
-    const double lower = is_column ? lp->col_lower_[index] : lp->row_lower_[index];
-    const double upper = is_column ? lp->col_upper_[index] : lp->row_upper_[index];
+  auto updatePrimalViolation = [&](double value, HighsInt index,
+                                   bool is_column) {
+    const double lower =
+        is_column ? lp->col_lower_[index] : lp->row_lower_[index];
+    const double upper =
+        is_column ? lp->col_upper_[index] : lp->row_upper_[index];
     double primal_infeasibility;
     if (value < lower - mip_feasibility_tolerance) {
       primal_infeasibility = lower - value;
@@ -1399,43 +1406,26 @@ bool HighsMipSolver::solutionFeasible(const HighsLp* lp,
       return;
     }
     if (is_column) {
-      updateViolation(primal_infeasibility, index,
-		      num_bound_violations,
-		      bound_violation,
-		      col_of_max_bound_violation);
+      updateViolation(primal_infeasibility, index, num_bound_violations,
+                      bound_violation, col_of_max_bound_violation);
     } else {
-      updateViolation(primal_infeasibility, index,
-		      num_row_violations,
-		      row_violation,
-		      row_of_max_row_violation);
-    }    
+      updateViolation(primal_infeasibility, index, num_row_violations,
+                      row_violation, row_of_max_row_violation);
+    }
   };
-  
+
+  // Check column integrality and feasibility
+  bool is_column = true;
   for (HighsInt i = 0; i != lp->num_col_; ++i) {
     const double value = col_value[i];
     obj += static_cast<HighsCDouble>(lp->col_cost_[i]) * value;
     if (lp->integrality_[i] == HighsVarType::kInteger)
-      updateViolation(fractionality(value), i,
-		      num_integrality_violations,
-		      integrality_violation,
-		      col_of_max_integrality_violation);
-    const double lower = lp->col_lower_[i];
-    const double upper = lp->col_upper_[i];
-    double primal_infeasibility;
-    if (value < lower - mip_feasibility_tolerance) {
-      primal_infeasibility = lower - value;
-    } else if (value > upper + mip_feasibility_tolerance) {
-      primal_infeasibility = value - upper;
-    } else {
-      continue;
-    }
-    updateViolation(primal_infeasibility, i,
-		    num_bound_violations,
-		    bound_violation,
-		    col_of_max_bound_violation);
+      updateViolation(fractionality(value), i, num_integrality_violations,
+                      integrality_violation, col_of_max_integrality_violation);
+    updatePrimalViolation(value, i, is_column);
   }
 
-  // Check row feasibility if there are a positive number of rows.
+  // Check row feasibility if there is a positive number of rows
   //
   // If there are no rows and pass_row_value is nullptr, then
   // row_value_p is also nullptr since row_value is not resized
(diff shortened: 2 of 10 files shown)


highs/lp_data/Highs.cpp
```diff
@@ -3999,7 +3999,7 @@ HighsPostsolveStatus Highs::runPostsolve() {
     return HighsPostsolveStatus::kNoPrimalSolutionError;
   const bool have_dual_solution =
       presolve_.data_.recovered_solution_.dual_valid;
-  const HighsInt report_3040_col = -21792;
+  const HighsInt report_3040_col = -578;
```
fwesselm, line 4002: Is this change needed?

highs/mip/HighsMipSolver.cpp
```diff
@@ -1368,8 +1384,14 @@ bool HighsMipSolver::solutionFeasible(const HighsLp* lp,
     obj += static_cast<HighsCDouble>(lp->col_cost_[i]) * value;
 
     if (lp->integrality_[i] == HighsVarType::kInteger) {
-      integrality_violation =
-          std::max(fractionality(value), integrality_violation);
+      double violation = fractionality(value);
```
fwesselm: Local variable `violation `shadows the `MipViolation& violation`. Not a bug, but maybe confusing.

highs/mip/HighsMipSolver.cpp
```diff
@@ -1382,7 +1404,14 @@ bool HighsMipSolver::solutionFeasible(const HighsLp* lp,
     } else
       continue;
 
-    bound_violation = std::max(bound_violation, primal_infeasibility);
+    if (primal_infeasibility) {
```
fwesselm: There are three similar blocks here. Maybe make a lambda like:

```
auto updateViolation = [&](double v, HighsInt index, HighsInt& num, double& max_v, HighsInt& max_index) {
      if (v <= 0) return;
      if (v > mip_feasibility_tolerance) num++;
      if (v > max_v) { max_v = v; max_index = index; }
  };

  Then the call sites are just:

  updateViolation(fractionality(value), i, num_integrality_violations, integrality_violation, col_of_max_integrality_violation);
  updateViolation(primal_infeasibility, i, num_bound_violations, bound_violation, col_of_max_bound_violation);
  updateViolation(primal_infeasibility, i, num_row_violations, row_violation, row_of_max_row_violation);
```

highs/presolve/HighsPostsolveStack.h
```diff
@@ -680,37 +752,52 @@ class HighsPostsolveStack {
...
+        printf("%s\n", ss.str().c_str());
+      }
     };
 
-    auto reportColLogging = [&](const HighsInt reduction) {
+    // Initialise to illegal values so that initial values are logged
+    double report_col_value = kHighsInf;
+
+    // Lambda for logging the solution for a specific column whenever its value
+    changes auto reportColLogging = [&](const HighsInt reduction) {
```
fwesselm: `changes ` belongs to previous line?

highs/presolve/HPresolve.cpp
```diff
@@ -101,6 +101,9 @@ void HPresolve::setInput(HighsLp& model_, const HighsOptions& options_,
                 static_cast<int>(this->reductionLimit));
   }
   this->in_initial_sweep_ = false;
+  // last_reduction_ is used to identify when HPresolve::checkLimits
+  // is called for the first time following a reduction
+  this->last_reduction_ = -kHighsIInf;
```
fwesselm: `last_reduction_ ` is declared as `size_t`, so this would wrap around to a huge value. Maybe use  `std::numeric_limits<size_t>::max()` instead?

Also, `last_reduction_ ` is set, but never used.

highs/presolve/HighsPostsolveStack.h
```diff
@@ -277,6 +327,28 @@ class HighsPostsolveStack {
...
 
+  HighsInt getPresolvedIndex(
+      const HighsInt original_index,
+      const std::vector<HighsInt>& original_indices) const {
+    auto it = std::find(original_indices.begin(), original_indices.end(),
+                        original_index);
+    // If found, calculate the index by subtracting the base iterator
+    if (it != original_indices.end()) {
+      HighsInt presolved_index = it - original_indices.begin();
+      assert(original_indices[presolved_index] == original_index);
```
fwesselm: Would you like to keep this `assert` (to see if `std::find` is working)?

If not, this could be simplied:

```
    if (it != original_indices.end())
      return static_cast<HighsInt>(it - original_indices.begin());
    return -1;
```

highs/presolve/HighsPostsolveStack.h
```diff
@@ -719,19 +806,26 @@ class HighsPostsolveStack {
...
+                                  utilBasisStatusToString(basis.col_status[report_col]).c_str());
+      printf("%s\n", ss.str().c_str());
       report_col_value = col_value;
-      report_col_dual = col_dual;
-      report_col_status = col_status;
     };
+
     if (report_col >= 0) reportColLogging(-1);
+
+    size_t check_reduction = -kHighsIInf;
```
fwesselm: This wraps around to a huge value, which may be the intention.

I think using ` std::numeric_limits<size_t>::max()` would be clearer.

fwesselm (approved):
@jajhall, thanks for this!

highs/lp_data/Highs.cpp
jajhall: No, it's just an artefact of debugging. So long as it's negative no logging occurs

highs/mip/HighsMipSolver.cpp
jajhall: Fixed now we have the lambda

highs/mip/HighsMipSolver.cpp
jajhall: Indeed, I'll add this

highs/presolve/HighsPostsolveStack.h
jajhall: Corrected now

highs/mip/HighsMipSolver.cpp
jajhall: Added a `updatePrimalViolation` that eliminates further duplicated code

highs/presolve/HighsPostsolveStack.h
jajhall: True, I added the assert to check my use of it. I'll use your simplification

highs/presolve/HighsPostsolveStack.h
jajhall: Good spot, and it wasn't my intention!

Since I'm only checking for `check_reduction` to be equal to a specific reduction, switching it off by making it infinitely large is fine.

In `highs/lp_data/HConst.h` we have 

> `const size_t kHighsSize_tInf = (std::numeric_limits<size_t>::max)();`

I had the "`check_col`" logic in my mind when some checking/logging is only switched off by making it negative.

highs/presolve/HPresolve.cpp
jajhall: Good spot, and it wasn't my intention! 

I wanted it to be large and negative (which is, of course, not possible with `size_t`) so `last_reduction_` would be less than `numreductions` in the first call to `HPresolve::checkLimits`. In practice it doesn't matter if `last_reduction_` is initialised to 0.

highs/mip/HighsMipSolver.cpp
```diff
@@ -1349,43 +1350,82 @@ std::array<char, 128> getGapString(const double gap_,
...
-    bound_violation = std::max(bound_violation, primal_infeasibility);
+  // Check column integrality and feasibility
+  bool is_column = true;
+  for (HighsInt i = 0; i != lp->num_col_; ++i) {
+    const double value = col_value[i];
+    obj += static_cast<HighsCDouble>(lp->col_cost_[i]) * value;
+    if (lp->integrality_[i] == HighsVarType::kInteger)
+      updateViolation(fractionality(value), i, num_integrality_violations,
+                      integrality_violation, col_of_max_integrality_violation);
+    updatePrimalViolation(value, i, is_column);
```
fwesselm, line 1425: Just pass `true` here?

highs/mip/HighsMipSolver.cpp
```diff
@@ -1401,21 +1441,9 @@ bool HighsMipSolver::solutionFeasible(const HighsLp* lp,
...
-      } else if (value > upper + mip_feasibility_tolerance) {
-        primal_infeasibility = value - upper;
-      } else
-        continue;
-
-      row_violation = std::max(row_violation, primal_infeasibility);
-    }
+    bool is_column = false;
+    for (HighsInt i = 0; i != lp->num_row_; ++i)
+      updatePrimalViolation(row_value_p[i], i, is_column);
```
fwesselm, line 1446: And pass `false` here? (And remove `is_column` altogether.)

fwesselm (approved):

highs/mip/HighsMipSolver.cpp
jajhall: I feel that it's clearer to pass a named identifier than just true/false

PR #3220, opened 2026-08-18
Fix #3216

### LINKED ISSUE #3216: Presolve appears to prevent feasible incumbents being reconstructed, leading to massive search tree growth
Hi HiGHS Team,

I'm seeing a significant behavioural difference between solving a binary MIP with presolve enabled vs disabled.

I am running this through the C# HiGHS.Native 1.15.1 interface but have also tested the attached MPS file with a local HiGHS exe built from the main branch.

With presolve enabled:
- HiGHS repeatedly reports:
```
WARNING: Solution with objective ... has untransformed violations:
bound = 0;
integrality = 0;
row = 1
```
- No feasible incumbent is ever accepted.
- BestSol remains inf.
- The branch-and-bound tree grows rapidly (>100k nodes).
- The solve eventually times out or is OOM-killed.

With presolve disabled:
- No untransformed violation warnings occur.
- Feasible incumbents are found immediately.
- The model solves to optimality in ~14 seconds.
- Only 32 nodes are explored.

Attached to this issue you will find
- Problem-causing "model.mps" (uploaded as "model.txt")
- Results expected with presolve on "solver-log-presolve-on.txt"
- Results expected with presolve off "solver-log-pesolve-off.txt"

Side note, you will notice that our solver logs start by spamming: 

"Rows:           1 upper bounds greater than or equal to        1e+20 are treated as +Infinity"

This is because we set NULLs to HiGHS `infinite_bound` option. Is this the right way of setting up an upper bound or should we be using Highs Infinity?

Many thanks,

Gbtank

---

[solver-log-presolve-on.txt](https://github.com/user-attachments/files/31022404/solver-log-presolve-on.txt)

[model.txt](https://github.com/user-attachments/files/31022428/model.txt)

[solver-log-presolve-off.txt](https://github.com/user-attachments/files/31022410/solver-log-presolve-off.txt)

fwesselm:
Fix #3216 
- Correctly fix last non-fixed entry in equality cliques.
- Do not exit early when making bound modifications in `HPresolve::finaliseProbing`.
- Added a unit test.

### DIFF
--- a/highs/mip/HighsCliqueTable.cpp
+++ b/highs/mip/HighsCliqueTable.cpp
@@ -836,6 +836,17 @@ void HighsCliqueTable::removeClique(HighsInt cliqueid) {
   numEntries -= len;
 }
 
+void HighsCliqueTable::fixLastActiveAndRemove(HighsDomain& globaldom,
+                                              HighsInt cliqueid) {
+  if (cliques[cliqueid].equality && cliques[cliqueid].numActive() == 1)
+    for (HighsInt i = cliques[cliqueid].start; i != cliques[cliqueid].end; ++i)
+      if (!globaldom.isFixed(cliqueentries[i].col)) {
+        fixCol(globaldom, cliqueentries[i].complement(), false);
+        break;
+      }
+  removeClique(cliqueid);
+}
+
 void HighsCliqueTable::extractCliques(
     const HighsMipSolver& mipsolver, std::vector<HighsInt>& inds,
     std::vector<double>& vals, std::vector<int8_t>& complementation, double rhs,
@@ -1519,7 +1530,8 @@ void HighsCliqueTable::processInfeasibleVertices(HighsDomain& globaldom) {
       // may be found by probing and will be deleted upon rebuild anyways
       vHashLists.for_each([&](HighsInt cliqueid) {
         cliques[cliqueid].numZeroFixed += 1;
-        if (cliques[cliqueid].numActive() <= 1) removeClique(cliqueid);
+        if (cliques[cliqueid].numActive() <= 1)
+          fixLastActiveAndRemove(globaldom, cliqueid);
       });
       continue;
     }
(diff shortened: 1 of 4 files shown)


highs/mip/HighsCliqueTable.cpp
```diff
@@ -836,6 +836,17 @@ void HighsCliqueTable::removeClique(HighsInt cliqueid) {
   numEntries -= len;
 }
 
+void HighsCliqueTable::fixLastActiveAndRemove(HighsDomain& globaldom,
+                                              HighsInt cliqueid) {
+  if (cliques[cliqueid].equality && cliques[cliqueid].numActive() == 1)
+    for (HighsInt i = cliques[cliqueid].start; i != cliques[cliqueid].end; ++i)
+      if (!globaldom.isFixed(cliqueentries[i].col)) {
+        fixCol(globaldom, cliqueentries[i].complement(), false);
```
Opt-Mucca, line 844: How do we know which way we can fix the clique variable? We don't check what the other variables in the clique have been fixed to?

Opt-Mucca (commented):
Changes look good. I've left a comment because I'm not sure how all cases are handled.

highs/mip/HighsCliqueTable.cpp
fwesselm: @Opt-Mucca I hope I am not missing something. My naive understanding was that we have equality cliques with one element `x = 1` or `(1 - x) = 1`. We then fix accordingly. I added a test for the opposite case where the last active variable can be fixed to zero.

Opt-Mucca: Ignore me. I think I'm being overly cautious. Happy with the change!

Opt-Mucca (approved):

jajhall (approved):
I can't comment on the technical side of clique removal, but if this fixes #3216 and has two unit tests to validate the fix and for code coverage, then it's good with me

PR #3228, opened 2026-08-24
Make Xcode 26.6 happy. Move HMW destructor to .cpp

### LINKED ISSUE #3227: HighsMipWorker.h: inline destructor calls reset() on unique_ptr of incomplete HighsSearch — build fails on Xcode 26 / newer libc++
Version: v1.15.1
Language: C++ (built via Conan)

MIP — the issue is in `src/highs/mip/HighsMipWorker.h`.

macOS with **Xcode 26.6** (Apple Clang, newer modular libc++). Not seen on older Xcode / your `macos-latest` CI, which uses an older libc++.

**What did you do?**
Steps to reproduce the behaviour:
1. Build HiGHS from source on macOS with Xcode 26.x (Debug, unity build as HiGHS configures by default).
2. Compilation of the MIP sources fails.

The cause: `HighsMipWorker.h` holds a `std::unique_ptr<HighsSearch>` where `HighsSearch` is only forward-declared, and defines the destructor **inline** in the header:

```cpp
class HighsSearch;                          // line 21 — forward declaration only
...
std::unique_ptr<HighsSearch> search_ptr_;   // line 71
...
~HighsMipWorker() {                         // line 89 — inline destructor
  search_ptr_.reset();                      // line 90 — requires complete HighsSearch
  sepa_ptr_.reset();
}
```

Destroying the `unique_ptr` instantiates `std::default_delete<HighsSearch>`, which requires a complete type. Xcode 26's libc++ enforces this with a hard `static_assert`; older libc++/libstdc++ only warned, which is why it built before.

A clean compile, as on older toolchains.

**What did you see instead?**
```
.../c++/v1/__memory/unique_ptr.h:75:19: error: invalid application of 'sizeof' to an incomplete type 'HighsSearch'
   75 |     static_assert(sizeof(_Tp) >= 0, "cannot delete an incomplete type");
note: in instantiation of member function 'std::default_delete<HighsSearch>::operator()' requested here
note: in instantiation of member function 'std::unique_ptr<HighsSearch>::reset' requested here
   90 |     search_ptr_.reset();
.../mip/HighsMipWorker.h:21:7: note: forward declaration of 'HighsSearch'
   21 | class HighsSearch;
make[3]: *** [highs/CMakeFiles/highs.dir/Unity/unity_3_cxx.cxx.o] Error 1
```

**Anything else we should know about your project / environment**
Suggested fix — declare the destructor in the header and define it out-of-line in `HighsMipWorker.cpp`, where `HighsSearch.h` is already included (canonical handling for `unique_ptr` to a forward-declared type):

```cpp
// HighsMipWorker.h
~HighsMipWorker();
```
```cpp
// HighsMipWorker.cpp  (HighsSearch is complete here)
HighsMipWorker::~HighsMipWorker() = default;
```

Alternatively `#include "mip/HighsSearch.h"` in the header.

Opt-Mucca:
## Description

Moves `HighsMipWorker` destructor from `HighsMipWorker.h` to `HighsMipWorker.cpp`. Destroying the `std::unique_pointer` to `HighsSearch` apparently needs more than a forward declaration for `Xcode 26.6`.

## Checklist

- [x] Tests are passing

Closes #3227

### DIFF
--- a/highs/mip/HighsMipWorker.cpp
+++ b/highs/mip/HighsMipWorker.cpp
@@ -35,6 +35,11 @@ HighsMipWorker::HighsMipWorker(const HighsMipSolver& mipsolver,
   sepa_ptr_->setLpRelaxation(lp_);
 }
 
+HighsMipWorker::~HighsMipWorker() {
+  search_ptr_.reset();
+  sepa_ptr_.reset();
+}
+
 const HighsMipSolver& HighsMipWorker::getMipSolver() const {
   return mipsolver_;
 }
--- a/highs/mip/HighsMipWorker.h
+++ b/highs/mip/HighsMipWorker.h
@@ -87,10 +87,7 @@ class HighsMipWorker {
                  HighsDomain* domain, HighsCutPool* cutpool,
                  HighsConflictPool* conflictpool, HighsPseudocost* pseudocost);
 
-  ~HighsMipWorker() {
-    search_ptr_.reset();
-    sepa_ptr_.reset();
-  }
+  ~HighsMipWorker();
 
   void resetSearch();
 


filikat (approved):
Great!
This was undefined behaviour before. From https://en.cppreference.com/cpp/memory/unique_ptr: "`unique_ptr` may be constructed for an incomplete type `T`[...]. If the default deleter is used, `T` must be complete at the point in code where the deleter is invoked, which happens in the [...] reset member function of `unique_ptr`."

Opt-Mucca:
@mathgeekcoder If I'm understanding the failing test correctly, then I think HiGHS is solving the instance too quickly and the `isRunning` check is against an already solved instance?

mathgeekcoder:
> @mathgeekcoder If I'm understanding the failing test correctly, then I think HiGHS is solving the instance too quickly and the `isRunning` check is against an already solved instance?

Yeah, unfortunately I believe that's the case.  Honestly, it's a bit of a hacky test (not my proudest work). We can either hack it (by increasing N to make the problem harder) or add forced delay via a callback handler (e.g., simplexInterrupt).

e.g., something like

```python
def test_solver_running_state(self):
    # build model as before (removed for brevity)

    entered = threading.Event()   # solver thread reached the callback
    release = threading.Event()   # test tells solver it may continue

    def hold(e, data):
        entered.set()
        release.wait(timeout=30)  # bounded delay, don't hang CI

    h.HandleUserInterrupt = True
    h.cbSimplexInterrupt.subscribe(hold)

    h.startSolve()
    self.assertTrue(entered.wait(timeout=30), "solver never reached callback")
    self.assertTrue(h.is_solver_running())     # solver is still inside hold()
    release.set()                              # solver can exit hold()

    h.cancelSolve()
    h.wait()
```

Opt-Mucca:
@mathgeekcoder I'm happy with that suggested change! I like it more than just increasing `N`. Should I stuff it into this PR or do you create a separate one?

mathgeekcoder:
> @mathgeekcoder I'm happy with that suggested change! I like it more than just increasing `N`. Should I stuff it into this PR or do you create a separate one?

Wonderful. I'm happy either way!  Though, earliest I can get to this will be next week.  If you choose to add it here, I can likely help review before then.

Opt-Mucca:
@mathgeekcoder we do a separate PR. No stress on the timing. I'm still not 100% confident on doing anything more than copying the suggestion for now (need to take the time to get more invested in `highspy` at some point......)

PR #3241, opened 2026-08-27
Avoid stalling in `HighsRedcostFixing::addRootRedcost`

fwesselm:
## Description

- Avoid stalling in `HighsRedcostFixing::addRootRedcost` for huge bounds (`>= kHighsIInf`) on integer variables.
- Add a unit test.

## Checklist

- [x] I have read the [contributing guidelines](https://github.com/ERGO-Code/HiGHS/blob/latest/CONTRIBUTING.md)
- [x] This PR targets the `latest` branch
- [x] Tests are passing
- [x] Documentation was updated where relevant
- [x] This PR is not primarily AI-generated (per the AI contributions policy in CONTRIBUTING.md)

PR #3243, opened 2026-08-27
Now mapping to `kHighsInf`, any bounds on (semi-)integer variables that are not less than 1e9

jajhall:
## Description

Ridiculous large finite bounds on integer variables caused a failure addressed by #3241 

Removed infinite bound of 1e9 for integers, but added separate continuous and noncontinuous bound analysis, and warning statement for badly-scaled models

## Checklist

- [x] Tests are passing


Opt-Mucca (commented):
I don't mind this change, but I do worry `1e9` is too small here (don't see an easy way that we can increase it).

We'd now risk giving "infeasible" solutions for people who modelled with such large bounds.

highs/lp_data/HighsLpUtils.cpp
```diff
@@ -386,9 +386,15 @@ HighsStatus assessBounds(const HighsOptions& options, const std::string& type,
...
   HighsInt num_infinite_upper_bound = 0;
+  HighsInt num_infinite_integer_lower_bound = 0;
+  HighsInt num_infinite_integer_upper_bound = 0;
   HighsInt local_ix;
   HighsInt ml_ix;
   HighsInt usr_ix = -1;
+  // Any bounds on integer or semi-integer variables that are not less
+  // than infinite_integer_bound are set to kHighsInf
+  const double infinite_integer_bound =
+      1e9;  // Approx static_cast<double>(kHighsIInf/2);
```
fwesselm: Would it be worth having a separate option `infinite_integer_bound` (instead of hard-coding this)?

fwesselm (commented):
@jajhall, just one question regarding hard-coding the 1e9.

jajhall:
> I don't mind this change, but I do worry `1e9` is too small here (don't see an easy way that we can increase it).
> 
> We'd now risk giving "infeasible" solutions for people who modelled with such large bounds.

No, as the change relaxes the bounds on such columns. So unboundedness is the "risk".

However, if someone sets 1e9 as a bound that might be active in an optimal solution they are asking for trouble. Even the simplex solver may fail as 1e9/primal_feasibility_tolerance is 1/double_precision.

The reason for the `infinite_bound` option is that some folk will not know the value to use for an "infinite" bound, so just put something large. To be honest, allowing them to vary the value of the `infinite_bound` option from its default value of 1e20 (with no upper limit!) gives the misleading impression that they can have meaningful bounds greater than 1e20. If it were not for the prospect of having some internal problem scaling facility, I'd be in favour of reducing the default value to 1e12, and only allowing users to reduce it - with the only lower limit being zero.

highs/lp_data/HighsLpUtils.cpp
jajhall: I don't think so. There's no way we'd allow a larger value, and we don't need it to be any less for the original purpose.

The reason for the `infinite_bound` option is that some folk will not know the value to use for an "infinite" bound, so just put something large. To be honest, allowing them to vary the value of the `infinite_bound` option from its default value of 1e20 (with no upper limit!) gives the misleading impression that they can have meaningful bounds greater than 1e20. If it were not for the prospect of having some internal problem scaling facility, I'd be in favour of reducing the default value to 1e12, and only allowing users to reduce it - with the only lower limit being zero.

Opt-Mucca:
> No, as the change relaxes the bounds on such columns. So unboundedness is the "risk".
However, if someone sets 1e9 as a bound that might be active in an optimal solution they are asking for trouble. Even the simplex solver may fail as 1e9/primal_feasibility_tolerance is 1/double_precision.

It's not that uncommon for users to model terms with `1e+9`. I'm quite afraid of such unbounded cases.
What I meant with "infeasible" solutions is that the integral bound doesn't have to be tight to change the optimal solution drastically. There may have been an optimal solution around `x ≈ 1e+5`, but now another "optimal" solution at `x ≈ 1e+10` can pop up despite the user having modelled that they don't want it with the `x <= 1e+9` bound  (not going to be common, but not so ridiculous to imagine).

Opt-Mucca (approved):
Happy with the change now! I'd still suggest using `kExcessivelyLargeBoundValue` instead of `1e4`, but can live with the compromise

highs/lp_data/HighsSolve.cpp
```diff
@@ -538,20 +551,33 @@ void assessExcessiveObjectiveBoundScaling(const HighsLogOptions log_options,
...
                  "%s has some excessively large Hessian values\n",
                  problem.c_str());
-  if (0 < min_col_bound && min_col_bound < small_bound)
+  if (0 < min_continuous_col_bound && min_continuous_col_bound < small_bound)
     highsLogUser(log_options, HighsLogType::kWarning,
-                 "%s has some excessively small column bounds\n",
-                 problem.c_str());
-  if (max_col_bound > large_bound)
+                 "%s has some excessively small bounds on%s variables\n",
+                 problem.c_str(), is_mip ? " continous" : "");
```
fwesselm: continuous?

highs/lp_data/HighsSolve.cpp
```diff
@@ -674,9 +703,16 @@ void assessExcessiveObjectiveBoundScaling(const HighsLogOptions log_options,
...
-  if (order_of_magnitude_message || dl_user_bound_scale)
+  if (order_of_magnitude_message || dl_user_bound_scale) {
     highsLogUser(log_options, HighsLogType::kWarning, "%s\n",
                  message.str().c_str());
+    warning_issued = true;
+  }
+  if (warning_issued)
+    highsLogUser(log_options, HighsLogType::kWarning,
+                 "%s is badly scaled, which may compromise the speed, accuracy "
+                 "and reliablilty of solvers in HiGHS\n",
```
fwesselm: reliability?

highs/lp_data/HighsSolve.cpp
```diff
@@ -538,20 +551,33 @@ void assessExcessiveObjectiveBoundScaling(const HighsLogOptions log_options,
...
-                 problem.c_str());
-  if (max_col_bound > large_bound)
+                 "%s has some excessively small bounds on%s variables\n",
+                 problem.c_str(), is_mip ? " continous" : "");
+  if (max_continuous_col_bound > large_bound)
     highsLogUser(log_options, HighsLogType::kWarning,
-                 "%s has some excessively large column bounds\n",
-                 problem.c_str());
+                 "%s has some excessively large bounds on%s variables\n",
+                 problem.c_str(), is_mip ? " continous" : "");
```
fwesselm: continuous?

highs/lp_data/HighsSolve.cpp
```diff
@@ -515,8 +527,9 @@ void assessExcessiveObjectiveBoundScaling(const HighsLogOptions log_options,
   // max_col_cost = 0
   assert(max_col_cost >= 0);
   // LPs with no columns or no finite nonzero bounds will have
-  // max_col_bound = 0
-  assert(max_col_bound >= 0);
+  // max_continuous_col_bound = 0 and max_noncontinuous_col_bound
```
fwesselm: and `max_noncontinuous_col_bound = 0`?

highs/lp_data/HighsSolve.cpp
fwesselm, line 479: `min_scalable_bound` and `max_scalable_bound` do not take into account the non-continuous bounds, correct?

fwesselm (approved):
Thanks, @jajhall! Just minor questions from me.

highs/lp_data/HighsSolve.cpp
jajhall: Yes, because these bounds aren't scalable.

highs/lp_data/HighsSolve.cpp
jajhall: Corrected

highs/lp_data/HighsSolve.cpp
jajhall: Corrected

highs/lp_data/HighsSolve.cpp
jajhall: Corrected

highs/lp_data/HighsSolve.cpp
jajhall: Corrected

jajhall:
> Happy with the change now! I'd still suggest using `kExcessivelyLargeBoundValue` instead of `1e4`, but can live with the compromise

OK, I've raised it to `1e5` ;-)

PR #3257, opened 2026-09-04
Various improvements for highspy

### LINKED ISSUE #3251: HandleUserInterrupt setter is not idempotent — re-assigning True leaks interrupt callbacks
`Highs.HandleUserInterrupt = True` subscribes `__user_interrupt_event` to the simplex/IPM/MIP interrupt events **every time it is assigned**, because `HighsCallback.subscribe` is a plain `list.append` with no membership check. Assigning `True` to a property that is already `True` should be a no-op.

This matters for the persistent-solver pattern — keep one `Highs` instance and re-solve it as the model changes. Pyomo's HiGHS interfaces assign the flag immediately before every `run()` (`pyomo/contrib/appsi/solvers/highs.py`, and the same line in `pyomo/contrib/solver/solvers/highs.py`), so the Nth solve carries N copies of the callback, and HiGHS calls into Python once per copy **per simplex iteration**. The cost is quadratic in the number of solves.

### Reproducer

Needs only `highspy` and `numpy`.

```python
"""highspy: HandleUserInterrupt is not idempotent, so re-asserting it leaks callbacks."""

import time

import highspy
import numpy as np

N_COL, N_ROW, NZ_PER_ROW, SOLVES = 4000, 3000, 8, 400


def build() -> highspy.Highs:
    rng = np.random.default_rng(0)
    h = highspy.Highs()
    h.setOptionValue("output_flag", False)
    # Presolve would answer this LP outright; the point here is simplex work.
    h.setOptionValue("presolve", "off")

    h.addCols(
        N_COL, rng.random(N_COL), np.zeros(N_COL), np.full(N_COL, 2.0),
        0, np.zeros(1, np.int32), np.zeros(1, np.int32), np.zeros(1),
    )
    starts = np.arange(N_ROW, dtype=np.int32) * NZ_PER_ROW
    indices = np.concatenate(
        [np.sort(rng.choice(N_COL, NZ_PER_ROW, replace=False)) for _ in range(N_ROW)]
    ).astype(np.int32)
    status = h.addRows(
        N_ROW, rng.uniform(8.0, 20.0, N_ROW), np.full(N_ROW, highspy.kHighsInf),
        N_ROW * NZ_PER_ROW, starts, indices, rng.random(N_ROW * NZ_PER_ROW) + 0.5,
    )
    assert status == highspy.HighsStatus.kOk and h.getNumRow() == N_ROW
    return h


def run(reassert_every_solve: bool) -> None:
    h = build()
    rng = np.random.default_rng(1)
    columns = np.arange(N_COL, dtype=np.int32)
    if not reassert_every_solve:
        h.HandleKeyboardInterrupt = True

    timings = []
    for _ in range(SOLVES):
        # A fresh objective and a cold basis, so every solve is real work.
        h.changeColsCost(N_COL, columns, rng.random(N_COL))
        h.clearSolver()
        if reassert_every_solve:
            h.HandleKeyboardInterrupt = True
        started = time.perf_counter()
        h.run()
        timings.append(time.perf_counter() - started)

    label = "assigned before every solve" if reassert_every_solve else "assigned once"
    print(
        f"{label:<28} subscribers={len(h.cbSimplexInterrupt.callbacks):>4}  "
        f"first 20 solves={np.mean(timings[:20]) * 1e3:>7.1f} ms  "
        f"last 20={np.mean(timings[-20:]) * 1e3:>7.1f} ms  "
        f"total={sum(timings):>6.1f} s"
    )


run(reassert_every_solve=False)
run(reassert_every_solve=True)
```

### Output

```
assigned once                subscribers=   1  first 20 solves=    3.0 ms  last 20=    2.9 ms  total=   1.1 s
assigned before every solve  subscribers= 400  first 20 solves=    3.2 ms  last 20=    6.9 ms  total=   2.0 s
```

Both start at the same per-solve cost; by solve 400 the second is 2.4x slower, and the gap keeps widening with the number of solves. `len(h.cbSimplexInterrupt.callbacks)` shows the cause directly.

`subscribe` and the `HandleUserInterrupt` setter are unchanged on master as of today, so this is not fixed in an unreleased version.

### Environment

highspy 1.15.1 (PyPI), Python 3.14, macOS arm64.

### Suggested fix

Make the setter idempotent:

```python
@HandleUserInterrupt.setter
def HandleUserInterrupt(self, value: bool):
    if value == self.__handle_user_interrupt:
        return
    ...
```

or add a membership check in `HighsCallback.subscribe`. Either keeps `x.flag = True` a no-op when the flag is already set.

mathgeekcoder:
## Description

Various improvements for highspy:
* fixed `HandleUserInterrupt` setter and changed callback handler to a unique `list` to avoid repeated calls to same code, but preserve order of callbacks (for deterministic behaviour)
* fixed flaky highspy threading unit tests
* cleaned up code per updated linter (Ruff) rules
* added HighsError and HighsStatusError exceptions (motivated by linter, also provides better error handling)
* fixed issue with min/max with constant objective (e.g., `h.minimize(4)` now works, even though it's not interesting)

## Checklist

- [x] Tests are passing

## Related issue
Closes #3251 and #3228 (flaky tests discussion)


mathgeekcoder:
Note to self: when changing callbacks from `list` to `set`, the order of callbacks is no longer guaranteed to be in the subscribed order.

This assumption was exploited in the `test_usercallbacks` call, which is failing depending how the system / python version sorts the callback functions. 

I don't think we should ever assume a guaranteed order, but I can also replace `set` with a unique `list`.

galabovaa (approved):
Looks good to me!

galabovaa:
Thank you @mathgeekcoder

PR #3259, opened 2026-09-08
Track row origins for clique merging

fwesselm:
## Description

When `HighsCliqueTable::extractCliques` processes a one-sided, all-binary inequality and the extracted clique covers all variables in the row, the row's information is fully captured by the clique. If `HighsCliqueTable::runCliqueMerging` later absorbs that clique into a larger one, the original row becomes redundant.

Previously, `HighsCliqueTable::extractCliques` did not pass a row origin when calling `HighsCliqueTable::addClique`, so `HighsCliqueTable::runCliqueMerging` could never identify and delete these subsumed rows. This change passes the row index as the origin when:

1. The row is one-sided.
2. The extracted clique covers all variables in the row (partial cliques don't represent the full row)

Rows meeting both conditions are equivalent to a clique, so once merged into a larger clique they are safely removed.

@Opt-Mucca, I hope I am not missing something here. It would be great if you could check the two "main" changes in `HighsCliqueTable::extractCliques`: Row removal is allowed if (a) all variables are binary, have equal coefficients and form a clique OR (b) all variables are binary, have unequal coefficients and form a clique. The question is whether we are not losing anything by removing those rows (after clique merging).

I also added unit tests for this.

With this change presolve reduces the number of rows from around 169k to 3k on `neos-631710`.

## Checklist

- [ ] Tests are passing


Opt-Mucca (changes requested):
I think there's a bug here after looking at the isolated change. We could now be extending a non setppc row, but when extending we'd be adding something with a coefficient of 1.

Example: `7x1 + 8x2 + 7x3 <= 8`. Want to extend with `x4`, but we can't just make it `7x1 + 8x2 + 7x3 + x4 <= 8` otherwise we'd allow solutions that violate the clique.

`x4` above would need to be added with a coefficient that's larger than `8 - min(a_i)`, e.g. `1 + feastol()`

We'd somehow have to normalise the row before doing clique merging (make all coefficients the same?). I think this can only strengthen the row as long as trivial fixings have been applied before. A term like `15x5` in the above example would just be fixed to 0, but would create fake solutions if the row were normalised first, e.g., `x1 + x2 + x3 + x5 <= 1`.

fwesselm:
> I think there's a bug here after looking at the isolated change. We could now be extending a non setppc row, but when extending we'd be adding something with a coefficient of 1.
> 
> Example: `7x1 + 8x2 + 7x3 <= 8`. Want to extend with `x4`, but we can't just make it `7x1 + 8x2 + 7x3 + x4 <= 8` otherwise we'd allow solutions that violate the clique.
> 
> `x4` above would need to be added with a coefficient that's larger than `8 - min(a_i)`, e.g. `1 + feastol()`
> 
> We'd somehow have to normalise the row before doing clique merging (make all coefficients the same?). I think this can only strengthen the row as long as trivial fixings have been applied before. A term like `15x5` in the above example would just be fixed to 0, but would create fake solutions if the row were normalised first, e.g., `x1 + x2 + x3 + x5 <= 1`.

Thank you, @Opt-Mucca! I should have read the lifting code more carefully! I think I need more time to investigate this and will close this PR for the time being!

PR #3260, opened 2026-09-09
Add snapshot so detection values match substituted values

Opt-Mucca:
## Description

This fixes a bug where the implied bounds on a column are updated after substitution in one equation, and are then used in substitution for a second equation. Currently [repro.mps.txt](https://github.com/user-attachments/files/32001897/repro.mps.txt) is shown to be incorrectly infeasible (only a few variables and columns). The instance was found by work with @odow and some LLM fuzzers.

## Checklist

- [x] Tests are passing


check/TestMipSolver.cpp
```diff
@@ -1782,3 +1782,14 @@ TEST_CASE("redcost-fixing-large-bounds", "[highs_test_mip_solver]") {
       mipsolver, mipsolver.mipdata_->getDomain());
   REQUIRE(!lurkingBounds.empty());
 }
+
+TEST_CASE("pr-3260", "[highs_test_mip_solver]") {
```
fwesselm, line 1808: Since the model is very small, I would recommend having it in the test file instead of using an MPS file:

```
TEST_CASE("pr-3260", "[highs_test_mip_solver]") {
  HighsLp lp;
  lp.num_col_ = 3;
  lp.num_row_ = 2;
  lp.col_cost_ = {0, 0, 0};
  lp.col_lower_ = {0, 0, -3};
  lp.col_upper_ = {0, 1, 0};
  lp.row_lower_ = {-1, -1};
  lp.row_upper_ = {-1, -1};
  lp.a_matrix_.start_ = {0, 2, 4, 6};
  lp.a_matrix_.index_ = {0, 1, 0, 1, 0, 1};
  lp.a_matrix_.value_ = {1, 1, 1, 1, 1, 1};
  lp.integrality_ = {HighsVarType::kInteger, HighsVarType::kInteger,
                     HighsVarType::kContinuous};
  Highs highs;
  highs.setOptionValue("output_flag", dev_run);
  REQUIRE(highs.passModel(lp) == HighsStatus::kOk);
  const HighsModelStatus require_model_status = HighsModelStatus::kOptimal;
  const double optimal_objective = 0.0;
  solve(highs, kHighsOnString, require_model_status, optimal_objective);
}
```

highs/presolve/HPresolve.cpp
```diff
@@ -4001,12 +4001,24 @@ HPresolve::Result HPresolve::rowPresolve(HighsPostsolveStack& postsolve_stack,
           // printf("simple probing case on row of size %" HIGHSINT_FORMAT
           // "\n", rowsize[row]);
 
+          // Snapshot bounds as they may change when removing or substituting
+          // a column, which would make future substitutions invalid
+          std::vector<std::pair<double, double>> implVarSnapshot;
```
fwesselm, line 4006: If we need this again, we could make a method to snapshot the bounds. But, it's fine for now.

fwesselm (approved):
Thank you for fixing this, @Opt-Mucca!

check/TestMipSolver.cpp
fwesselm: I can add this!

PR #3261, opened 2026-09-09
Add fix

Opt-Mucca:
## Description

This stops the case where the post-solve stack rejects merging a parallel column yet the code continues to modify the problem despite the fail. I think the `break` was accidentally gated behind a debug check, although I think it can be downgraded to ` continue`. It seems that the tolerance checks can be slightly different at different points (probably some scaling). I don't see this affecting any models within our test set. This fix also came from an instance* that @odow generated with an LLM fuzzer.

## Checklist

- [x] Tests are passing


fwesselm (approved):
Nice @Opt-Mucca!!

Opt-Mucca:
@fwesselm I've just added the tiny MIP model that was failing as a test case. Will merge when tests pass.

fwesselm:
> @fwesselm I've just added the tiny MIP model that was failing as a test case. Will merge when tests pass.

Let me put the test into the test file directly instead of using MPS.

PR #3264, opened 2026-09-10
Aggregate implications

fwesselm:
## Description

This PR adds aggregations for implied bounds (implications) as described by [Chen et al., Exploiting Variable Implications in Presolve for Mixed Integer Programming](https://doi.org/10.48550/arXiv.2607.04313).

Note that the current code does not consider cliques, i.e. "implications" between binary variables.

Unfortunately, the effect on performance is relatively small (ratios of shifted geometric means are close to 1.0). However, the models `gfd-schedulen180f7d50m30k18` and `neos-4763324-toguru` (where 50% of the rows are eliminated) can be solved to optimality within a time limit of two hours with the new technique.

I added unit tests for the new technique.

Note that the paper also describes "implication aware" constraint propagation, which I will investigate next.

## Checklist

- [x] Tests are passing


highs/mip/HighsCliqueTable.cpp
```diff
@@ -987,6 +987,33 @@ void HighsCliqueTable::extractCliques(
   }
 }
 
+template <bool Sort, typename Comparator>
+HighsInt HighsCliqueTable::extendClique(
+    std::vector<HighsInt>& neighbourhoodInds, std::vector<CliqueVar>& clqVars,
+    HighsInt seedPos, HighsInt extensionEnd, const Comparator& candidateOrder) {
+  for (HighsInt i = seedPos; i < extensionEnd; ++i) {
+    if (Sort)
```
Opt-Mucca, line 995: I'm not sure on the C++ here: So this wouldn't sort if there's a `lambda` that always returns `false`? If not, doesn't this introduce some slight overhead that didn't used to exit?

highs/presolve/HPresolve.cpp
```diff
@@ -9113,25 +9119,203 @@ void HPresolve::extractVarBounds(HighsInt row) {
...
+            HighsHashTree<HighsInt, colImpliedBounds>& boundsMap,
+            double baseBound, HighsInt direction) {
+          for (const auto& clique : cover) {
+            if (clique.size() < 2) continue;
+
+            // find an unconsumed row to reuse and remove the rest
+            HighsInt row = -1;
+            for (const auto& var : clique) {
+              const auto* bounds = boundsMap.find(var.col);
+              HighsInt currentrow = bounds->originalBound.origin;
```
Opt-Mucca, line 9259: I think there needs to be a check here that `currentrow` is still single sided.

Opt-Mucca (approved):
@fwesselm I've got one C++ question and one potential correctness error that I believe is a one line fix. The rest looks good!

highs/presolve/HPresolve.cpp
fwesselm: @Opt-Mucca, currently, the code only stores `origin` with a VLB/VUB if the source row is not ranged:

```
   // only record origin for size-2 one-sided rows that can be safely rewritten
    HighsInt origin = rowsize[row] == 2 && !isRanged(row) ? row : -1;
```

Therefore, all rows will be one-sided in this loop.

highs/mip/HighsCliqueTable.cpp
fwesselm: I know this looks weird, but `Sort` is a template parameter, so `if (Sort)` is resolved at compile time. With `extendClique<false>`, the sort branch is eliminated entirely - the comparator is dead code.

Opt-Mucca: Thank you for the explanation!

highs/presolve/HPresolve.cpp
Opt-Mucca: Can't the source row get updated to become double-sided though, e.g., parallel row merging? Or all origins invalidated by such a solution path?

fwesselm: @Opt-Mucca, any modification made to the row (via `addToMatrix`) will reset the origin. I will have a closer look and double-check.

fwesselm: Helloooo @Opt-Mucca! Looks like there are a couple of cases where just the row side(s) are manipulated in presolve, which should also trigger a reset of the origin. Also, I noticed that dual row bounds are not updated in these cases, so there is more to investigate. Hoooooray!

Opt-Mucca: I was also completely missing the case where the RHS value changes.....

fwesselm: @Opt-Mucca, this should be fixed by #3270.

Opt-Mucca:
@fwesselm I'm happy for this to be merged whenever now provided that it's performance positive

PR #3269, opened 2026-09-11
Make feasibility jump code more robust (Fix #3262)

### LINKED ISSUE #3262: Segfault in feasibility jump
Version: v1.15.1
Language: Julia

MIP

macOS

**What did you do?**
```Julia
julia> using JuMP, HiGHS

julia> model = read_from_file("/Users/oscardowson/Claude/highs-fuzzer/results/issue-12/minimized/model.mps")
A JuMP Model
├ solver: none
├ objective_sense: MIN_SENSE
│ └ objective_function_type: AffExpr
├ num_variables: 2
├ num_constraints: 4
│ ├ AffExpr in MOI.GreaterThan{Float64}: 2
│ ├ AffExpr in MOI.LessThan{Float64}: 1
│ └ VariableRef in MOI.Integer: 1
└ Names registered in the model: none

julia> set_optimizer(model, HiGHS.Optimizer)

julia> set_attribute(model, "presolve", "off")

julia> optimize!(model)
Using BLAS: Apple Accelerate 
MIP has 3 rows; 2 cols; 5 nonzeros; 1 integer variables (0 binary)
Coefficient ranges:
  Matrix  [1e+00, 1e+02]
  Cost    [0e+00, 0e+00]
  Bound   [0e+00, 0e+00]
  RHS     [3e+02, 3e+02]

Presolve is switched off

Solving MIP model with:
   3 rows
   2 cols (0 binary, 1 integer, 0 implied int., 1 continuous, 0 domain fixed)
   5 nonzeros
   Thread count 4 (of 8 threads). Using 1 max workers. Parallel search off

[2864] signal (10.1): Bus error: 10
in expression starting at REPL[16]:1
_ZN24external_feasibilityjump7Problem8setValueIZNS_21FeasibilityJumpSolver14doVariableMoveEjEUlNS_15LhsModificationEE_EEmjdT_ at /Users/oscardowson/.julia/artifacts/5a3d28a9c8ef24dab596551d0b991b2f9d029ee7/lib/libhighs.1.15.dylib (unknown line)
```

```
NAME        
ROWS
 N  NoObj   
 G  r0      
 L  r1      
 G  r2      
COLUMNS
    MARK0000  'MARKER'                 'INTORG'
    x0        r1        -100
    x0        r2        1
    MARK0001  'MARKER'                 'INTEND'
    x1        r0        -1
    x1        r1        1
    x1        r2        -1
RHS
    RHS_V     r1        308
BOUNDS
 FR BOUND     x0      
 FR BOUND     x1      
ENDATA
```

Here's the LP file, but it doesn't reproduce.
```
\ File written by HiGHS .lp file handler
min
 obj: 
st
 r0: -1 x1 >= +0
 r1: -100 x0 +1 x1 <= +308
 r2: +1 x0 -1 x1 >= +0
bounds
 x0 free
 x1 free
bin
gen
 x0
semi
end
```

Optimal solution

**What did you see instead?**
Segfault

BenChampion:
## Description

This fixes the issue that @odow identified in #3262.

Details:
* The ultimate cause is a move score that eventually hits minus infinity, at which point minus infinity no longer works as "a value less than any value that appears" when taking a maximum.
* This PR fixes a proximate cause (variable index equal to sentinel value). This could occur for other reasons (for example, an empty row), and is a lot easier than fixing the fundamentals of the algorithm (which one way or another we may replace anyhow).
* The new test problem has a slight modification so that the origin isn't feasible. This is just to future-proof the test a bit in case we expand the scope of the trivial heuristics. (At the moment they don't run for this problem [because it contains a continuous variable](https://github.com/ERGO-Code/HiGHS/blob/5a91a74429b2e4feddf929999c07439181863744/highs/mip/HighsMipSolverData.cpp#L272). Any thoughts on this @jajhall?)

## Checklist

- [x] Tests are passing

## Related issue
Closes #3262

### DIFF
--- a/check/TestMipSolver.cpp
+++ b/check/TestMipSolver.cpp
@@ -1730,6 +1730,33 @@ TEST_CASE("issue-3171", "[highs_test_mip_solver]") {
   solve(highs, kHighsOnString, require_model_status, optimal_objective);
 }
 
+TEST_CASE("issue-3262", "[highs_test_mip_solver]") {
+  Highs highs;
+  highs.setOptionValue("output_flag", dev_run);
+
+  HighsLp lp;
+  lp.sense_ = ObjSense::kMinimize;
+  lp.num_col_ = 2;
+  lp.num_row_ = 3;
+  lp.col_lower_ = {-kHighsInf, -kHighsInf};
+  lp.col_upper_ = {kHighsInf, kHighsInf};
+  lp.col_cost_ = {0., 0.};
+  lp.integrality_ = {HighsVarType::kInteger, HighsVarType::kContinuous};
+  lp.row_lower_ = {1., -kHighsInf, 1.};
+  lp.row_upper_ = {kHighsInf, 307., kHighsInf};
+  lp.a_matrix_.format_ = MatrixFormat::kColwise;
+  lp.a_matrix_.start_ = {0, 2, 5};
+  lp.a_matrix_.index_ = {1, 2, 0, 1, 2};
+  lp.a_matrix_.value_ = {-100., 1., -1., 1., -1.};
+
+  highs.passModel(lp);
+  highs.setOptionValue("presolve", "off");
+
+  REQUIRE_NOTHROW(highs.run());
+
+  highs.resetGlobalScheduler(true);
+}
+
 TEST_CASE("issue-2900", "[highs_test_mip_solver]") {
   std::string filename =
       std::string(HIGHS_DIR) + "/check/instances/issue-2900.mps";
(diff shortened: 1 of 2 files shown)


jajhall (approved):
Looks fine to me. 

As for the trivial heuristic not being applied to problems with continuous variables, my instinct is that the all-zero integer assignment is only worth trying "for free" as it's unlikely to give a good primal bound (even if feasible), so it's not worth solving an LP when, in most cases, the task of the LP would be to determine infeasibility of the assignment.

PR #3270, opened 2026-09-11
Reset dual-implied bounds when upgrading equalities

Opt-Mucca:
## Description

This is another instance found by @odow and his LLM fuzzer. The issue seems to be that when upgrading a row to an equality, the implied dual bounds on other rows are not notified. This means that there can be a stale implied bound lying around, and in this three column LP example, it is used for some additional reduction (I think it was a substitution). This lead to an incorrect return status in the test I added (instance directly from @odow )

The fix I added was to reset all the dependent implied bounds when upgrading the row to an equality (I will test this over the weekend). @fwesselm This is another presolve change where I'm a bit out of my depth.

## Checklist

- [x] Tests are passing

### DIFF
--- a/check/TestPresolve.cpp
+++ b/check/TestPresolve.cpp
@@ -1236,3 +1236,25 @@ TEST_CASE("issue-3140", "[highs_test_presolve]") {
 
   highs.resetGlobalScheduler(true);
 }
+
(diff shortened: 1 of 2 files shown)


Opt-Mucca:
A nicer and more robust fix than my original one! This should be merged when the tests pass.

fwesselm (approved):
Thanks for finding the culprit, @Opt-Mucca !

PR #3272, opened 2026-09-13
Fix 3271

### LINKED ISSUE #3271: Bug with simple MIP
Clear bug with this simple MIP

```Julia
julia> using JuMP, HiGHS

julia> model = Model(HiGHS.Optimizer)
A JuMP Model
├ solver: HiGHS
├ objective_sense: FEASIBILITY_SENSE
├ num_variables: 0
├ num_constraints: 0
└ Names registered in the model: none

julia> @variable(model, x0 <= 4.5, Int)
x0

julia> @variable(model, x1, Int)
x1

julia> @variable(model, x2 == 0, Int)
x2

julia> @objective(model, Min, 3 * x2)
3 x2

julia> @constraint(model, x0 >= -15.5)
x0 ≥ -15.5

julia> optimize!(model)
Using BLAS: Apple Accelerate 
MIP has 1 row; 3 cols; 1 nonzero; 3 integer variable (0 binary)
Coefficient ranges:
  Matrix  [1e+00, 1e+00]
  Cost    [3e+00, 3e+00]
  Bound   [4e+00, 4e+00]
  RHS     [2e+01, 2e+01]
Presolving model
0 rows, 0 cols, 0 nonzeros 0s
0 rows, 0 cols, 0 nonzeros 0s
Presolve reductions: rows 0(-1); columns 0(-3); nonzeros 0(-1) - Reduced to empty
Presolve: Optimal
WARNING: Solution with objective 0 has untransformed violations: bound = 0; integrality = 0.5; row = 0

         0       0         0   0.00%   0               0                  0.00%        0      0      0         0     0.0s

Solving report
  Status            Optimal
  Primal bound      0
  Dual bound        0
  P-D integral      0
  Solution status   infeasible
                    0 (objective)
                    0 (bound viol.)
                    0.5 (int. viol.)
                    0 (row viol.)
  Timing            0.00
                    0.00 (Presolve)
                    0.00 (Solve)
                    0.00 (Postsolve)
  Max sub-MIP depth 0
  Nodes             0
  Repair LPs        1 (0 feasible; 0 iterations)
  LP iterations     0

julia> value(x0)
4.5

julia> solution_summary(model)
solution_summary(; result = 1, verbose = false)
├ solver_name          : HiGHS
├ Termination
│ ├ termination_status : OPTIMAL
│ ├ result_count       : 1
│ ├ raw_status         : kHighsModelStatusOptimal
│ └ objective_bound    : 0.00000e+00
├ Solution (result = 1)
│ ├ primal_status        : INFEASIBLE_POINT
│ ├ dual_status          : NO_SOLUTION
│ ├ objective_value      : 0.00000e+00
│ ├ dual_objective_value : NaN
│ └ relative_gap         : 0.00000e+00
└ Work counters
  ├ solve_time (sec)   : 3.32083e-04
  ├ simplex_iterations : 0
  ├ barrier_iterations : -1
  └ node_count         : 0
```

jajhall:
## Description

Adds `HPresolve::roundIntegerVariableBounds()` to round bounds on integer variables to integer values so that, for example, empty columns are not fixed to non-integer values

I'm rather surprised that this isn't already done. Maybe it happens after `HPresolve::initialRowAndColPresolve` where the empty column in `issue-3271` (after the singleton row is removed) is fixed to its upper bound of 4.5. However, I didn't spot it

## Checklist

- [ ] Tests are passing

## Related issue
Closes #3271


odow:
The fuzzer found a whole suite of very similar problems. Does this also fix:
```
\ File written by HiGHS .lp file handler
min
 obj: 
st
 r0: -1 x0 -1 x1 >= -0
bounds
 -0.5 <= x0 <= 0.5
bin
gen
 x0
semi
end
```

```Julia
julia> using JuMP, HiGHS

julia> begin
           model = Model(HiGHS.Optimizer)
           @variable(model, -0.5 <= x0 <= 0.5, Int)
           @variable(model, x1 >= 0)
           @constraint(model, x0 + x1 <= 0)
           optimize!(model)
       end
Using BLAS: Apple Accelerate 
MIP has 1 row; 2 cols; 2 nonzeros; 1 integer variables (0 binary)
Coefficient ranges:
  Matrix  [1e+00, 1e+00]
  Cost    [0e+00, 0e+00]
  Bound   [5e-01, 5e-01]
  RHS     [0e+00, 0e+00]
Presolving model
Presolve: Infeasible

         0       0         0   0.00%   -inf            inf                  inf        0      0      0         0     0.0s

Solving report
  Status            Infeasible
  Primal bound      inf
  Dual bound        -inf
  Gap               inf
  P-D integral      0
  Solution status   -
  Timing            0.00
                    0.00 (Presolve)
                    0.00 (Solve)
                    0.00 (Postsolve)
  Max sub-MIP depth 0
  Nodes             0
  Repair LPs        0
  LP iterations     0
```

Opt-Mucca (approved):
I'm also a bit amazed that this bug has managed to survive for so long. I never considered that we wouldn't be fixing this first...... Change looks correct!

Suggestion: I'd change `variable` to `col` in the function name.

fwesselm:
> I'm also a bit amazed that this bug has managed to survive for so long. I never considered that we wouldn't be fixing this first...... Change looks correct!
> 
> Suggestion: I'd change `variable` to `col` in the function name.

@Opt-Mucca, I need to check why this is not detected in `initialRowAndColPresolve`. I added handling for inconsistent bounds there. Maybe it is not readched due to the new initial sweep.

jajhall:
> @Opt-Mucca, I need to check why this is not detected in initialRowAndColPresolve. I added handling for inconsistent bounds there. Maybe it is not readched due to the new initial sweep.

No, initialSweep isn't called for MIPs, and the incorrect fixing is in initialRowAndColPresolve

This bound rounding should be done before initial sweep, in case we apply the latter to MIPs. The inconsistent bound checking should also be done sooner so that the presolve data structures aren't formed unnecessarily

Opt-Mucca:
> @Opt-Mucca, I need to check why this is not detected in `initialRowAndColPresolve`. I added handling for inconsistent bounds there. Maybe it is not readched due to the new initial sweep.

I think the issue is that the rows are presolved first in that sweep. I don't see any better method than adding a new checking function (I was also thinking of `initialSweep`, but if that's not called for MIPs then it makes no sense)

jajhall:
> > @Opt-Mucca, I need to check why this is not detected in `initialRowAndColPresolve`. I added handling for inconsistent bounds there. Maybe it is not readched due to the new initial sweep.
> 
> I think the issue is that the rows are presolved first in that sweep. I don't see any better method than adding a new checking function (I was also thinking of `initialSweep`, but if that's not called for MIPs then it makes no sense)
> 
> 


Read my comment. 

I'll change my method to checkModel, and add trivial bound violation checking

jajhall:
> The fuzzer found a whole suite of very similar problems. Does this also fix:

Yes

jajhall:
> @Opt-Mucca, I need to check why this is not detected in `initialRowAndColPresolve`. I added handling for inconsistent bounds there. 

I see this, but what happens is that the loop through the rows identifies a singleton, and when this is removed, the corresponding column is empty, so is fixed at its upper bound. This is non-integer, because the rounding takes place in the loop over the columns. 

Your call to `changeColBounds` in `initialRowAndColPresolve` is, I believe, unnecessary, since it's passing the bounds for the column in the model. If these had changed as a result of some other presolve operation, that operation would have already called `changeColBounds`, so any rounding and bound updating would already have taken place.

Opt-Mucca (approved):
Happy with the changes!

PR #3275, opened 2026-09-14
Fix 3273

### LINKED ISSUE #3273: Optimal solution illegally cut off in MIP solver
Consider this problem:
```
max
 obj: x1 + x2 
st
 r0: - x1 + x3 >= 0
 r1: 1000 x0 - x2 >= 0
 r2: x0 + x2 <= 12
bounds
 -inf <= x3 <= 1
bin
 x0
end
```

* presolve=on finds objective=12
* presolve=off finds objective=11

Both satisfy primal feasibility, so presolve is hiding a bug somewhere in the
MIP search.

## Patch

Claude tells me the patch is
```patch
diff --git a/highs/mip/HighsDomain.cpp b/highs/mip/HighsDomain.cpp
index f09d0f5b81..c163ff95ff 100644
--- a/highs/mip/HighsDomain.cpp
+++ b/highs/mip/HighsDomain.cpp
@@ -3540,10 +3540,15 @@ bool HighsDomain::ConflictSet::explainBoundChange(
       HighsInt ninfmin;
       HighsCDouble minAct;
       globaldom.computeMinActivity(0, len, inds, vals, ninfmin, minAct);
-      assert(ninfmin <= 1);
-      // todo: treat case with a single infinite contribution that propagated a
-      // bound
-      if (ninfmin == 1) return false;
+      // The objective propagation may have used local bounds of columns that
+      // are unbounded in the global domain (e.g. bounds from a branching or
+      // heuristic fixing). Then the global minimal activity is -infinity and
+      // the bound change cannot be explained from the global residual
+      // activity: minAct only holds the finite contributions, so passing it
+      // on would give an explanation that is too weak (an empty conflict,
+      // i.e. global infeasibility, in the worst case).
+      // todo: treat case with infinite contributions that propagated a bound
+      if (ninfmin > 0) return false;
 
       return explainBoundChangeLeq(currentFrontier, domchg, inds, vals, len,
                                    rhs, double(minAct));
```

## Log with presolve
```julia
julia> begin
           model = read_from_file("/tmp/bug-027-min-b.lp")
           optimize!(model)
       end
Using BLAS: Apple Accelerate 
MIP has 3 rows; 4 cols; 6 nonzeros; 1 integer variables (1 binary)
Coefficient ranges:
  Matrix  [1e+00, 1e+03]
  Cost    [1e+00, 1e+00]
  Bound   [1e+00, 1e+00]
  RHS     [1e+01, 1e+01]
Presolving model
3 rows, 3 cols, 5 nonzeros 0s
0 rows, 0 cols, 0 nonzeros 0s
Presolve reductions: rows 0(-3); columns 0(-4); nonzeros 0(-6) - Reduced to empty
Presolve: Optimal

         0       0         0   0.00%   12              12                 0.00%        0      0      0         0     0.0s

Solving report
  Status            Optimal
  Primal bound      12
  Dual bound        12
  P-D integral      0
  Solution status   feasible
                    12 (objective)
                    0 (bound viol.)
                    0 (int. viol.)
                    0 (row viol.)
  Timing            0.00
                    0.00 (Presolve)
                    0.00 (Solve)
                    0.00 (Postsolve)
  Max sub-MIP depth 0
  Nodes             0
  Repair LPs        0
  LP iterations     0
```

## Log without presolve
```julia
julia> begin
           model = read_from_file("/tmp/bug-027-min-b.lp")
           set_attribute(model, "presolve", "off")
           optimize!(model)
       end
Using BLAS: Apple Accelerate 
MIP has 3 rows; 4 cols; 6 nonzeros; 1 integer variables (1 binary)
Coefficient ranges:
  Matrix  [1e+00, 1e+03]
  Cost    [1e+00, 1e+00]
  Bound   [1e+00, 1e+00]
  RHS     [1e+01, 1e+01]

Presolve is switched off

Solving MIP model with:
   3 rows
   4 cols (1 binary, 0 integer, 0 implied int., 3 continuous, 0 domain fixed)
   6 nonzeros
   Thread count 5 (of 10 threads). Using 1 max workers. Parallel search off

 J       0       0         0   0.00%   inf             11                 Large        0      0      0         0     0.0s
         1       0         1 100.00%   11              11                 0.00%        0      0      1         2     0.0s

Solving report
  Status            Optimal
  Primal bound      11
  Dual bound        11
  P-D integral      5.65538909237e-06
  Solution status   feasible
                    11 (objective)
                    0 (bound viol.)
                    0 (int. viol.)
                    0 (row viol.)
  Timing            0.01
                    0.00 (Presolve)
                    0.01 (Solve)
                    0.00 (Postsolve)
  Max sub-MIP depth 0
  Nodes             1
  Repair LPs        0
  LP iterations     2
                    0 (strong br.)
                    0 (separation)
                    0 (heuristics)
```

## Claude's explanation (from a slightly different problem)

Branch `fix/v2-presolve_off_mip_suboptimal` in `worktrees/v2-presolve_off_mip_suboptimal`, based on `5d6c11e1a5` (branch `fix/combined`). Patch: `fix.patch` (only `highs/mip/HighsDomain.cpp`, one condition in `HighsDomain::ConflictSet::explainBoundChange`).

### Root cause

This is not a presolve/LP-presolve or bound-bookkeeping problem: the root LP is solved correctly (on the minimized model of issue-027 it gives -12.988 with x0 = 0.012 fractional, the true LP relaxation value) and `lower_bound`, `upper_limit` and `optimality_limit` are all right. The tree is killed by a **wrong conflict cut with zero entries** ("always infeasible") that the conflict analysis derives inside a rounding heuristic at the root. It is only visible with `presolve=off` because presolve either removes the unbounded continuous columns that trigger it or makes the incumbent found by the heuristics optimal already.

Trace on the minimized model of issue-027 (`min -x1 - x2; -x1 + x3 >= 0; 1000 x0 - x2 >= 0; x0 + x2 <= 12; x0 binary; x1, x2 >= 0; x3 <= 1`, optimum -12):

1. The feasibility jump finds the incumbent -11 (x0 = 1, x1 = 0, x2 = 11), so `upper_limit = -11`.
2. The root LP gives -12.988 with x0 = 0.012, status optimal (`HighsMipSolverData::evaluateRootLp`).
3. `evaluateRootNode` runs `heuristics.randomizedRounding(worker, firstlpsol)` (`HighsMipSolverData.cpp:2118`), which calls `HighsPrimalHeuristics::tryRoundedPoint` (`HighsPrimalHeuristics.cpp:971`). In a local copy of the global domain it fixes x0 = 0 with reason `branching` and propagates. The local domain-change stack becomes
   - pos 0: `x0 <= 0` (branching),
   - pos 1: `x2 <= 0` (row 1: `1000 x0 - x2 >= 0`),
   - pos 2: `x1 >= 11` (reason `kObjective`: `-x1 - x2 <= upper_limit = -11` with the local bound `x2 <= 0`),
   - infeasible with reason row 0 (`-x1 + x3 >= 0` has max activity -10 < 0).
   The local domain is correctly infeasible: x0 = 0 cannot beat -11. `tryRoundedPoint` then calls `localdom.conflictAnalysis` (`HighsPrimalHeuristics.cpp:1002`); the correct conflict is `x0 >= 1`.
4. `ConflictSet::explainInfeasibility` (`HighsDomain.cpp:3205`) explains the row-0 infeasibility by `x1 >= 11` (pos 2), and `computeCuts` → `resolveDepth` tries to resolve pos 2 through `explainBoundChange` (`HighsDomain.cpp:3469`), case `Reason::kObjective` (`:3533`). It fetches the objective propagation constraint `-x1 - x2 <= -11` and computes its minimal activity in the **global** domain, where both x1 and x2 are unbounded above: `ninfmin = 2`, `minAct = 0` (`computeMinActivity` only sums the finite contributions). The code had
   ```
   assert(ninfmin <= 1);
   // todo: treat case with a single infinite contribution that propagated a bound
   if (ninfmin == 1) return false;
   ```
   so with `ninfmin == 2` (assertion inactive in a release build) it falls through and calls `explainBoundChangeLeq(..., rhs = -11, minAct = 0)` (`HighsDomain.cpp:2878`) with a global residual activity of 0 instead of -infinity. Inside, `M = 0 - (-1) * col_upper_[x1] = +inf`, so `resolveLinearLeq` (`:3094`) sees `covered = M - Mlower = +inf >= 0` and returns `true` with an **empty** `resolvedDomainChanges`: the bound change `x1 >= 11` is "explained" by nothing.
5. `resolveDepth` erases pos 2 from the reason-side frontier, inserts nothing, `numResolved = 1 > 0`, and `computeCuts` adds the empty frontier as a conflict cut (`HighsConflictPool::addConflictCut`, 0 entries; the log's `Confl. 1`). An empty conflict is unconditionally violated.
6. The next `evaluateRootLp` call propagates the global domain, the conflict pool marks it infeasible without any bound change (`propagate infeasible=1, changedcols=0` in the trace), `updateLowerBound(upper_bound)` sets the dual bound to -11 and the search ends with "Optimal -11".

The same mechanism (an objective-reason bound change on a column that is globally unbounded together with at least one other globally unbounded objective column, bounded locally by a heuristic fixing) explains all instances of the class: the original issue-027 model (`ninfmin = 2` for x1, -20 → -50), `findings-v2-run1/issue-016` and `issue-019`, `findings-v3/issue-010` (conflict from `x2` with `ninfmin = 2`, prunes the -2.6667 solution and reports 3) and the bigm-mode regression finding seed 3906 (see below).

Why the objective case is different from rows/cuts: for model rows and cuts `explainBoundChange` passes `globaldom.getMinActivity(row)`, which is -infinity as soon as one contribution is infinite, and `explainBoundChangeLeq` returns `false` immediately (`if (minAct == -kHighsInf) return false;`). For the objective the code asks `computeMinActivity` for the finite part and the infinite count separately and only rejected the count 1. The objective propagation itself is correct (it propagates when the *local* number of infinite contributions is at most one), so `ninfmin` in the global domain can be any number; the assertion `ninfmin <= 1` was simply wrong.

### What the patch changes

`highs/mip/HighsDomain.cpp`, `HighsDomain::ConflictSet::explainBoundChange`, case `Reason::kObjective`: the bound change is declared unexplainable (`return false`) whenever the objective propagation constraint has **any** infinite contribution in the global domain (`ninfmin > 0`), instead of only for `ninfmin == 1`; the wrong assertion is removed. This is the same rule `explainInfeasibility` already applies for the objective (`HighsDomain.cpp:3282`) and the same behaviour rows and cuts get through `getMinActivity() == -inf`.

Correctness: returning `false` leaves the bound change in the frontier unresolved (`resolveDepth` skips it), so every conflict that is still produced consists only of bound changes that were validly derived; nothing that was valid before is lost. On the minimized model the conflict analysis now produces the (valid, if weak) conflict `{x1 >= 1.00001}` (i.e. x1 <= 1 globally), the domain stays feasible, and branching on x0 finds -12. A full explanation through the local bounds of the globally unbounded columns (which would yield the stronger conflict `x0 >= 1`) would need `explainBoundChangeLeq/Geq` to handle infinite base bounds; that is left as the existing todo.

### Replay results after the fix

`./fuzz-v2-presolve_off_mip_suboptimal --replay-file <model.txt>` (no `[FINDING]` line):

| instance | presolve=on | presolve=off | exact |
|---|---|---|---|
| `results-v2/presolve_off_mip_suboptimal/issue-027` | Optimal, -50, x = (1, 10, 0, 24, 1) | Optimal, -50, x = (1, 10, 0, 24, 1) (was -20) | Optimal, -50 |
| `findings-v2-run1/issue-016` | Optimal, 59 | Optimal, 59 (was 53) | Optimal, 59 |
| `findings-v2-run1/issue-019` | Optimal, 72.5909 | Optimal, 72.5909 (was 41) | Optimal, 72.5909 |
| regression seed 3906 (`regression-seed-3906/`, bigm mode, found by the baseline run below) | Optimal, -85 | Optimal, -85 (was -67) | Optimal, -85 |

`highs --options_file` (fuzzer options, `presolve=off`, `time_limit=5`) on `model.mps`, answer from `--solution_file`:

- `issue-027/minimized/model.mps`: Optimal, -12 (was -11); `issue-027/model.mps`: Optimal, -50.
- `findings-v2-run1/issue-016/minimized/model.mps`: Optimal, -59; `issue-019/minimized/model.mps`: Optimal, -20 (were -53 and -14). The unminimized `model.mps` of 016/019 carry no OBJSENSE, so they are the minimisation variants; baseline and patched binaries agree on them (-12 and 24.3333); the fuzzer replays above use the intended maximisation.
- `findings-v3/issue-010/model.mps`: **Time limit reached**, primal bound 2.66666666667 (the true optimum), dual bound 1.4 after 8156 nodes (was "Optimal 3", which the wrong conflict produced by pruning the tree). With a 120 s limit: ISSUE010_120. This model has three unbounded integer columns (x1, x2, x4 with infinite upper or lower bounds), and without presolve HiGHS's branch-and-bound does not close the gap; the patched binary no longer claims a wrong optimum, which is the behaviour a correct solver must have here. With `presolve=on` it is solved in 0.01 s as before.

### Fuzzer regression check

`--seed 777 --count 4000 --maxsave 1 --no-minimize`, baseline `./fuzz-v2-comb2` (combined build) vs patched `./fuzz-v2-presolve_off_mip_suboptimal`:

| run | baseline | patched |
|---|---|---|
| `--mode all` | metamorphic_mismatch=1 reduction_limit_prefix_mismatch=2 rule_mask_mismatch=9 soft_status_mismatch=311 | metamorphic_mismatch=1 reduction_limit_prefix_mismatch=2 rule_mask_mismatch=9 soft_status_mismatch=311 (identical seed list of non-soft findings) |
| `--mode bigm` | **exact_mismatch_presolve_off=1** (seed 3906) rule_mask_mismatch=6 soft_status_mismatch=357 | rule_mask_mismatch=6 soft_status_mismatch=357 |
| `--mode dual_fixing` | DUALFIX_BASE | DUALFIX_PATCHED |

No new findings of any kind; the one instance of this class found by the baseline (bigm mode, seed 3906) disappears (its replay is in the table above).

### Earlier issues

`python3 verify.py --highs worktrees/v2-presolve_off_mip_suboptimal/build/bin/highs --issue NN` for NN = 01..12 gives the same answers as the combined build (issues 01: Unbounded; 02: 6 / 0; 03: -2 / -0.5; 04: -468 / -9; 05: 0 / 0; 06: -4 / 0; 07: Infeasible; 08: 0 / 0; 09: Infeasible for all five models; 10: Primal infeasible or unbounded / Unbounded; 11: Infeasible / Time limit reached with presolve off, as on the combined build; 12: 11 / 0). The issue-13 examples (`results/issue-13/example-*` and `minimized-seed*`, presolve on and off) give 006 Infeasible, 015 Infeasible, 019 0, 032 -25.6667, 040 -58.3333, seed126 0, seed21 -28, seed90 -2, matching the expected column of its README.

### Doubts

- The fix is conservative: it stops the conflict analysis at objective-derived bound changes that involve globally unbounded columns rather than explaining them properly, so conflicts can be weaker than they could be in such models. This only affects conflict strength, never correctness.
- `explainBoundChangeLeq/Geq` also contain arithmetic on `cand.delta` with infinite global bounds (`vals * (ub - inf)`) that is now unreachable from the objective case; the row and cut cases were already guarded by the `-inf` activity test.

Opt-Mucca:
## Description

This fixes the issue described in #3273 TLDR: Local domain has added some bound changes for some unbounded variables, e.g., for a LNS heuristic. The LP with this more restrictive local domain is infeasible, and conflict analysis is called. This gets resolved using the actual global domain. There was an incorrect assumption that no conflict could be resolved via an objective propagation when there were multiple unbounded variables in the objective. This PR adds a test and rejects any explanation if there's an unbounded term.    

## Checklist

- [ ] Tests are passing

## Related issue
Closes #3273


fwesselm (approved):
Thanks, @Opt-Mucca!

highs/mip/HighsDomain.cpp
```diff
@@ -3547,10 +3546,8 @@ bool HighsDomain::ConflictSet::explainBoundChange(
       HighsInt ninfmin;
       HighsCDouble minAct;
       globaldom.computeMinActivity(0, len, inds, vals, ninfmin, minAct);
-      assert(ninfmin <= 1);
```
odow, line 3550: @galabovaa so this assert would have caught the bug, since it would have fired when `ninfmin == 2`. Should we compiled HiGHS with asserts enabled? Or should we just transition cheap asserts like this to some other error?

PR #3276, opened 2026-09-14
Fix 3274

### LINKED ISSUE #3274: Incorrect solution with presolve
Consider this problem:
```
min
 obj: +1 x1 
st
 r0: +1 x1 -1 x2 >= -0
 r1: +1 x0 +1 x2 -1 x3 = +0
 r2: -68 x2 -1 x3 >= -12
bounds
 x0 free
 x1 free
bin
gen
 x0
 x1
 x3
semi
end
```

* presolve=on finds objective=1
* presolve=off finds objective=0

This is a bug in presolve.

## Claude's patch

```patch
diff --git a/highs/presolve/HPresolve.cpp b/highs/presolve/HPresolve.cpp
index 3af6cea3dd..77702ff7a9 100644
--- a/highs/presolve/HPresolve.cpp
+++ b/highs/presolve/HPresolve.cpp
@@ -4307,6 +4307,14 @@ HPresolve::Result HPresolve::rowPresolve(HighsPostsolveStack& postsolve_stack,
           // skip continuous variables
           if (model->integrality_[col] == HighsVarType::kContinuous) continue;
 
+          // the coefficient tightening below relies on the integer column
+          // taking the values col_upper, col_upper - 1, ... (or col_lower,
+          // col_lower + 1, ...). Implied bounds, and the bounds of columns
+          // that were converted to implied integers, need not be integral,
+          // so round them to the nearest integral bound first.
+          col_lower = std::ceil(col_lower - primal_feastol);
+          col_upper = std::floor(col_upper + primal_feastol);
+
           if (val > maxAbsCoefValue + primal_feastol) {
             assert(col_upper != kHighsInf);
             // new matrix coefficient is direction * maxAbsCoefValue; subtract
```

## Log with presolve=on

```julia
julia> begin
           model = read_from_file("/tmp/bug-018.lp")
           optimize!(model)
       end
Using BLAS: Apple Accelerate 
MIP has 3 rows; 5 cols; 7 nonzeros; 4 integer variables (0 binary)
Coefficient ranges:
  Matrix  [1e+00, 7e+01]
  Cost    [1e+00, 1e+00]
  Bound   [0e+00, 0e+00]
  RHS     [1e+01, 1e+01]
Presolving model
3 rows, 4 cols, 7 nonzeros 0s
0 rows, 0 cols, 0 nonzeros 0s
Presolve reductions: rows 0(-3); columns 0(-5); nonzeros 0(-7) - Reduced to empty
Presolve: Optimal

         0       0         0   0.00%   1               1                  0.00%        0      0      0         0     0.0s

Solving report
  Status            Optimal
  Primal bound      1
  Dual bound        1
  P-D integral      0
  Solution status   feasible
                    1 (objective)
                    0 (bound viol.)
                    0 (int. viol.)
                    0 (row viol.)
  Timing            0.00
                    0.00 (Presolve)
                    0.00 (Solve)
                    0.00 (Postsolve)
  Max sub-MIP depth 0
  Nodes             0
  Repair LPs        0
  LP iterations     0
```

## Log with presolve=off
```julia
julia> begin
           model = read_from_file("/tmp/bug-018.lp")
           set_attribute(model, "presolve", "off")
           optimize!(model)
       end
Using BLAS: Apple Accelerate 
MIP has 3 rows; 5 cols; 7 nonzeros; 4 integer variables (0 binary)
Coefficient ranges:
  Matrix  [1e+00, 7e+01]
  Cost    [1e+00, 1e+00]
  Bound   [0e+00, 0e+00]
  RHS     [1e+01, 1e+01]

Presolve is switched off

Solving MIP model with:
   3 rows
   5 cols (0 binary, 4 integer, 0 implied int., 1 continuous, 0 domain fixed)
   7 nonzeros
   Thread count 5 (of 10 threads). Using 1 max workers. Parallel search off

 J       0       0         0 100.00%   -inf            0                  Large        0      0      0         0     0.0s
         1       0         1 100.00%   0               0                  0.00%        0      0      0         0     0.0s

Solving report
  Status            Optimal
  Primal bound      0
  Dual bound        0
  P-D integral      0
  Solution status   feasible
                    0 (objective)
                    0 (bound viol.)
                    0 (int. viol.)
                    0 (row viol.)
  Timing            0.01
                    0.00 (Presolve)
                    0.01 (Solve)
                    0.00 (Postsolve)
  Max sub-MIP depth 0
  Nodes             1
  Repair LPs        0
  LP iterations     0
```

## Claude's explanation

Worktree `worktrees/v2-freeint`, branch `fix/v2-freeint`, commit b0ca2d32ff on top of
5d6c11e1a5 (`fix/combined`). Patch: `fix-freeint.patch` (applies cleanly to
`worktrees/combined` with `git apply --check`).

### Root cause

`highs/presolve/HPresolve.cpp`, `HPresolve::rowPresolve`, lambda `strengthenCoefs`
(lines 4291-4331 of the base commit; called for `<=` and `>=` rows that contain
integer or implied-integer columns). This is the coefficient tightening of
Achterberg et al. (section 3.3): for a `>=` row with minimal activity `L` and
`d = b - L > 0`, an integer column with coefficient `|a_j| > d` gets the
coefficient `d` and the side is moved by `(|a_j| - d) * bound_j`. The argument
requires that the column only takes the values `bound_j, bound_j +/- 1, ...`,
i.e. `bound_j` must be integral. The code takes the bound from
`impliedRowBounds.getImplVarLower/Upper(row, col)`, which is the tighter of the
column bound and the *implied* bound; implied bounds are stored unrounded
(`updateColImpliedBounds` calls `changeImplColLower/Upper` with the raw value),
and a continuous column that `convertImpliedInteger` turns into an implied
integer keeps whatever fractional implied bound it already had. With a
fractional bound the new row cuts off feasible integer points.

What happens on the minimized models (traced with temporary printf):

- issue-029 (`min x0 + x3; x0 + x1 + x2 = 0; x0 - x1 >= 0; -10 x1 + x3 >= 0`,
  `x0, x2` free integers, `x1 >= 0`, `x3 <= 1`): row 2 gives `x1` the implied
  upper bound 0.1 while it is still continuous; `colPresolve` then converts
  `x1` to an implied integer through the equation. `rowPresolve` on
  `x0 - x1 >= 0` computes `L = 0 - 0.1`, `d = 0.1`, and rewrites the row to
  `0.1 x0 - 0.1 x1 >= 0.09`, i.e. `x0 - x1 >= 0.9`, using `u_1 = 0.1`. The
  true next value of `x1` below 0.1 is 0, not -0.9, so `x0 >= 1` is forced and
  presolve returns objective 1 (Infeasible on the full instance).
- issue-018 (`min x1; x1 - x2 >= 0; x0 + x2 - x3 = 0; -68 x2 - x3 >= -12`):
  `x2` gets the implied upper bound 12/68 from row 2, becomes an implied
  integer through row 1, and `x1 - x2 >= 0` is tightened to
  `x1 - x2 >= 0.824`, so `x1 >= 1` (objective 1 instead of 0; 8 instead of 0 on
  the full instance).
- issue-026 (`x1 + x2 >= 1; 2 x0 - 2 x1 + 8 x2 + 7 x3 >= 16; -x2 + x3 >= 0;
  -x0 + x2 + 2 x3 = 2`): `x2` (implied integer via the equation) has the
  implied lower bound 7/8 from row 1; `x1 + x2 >= 1` becomes
  `x1 + x2 >= 1.875`; with `x2 = 1` (forcing row) this gives `x1 >= 1`, which
  contradicts `x1 <= 0.5`, hence Infeasible.

### The patch

In `strengthenCoefs`, after skipping continuous columns, round the bounds of
the integer column before they are used:
`col_lower = ceil(col_lower - primal_feastol)`, `col_upper = floor(col_upper + primal_feastol)`.

Why this is correct: the rounded bound `l'` (`u'`) is a valid bound for the
integer column, so `x_j` really takes values `l', l'+1, ...`. The value
`d = b - L` still uses the activity bound computed with the unrounded bound
`l <= l'`, i.e. `L <= L'` and `d >= d'`. For a column with `a_j > d` the new
row `d x_j + rest >= b - (a_j - d) l'` is equivalent to the old one at
`x_j = l'`, and for `x_j >= l' + 1` it is redundant because
`rest >= min(rest) - a_j (l' - l) - d (x_j - l' - 1)` holds for every feasible
`rest`; conversely the old row is redundant there because `a_j + L > b`. The
same argument applies to upper bounds and to `<=` rows (the code handles them
by the `direction` sign). Eight lines, no behaviour change when the bounds are
already integral.

Opt-Mucca:
## Description

This fixes #3274 The issue was that a continuous column may have had a tighter (non-integer) implied bound than its actual bound, which got carried over in the implied integer upgrade. This caused issued downstream in coefficient strengthening, where the non-integer implied bound caused on incorrect deduction. Fix: Use implied bounds to tighten the columns bound when upgrading implied integrality.

## Checklist

- [x] Tests are passing

## Related issue
Closes #3274


odow:
Nice nice. The lesson is definitely that the patches are a good starting point to find where the offending code is, but they can't be applied blindly.

Opt-Mucca:
> Nice nice. The lesson is definitely that the patches are a good starting point to find where the offending code is, but they can't be applied blindly.

In this case I've essentially applied the second patch exactly (only re-worded the comment)! The first patch in the issue was treating the symptom instead of the actual cause.

odow:
> The first patch in the issue was treating the symptom instead of the actual cause.

Yeah this is what I meant. The usual story. The patches _look_ correct, but they sometimes struggle to understand the deeper story.

check/TestMipSolver.cpp
```diff
@@ -1719,6 +1719,7 @@ TEST_CASE("issue-3170", "[highs_test_mip_solver]") {
   const HighsModelStatus require_model_status = HighsModelStatus::kOptimal;
   const double optimal_objective = -21446579.3647;
   solve(highs, kHighsOffString, require_model_status, optimal_objective);
+  highs.setOptionValue("output_flag", dev_run);
```
fwesselm, line 1722: Was this added for debugging?

highs/presolve/HPresolve.cpp
```diff
@@ -481,9 +481,15 @@ HPresolve::StatusResult HPresolve::convertImpliedInteger(HighsInt col,
       ++rowsizeImplInt[nonzero.index()];
   }
 
+  // Potentially strengthen existing bound using implied bounds
```
fwesselm, line 484: Alternatively, I think you could call `updateColImpliedBounds` for all column elements (which would re-do some of the work).

Sorry, this may be a weird request, but would you consider adding a comment as a reminder why this is needed here (- I tend to forget). 

For example:

> // updateColImpliedBounds updates and rounds col_lower_/col_upper_ for
> // non-continuous variables, so tighten using implied bounds before rounding

fwesselm (approved):
Thanks, @Opt-Mucca! Just two questions from me.

check/TestMipSolver.cpp
Opt-Mucca: No. This is me being sloppy and not double checking the entire file..... I was trying my hand at using an LLM to help with test creation. My bad. Edit: Will change this line and some of the others.

Double edit: This was added in the merge from updates on `latest` @jajhall . I'm not sure why they're here, but I'd now probably leave them as I don't think they hurt.

highs/presolve/HPresolve.cpp
Opt-Mucca: Do we want to re-do all of the work here? Is there any extra benefit to calling `updateColImpliedBounds` (we'd have to call it for all rows featuring this column)?

Not a weird request at all. I'll add a comment based on yours!

check/TestMipSolver.cpp
jajhall: > No. This is me being sloppy and not double checking the entire file..... I was trying my hand at using an LLM to help with test creation. My bad. Edit: Will change this line and some of the others.
> 
> Double edit: This was added in the merge from updates on `latest` @jajhall . I'm not sure why they're here, but I'd now probably leave them as I don't think they hurt.

This call is used for debugging by setting 'dev_run' true, or by commenting it out temporarily. It's added because 'solve' no longer sets 'output_flag' to 'dev_run'

PR #3278, opened 2026-09-14
test case of model file 1451.lp bug fixed

aryanshukla4:
## Checklist

- [x] Tests are passing


aryanshukla4:
Fixes a bug where constants on the LHS of a constraint (e.g., x - 1 >= 2) were ignored.
Previously, con->expr->offset was only applied to the objective row. This caused x - 1 >= 2 to incorrectly evaluate as x >= 2 instead of x >= 3. This PR subtracts the LHS offset from the row bounds, which safely corrects the evaluation (even for unbounded constraints).

jajhall (commented):
Thanks, but it's correct to ignore the constant in the LHS, as this is done by the commercial solver against which we compare. However, as it does, we should issue a warning.

PR #3281, opened 2026-09-14
Fix 3221

### LINKED ISSUE #3221: Apple silicon: default thread count halves core count; cache-line padding assumes 64 bytes
## Summary

Two defaults in `highs/parallel/` assume x86 characteristics and cost performance on Apple silicon (and, for one of them, AArch64 generally). Reporting as an issue rather than a PR, per CONTRIBUTING.md.

**Disclosure:** this investigation was carried out with AI assistance (Claude). I'm aware AI-generated PRs aren't accepted, which is why this is an issue — the findings and measurements are below so the team can decide whether they're worth acting on, and implement independently if so.

Environment: Apple M5, 10 physical cores (4P + 6E, no SMT), macOS 26.6, Apple clang, CMake 4.4.2, Release builds.

---

### 1. Default thread count assumes SMT, halving core count on Apple silicon

`highs/parallel/HighsParallel.h:25`

```cpp
numThreads = (std::thread::hardware_concurrency() + 1) / 2;
```

This halves the logical count to recover physical cores, which is right for SMT-enabled x86. Apple silicon has no SMT, so `hardware_concurrency()` already *is* the physical core count and this discards half the machine — 5 workers on a 10-core M5. Since `getMaxNumWorkers()` returns `ceil(1.7 * num_threads())`, the MIP solver then runs 9 workers instead of 17.

Querying `hw.physicalcpu` via `sysctlbyname` under `__APPLE__`, leaving the existing heuristic elsewhere, gives (MIPLIB 2017, 10 benchmark instances solvable in 10–60s, 5 reps, paired and interleaved, `--parallel=on`):

| metric | result |
|---|---|
| node throughput | **+27% median**, higher on 9 of 10 instances |
| shifted geometric mean (s=10) | **+16.2%** (35.77s → 30.77s) |
| median per-instance wall-clock ratio | −0.4%, faster on 5 of 10 |
| runs hitting a 300s limit | 1 before, 0 after |

The throughput figure is the reliable one. Wall-clock-to-optimality varies a lot because a different worker count explores a different tree: `enlight_hard` runs 1.65x the node rate yet finishes 41% slower, having entered a tree twice the size. The geometric mean is carried by a few large wins (`gmu-35-40` 2.95x, `swath3` 1.69x) while the median is flat, so I'd read +16% as an aggregate over this set, not a per-instance expectation.

I scoped this to `__APPLE__` rather than all of AArch64 deliberately — some ARM server parts (e.g. ThunderX2) do have SMT, so the halving may be correct there.

HiPO gains about +1.9% from the same change; its parallel scaling saturates near 4 threads (1.33x going from 1 to 10 on 3D min-cost-flow LPs), so doubling 5→10 lands on the flat part of the curve.

### 2. Cache-line padding is 64 bytes; AArch64 uses 128

`highs/parallel/HighsCacheAlign.h:19` returns 64, and `HighsSplitDeque.h` pads its concurrently-written members with `alignas(64)` (`ownerData`, `splitRequest`, `stealerData`, `workerBunkData`, `taskArray`, plus `haveJobs`/`sleeperStack` in `WorkerBunk`); `HighsBinarySemaphore.h:25` does the same.

AArch64 (Apple silicon included) and 64-bit POWER use 128-byte cache lines — `sysctl hw.cachelinesize` reports 128 here. So on those targets the padding does not place owner and stealer data on distinct lines, and the false sharing the padding exists to prevent still occurs.

Related: the assertions in the `HighsSplitDeque` constructor

```cpp
static_assert(offsetof(HighsSplitDeque, stealerData) == 128, ...);
```

pass vacuously on AArch64 — they check 64-byte multiples regardless of the real line size, so nothing verifies the padding does its job.

In fairness this measured **neutral** for me (+0.3%, faster in 13 of 24 paired runs), so I'd put it as a correctness issue rather than a performance one. Deriving the padding from an architecture-dependent constant also makes those `offsetof` assertions meaningful on both architectures.

### 3. The `HighsSpinMutex` ToDo is a trap worth documenting

`highs/parallel/HighsSpinMutex.h:30`

```cpp
// ToDo: See if this is OK on Mac M1
std::this_thread::yield();
```

I checked, and the obvious answer is wrong. Substituting the usual PAUSE stand-in on AArch64 (`isb sy` — `yield` is a no-op on many implementations) is a **regression**: −4.6% median, winning only 7 of 24 paired runs on parallel MIP.

The reason is that the callers are time-budgeted spin loops — `HighsBinarySemaphore::acquire`, `HighsTask::getStealerIfUnfinished`, `HighsSplitDeque::runStolenTask` — which spin for a fixed number of microseconds, doubling the iteration count, before sleeping. A backoff primitive two orders of magnitude cheaper doesn't shorten the wait; it spins far more times on a contended atomic within the same budget and stops handing the core to the thread being waited on. The constants are calibrated against the cost of the yield, so the primitive and the constants would need revisiting together.

Might be worth replacing the ToDo with a note to that effect, since it invites a change that costs ~5%.

---

### Unrelated, happy to split into its own issue

`SFLOAT` (single-precision cuPDLP) does not compile. In `highs/pdlp/cupdlp/glbopts.h`, `cupdlp_int` is typedef'd only inside the `#ifndef SFLOAT` branch, so `-DSFLOAT` fails immediately (~40 errors). Adding the missing typedef exposes ~15 more: two header/definition mismatches (`cupdlp_proj.h:11` declares `double *r` where the definition uses `cupdlp_float *r` — identical in fp64, conflicting in fp32), and the rest structural in `CupdlpWrapper.cpp`, which passes HiGHS's `double` arrays straight into cuPDLP with no conversion layer.

Separately, `kDefaultKktTolerance` is 1e-7 (`HConst.h:345`), which is below float32 machine epsilon (1.19e-7) — so an fp32 PDLP could not reach the default tolerance regardless. Possibly the mode should just be removed rather than repaired.

---

### Also checked, negative results

- `-mcpu=native` / `-mcpu=apple-mN`: **+0.1%** (shifted geometric mean, 6 MIPLIB instances, 4 reps). Node counts were identical between builds, so this is a clean comparison — there's simply nothing there.
- Hand-written NEON in the simplex/factor kernels doesn't look worthwhile: profiling a MIP solve, self time concentrates in `cmirCutGenerationHeuristic`, `HFactor::btranU`/`ftranU`/`btranL`, and malloc/free traffic — sparse, indirect-indexed, branch-heavy code that doesn't vectorise. Allocator pressure ranks alongside the hottest compute kernel, which may be a more interesting target and is architecture-neutral.

### Reference implementation

A branch with (2) and a comment for (3) is at https://github.com/subsriram/HiGHS/tree/claude/apple-silicon-optimization-05fa32 — 4 files, `highs/parallel/` only, on top of `latest`, full suite passing (168/168, `-DALL_TESTS=ON`). Linking purely as a reference for the measurements; not proposing it for merge. _(Edited: this originally read 7 files branched from `master`. After @filikat pointed out #3137, the thread-count part was dropped and the branch rebased onto `latest` — see comment below.)_

Caveats on the numbers: one machine, 10 MIPLIB instances (12 of my 24 readable candidates were too hard for HiGHS within 100s), 5 reps. Variance between instances is high. Happy to run a wider sweep if that would help.

filikat:
## Description
- Substitute hard coded value of 64 for cache line size with a constant `kCacheLineSize`, so that it is easier to modify it and recompile Highs on a machine with different cache line size.
- Update the "ToDo" comment in `HighsSpinMutex` to highlight that `std::this_thread::yield()` is the correct instruction to use on AArch64.

## Checklist

- [x] Tests are passing

## Related issue
Closes #3221


highs/parallel/HighsSchedulerConstants.h
mathgeekcoder, line 11: I'm surprised I haven't seen this code before. Using `enum` is fine, but doesn't let us specify different types (e.g., what if we needed a `double` constant?).

Perhaps we can replace existing code with the following:

```cpp
struct HighsSchedulerConstants {
  static constexpr int kNumTryFac = 16;
  static constexpr int kMicroSecsBeforeSleep = 5000;
  static constexpr int kMicroSecsBeforeGlobalSync = 1000;
  static constexpr size_t kCacheLineSize = 64;
};
```

highs/parallel/HighsSplitDeque.h
```diff
@@ -49,6 +49,9 @@ extern "C" void AnnotateHappensAfter(const char* f, int l, void* addr);
 class HighsSplitDeque {
   using cache_aligned = highs::cache_aligned;
 
+  static constexpr size_t cacheLineSize =
```
mathgeekcoder: If we make the change to constants, this code should no longer be necessary

highs/parallel/HighsSplitDeque.h
```diff
@@ -191,18 +194,18 @@ class HighsSplitDeque {
   };
 
  private:
-  static_assert(sizeof(OwnerData) <= 64,
+  static_assert(sizeof(OwnerData) <= cacheLineSize,
```
mathgeekcoder: Can possibly add check for power-of-two:

```cpp
static_assert(cacheLineSize > 0 && (cacheLineSize & (cacheLineSize - 1)) == 0, "cache line size must be a power of two");
```

mathgeekcoder (commented):
Looks good! Minor suggestions if you wanted some improvements.

highs/parallel/HighsSchedulerConstants.h
filikat: Good idea, I will make the change. There are probably some other hard coded numbers that I can include here.

highs/parallel/HighsSplitDeque.h
filikat: I had that exact code already, but then I realised that `alignas` spits out lots of errors if the parameter is not a power of two, so that static assert would be swamped by many other messages saying the exact same thing.

mathgeekcoder: Makes sense!

mathgeekcoder (approved):
Looks good!

PR #3282, opened 2026-09-14
Added a warning for nonzero LHS constants

aryanshukla4:
## Checklist

- [x] I have read the [contributing guidelines](https://github.com/ERGO-Code/HiGHS/blob/latest/CONTRIBUTING.md)
- [ ] This PR targets the `latest` branch
- [x] Tests are passing
- [x] Documentation was updated where relevant
- [x] This PR is not primarily AI-generated (per the AI contributions policy in CONTRIBUTING.md)

PR #3283, opened 2026-09-14
Added warning about constant LHS entries in LP file constraints

jajhall:
## Description

Prompted by #3279, these are viewed as illegal numeric variable names, so are ignored. Unit test `filereader-edge-cases` exposed this with `1451.lp`, which has now been extended for code coverage.

When processing the LP file reader `Model` to create a `HighsLp` instance, matrix entries that are explicit zeros and/or sum to a nonzero value always were reported, so the illegal numeric variable names are handled in the same way. Not all occurrences of such are logged - to avoid verbosity. A maximum of 10 of any such reports are logged. Maybe this is excessive, but I don't want to put more time into this

## Checklist

- [x] Tests are passing

## Related PR

Improvement on #3282, so closes it


odow:
JuMP does
```Julia
julia> using JuMP

julia> model = read_from_file("/tmp/bug.lp");

julia> print(model)
Min 2 x
Subject to
 x + 1 ≥ 0
 x ≥ 0

julia> print(read("/tmp/bug.lp", String))
min
    obj: 2 x
s.t.
    x + 1 >= 0
end

julia> using HiGHS

julia> set_optimizer(model, HiGHS.Optimizer)

julia> optimize!(model)
Using BLAS: Apple Accelerate 
ERROR: In `MathOptInterface.ScalarAffineFunction{Float64}`-in-`MathOptInterface.GreaterThan{Float64}` constraint: Constant 1.0 of the function is not zero. The function constant should be moved to the set. You can use `MOI.Utilities.normalize_and_add_constraint` which does this automatically.
```

odow:
I took another look at this. Gurobi doesn't ignore the constant. It interprets the constant _as the name of a new decision variable_. That effectively means there is a free variable in the constraint. It's clearly the wrong thing to do. 

I think we should keep the error here.

PR #3285, opened 2026-09-14
Add fix for dominatedColumns and DuplicateColumn::undo

Opt-Mucca:
## Description

This fixes another instance from @odow and his LLM fuzzer. The instance raised two issues and has been added as a test.
1. `dominatedColumns` did not guard against the case where it tries to fix a value to an infinite bound. This resulted in a `dualInfeasible` status for a feasible instance. I think in general we fixed some problems earlier to avoid this case because it wasn't post-solvable, so I think this has just slipped under the radar?
2. When post-solving a duplicate column that was a combination of an integer and continuous column, the integer column was always rounded up, which sometimes resulted in a solution that the continuous column couldn't satisfy without violating its bounds. I've added a check that rounds down if rounding up is potentially unsafe. I think the logic is safe here, but postsolve is always confusing as hell.

## Checklist

- [x] Tests are passing


fwesselm (approved):
Thanks again, @Opt-Mucca! Good to merge!

Opt-Mucca:
@fwesselm please merge your comments first. Thanks for reviewing!

fwesselm:
> @fwesselm please merge your comments first. Thanks for reviewing!

Let me try to add one more test for the uncovered code in `HighsPostsolveStack` before merging.

PR #3286, opened 2026-09-14
Enhance clique extraction

fwesselm:
## Description

1. `HPresolve::normaliseCliqueRows` scales all-binary rows to `+-1` coefficients and `<=` form for clique extraction if possible.
2. On `neos-631710` the normalisation enables clique merging to detect and remove 166500 binary-binary implication (clique) rows, reducing the presolved model to 3076 rows.
3. Added unit tests.

## Checklist

- [x] Tests are passing

### DIFF
--- a/highs/mip/HighsCliqueTable.cpp
+++ b/highs/mip/HighsCliqueTable.cpp
@@ -851,7 +851,7 @@ void HighsCliqueTable::extractCliques(
     const HighsMipSolver& mipsolver, std::vector<HighsInt>& inds,
     std::vector<double>& vals, std::vector<int8_t>& complementation, double rhs,
     HighsInt nbin, std::vector<HighsInt>& perm, std::vector<CliqueVar>& clique,
-    double feastol) {
+    double feastol, HighsInt origin) {
   HighsImplications& implics = mipsolver.mipdata_->implications;
   HighsDomain& globaldom = mipsolver.mipdata_->getDomain();
 
@@ -939,7 +939,8 @@ void HighsCliqueTable::extractCliques(
         clique.emplace_back(inds[pos], 1);
     }
 
-    addClique(mipsolver, clique.data(), nbin);
+    addClique(mipsolver, clique.data(), nbin, false,
+              nbin == ntotal ? origin : kHighsIInf);
     if (globaldom.infeasible()) return;
     // printf("extracted this clique:\n");
     // printClique(clique);
@@ -1334,7 +1335,7 @@ void HighsCliqueTable::extractCliques(HighsMipSolver& mipsolver,
       entries[col] += val;
     }
 
-    auto checkRow = [&](double rhs, HighsInt direction) {
+    auto checkRow = [&](HighsInt row, double rhs, HighsInt direction) {
       if (direction * rhs == kHighsInf) return;
       rhs = direction * (rhs - offset);
       inds.clear();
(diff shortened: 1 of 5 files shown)


highs/presolve/HPresolve.cpp
```diff
@@ -1584,13 +1584,145 @@ HPresolve::Result HPresolve::dominatedColumns(
...
   HighsDomain& domain = mipsolver->mipdata_->getDomain();
   HighsCliqueTable& cliquetable = mipsolver->mipdata_->cliquetable;
 
   shrinkProblem(postsolve_stack);
 
+  // first call?
+  firstCall = !mipsolver->mipdata_->cliquesExtracted;
+
+  // todo: rows that become setppc after domain propagation (below) are
+  // not normalised and thus cannot be deleted by clique merging
```
Opt-Mucca, line 1723: I think this is fine! The change is going to cover the vast majority of cases already.

Opt-Mucca (approved):
This is a great change! It's surely going to be a nice performance bump.

highs/presolve/HPresolveTest.cpp
```diff
@@ -48,6 +50,13 @@ HPresolve::Result HPresolve::presolveRuleTestParallelRowsAndCols(
                "HPresolve::presolveRuleTestParallelRowsAndCols\n");
   return detectParallelRowsAndCols(postsolve_stack);
 }
+HPresolve::Result HPresolve::presolveRuleTestProbing(
+    HighsPostsolveStack& postsolve_stack) {
+  assert(options->presolve_rule_test == kPresolveRuleProbing);
+  highsLogUser(options->log_options, HighsLogType::kInfo,
+               "HPresolve::presolveRuleTestProbing\n");
+  return runProbing(postsolve_stack);
```
Opt-Mucca, line 59: Does this need to test whether `mipsolver != nullptr`?

highs/presolve/HPresolve.cpp
```diff
@@ -1620,13 +1620,145 @@ HPresolve::Result HPresolve::dominatedColumns(
...
+    if (computeDynamism(getStoredRow()) > maxDynamism) continue;
+
+    // compute scaling factor that makes all coefficients integral
+    double intScale = HighsIntegers::integralScale(
+        rowCoefsInt, options->small_matrix_value, options->small_matrix_value);
+    if (intScale == 0 || intScale > 1e3) continue;
+
+    // scale so that we have a <= inequality or equation
+    if (model->row_upper_[row] == kHighsInf) intScale = -intScale;
+    scaleStoredRow(row, intScale, true);
```
Opt-Mucca, line 1683: Can this be potentially incorrect? We scale a row to make it integral, but the scaled RHS may still be integral, and this function is going to call `std::round`. Could we end up with something like `2x1 + 2x2 <= 3.7`, which should only permit `x1 + x2 <= 1`, but after rounding the RHS permits `x1 + x2 = 2`.

highs/presolve/HPresolveTest.cpp
fwesselm: Good suggestion. It should not happen, but I will add a check.

highs/presolve/HPresolve.cpp
fwesselm: @Opt-Mucca, you are right. I did not even check the details of `scaleStoredRow` and was assuming it would cover all the cases. I am going to use `std::floor` and `std::ceil` in `scaleStoredRow`.

highs/mip/HighsCliqueTable.cpp
```diff
@@ -1294,35 +1292,60 @@ void HighsCliqueTable::extractCliques(HighsMipSolver& mipsolver,
...
+          if (!globaldom.isBinary(col) || globaldom.isFixedToVal(col, 1 - dir))
+            continue;
+
+          // add to clique
+          clique.emplace_back(col, dir);
+        }
+
+        // add clique to clique table
+        if (clique.size() >= 2) {
+          addClique(mipsolver, clique.data(),
```
Opt-Mucca: I've spent way too long looking at this, but I think the current code (and original code) is missing a case here. If you had some row like `0.5 <= x1 + x2 <= 1` then it'd be stored as a non-equality clique. Assuming the clique then gets merged into something else, the entire row will be deleted. The equality information is now missing however and the resulting problem could admit incorrect solutions! I am guessing this never happens because all (my concern is `near all`) solution paths have gone through row presolve and upgraded the row to an equality.

highs/presolve/HPresolve.cpp
Opt-Mucca: @fwesselm I don't why I didn't suggest that change........ I was somehow thinking of rounding them before.

highs/mip/HighsCliqueTable.cpp
fwesselm: @Opt-Mucca, thanks! `extractCliques` should handle this. I will update the code shortly.

PR #3287, opened 2026-09-15
Fix bug in strengthen coefficients

Opt-Mucca:
## Description

This fixes another instance that @odow generated with his LLM fuzzer. The issue is that a non-continuous column is using some fractional implied bound, which dominoes into an error elsewhere. I've marked this as a draft PR as I'm not 100% confident on the generalisation of the fix here yet (@fwesselm nearly forgot about this, so wanted to create a draft PR)

## Checklist

- [x] Tests are passing


Opt-Mucca:
I believe this issue has been solved by #3276 
I would close the PR and re-open if @odow shows with his fuzzer that it's a separate issue.

PR #3290, opened 2026-09-15
Immediately return after fixing column

Opt-Mucca:
## Description

There's an issue with `dominatedColumns`, where a truly redundant column may be able to be fixed to either its lower or its upper bound, and the code will try to fix to both one-after-another. This raises an error when it tries to fix an already deleted column. The code-fix is to return after fixing the column to one side.

This was found by looking at a failing instance on my clique table changes. 

## Checklist

- [x] Tests are passing


highs/presolve/HPresolve.cpp
fwesselm: I think this the symmetric case. It should also return immediately.

fwesselm (approved):
Sorry, @Opt-Mucca, this is my fault. 

Please merge (after making the similar change for the upper bound case)!

EDIT: I can just add this. I hope that is OK.

highs/presolve/HPresolve.cpp
Opt-Mucca: I didn't change this because there was no code that would be executed afterwards, but that's laziness on my part (puts a nice landmine there if the function ever got extended). Thanks for updating it.

PR #3291, opened 2026-09-15
Fix 3288

### LINKED ISSUE #3288: High_getBasis should return whether the basis is valid instead of Ok
We should return kHighsStatusErr if the basis is not valid:
 
https://github.com/ERGO-Code/HiGHS/blob/73cac48c5340d775a477087198611862559be250/highs/interfaces/highs_c_api.cpp#L571-L581

Ditto with getSolution if the primal solution is not valid

jajhall:
## Description

C API methods `Highs_getBasis` and `Highs_getSolution` both return `kHighsStatusError` if there is, respectively, no valid basis or no primal solution. Note that the validity of the basis was always available in `HighsInfo`, so this gives another prompt to users to identify when there is no valid basis.

Also prompted significant rationalisation of the `getKktFailures` family of methods in `HighsSolution.h` and comments on the methods that remain.

## Checklist

- [x] Tests are passing

## Related issue
Closes #3288


odow (commented):
LGTM.

We should probably update the docstrings:

https://github.com/ERGO-Code/HiGHS/blob/6e7128a5aea56cadd3b68d7343b5f1031c8f8e11/highs/interfaces/highs_c_api.h#L1051

"call succeeded" -> "basis is valid"

https://github.com/ERGO-Code/HiGHS/blob/6e7128a5aea56cadd3b68d7343b5f1031c8f8e11/highs/interfaces/highs_c_api.h#L462

"the call succeeded" -> "a primal solution exists"

jajhall:
> LGTM.
> 
> We should probably update the docstrings:
> 
> https://github.com/ERGO-Code/HiGHS/blob/6e7128a5aea56cadd3b68d7343b5f1031c8f8e11/highs/interfaces/highs_c_api.h#L1051
> 
> "call succeeded" -> "basis is valid"
> 
> https://github.com/ERGO-Code/HiGHS/blob/6e7128a5aea56cadd3b68d7343b5f1031c8f8e11/highs/interfaces/highs_c_api.h#L462
> 
> "the call succeeded" -> "a primal solution exists"

OK, although we don't do that for other calls

PR #3293, opened 2026-09-16
Add test for number of threads of OpenBLAS

filikat:
## Description

- Add `openblas_get_num_threads`, to query the number of threads used by OpenBLAS.
- After setting the number of threads for OpenBLAS to 1, check that indeed OpenBLAS is using 1 thread. This may catch errors in how the flag `HIPO_USES_OPENBLAS` is passed to HiPO.
- If `BLAS_LIBRARIES` is used and the provided library contains the string `openblas`, set the BLAS metadata so that HiPO knows that OpenBLAS is being used.

## Checklist

- [x] Tests are passing


mathgeekcoder (commented):
**Quick question:** are you wanting to use `openblas_get_num_threads` for anything else?

An alternate implementation would be to modify the `set_num_threads` to also return the number of threads:

```cpp
int openblas_set_num_threads(int num_threads) {
#if defined(HIPO_USES_OPENBLAS)
  openblas_set_num_threads(num_threads);
  return openblas_get_num_threads();
#else
  return num_threads;
#endif
}
```

This would make most of the other changes unnecessary.

filikat:
`openblas_get_num_threads` is not used anywhere else at the moment, so I can remove it and follow your suggestion. I would still leave the check of whether the BLAS provider and `HIPO_USES_OPENBLAS` agree. If OpenBLAS is used, but `HIPO_USES_OPENBLAS` is not set, the performance deteriorates considerably because the number of threads is not set to 1. These changes are meant to guarantee that it doesn't happen.

highs/ipm/hipo/ipm/FactorHighsSolver.cpp
```diff
@@ -593,7 +594,10 @@ void FactorHighsSolver::setParallelBeforeSymbolic() {
 }
 
 static bool usingAppleBlas() {
-  return strstr(HighsExtras::blas::getInfo()->provider, "Apple") != nullptr;
+  std::string provider = HighsExtras::blas::getInfo()->provider;
```
mathgeekcoder, line 597: Perhaps we can include something like this in `stringutil.h`?

```cpp
#if defined(_WIN32) || defined(_WIN64)
#include <string.h>  // _stricmp
#else
#include <strings.h>  // strcasecmp
#endif

inline int stringcasecmp(const char* lhs, const char* rhs) {
#if defined(_WIN32) || defined(_WIN64)
  return _stricmp(lhs, rhs);
#else
  return strcasecmp(lhs, rhs);
#endif
}
```

Then this becomes:
```cpp
static bool usingAppleBlas() {
   return stringcasecmp(HighsExtras::blas::getInfo()->provider, "apple") == 0;
}
```

It should be faster and cleaner than creating a string, converting to lower and comparing.

highs/ipm/IpxWrapper.cpp
```diff
@@ -420,6 +420,25 @@ HighsStatus solveQpHipo(HighsQpSolverObject& solver_object) {
                    solver_object.callback_);
 }
 
+static bool usingOpenBLAS() {
+  std::string provider = HighsExtras::blas::getInfo()->provider;
```
mathgeekcoder, line 424: As above

```cpp
static bool usingOpenBLAS() {
   return stringcasecmp(HighsExtras::blas::getInfo()->provider, "openblas") == 0;
}
```

cmake/FindHipoDeps.cmake
```diff
@@ -31,6 +31,11 @@ endfunction()
 function(highs_configure_blas_metadata)
     set(HIGHS_BLAS_COMPILE_DEFINITION "" PARENT_SCOPE)
 
+    string(TOLOWER "${BLAS_LIBRARIES}" BLAS_LIBRARIES_LOWER)
```
mathgeekcoder, line 34: Sorry, can you explain why this is needed?

highs/ipm/hipo/ipm/FactorHighsSolver.cpp
mathgeekcoder: Alternatively, we could modify `HighsExtrasFeatureInfo`:
```cpp
bool HighsExtrasFeatureInfo::isProvider(const char* check) {
   return stringcasecmp(provider, check) == 0;
}
```
such that:
```cpp
HighsExtras::blas::getInfo()->isProvider("apple");
```

mathgeekcoder (changes requested):
Looking better! I've added a couple more optional suggestions

highs/ipm/hipo/ipm/FactorHighsSolver.cpp
filikat: Yes, this looks neat, I will make the changes

cmake/FindHipoDeps.cmake
filikat: If I use OpenBLAS by passing `libopenblas.so` to `BLAS_LIBRARIES`, CMake doesn't know which vendor is being used and so it doesn't set the BLAS metadata (i.e., the provider is unknown, etc). Crucially, this means that `HIPO_USES_OPENBLAS` is not set, and OpenBLAS runs multi-threaded, with a substantial reduction in performance.
Maybe there are better ways of achieving this.

highs/ipm/hipo/ipm/FactorHighsSolver.cpp
filikat: We need to check for substrings, not exact match though (so that "Apple Accelerate" matches "apple"). I replaced `strcasecmp` with `strcasestr`. Unfortunately there seems to be no equivalent on Windows. One possible solution is to use `_strnicmp` iteratively. I am not sure of whether this is fast, but it shouldn't be a big deal. The providers are short and need to be checked only a couple of times.

filikat: I'm having some issues to include `stringutil.h` in `HighsExtrasApiBinding.h` in a way that works everywhere, in order to implement `isProvider`. I will remove `isProvider` for now.

PR #3298, opened 2026-09-17
Fix initial sweep interrupt

jajhall:
## Description

This is my fix for `HighsPostsolveStack::ZeroObjSingletonContinuousCol::undo` that is only WIP for the primal-dual case, and a fix for the failure when `HPresolve::initialSweep` is interrupted by the time or reductions limit. However, that triggered work on handling failures better - in terms of re-running internally and communicating better with users. 

When exceptions are caught, if they result from memory allocation errors, this is identified and the model status `HighsModelStatus::kMemoryLimit` is set. The try-catch block in IPX has been removed since LP solving takes place inside its own try-catch block. Please review this @filikat 

The "active cost norm" used for relative dual feasibility doesn't take into consideration any free column singletons used to transfer the objective from "row 0" of the matrix to the "c" vector. Hence "getEffectiveCosts" determines the objective function after all free column singletons have been substituted (possibly recursively) and it's the norm of the active costs for this vector that is used in getKktFailures. This norm and the corresponding bound norm for relative primal feasibility are made available in HighsInfo. Please review this @filikat 

Please review the presolve modifications, @fwesselm 

I felt that this needed merging before it grew too big!

## Checklist

- [ ] Tests are passing

### DIFF
--- a/check/TestModelProperties.cpp
+++ b/check/TestModelProperties.cpp
@@ -141,3 +141,23 @@ TEST_CASE("afiro-ill-conditioning", "[highs_model_properties]") {
   highs.getIllConditioning(ill_conditioning, constraint);
   highs.getIllConditioning(ill_conditioning, !constraint);
 }
+
+TEST_CASE("infinite-bounds", "[highs_model_properties]") {
+  Highs h;
+  h.setOptionValue("output_flag", dev_run);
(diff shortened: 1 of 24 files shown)


jajhall:
I'll try to split this up into separate PRs

PR #3300, opened 2026-09-17
Non stop initial sweep

jajhall:
## Description

`HPresolve::initialSweep` cannot be interrupted by the time or reductions limit, otherwise the reduced LP does not correspond to the reductions that have been performed. By setting infinite time and reductions limits for the duration of  `HPresolve::initialSweep`, this interruption is prevented. Since `HPresolve::initialSweep` has minimal computational overhead (just two passes through the constraint matrix) presolve will not overrun any time limit meaningfully. Although it's possible to recover from an interruption, writing the code to do so is not justified. Please review this @fwesselm 

That triggered work on handling failures better - in terms of re-running internally and communicating better with users. When exceptions are caught, if they result from memory allocation errors, this is identified and the model status `HighsModelStatus::kMemoryLimit` is set, otherwise it's set to `HighsModelStatus::kSolveError`. The try-catch block in IPX has been removed since LP solving takes place inside its own try-catch block. Please review this @filikat

## Checklist

- [ ] Tests are passing


highs/presolve/HPresolve.cpp
```diff
@@ -570,6 +573,9 @@ void HPresolve::chooseRules() {
       }
       bit *= 2;
     }
+  } else if (presolve_light_off) {
+    highsLogUser(options->log_options, HighsLogType::kInfo,
```
fwesselm: So, `presolve_light = off` means that initial sweep is run? This is fine, but maybe a little counterintuitive. Isn't the initial sweep controlled by `this->allow_rule_[kPresolveRuleInitialSweep]`?

highs/presolve/HPresolve.cpp
```diff
@@ -6734,13 +6754,14 @@ HPresolve::Result HPresolve::presolve(HighsPostsolveStack& postsolve_stack) {
   HPRESOLVE_CHECKED_CALL(checkOriginalModelBounds());
 
   if (options->presolve != kHighsOffString && mipsolver == nullptr &&
-      !options->presolve_rule_test) {
+      this->allow_rule_[kPresolveRuleInitialSweep]) {
```
fwesselm, line 6285: If a presolve rule is supposed to be tested in isolation, does it make sense to run the initial sweep?

highs/presolve/HPresolve.cpp
fwesselm: The limits were just set to infinite values, so this call to `checkLimits` could be removed, correct?

fwesselm: I noticed that the same comment is also used above. Is this one correct?

fwesselm (approved):
Looks good, @jajhall!

I am wondering whether the initial sweep could live in its own class? Are you considering to add more features to it?

highs/presolve/HPresolve.cpp
jajhall: This was a hack for testing: it should have been edited out

highs/presolve/HPresolve.cpp
jajhall: Again, I think that I removed this for some reason when debugging. The `!options->presolve_rule_test &&` should be reinserted

highs/presolve/HPresolve.cpp
jajhall: Yes, but this is a point where the initial sweep could be interrupted once it no longer has checked calls elsewhere. Not a reason to complicate matters by allowing initial sweep to be interrupted

jajhall: No, second comment should be deleted

filikat (approved):

fwesselm:
@jajhall, I have re-opened the PR and pushed my changes to it. Interestingly, there are test failures on macos that seem unrelated...

fwesselm:
`HighsExternalDeps.cpp` included `zlib.h` inside a `namespace HighsExtras {}` block. On `macOS`, `zlib.h` transitively includes system headers. The speculation is that this was "hidden" because `macOS` uses unity build (batch size 8), and the batch grouping happened to work. Adding new files shifted the batch boundaries, changing which files were compiled together.

jajhall:
> @jajhall, I have re-opened the PR and pushed my changes to it. Interestingly, there are test failures on macos that seem unrelated...

Indeed, very strange. Worth flagging this up with @galabovaa and @mathgeekcoder

mathgeekcoder:
@fwesselm good spotting! @jajhall thanks for bringing this to my attention.

Definitely sounds like an issue exposed by the unity build on macos (needed to significantly reduce build times). 

FYI: I believe it gets a little confused since `reader.cpp` and `HMpsFF.cpp` directly include `zstr.hpp` (which includes `zlib.h`) in global space, whereas `HighsExternalDeps.cpp` was including it in the `HighsExtras` space. If those source files were not in the same unity batch the `#ifdef` guard might include `zlib.h` in one space but not the other.

PR #3297, opened 2026-09-17
Test 3295

jajhall:
## Description

This merge is from a local copy of https://github.com/LaurentPlagne/HiGHS/tree/perf/unit-diagonal-pivots with which I've done some regression testing and translated some comments into English

## Checklist

- [x] I have read the [contributing guidelines](https://github.com/ERGO-Code/HiGHS/blob/latest/CONTRIBUTING.md)
- [x] This PR targets the `latest` branch
- [ ] Tests are passing
- [x] Documentation was updated where relevant
- [x] This PR is not primarily AI-generated (per the AI contributions policy in CONTRIBUTING.md)

PR #3295, opened 2026-09-17
Perf/unit diagonal pivots

LaurentPlagne:
## Description

Hi Julian, Ivet and HiGHS team,

I work at EDF labs and had the pleasure to meet Ivet at the last local Julia conference in Paris. I do use HiGHS in the context of dynamic programming for Hydro planning. Ivet and I discussed the performance of HiGHS when applied to a large collection of small very similar LPs (warm start). I asked about SIMD optimization and Ivet told me that HiGHS was not strongly optimized for this use case. 

With the help of LLMs I did a direct translation of the simplex part of HiGHS (https://github.com/LaurentPlagne/TinyHIGHS.jl) in julia (because I am more productive in micro-optimization with this language). Profiling and optimizing this toy model, I have implemented some optimization that allow TinyHiGHS to achieve better performance than HiGHS (and Clp) for my use case. 

- Some of these performance optimizations can be easily applied to HiGHS -> this PR (vectorization of FP divisions..)
- Some of these performance optimization involve deeper changes that could be discussed by your teams (reuse of cached workspace in place of multiple creations/destructions). If you are interested by the later you may have a look to the TinyHiGHS.jl project.

Thank you very much for providing such an amazing tool for the community !

Best regards,

Laurent Plagne

The rest of the message has been generated by a LLM (the english level is definively better :)

During our work on porting and benchmarking HiGHS's revised simplex solver to Julia in the open-source project [TinyHiGHS.jl](https://github.com/LaurentPlagne/TinyHIGHS.jl), we analyzed micro-architectural bottlenecks during FTRAN and BTRAN on real-world LP instances.

This PR introduces an exact, zero-overhead short-circuit for unit diagonal pivots in `HFactor` and includes an official Catch2 warm-start sequence regression test.

---

### 1. Motivation & Micro-architectural Impact
In network flow, transportation, scheduling, and multi-commodity problems, constraint matrices are predominantly composed of $\{0, \pm 1\}$ coefficients. During LU factorization with Forrest-Tomlin updates:
- **82.8% of the diagonal pivots in $U$ are exactly $+1.0$**, and **7.3% are $-1.0$**.
- Over **90.1% of all floating-point divisions during FTRAN and BTRAN are divisions by $\pm 1.0$**.

On modern x86-64 and AArch64 architectures:
- Floating-point division (`FDIV` / `vdivsd`) has a latency of 10 to 15 clock cycles and cannot pipeline like addition or multiplication.
[...]

PR #3301, opened 2026-09-18
HFactor: portable vectorized pre-inversion and branchless substitutio…

LaurentPlagne:
Hi, 

I hope that this PR correct the flaws of the previous one (no more branching).

My previous timings were totally off because I did compare the result of a native compiled HiGHS binary on my machine with portable HIGHS artifact binary provided by Julia. I was a bit shocked to discover that it was nor a fat binary but a generic one. The difference is quite substantial (40%) in my case. I do believe that this should better emphasized in the Julia/JuMP community (you can use a locally compiled HiGHS binary with JuMP). It would be great if a fat binary artifact could be directly accessible in Julia (It is done for BLAS for example) but I do not enough about this topic and its consequences.

Best regards,

Laurent Plagne

The rest is AI generated.

### Summary of Changes
Following the discussion and test results from PR #3295 and #3297, this PR introduces a **strictly branchless** approach to eliminate the floating-point division (`FDIV`) bottleneck in triangular substitutions (`ftranU`, `btranU`, and `solveHyper`), avoiding the branch misprediction penalties observed previously on MIP relaxations.

1. **Portable Auto-Vectorized Pre-Inversion**:
   In `HFactor::buildFinish`, diagonal pivot inverses ($D^{-1} = 1.0 / U_{ii}$) are precomputed in a single contiguous loop with standard compiler vectorization directives (`HIGHS_RESTRICT`, `#pragma clang loop vectorize(enable)`, `#pragma GCC ivdep`, `#pragma loop(ivdep)`, `#pragma ivdep`).
   - Cleanly auto-vectorizes into SIMD instructions across Clang/AppleClang, GCC, MSVC, and Intel oneAPI without raw platform-dependent intrinsics.
2. **Dynamic Updates & Representation Sync**:
   In `HFactor::updateFT`, `updateMPF`, and `addRows`, reciprocal pivots are tracked alongside `u_pivot_value`. `InvertibleRepresentation` preserves and restores `u_pivot_inv_value`.
3. **Branchless Substitution**:
   In `solveHyper`, `ftranU`, and `btranU`:
   ```cpp
   pivot_multiplier *= u_pivot_inv_value[i_logic];


odow:
> I did compare the result of a native compiled HiGHS binary on my machine with portable HIGHS artifact binary provided by Julia. I was a bit shocked to discover that it was nor a fat binary but a generic one. The difference is quite substantial (40%) in my case.

What machine and compiler? What BLAS did you use?

Could you try one of the binaries from https://github.com/ERGO-Code/HiGHS/releases/tag/v1.15.1

LaurentPlagne:
Hi Oscar, my machine is m1 max. No specific Blas for this test (I do not know if HiGhs relies on any BLAS function)

 It is bit late for a detailed description but I was comparing  natively compiled HiGHs binary with Julia artifact and (wrongly) concluded that my patch (the previous branching one) was responsible for the speed up...

LaurentPlagne:
Argh, the tests failed (I had no failure on my fork...).

LaurentPlagne:
Since you are here, you can have a Look to the Julia translation of the simplex (TinyHiGHS.jl). It leads to a rather large speedup on my hydro case (mainly because there is zero allocations for warm starts)

odow:
What compiler (and version)?

odow:
> you can have a Look to the Julia translation of the simplex (TinyHiGHS.jl)

1. It's pretty cool that this is now possible.
2. HiGHS is definitely geared towards large complex MIPs. Tiny LPs like this don't really get measured in benchmarking decisions (@Opt-Mucca). Have you tried GLPK? I'd be interested to see how it compared. In some examples I found that GLPK was much faster than HiGHS, mostly because it did fewer complicated things (so it was less robust and slower on larger models).

LaurentPlagne:
Hi Oscar,

Some steps to reproduce

### Environment & Setup
- **Machine**: Apple M1 Max (macOS arm64).
- **Compiler**: Apple Clang (`clang++ -O3`).
- **BLAS**: None (`BUILD_OPENBLAS=OFF`), this is pure revised simplex warm-start.

---

### Benchmark Results on `sequence_small` 
| Build / Engine | Total Time | Time / solve | Notes |
| :--- | :---: | :---: | :--- |
| **`HiGHS_jll` v1.15.1 (Yggdrasil)** | **16.78 ms** | 220.8 µs | Standard Julia artifact |
| **Official GitHub Release v1.15.1** | **6.00 ms** | 79.0 µs | `highs-1.15.1-arm-apple-static-mit.tar.gz` |
| **Local native compile (`clang++ -O3`)** | **4.03 ms** | 53.0 µs | Compiled locally on M1 |
| **TinyHiGHS.jl (pure Julia prototype)** | **2.00 ms** | 26.3 µs | Zero-allocation persistent buffers |


---

###Reproduce


```bash
git clone https://github.com/LaurentPlagne/TinyHIGHS.jl.git
cd TinyHIGHS.jl

# 1. To run the C++ replay comparing your installed libhighs:
./contrib_highs/run_bench_cpp.sh

# 2. To run the full 3-way benchmark in Julia:
julia --project=. bench/bench_3way.jl

LaurentPlagne:
... And I cannot use GLPK because it is not thread safe.

jajhall:
This pre-division is the obvious alternative to branching, but I did wonder whether the overhead of performing divisions corresponding to pivots that are never used (due to hyper-sparsity) might offset any advantage. 

I'll do some testing on my benchmark LPs, as there must be no regression there for this to be included in HiGHS 

In passing, when studying individual LPs yesterday, I didn't reproduce the  performance gains I quoted on full set testing. This is unsurprising given the minimal change in the code and branching that was introduced

jajhall:
> HiGHS is definitely geared towards large complex MIPs. Tiny LPs like this don't really get measured in benchmarking decisions (@Opt-Mucca). Have you tried GLPK? I'd be interested to see how it compared. In some examples I found that GLPK was much faster than HiGHS, mostly because it did fewer complicated things (so it was less robust and slower on larger models).

That said, we have have discussions with potential clients who solve lots of tiny problems and would want HiGHS to be faster.

Some day it would be interesting to see what GLPK does that can make it faster on tiny problems. Whilst what @odow says is true for the MIP solver, it's harder to see for the simplex solver.

LaurentPlagne:
> This pre-division is the obvious alternative to branching, but I did wonder whether the overhead of performing divisions corresponding to pivots that are never used (due to hyper-sparsity) might offset any advantage.
> 
> I'll do some testing on my benchmark LPs, as there must be no regression there for this to be included in HiGHS
> 
> In passing, when studying individual LPs yesterday, I didn't reproduce the performance gains I quoted on full set testing. This is unsurprising given the minimal change in the code and branching that was introduced

Yes, I fell into various perf traps during this PR process (target, compiler, artifacts,...). I am not very experienced with large open source software contribution. The LLM agent assistance help a lot to overcome some technical steps but also make stupid mistakes... Nevertheless, the Julia translation of the HiGHS simplex would have been totally out of my reach without agents : and the gain is very substantial on my specific Hydro project. 

You may try to  relaunch the perf test in this PR to see if it is generally beneficial or not. I would be super happy if it helps but not that surprised if it is not the case : not that easy to make useful contributions on important software.

Concerning, the specific case of numerous similar tiny LPs (pretty commonly encountered in dynamic programming), the LP driver/modeler was in my case a very important topic (more important than the LP solver performance). The JuMP modeler is super expressive and general, but Its design is not primarily tuned for this kind of workflow. I had to write a simple "reactive" modeler that allows to define your problem once (providing function definition for variables, constraints and objective), and the underlying LP solver instance is automatically updated (with no allocations) when the parameters are modified.

LaurentPlagne:
A final word on the topic of small LPs : the main improvement in TinyHIGHS is to avoid contruction/elimination of temporary vectors (replaced by reuse of workspace). These allocations are of special importance in Julia because the GC tends to kill the MT performances. But I think that it is also relevant in C++ because of system calls (unless you use specific specialized pool allocators). This changes make it faster than Clp and GLPK -> I guess that it should be the way to go for HiGHS too (for this specific class of problem).

jajhall:
> A final word on the topic of small LPs : the main improvement in TinyHIGHS is to avoid contruction/elimination of temporary vectors (replaced by reuse of workspace). These allocations are of special importance in Julia because the GC tends to kill the MT performances. But I think that it is also relevant in C++ because of system calls (unless you use specific specialized pool allocators). This changes make it faster than Clp and GLPK -> I guess that it should be the way to go for HiGHS too (for this specific class of problem).

Thanks, I can look into these. Presumably your unit test in `check/TestSequenceWarmStart.cpp` exposes these

As for the pre-inversion, I've started testing over my benchmark LPs, but interactive experiments on individual LPs show a slow-down, if anything. I'm using gcc on Linux, and assume that the loop in invertPivots is vectorised. I'm not sufficiently literate when it comes to such matters to check, though. Clearly if this isn't happening, there will be a performance hit. 

It's also interesting to see the extent to which pre-inversion leads to changes in simplex iteration count - and the failure to solve one problem! It's due to different rounding causing changes in tie-breaking, so random. There's no suggestion that there is any error in what you've done.

LaurentPlagne:
## Correction: initial replay sequences were corrupted

I identified an issue in the first generated warm-start replay corpus used
for the benchmark.

The LP serializer omitted zero-cost objective variables, and the reader did
not consistently restore the canonical `c0..cN` column order. This affected
the anonymized `sequence_small` and `sequence_medium` replay assets.

Therefore, the previous iteration counts, objectives, and timing results
based on those files must be disregarded. The Netlib instances were not
affected.

The issue is fixed in TinyHiGHS commit
[`c084a8e`](https://github.com/LaurentPlagne/TinyHIGHS.jl/commit/c084a8e).
The sequence assets were regenerated and the replay tests now pass.

Please use the corrected assets from:
https://github.com/LaurentPlagne/TinyHIGHS.jl

@odow

LaurentPlagne:
> As for the pre-inversion, I've started testing over my benchmark LPs, but interactive experiments on individual LPs show a slow-down, if anything. I'm using gcc on Linux, and assume that the loop in invertPivots is vectorised. I'm not sufficiently literate when it comes to such matters to check, though. Clearly if this isn't happening, there will be a performance hit.

Ah, deeply sorry for the noise (and the time you have wasted on this). I will try my best to avoid that in the future. 
Though, I am pretty confident that the tempo elimination has a substantial effect on the performance. But it is more demanding and probably require a deeper understanding of the HiGHS achitecture. You can check the effect on TinyHIGHS.jl

Thank you again for your patience.

jajhall:
> Ah, deeply sorry for the noise (and the time you have wasted on this). I will try my best to avoid that in the future.
> Though, I am pretty confident that the tempo elimination has a substantial effect on the performance. But it is more demanding and probably require a deeper understanding of the HiGHS achitecture. You can check the effect on TinyHIGHS.jl

No problem @LaurentPlagne, I'm glad you shared the enhancement that worked for you. Few people have anything to contribute in simplex, so it was interesting to experiment.

Opt-Mucca:
> Have you tried GLPK? I'd be interested to see how it compared. In some examples I found that GLPK was much faster than HiGHS, mostly because it did fewer complicated things (so it was less robust and slower on larger models).

I have only really used GLPK once years ago and never really explored it for small instances, although I've heard this anecdote from @odow and @jajhall 

> Some day it would be interesting to see what GLPK does that can make it faster on tiny problems. 

Agree. I hope some of the recent presolve changes now provide options for such instances, e.g., light presolve -> trivial heuristics -> root LP -> proven bound within 0.01%.

LaurentPlagne:
I tried GLPK and it could be faster on some instances but more brittle (failure on other instances). More importantly, I cannot use GLPK because it is not thread safe (it looks difficult to make it so) and I solve numerous LPs in parallel.

LaurentPlagne:
Have a look at section 2 here https://laurentplagne.github.io/TinyHIGHS.jl/dev/porting_optimizations/

LaurentPlagne:
One quick test would be to see if a pool allocator has some effect one the perf (e.g. https://www.boost.org/doc/libs/1_38_0/libs/pool/doc/interfaces/pool_alloc.html). Intel TBB have some fast allocators too.

PR #3306, opened 2026-09-20
Use the thread-safe aliases in the kZeroObjSingletonContinuousCol undo case

### LINKED ISSUE #3305: Parallel MIP: kZeroObjSingletonContinuousCol undo bypasses the thread-safe aliases, aborting the process
Version: `latest` (reproduced at `a740ff7c78`). Not present on `master` — the reduction type has not been released yet.
Language: C++ (surfaced via the Rust `highs-sys` bindings, but the defect is in the C++ core).

MIP, with `parallel=on` (the shipped default).

macOS 15 (arm64) and Linux x86-64; not OS-specific.

**What did you do?**
Solved a MIP large enough for the parallel search to find an incumbent on a worker thread, where presolve had applied the `kZeroObjSingletonContinuousCol` reduction (a zero-cost continuous singleton in a ranged row).

The model solved to optimality, as it is with `parallel=off` or `threads=1`.

**What did you see instead?**
The process aborts. Depending on the run: `std::length_error`, `std::bad_alloc`, `SIGABRT`, or `SIGBUS`. When it does not abort it silently returns a solution that violates the ranged row the singleton was removed from.

**Diagnosis**

`HighsPostsolveStack::undo` aliases its three mutable members to thread-local copies when `thread_safe` is set (`highs/presolve/HighsPostsolveStack.h:1069-1071`):

```cpp
HighsDataStack& reductionValues_ = thread_safe ? reductionValuesCopy : reductionValues;
std::vector<Nonzero>& colValues_  = thread_safe ? colValuesCopy      : colValues;
std::vector<Nonzero>& rowValues_  = thread_safe ? rowValuesCopy      : rowValues;
```

Fifteen of the sixteen `switch` cases use those aliases. `kZeroObjSingletonContinuousCol` (`:1294`) pops from the members instead:

```cpp
case ReductionType::kZeroObjSingletonContinuousCol: {
  ZeroObjSingletonContinuousCol reduction;
  reductionValues.pop(rowValues);      // members, not reductionValues_ / rowValues_
  reductionValues.pop(reduction);
  reduction.undo(options, rowValues, solution, basis);
  break;
}
```

With `thread_safe = true` this reads the member stack at a position nobody reset — `undoPrimal` only calls `resetPosition()` on the member stack in the serial path — and mutates state shared with the other workers.

`HighsDataStack::pop(std::vector<T>&)` reads the element count from the stack at the current position and resizes to it, so a stale position yields an arbitrary count: `std::length_error` when it exceeds `max_size()`, `std::bad_alloc` when it does not, and an out-of-bounds read when `position` underflows past zero.

`HighsMipWorker::transformNewIntegerFeasibleSolution` is the only caller that passes `thread_safe = true`, which is why this appears only in a parallel MIP search. The throw lands on a `HighsTaskExecutor` worker thread with no handler above it, so `std::terminate` runs and the process dies. `Highs::optimizeModelTryCatch` only guards the main-thread chain, which is why the same corruption sometimes surfaces as a returned `Error` and sometimes kills the process.

The case was added in `9cce7235d9` ("Remove continuous singletons from double-sided rows", 2026-04-10), seven months after `848ced1785` ("Add thread safe undoPrimal option", 2025-09-11) introduced `thread_safe`, so the new case body was written against the pre-`thread_safe` shape.

**Suggested fix**

Three lines, using the aliases that are already in scope:

```diff
         case ReductionType::kZeroObjSingletonContinuousCol: {
           ZeroObjSingletonContinuousCol reduction;
-          reductionValues.pop(rowValues);
-          reductionValues.pop(reduction);
-          reduction.undo(options, rowValues, solution, basis);
+          reductionValues_.pop(rowValues_);
+          reductionValues_.pop(reduction);
+          reduction.undo(options, rowValues_, solution, basis);
           break;
         }
```

**Minimal reproducer**

This does not need a MIP — it is reachable directly through `undoPrimal`:

```
min  x0 + x1
s.t.  2 <= x0 + x1 + s <= 6
      3 <= x0 + 2 x1
      0 <= x0, x1 <= 10,  0 <= s <= 4
```

`s` is a zero-cost continuous singleton in ranged row 0, so presolve relaxes it out and its value is recovered only in postsolve. Calling `undoPrimal(..., thread_safe = true)` on the resulting stack returns `s = 0` and row 0 comes back violated (activity 1.5 against a lower bound of 2); with `thread_safe = false` it is correct. Running the thread-safe call concurrently on eight tasks is what produces the throws.

**Verification against a real model**

These runs were made on our fork, which is upstream `latest` at `939eba361d` plus some local MIP-option commits, rather than on a clean `latest` build — flagging that so the numbers are read for what they are. A 12-product-location dispatch model (2,982 rows / 3,450 columns / 540 integers):

| config | unpatched | with the fix |
|---|---|---|
| `parallel=on`, `threads=0` (shipped default) | SIGABRT `std::length_error` 3/3 | optimal 5/5 |
| `parallel=on`, `threads=4` | SIGABRT / SIGBUS | optimal 3/3 |
| `parallel=on`, `threads=1` | optimal | optimal 3/3 |
| `parallel=off` | optimal | optimal 3/3 |

A larger 27-product-location model that aborted with `std::bad_alloc` via `HighsPrimalHeuristics::RINS` → `solveSubMip` → `trySolution` also solves to optimality 3/3. In both cases the patched objective matches the single-threaded and `parallel=off` runs, so this restores correct postsolve rather than merely suppressing the throw.

What I verified directly on a clean `latest` at `a740ff7c78`: the minimal reproducer above, as a `check/TestPresolveRules.cpp` case. It fails without the fix (`REQUIRE( rowZeroActivity(thread_safe) >= 2 )` → `1.5 >= 1.999999999`) and passes with it, and the full unit suite is green with the fix (378 cases / 1,260,264 assertions).

**Anything else we should know**

Secondary and separate from the above: `HighsTaskExecutor::run_worker` / `HighsSplitDeque::runStolenTask` let any `std::exception` escape a task to `std::terminate`. That is what turns this particular bug into a process death rather than an error return, and would do the same for any other throw on a worker.

Per CONTRIBUTING.md I am raising this before opening a PR. I have the fix and a regression test ready on a branch (full unit suite green: 378 cases / 1,260,264 assertions) and will link the PR here — happy for you to take the one-line change yourselves instead if that is easier.

The analysis and patch above were AI-assisted; I have reviewed and verified them.

MathieuDutSik:
## Description

Fixes a crash in the parallel MIP search: `kZeroObjSingletonContinuousCol` is the one `HighsPostsolveStack::undo` case that does not use the thread-safe aliases.

`undo` aliases its three mutable members to thread-local copies when `thread_safe` is set (`HighsPostsolveStack.h:1069-1071`):

```cpp
HighsDataStack& reductionValues_ = thread_safe ? reductionValuesCopy : reductionValues;
std::vector<Nonzero>& colValues_  = thread_safe ? colValuesCopy      : colValues;
std::vector<Nonzero>& rowValues_  = thread_safe ? rowValuesCopy      : rowValues;
```

Fifteen of the sixteen `switch` cases go through those aliases. `kZeroObjSingletonContinuousCol` pops from the members instead, so with `thread_safe = true` it reads the member stack at a position nobody reset — `undoPrimal` calls `resetPosition()` on the member stack only in the serial path — and mutates state shared with the other workers.

`HighsDataStack::pop(std::vector<T>&)` reads the element count from the stack at the current position and resizes to it, so a stale position yields an arbitrary count: `std::length_error` when it exceeds `max_size()`, `std::bad_alloc` when it does not, and an out-of-bounds read when `position` underflows past zero. When it does not throw, the singleton comes back as 0 and the ranged row it was removed from is left violated.

`HighsMipWorker::transformNewIntegerFeasibleSolution` is the only caller passing `thread_safe = true`, which is why this appears only in a parallel MIP search. The throw lands on a `HighsTaskExecutor` worker with no handler above it, so `std::terminate` runs and the process dies; `Highs::optimizeModelTryCatch` only guards the main-thread chain, which is why the same corruption sometimes surfaces as a returned `Error` instead.

The case was added in `9cce7235d9` ("Remove continuous singletons from double-sided rows", 2026-04-10), seven months after `848ced1785` ("Add thread safe undoPrimal option", 2025-09-11) introduced `thread_safe`, so the new case body was written against the pre-`thread_safe` shape.

The fix is three lines, using the aliases already in scope.

## Verification

- The full unit suite is green.
- `clang-format --dry-run -Werror` is clean on the changed file.

Separately, on our fork (upstream `latest` at `939eba361d` plus some local MIP-option commits, so not a clean `latest` build), a 2,982-row / 3,450-column / 540-integer dispatch model that aborted with `std::length_error` 3/3 at the shipped `parallel=on`, `threads=0` default now solves to optimality 5/5, and a larger model that aborted with `std::bad_alloc` via `RINS` → `solveSubMip` → `trySolution` solves 3/3. The patched objectives match the `threads=1` and `parallel=off` runs, so this restores correct postsolve rather than only suppressing the throw.

## Checklist

- [x] Tests are passing
- [ ] Documentation was updated where relevant — N/A
- [ ] This PR is not primarily AI-generated (per the AI contributions policy in CONTRIBUTING.md)

The analysis and patch were AI-assisted; I have reviewed and verified them. Leaving that box unchecked deliberately rather than misreporting it — the source change is the three-line alias swap, so please take it directly if that is the easier path for you.

## Related issue

Closes #3305

Secondary and out of scope here: `HighsTaskExecutor::run_worker` / `HighsSplitDeque::runStolenTask` let any `std::exception` escape a task to `std::terminate`, which is what turns this bug into a process death rather than an error return.


jajhall:
Thanks, even to me this looks OK and acceptable. Pattern-matching (at least) tells me that `case ReductionType::kZeroObjSingletonContinuousCol` should use `reductionValues_` and `rowValues_` rather than `reductionValues` and `rowValues`. 

I see that `thread_safe` is only set true for `undoPrimal` in `HighsMipWorker`. I guess this is to avoid the overhead of allocating `reductionValuesCopy` in every undo operation where `reductionValues` are used. Otherwise it would look simpler to always use `thread_safe`.

Opt-Mucca:
@jajhall I should've spotted this when we merged `kZeroObjSingletonContinuousCol`, although a missing `_` is truly nasty....... This slipped through because we made those changes before merging parallel MIP (and then revived them a few weeks ago). 

@MathieuDutSik great spot! That bug currently completely corrupts some parallel MIP runs. Can you please remove the new test? I believe it is a bit overkill for this specific bit of the code and the parallel interaction. I'd then happily merge the change!

MathieuDutSik:
> @jajhall I should've spotted this when we merged `kZeroObjSingletonContinuousCol`, although a missing `_` is truly nasty....... This slipped through because we made those changes before merging parallel MIP (and then revived them a few weeks ago).
> 
> @MathieuDutSik great spot! That bug currently completely corrupts some parallel MIP runs. Can you please remove the new test? I believe it is a bit overkill for this specific bit of the code and the parallel interaction. I'd then happily merge the change!

Done.

Opt-Mucca (approved):

jajhall:
Spurious CI failures

PR #3309, opened 2026-09-21
Remove `removeSlacks`

jajhall:
## Description

Fuzzing identified an error in a dual value when the `removeSlacks` presolve reduction was used.

The `removeSlacks` presolve reduction cannot be used when the cost is nonzero, since dual postsolve cannot be computed (due to the objective substitution changing basic costs so the row duals can't be maintained in postsolve). When `removeSlacks` is restricted to zero cost column singletons, it's a duplicate of the `zeroCostSingleton` reduction.

Note that the `removeSlacks` presolve reduction was off by default - controlled by the `presolve_remove_slacks` option - because it didn't seem to be valuable for LPs. The `presolve_remove_slacks` option has been retained - albeit with "redundant option!" added to its description, since removing it would be a breaking change. 

## Checklist

- [ ] Tests are passing


Opt-Mucca (approved):
Looks good! I support removing dead code.

jajhall:
Spurious CI failures, since changes are removal of code

PR #3310, opened 2026-09-21
Feasibility jump: fix #3302 and #3303

### LINKED ISSUE #3302: Feasibility Jump: rows with a negative coefficient are skipped when computing the jump value
**Version / language:** `master` (73cac48, 2026-08-19); the same code is in v1.15.1 and on `latest`. C++.
**Solver:** MIP — Feasibility Jump, `highs/mip/feasibilityjump.hh`
**OS:** Linux x86-64; not platform-specific.

### Summary

`JumpMove::updateValue` skips every `Lte`/`Gte` row in which the variable has a negative coefficient, so the jump value is computed as if those rows did not exist. HiGHS only ever passes `Lte`/`Gte` rows to FJ (`HighsFeasibilityJump.cpp` adds a `Gte` row for a finite row lower bound and an `Lte` row for a finite upper bound), so inside HiGHS this is every row with a negative coefficient, ranged and equality rows included.

### Detail

For each row containing the variable, [`updateValue`](https://github.com/ERGO-Code/HiGHS/blob/73cac48c5340d775a477087198611862559be250/highs/mip/feasibilityjump.hh#L362-L373) divides both endpoints of the row's bound interval by the coefficient:

```cpp
std::pair<double, double> validRange = {
    ((1.0 / cell.coeff) * (bound.first  - residualIncumbent)),
    ((1.0 / cell.coeff) * (bound.second - residualIncumbent)),
};
// ... integer rounding ...
if (validRange.first > validRange.second) continue;
```

Dividing by a negative number reverses the inequality, but the endpoints are not swapped. For a negative coefficient an `Lte` row gives `(+inf, t)` and a `Gte` row gives `(t, -inf)`, so `first > second` always holds and the row is skipped. It adds neither a critical value to `bestShiftBuffer` nor a term to `currentSlope`. (`RowType::Equal` is only partly affected, since its `(rhs, rhs)` segment survives, but HiGHS never emits it.)

In the paper (Sect. 2.2), the critical value in eq. (5) and the penalty in eq. (6) each have a separate case for `a_ij < 0`, and Algorithm 1 accumulates the slope from those penalties.

`resetMoves` scores whatever value `updateValue` returns, so the score is correct for the value chosen; the defect is that the wrong value is proposed. Binaries are barely affected, because their jump value is the opposite bound either way. General integers and continuous columns are what change.

### What did you do

One integer variable `x0` in `[0,10]`, incumbent 0, and the constraint `x0 >= 3` written two equivalent ways. Header-only, no linking needed; from the root of an unmodified `master` checkout after a CMake configure: `g++ -std=c++11 -I highs -I build repro.cpp`.

```cpp
#include <cstdio>
#include "mip/feasibilityjump.hh"
namespace fj = external_feasibilityjump;

double jump(fj::RowType sense, double rhs, double coeff) {
  fj::Problem p(1e-5, 1e-5);
  p.addVar(fj::VarType::Integer, 0.0, 10.0, 0.0);
  int idx = 0;
  p.addConstraint(sense, rhs, 1, &idx, &coeff, 0);
  double x0 = 0.0;
  p.resetIncumbent(&x0);
  fj::JumpMove jm(1e-5);
  jm.init(p);
  jm.updateValue(p, 0);
  double v = 0;
  jm.forEachVarMove(0, [&](fj::Move& m) { v = m.value; });
  return v;
}

int main() {
  printf("%g\n", jump(fj::RowType::Gte,  3.0,  1.0));  //  x0 >=  3
  printf("%g\n", jump(fj::RowType::Lte, -3.0, -1.0));  // -x0 <= -3
  printf("%g\n", jump(fj::RowType::Gte,  7.0,  2.0));  //  2 x0 >=  7
  printf("%g\n", jump(fj::RowType::Lte, -7.0, -2.0));  // -2 x0 <= -7
}
```

### Expected

`3 3 4 4`: the jump value should not depend on which way the row is written.

### Saw

`3 10 4 10`. With a negative coefficient the row is ignored and the variable jumps to a bound. The same shows end to end: `FeasibilityJumpSolver::solve` on the one-row model reports a first feasible `x0 = 3` for the `+1` spelling and `x0 = 10` for the `-1` spelling.

### Suggested fix

Swap the endpoints for a negative coefficient, before the integer rounding:

```cpp
if (cell.coeff < 0.0) std::swap(validRange.first, validRange.second);
```

With that line the program prints `3 3 4 4`, and a continuous `x0` with `-2 x0 <= -7` gives 3.5. The positive-coefficient cases are unchanged. If the swap is placed after the rounding instead, `ceil`/`floor` have been applied to the wrong ends, and a fractional critical value comes out one below the smallest integer satisfying the row.

### Impact

I have not measured the effect of this change alone on HiGHS's MIP performance. The repro shows a wrong jump value, not an infeasible or incorrect result.

### Context

The same lines are in the [SINTEF reference implementation](https://github.com/sintef/feasibilityjump/blob/a293cdd4f4e7575ea38de387bcdf950b5f67b847/feasibilityjump.hh#L374-L385), so the authors may want it too. Filed as an issue rather than a PR, per CONTRIBUTING.md.

Reference: Luteberget & Sartor, *Feasibility Jump: an LP-free Lagrangian MIP heuristic*, Math. Prog. Comp. 15, 365-388, 2023. [doi:10.1007/s12532-023-00234-8](https://doi.org/10.1007/s12532-023-00234-8)

### LINKED ISSUE #3303: Feasibility Jump: the objective term of the move score has the wrong sign
**Version / language:** `master` (73cac48, 2026-08-19); the same code is in v1.15.1 and on `latest`. C++.
**Solver:** MIP — Feasibility Jump, `highs/mip/feasibilityjump.hh`
**OS:** Linux x86-64; not platform-specific.

### Summary

The objective term in an FJ move's score has the wrong sign: a move that worsens the objective scores positively. After first feasibility, FJ's improving mode therefore prefers worse solutions.

### Detail

Sect. 2.6 of the paper extends the score to a minimized sum of an objective term and the violation terms. In the implementation every other part of the score is improvement-positive:

- a constraint term is `weight * (score(new) - score(old))`, and `Constraint::score` returns minus the violation;
- `selectVariable` takes the maximum score;
- `updateGoodMoves` calls a move good when `bestMove(varIdx).score > 0`.

The objective term is added rather than subtracted, at two sites:

```cpp
// resetMoves
move.score += objectiveWeight * problem.vars[varIdx].objectiveCoeff *
              (move.value - problem.incumbentAssignment[varIdx]);

// updateWeights, the branch taken when no constraint is violated
move.score += weightUpdateIncrement *
              problem.vars[varIdx].objectiveCoeff *
              (move.value - problem.incumbentAssignment[varIdx]);
```

([resetMoves](https://github.com/ERGO-Code/HiGHS/blob/73cac48c5340d775a477087198611862559be250/highs/mip/feasibilityjump.hh#L755-L756), [updateWeights](https://github.com/ERGO-Code/HiGHS/blob/73cac48c5340d775a477087198611862559be250/highs/mip/feasibilityjump.hh#L654-L656).) `HighsMipSolverData::feasibilityJump` passes `addVar` a coefficient already multiplied by the model sense, so FJ always minimizes. A move that increases the objective then produces a positive term and can be selected as a good move. Nothing else compensates: `incumbentObjective` and the `bestObjective` test minimize, and `objectiveWeight` is never negative.

Reported solutions are still feasible and correctly valued, because the callback fires only when `violatedConstraints.empty()` and `incumbentObjective` is maintained independently of the score. Only the search direction is affected. Time to first feasibility is not affected either: `objectiveWeight` starts at `0.0` and is incremented only on the `violatedConstraints.empty()` branch of `updateWeights`, so the term is zero until a feasible point has been reached. The effect is partly masked, since with no violated constraint and no good move `selectVariable` falls through to `rng() % problem.vars.size()`, a score-blind pick that can improve the objective by luck.

### What did you do

Drove `FeasibilityJumpSolver` directly on `min sum_{j=1..20} j*x_j  s.t.  sum x_j >= 3`, x binary, starting from x = 0, up to 2e6 effort. The optimum is 6. Built from an unmodified `master` checkout: `g++ -std=c++11 -I highs -I build repro.cpp build/lib/libhighs.a -lz -pthread`.

```cpp
#include <cstdio>
#include <cstdlib>
#include <vector>
#include "mip/feasibilityjump.hh"
using namespace external_feasibilityjump;

int main(int argc, char** argv) {
  int seed = argc > 1 ? atoi(argv[1]) : 0;
  HighsLogOptions lo;
  lo.clear();
  bool off = false;
  HighsInt dev = 0;
  lo.output_flag = &off;
  lo.log_to_console = &off;
  lo.log_dev_level = &dev;
  FeasibilityJumpSolver s(lo, seed, 1e-9, 1e-6);
  const int n = 20;
  std::vector<int> idx(n);
  std::vector<double> coeff(n, 1.0);
  for (int j = 0; j < n; j++) {
    s.addVar(VarType::Integer, 0, 1, j + 1.0);
    idx[j] = j;
  }
  s.addConstraint(RowType::Gte, 3.0, n, idx.data(), coeff.data(), 0);
  std::vector<double> x0(n, 0.0);
  double best = 1e300;
  int nsol = 0;
  s.solve(x0.data(), [&](FJStatus st) {
    if (st.solution) {
      best = st.solutionObjectiveValue;
      nsol++;
    }
    return st.totalEffort > 2000000 ? CallbackControlFlow::Terminate
                                    : CallbackControlFlow::Continue;
  });
  printf("seed %d best %g solutions %d\n", seed, best, nsol);
}
```

### Expected

After the first feasible point, the objective improves toward 6.

### Saw

| seed | 0 | 1 | 2 | 3 | 4 |
|---|---|---|---|---|---|
| `master` | 35 | 22 | 26 | 53 | 40 |
| both signs negated | 6 | 6 | 6 | 6 | 6 |

The unmodified header reports exactly one solution per run and never improves on it. With both signs negated every seed reaches the optimum, with 2-3 improving solutions.

### Suggested fix

Change `+=` to `-=` at both sites. They should change together: `updateWeights` applies incrementally what `resetMoves` recomputes, so opposite signs would make the stored scores drift from their definition on every weight bump.

### Why it may have gone unnoticed

The paper says "the objective function was not taken into account in any of the computational results" (Sect. 2.6) and that the authors "experimented with giving the original objective function some importance through a simple weight update function" (Sect. 5), so this path may never have been benchmarked. I have not measured the effect on a full HiGHS MIP solve, where FJ stops at `nnz << 8` effort after its last improvement.

### Context

The same two lines are in the SINTEF reference implementation ([updateWeights](https://github.com/sintef/feasibilityjump/blob/a293cdd4f4e7575ea38de387bcdf950b5f67b847/feasibilityjump.hh#L629-L631), [resetMoves](https://github.com/sintef/feasibilityjump/blob/a293cdd4f4e7575ea38de387bcdf950b5f67b847/feasibilityjump.hh#L745-L747)), so the authors may want it too. Filed as an issue rather than a PR, per CONTRIBUTING.md.

Reference: Luteberget & Sartor, *Feasibility Jump: an LP-free Lagrangian MIP heuristic*, Math. Prog. Comp. 15, 365-388, 2023. [doi:10.1007/s12532-023-00234-8](https://doi.org/10.1007/s12532-023-00234-8)

Related: #3302, a second defect in the same file.

BenChampion:
## Description

1. Handles jump candidates with negative constraint coefficients #3302
1. Corrects sign of objective contribution #3303
1. Removes double-counting for equality constraints in jump candidate handling (noticed when fixing the other two)

Also amends comments.

## Checklist

- [ ] Tests are passing

## Related issues
Closes #3302
Closes #3303


BenChampion:
@jajhall I'd like to performance test this - we can chat about it tomorrow

jajhall (approved):
Good spot. Were these errors in the original code?

highs/mip/feasibilityjump.hh
```diff
@@ -359,9 +358,17 @@ class JumpMove {
         double residualIncumbent =
             constraint.incumbentLhs - cell.coeff * varIncumbentValue;
 
+        double boundForValidLower, boundForValidUpper;
```
fwesselm, line 361: I am not sure if it is any clearer that what you have, but you could also write:

```
auto localBound = bound;
if (cell.coeff < 0.)
  std::swap(localBound.first, localBound.second);
```

highs/mip/feasibilityjump.hh
```diff
@@ -754,7 +763,8 @@ class FeasibilityJumpSolver {
 
     forEachMove(varIdx, [&](Move& move) {
       move.score = 0.0;
-      move.score += objectiveWeight * problem.vars[varIdx].objectiveCoeff *
+      // -= to align objective minimization with score maximization
+      move.score -= objectiveWeight * problem.vars[varIdx].objectiveCoeff *
```
fwesselm, line 767: How important is accuracy in the score computation? I guess using `HighsCDouble` would be overkill since it is not needed for correctness.

highs/mip/feasibilityjump.hh
fwesselm: `int varIdx`?

fwesselm (approved):
Looks good, @BenChampion!

BenChampion:
> Good spot. 

Credit to @ spoorendonk!

>Were these errors in the original code?

Yes. I did check for interactions with new functionality, though.

highs/mip/feasibilityjump.hh
BenChampion: I think not very...but worth considering for any new heuristic in this family.

highs/mip/feasibilityjump.hh
BenChampion: Oh, nice spot! (I think we never hit this code.) I'll change.

highs/mip/feasibilityjump.hh
BenChampion: Thanks for the suggestion. After considering this and two other `swap` possibilities I'm going to stick with what I have. (The problem with `std::pair` is that `.first` and `.second` don't provide much context...)

Opt-Mucca (approved):
I guess this would mostly change the quality of solutions FJ was returning (not the general ratio of success)?

BenChampion:
> I guess this would mostly change the quality of solutions FJ was returning (not the general ratio of success)?

It's conceivable that we might see a bit of both, because the bug meant it was basically missing constraints when trying to reduce infeasibilities.

The FJ paper didn't test objective improvement of the reference implementation, but it's not clear to me whether that meant using a zero objective or something else, which would influence the size of the impact we'll see on solution quality.

We'll find out!

BenChampion:
This is good to merge!

Benchmarking (with a focus on early feasible solutions) shows decent improvements on both time and solution quality.

Thanks to @Opt-Mucca for the scripts that made this very easy!

<details>
<summary>
Details (for those interested)
</summary>

With the fork-point `fd860f9b` as baseline, I ran the MIPLIB2017 benchmark 240 (3 random seeds) with
```
time_limit=120
mip_rel_gap=1
```
**Note that these are quite different settings from typical performance runs, on which the overall impact may be much smaller.**

|Subset|# Instance-Seed Pairs|
|--|--|
|All|720|
|Gap < 1 (test and baseline)|514|
|Gap < 1 (test and baseline) **AND** test/baseline objectives differ|146|

On the subset of 514, the time ratio (shifted geometric means, shift 1) was 0.94 in favour of the change
(Node ratio has little meaning here.) Nearly 2/3 of the objectives didn't change, but among the subset of 146 above that did, the gap difference (shifted geometric means, shift 1) was 0.15 in favour of the change.
</details>

